PR #6, opened 2019-01-29
build: update build tools with conventional commits

markov00:
This PR add the following features on the DX:
- add Husky as hook manager
- add commitlint as hook for linting commit following conventional-commits
- add commitizen as cli tool for "learn how to write conventional commits"
- update README and FAQs


.nvmrc
```diff
@@ -0,0 +1 @@
+8
```
emmacunningham: Could we change this to `8.0.0`?  (I use asdf as my version manager and it has a hard time unless the full version is explicitly spelled out.)

markov00: I've update it to latest LTS

PR #29, opened 2019-02-01
fix(axis): add axisTitleHeight to axis increments

### LINKED ISSUE #26: Fix issues with spacing of multiple axes with same position
Left axes are overflowing into each other 
Chart padding has been increased (should match right padding) 

image

emmacunningham:
fixes #26

Previously, without the axisTitleHeight offset, multiple axes with the same position would overflow
into each other with margins less than the title height.  Also it would appear to give the chart additional padding because the axis nearest to the chart wasn't positioned correctly.

Now, the title won't overflow  into the next axis:

image

(Has been tested with multiple axes for each of the positions)

### DIFF
--- a/src/lib/axes/axis_utils.ts
+++ b/src/lib/axes/axis_utils.ts
@@ -464,23 +464,24 @@ export function getAxisPosition(
 
   if (isVertical(position)) {
     if (position === Position.Left) {
-      leftIncrement = maxLabelBboxWidth + tickSize + tickPadding + chartMargins.left;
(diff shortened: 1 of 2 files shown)


markov00:
@emmacunningham could you add a story with multiple Y and X axes on top, bottom, left and right with debug knobs so we can add a test for this in the future?

markov00:
and maybe some knobs to show and hide the axes

emmacunningham:
@markov00 Yep, see commit 4607394434d98e25db14f54b534b4445f4fa04bc

There's a toggle to show/hide each of the 8 axes (knobs grouped by position) and also a theme group for knobs to adjust margins/paddings as that might also affect how the axes stack on top of each other:

image

Issue #16, opened 2019-01-31
Move axis grid on the background


markov00:
The axis grid lines should be moved on the background, behind chart elements to clean up the chart a bit.
We can make that grid appear maybe decreasing the opacity of the elements but I think it's better to move the grid behind elements.

PR #30, opened 2019-02-01
refactor(axis/grid): render grid lines in separate layer from axis

emmacunningham:
addresses #16 

Previously, we were rendering the grid lines relative to each axis in the same layer as the axis.  Now, each axis' grid lines are rendered in a separate layer just for grid lines and the coordinates for the lines are relative to the chart (instead of the axis).  This simplifies much of the positioning logic and also gives us the option to draw anything we might want onto the grid layer without affect the axes.

Also, in combination with PR #29, this fixes the issues with rendering grid lines for multiple axes with the same position (there was both overflow and spacing issues previously). 

Since we render the grid layer first, the lines render behind the chart elements:

image

### DIFF
--- a/src/components/react_canvas/grid.tsx
+++ b/src/components/react_canvas/grid.tsx
@@ -30,39 +30,29 @@ export class Grid extends React.PureComponent<GridProps> {
     }
 
     const {
-      axisSpec: { tickSize, tickPadding, position, gridLineStyle },
-      axisTicksDimensions: { maxLabelBboxHeight },
+      axisSpec: { position, gridLineStyle },
       chartDimensions,
-      chartTheme: { chart: { paddings } },
     } = this.props;
 
     const config = gridLineStyle ? mergeWithDefaultGridLineConfig(gridLineStyle) : DEFAULT_GRID_LINE_CONFIG;
 
     const lineProps = isVertical(position) ?
       getVerticalAxisGridLineProps(
-        position,
-        tickPadding,
-        tickSize,
         tick.position,
         chartDimensions.width,
-        paddings,
       ) : getHorizontalAxisGridLineProps(
-        position,
-        tickPadding,
-        tickSize,
         tick.position,
-        maxLabelBboxHeight,
         chartDimensions.height,
-        paddings,
       );
 
     return <Line key={`tick-${i}`} points={lineProps} {...config} />;
   }
 
   private renderGrid = () => {
-    const { ticks, axisPosition } = this.props;
+    const { ticks, chartDimensions } = this.props;
+
     return (
-      <Group x={axisPosition.left} y={axisPosition.top}>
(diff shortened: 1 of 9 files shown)


src/components/react_canvas/grid.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const config = gridLineStyle ? mergeWithDefaultGridLineConfig(gridLineStyle) : DEFAULT_GRID_LINE_CONFIG;
+
+    const lineProps = isVertical(position) ?
+      getVerticalAxisGridLineProps(
+        tick.position,
+        chartDimensions.width,
+      ) : getHorizontalAxisGridLineProps(
+        tick.position,
+        chartDimensions.height,
+      );
```
markov00: What if we move this to the state manager? It will keep the rendering layer clean from any computation.

src/lib/axes/axis_utils.ts
```diff
@@ -271,42 +271,17 @@ export function getHorizontalAxisTickLineProps(
 }
 
 export function getVerticalAxisGridLineProps(
-  position: Position,
-  tickPadding: number,
-  tickSize: number,
   tickPosition: number,
   chartWidth: number,
-  paddings: Margins,
 ): number[] {
```
markov00: Can we type this as `[number, number, number, number]` that is a bit more "precise" than an array of number? I we always have the following quadruple: `[x1,y1,x2,y2]` to be used as a points on a line, right?

src/lib/axes/axis_utils.ts
```diff
@@ -271,42 +271,17 @@ export function getHorizontalAxisTickLineProps(
...
 
 export function getHorizontalAxisGridLineProps(
-  position: Position,
-  tickPadding: number,
-  tickSize: number,
   tickPosition: number,
-  labelHeight: number,
   chartHeight: number,
-  paddings: Margins,
 ): number[] {
```
markov00: same as previous comment

src/components/react_canvas/grid.tsx
emmacunningham: Yes, great suggestion.  I've handled this in this commit: 1791f6af505069c28521e29f5485c85e2d72a647

src/lib/axes/axis_utils.ts
emmacunningham: See https://github.com/elastic/elastic-charts/commit/1791f6af505069c28521e29f5485c85e2d72a647; since this gets used in a few different places now, I created a type `AxisLinePosition` for convenience.

markov00:
I've only a minor stylistic comment:
chart elements are on a layer, clipped by the dimension.
Axis are on their own layer, not clipped, this means that we will see half line stroke on top and on bottom lines. @emmacunningham can we add clipping also on grids?

emmacunningham:
Added clipping props to the Grid layer:

image

Issue #24, opened 2019-01-31
Add on hover interaction on legend items


markov00:
Add on hover interaction on each legend item and highlight series on chart

PR #31, opened 2019-02-04
feat(legend/series): add hover interaction on legend items

emmacunningham:
This PR implements the feature (see #24) to add a hover interaction with legend items that highlights the corresponding series in the chart.  Currently, this is implemented to set the opacity of the non-highlighted series to 0.25 and the highlighted series to 1.  This renders like below:

bar series:
bar_legend

line series:
line_legend

area series:
area_legend

mixed series:
mixed_legend

Per the discussion below, there is also a refactoring of how `specId` and `seriesKey` are associated with a Geometry.  All Geometries now have will have these properties as they are required to be able to identify a specific geometry belonging to a spec within a series.

Also, per the discussion below, as we will likely want to share some common styles across all the different geometries, while this PR only introduces the stylistic feature of changing the opacity based on a highlighted legend item, there is the ability to update these shared styles in a way that allows for re-use across geometries.

### DIFF
--- a/src/components/react_canvas/bar_geometries.tsx
+++ b/src/components/react_canvas/bar_geometries.tsx
@@ -91,7 +91,7 @@ export class BarGeometries extends React.PureComponent<
         return 1;
       }
 
-      return 0.6;
+      return 0.25;
     }
     return 1;
   }
--- a/src/components/react_canvas/bar_geometries.tsx
+++ b/src/components/react_canvas/bar_geometries.tsx
@@ -73,22 +73,20 @@ export class BarGeometries extends React.PureComponent<
     onElementOut();
   }
 
-  private computeBarOpacity = (bar: BarGeometry, overBar: BarGeometry | undefined): number => {
-    if (overBar && overBar !== bar) {
-      return 0.6;
-    }
-    return 1;
-  }
-
   private renderBarGeoms = (bars: BarGeometry[]): JSX.Element[] => {
     const { overBar } = this.state;
     return bars.map((bar, i) => {
       const { x, y, width, height, color, value } = bar;
 
-      // This sets the opacity if any bars within the chart are hovered over
-      const opacity = this.computeBarOpacity(bar, overBar);
+      // Properties to determine if we need to highlight individual bars depending on hover state
+      const hasGeometryHover = overBar != null;
+      const hasHighlight = overBar === bar;
+      const individualHighlight = {
+        hasGeometryHover,
+        hasHighlight,
+      };
 
-      const geometryStyle = getGeometryStyle(bar.geometryId, this.props.highlightedLegendItem);
+      const geometryStyle = getGeometryStyle(bar.geometryId, this.props.highlightedLegendItem, individualHighlight);
 
       if (this.props.animated) {
         return (
@@ -103,7 +101,6 @@ export class BarGeometries extends React.PureComponent<
                   height={props.height}
                   fill={color}
                   strokeWidth={0}
-                  opacity={opacity}
                   perfectDrawEnabled={true}
                   onMouseOver={this.onOverBar(bar)}
                   onMouseLeave={this.onOutBar}
@@ -124,7 +121,6 @@ export class BarGeometries extends React.PureComponent<
             height={height}
             fill={color}
             strokeWidth={0}
-            opacity={opacity}
             perfectDrawEnabled={false}
             onMouseOver={this.onOverBar(bar)}
             onMouseLeave={this.onOutBar}
--- a/src/components/react_canvas/bar_geometries.tsx
+++ b/src/components/react_canvas/bar_geometries.tsx
@@ -85,7 +85,7 @@ export class BarGeometries extends React.PureComponent<
       }
       return 1;
     } else if (highlightedLegendItem != null) {
-      const isPartOfHighlightedSeries = belongsToDataSeries(bar.value, highlightedLegendItem.value);
+      const isPartOfHighlightedSeries = belongsToDataSeries(bar.geometryId, highlightedLegendItem.value);
 
       if (isPartOfHighlightedSeries) {
         return 1;
--- a/src/components/react_canvas/bar_geometries.tsx
+++ b/src/components/react_canvas/bar_geometries.tsx
@@ -4,8 +4,7 @@ import React from 'react';
 import { Group, Rect } from 'react-konva';
 import { animated, Spring } from 'react-spring/konva';
 import { LegendItem } from '../../lib/series/legend';
-import { BarGeometry, GeometryValue } from '../../lib/series/rendering';
-import { belongsToDataSeries } from '../../lib/series/series_utils';
+import { BarGeometry, GeometryValue, getGeometryStyle } from '../../lib/series/rendering';
 import { ElementClickListener, TooltipData } from '../../state/chart_state';
 
 interface BarGeometriesDataProps {
@@ -75,23 +74,8 @@ export class BarGeometries extends React.PureComponent<
   }
 
   private computeBarOpacity = (bar: BarGeometry, overBar: BarGeometry | undefined): number => {
-    const { highlightedLegendItem } = this.props;
-
-    // There are two elements that might be hovered over that could affect this:
-    // a specific bar element or a legend item; thus, we handle these states as mutually exclusive.
-    if (overBar) {
-      if (overBar !== bar) {
-        return 0.6;
-      }
-      return 1;
-    } else if (highlightedLegendItem != null) {
-      const isPartOfHighlightedSeries = belongsToDataSeries(bar.geometryId, highlightedLegendItem.value);
-
-      if (isPartOfHighlightedSeries) {
-        return 1;
-      }
-
-      return 0.25;
+    if (overBar && overBar !== bar) {
+      return 0.6;
     }
     return 1;
   }
@@ -100,7 +84,12 @@ export class BarGeometries extends React.PureComponent<
     const { overBar } = this.state;
     return bars.map((bar, i) => {
       const { x, y, width, height, color, value } = bar;
+
+      // This sets the opacity if any bars within the chart are hovered over
       const opacity = this.computeBarOpacity(bar, overBar);
+
+      const geometryStyle = getGeometryStyle(bar.geometryId, this.props.highlightedLegendItem);
+
       if (this.props.animated) {
         return (
           <Group key={i}>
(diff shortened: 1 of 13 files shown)


markov00:
> * Currently, this is implemented to affect opacity.  Is this correct or is there another visual change that should be implemented to achieve the "highlighting" effect?

I think it's fine for now adding the opacity. We can add a border later, but I will let the design team decide 

> * I assume that whatever the highlighting effect is, it should be the same for all series visualizations.  So once it is settled what the highlighting effect should be, I will move the rendering logic for this to a shared utility rather than how it is now (with each geometries component handling the logic itself).

yes you are right.

>* I noticed that for both AreaGeometry and LineGeometry, the GeometryValue (which contains the specId and seriesKey, which is used to determine what should be highlighted or not) does not exist on the Geometry, but rather each of the points' PointGeometry.  However, an entire Area or Line would be highlighted as we hover over each legend item, so right now to get around this, I check the first element in `points` on AreaGeometry and LineGeometry (for BarGeometry, the GeometryValue is directly there as `value` on each bar), and use that to determine if an area or line should be highlighted.  This will technically work, but I'm wondering if we might want AreaGeometry and LineGeometry to also directly have `value` like BarGeometry to avoid having to check the first point.  

Yes please, add a GeometryValue also on AreaGeometry and LineGeometry.  it can be a `Partial<GeometryValue>` with only the specId set or we can split the GeometryValue into something like and use `GeometryId` on Area and Line:
```
interface GeometryId {
  specId: SpecId;
}
export type GeometryValue = GeometryId & {

  datum: any;
  seriesKey: any[];
}
```
Or another possibility is to add on the datum an array of GeometryValues? I will reduce that to the minimum required for the highlight for the moment and than we can add more data if we need it

emmacunningham:
@markov00 Thank you for the input!  

See https://github.com/elastic/elastic-charts/pull/31/commits/efd482d6c40af771416f0d334843feb3f014a915 for how the necessary `specId` and `seriesKey` are being added to each type of Geometry so we don't have to access individual points.

And to address the shared styles (as well as the likelihood that design team will offer more feedback on ways to distinguish the highlighted series), see these commits: https://github.com/elastic/elastic-charts/pull/31/commits/ae908ae907dae4b6450266e72b43cd8f24302982 and https://github.com/elastic/elastic-charts/pull/31/commits/647edd6b9dc6882b4b4c399dda058126e6065957

src/components/react_canvas/reactive_chart.tsx
```diff
@@ -315,6 +326,16 @@ class Chart extends React.Component<ReactiveChartProps, ReactiveChartState> {
...
+
+  private getHighlightedLegendItem = () => {
+    const {
+      highlightedLegendItemIndex,
+      legendItems,
+    } = this.props.chartStore!;
+    const highlightedLegendItem = highlightedLegendItemIndex != null ? legendItems[highlightedLegendItemIndex] : null;
+
+    return highlightedLegendItem;
+  }
```
markov00: Can we move this function to the store?

src/lib/series/rendering.ts
```diff
@@ -148,5 +170,34 @@ export function renderArea(
...
+  if (highlightedLegendItem != null) {
+    const isPartOfHighlightedSeries = belongsToDataSeries(geometryId, highlightedLegendItem.value);
+
+    return isPartOfHighlightedSeries ? GEOMETRY_STYLES.highlighted : GEOMETRY_STYLES.unhighlighted;
+  }
+
+  return GEOMETRY_STYLES.default;
+}
+
+const GEOMETRY_STYLES: { [key: string]: GeometryStyle } = {
```
markov00: What it this is part of the chart Theme?

src/lib/series/series_utils.ts
```diff
@@ -0,0 +1,31 @@
...
+    return false;
+  }
+
+  let ret = true;
+
+  a.forEach((aVal: any, idx: number) => {
+    if (aVal !== b[idx]) {
+      ret = false;
+    }
+  });
```
markov00: minor minor nit: an old school for loop can do this work a bit faster and can interrupt is cycle at the first difference. I know it's a unnecessary optimization :)

src/state/chart_state.ts
```diff
@@ -227,6 +228,13 @@ export class ChartStore {
     return this.xScale.type !== ScaleType.Ordinal && Boolean(this.onBrushEndListener);
   }
 
+  updateHighlightedLegendItem(legendItemIndex: number | null) {
+    if (legendItemIndex !== this.highlightedLegendItemIndex) {
+      this.highlightedLegendItemIndex = legendItemIndex;
+      this.computeChart();
```
markov00: You don't have to force the rendering again in that way.
You can make the `highlightedLegendItemIndex` variable an observable variable. This will trigger a force render on any components that use that variable from the store (mobx).
If you move the `getHighlightedLegendItem` into the store, as suggested, you can wrap that function into a mobx `computed` value, that works like an observable value and trigger renrender on every component that call it when some of it's internal value is changed.

src/components/react_canvas/bar_geometries.tsx
```diff
@@ -70,14 +72,24 @@ export class BarGeometries extends React.PureComponent<
     });
     onElementOut();
   }
+
+  private computeBarOpacity = (bar: BarGeometry, overBar: BarGeometry | undefined): number => {
+    if (overBar && overBar !== bar) {
+      return 0.6;
+    }
+    return 1;
```
markov00: why this and the `0.6` values differs from the opacity used in `GEOMETRY_STYLES`?
And is this method still neccesary as we overwrite it by the `geometryStyle` ?

src/components/react_canvas/bar_geometries.tsx
```diff
@@ -70,14 +72,24 @@ export class BarGeometries extends React.PureComponent<
...
     const { overBar } = this.state;
     return bars.map((bar, i) => {
       const { x, y, width, height, color, value } = bar;
-      let opacity = 1;
-      if (overBar && overBar !== bar) {
-        opacity = 0.6;
-      }
+
+      // This sets the opacity if any bars within the chart are hovered over
+      const opacity = this.computeBarOpacity(bar, overBar);
```
markov00: See comment above on line 80

src/components/react_canvas/bar_geometries.tsx
```diff
@@ -96,6 +108,7 @@ export class BarGeometries extends React.PureComponent<
                   onMouseOver={this.onOverBar(bar)}
                   onMouseLeave={this.onOutBar}
                   onClick={this.onElementClick(value)}
+                  {...geometryStyle}
```
markov00: You can also remove the `opacity={opacity}` as it's overwritten by this

src/components/react_canvas/bar_geometries.tsx
```diff
@@ -116,6 +129,7 @@ export class BarGeometries extends React.PureComponent<
             onMouseOver={this.onOverBar(bar)}
             onMouseLeave={this.onOutBar}
             onClick={this.onElementClick(bar.value)}
+            {...geometryStyle}
```
markov00: You can also remove the `opacity={opacity}` as it's overwritten by this

src/components/react_canvas/reactive_chart.tsx
emmacunningham: See eec61b31cdc8c8981958ce6661967945a538bdc4 (this was further modified in this commit once the value became observable 7d12db7382912bcc2de6daef67f1676593d0925a)

src/lib/series/rendering.ts
emmacunningham: See 98d5cece7ec29c829bf89ffced3f2a910a6090b8

src/lib/series/series_utils.ts
emmacunningham: I was actually thinking of this as I was writing this that it might be slightly better to write a for-loop but then went with `forEach` for the time being (I tend to prefer `forEach` when the performance hit isn't too bad because I find it easier to read).  I agree in this case it's also nice to be able to return early as soon as it's false! :)

See 61a24431115939bfdec933fa33d2b3b489e3b46a

src/components/react_canvas/bar_geometries.tsx
emmacunningham: Ah yes, this was leftover from the hover interaction over just a single bar in the geometry and I wasn't yet sure how best to account for both individual bar highlights as well as when a whole series should be highlighted, but now we pass in an additional optional parameter in case a component has individual hover states as well 77e101f83c11280e0d07ec944aeee2ad988eb10e

src/components/react_canvas/bar_geometries.tsx
emmacunningham: See response above and 77e101f83c11280e0d07ec944aeee2ad988eb10e

src/components/react_canvas/bar_geometries.tsx
emmacunningham: Removed in 77e101f83c11280e0d07ec944aeee2ad988eb10e

src/components/react_canvas/bar_geometries.tsx
emmacunningham: Removed in 77e101f83c11280e0d07ec944aeee2ad988eb10e

src/state/chart_state.ts
emmacunningham: See 4e4dafd020197bfd1d27fd13a6096823ae770ef5

emmacunningham:
>     * minor: do we always want to have the highlight on legend enabled? we can maybe add a general strategy to deal with defaults interactions like these ones (on a different PR)

👍 Opened a separate issue to track this for a new PR. (https://github.com/elastic/elastic-charts/issues/56)


>     * major: I've see that the over, out listeners fires few times hovering the same element (see gif) seems that between the two elements, on the padding, the pointer pass through and a mouseout is fired between the dot and the text. I think we should check this and have a consistent mouseover-mouseout event.
>       feb-11-2019 15-38-21

Using mouseenter/leave fixes this.  See 3fe141a5592425ce10962dda43543c6cae5c9e9e

legend_item_hover_event


>       -nit: do you mind adding a hover style on the legend item? an underline for the moment is fine, just to have a feedback on the hovered element.

Yes, I had ended up handling this in the PR for the legend click feature, which you can see here: https://github.com/elastic/elastic-charts/pull/51/files#diff-f6c762a3a7e1c32b9339b3e7cdc3fff1R101.  I cherry-picked the hover style (669da2a912f4d021f59fb260952aa64824c463cc).

emmacunningham:
@markov00 One more thing: I think there will be some big conflicts with the new Theme structure from #44, so it is probably best first to merge that one and then I can resolve those conflicts here (and in the other legend item PR that is open, which is marked WIP until this one is merged in because that was branched off of this one).

PR #37, opened 2019-02-05
fix: add missing min interval compare function

markov00:
Min interval is used on linear scales to compute the minimum interval between two adjacent values on X axis.
It currently works using sorting the unique list of x values.
The sort function was not implemented and the x rendering was based on string comparison of numbers: `0, 1, 10, 11, 2, 3, 4, 5, 6, 7, 8, 9`
I've added a numeric ascending compare function with tests.
I've also fixed the case when we have a single x value on a barchart with linear axis

### DIFF
--- a/src/state/utils.ts
+++ b/src/state/utils.ts
@@ -65,7 +65,7 @@ export function computeSeriesDomains(
   const xDomain = mergeXDomain(specsArray, xValues);
   const yDomain = mergeYDomain(splittedSeries, specsArray);
   const formattedDataSeries = getFormattedDataseries(specsArray, splittedSeries);
-
+  // tslint:disable-next-line:no-console
   // console.log({ formattedDataSeries, xDomain, yDomain });
 
   return {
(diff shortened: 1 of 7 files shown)


src/state/utils.ts
```diff
@@ -65,7 +65,7 @@ export function computeSeriesDomains(
   const xDomain = mergeXDomain(specsArray, xValues);
   const yDomain = mergeYDomain(splittedSeries, specsArray);
   const formattedDataSeries = getFormattedDataseries(specsArray, splittedSeries);
-
+  // tslint:disable-next-line:no-console
```
emmacunningham: minor nit, but if the following line is also commented out, perhaps we could remove both of these as they aren't being used?

markov00: @emmacunningham I'm actually leaving these here just because, when I need to debug a bit more the current code I can easily uncomment the console log lines.
For sure we will remove it before release

emmacunningham: 👍

PR #38, opened 2019-02-05
fix(axis): scale tick labels to fix text truncation on chrome

### LINKED ISSUE #18: Bottom axis labels not partially hidden when using linear scale
If you use a dataset with more than 10 x values, seems that the we don't correctly computed the bottom axis label width.
Can be reproduced using the following to generate a simple dataset:

```
import { DataGenerator } from '../utils/data_generators/data_generator';

const dg = new DataGenerator();
const data = dg.generateSimpleSeries();

...
<AreaSeries
    id={getSpecId('area')}
    xScaleType={ScaleType.Linear}
    yScaleType={ScaleType.Linear}
    xAccessor="x"
    yAccessors={['y']}
    data={data}
/>
```

screenshot 2019-01-31 at 22 13 47

The axis values are correct, but seems to be cutted off:
`0, 5,   1,   1,   2,   2,    3,   3,   4,   4` instead of
`0, 5, 10, 15, 20, 25, 30, 35, 40, 45`

emmacunningham:
fixes #18 

tl;dr the `measureText` function returns very different results from browser to browser and also possibly within the same browser as well, especially with very small font sizes.  We employ a strategy that scales up the font, measures the text size, then divides the result with the scaling factor to get a more accurate measurement of the text width.  Labels should now render fully in Chrome (the labels remain the same in Firefox):

image

### DIFF
--- a/src/lib/axes/canvas_text_bbox_calculator.ts
+++ b/src/lib/axes/canvas_text_bbox_calculator.ts
@@ -6,9 +6,7 @@ export class CanvasTextBBoxCalculator implements BBoxCalculator {
   private offscreenCanvas: HTMLCanvasElement;
   private context: CanvasRenderingContext2D | null;
   private fontScale: number;
-  // TODO specify styles for text
-  // TODO specify how to hide the svg from the current dom view
-  // like moving it a -9999999px
+
   constructor(rootElement?: HTMLElement) {
     this.offscreenCanvas = document.createElement('canvas');
     this.context = this.offscreenCanvas.getContext('2d');
--- a/src/lib/axes/canvas_text_bbox_calculator.ts
+++ b/src/lib/axes/canvas_text_bbox_calculator.ts
@@ -5,6 +5,7 @@ export class CanvasTextBBoxCalculator implements BBoxCalculator {
   private attachedRoot: HTMLElement;
   private offscreenCanvas: HTMLCanvasElement;
   private context: CanvasRenderingContext2D | null;
+  private fontScale: number;
   // TODO specify styles for text
   // TODO specify how to hide the svg from the current dom view
   // like moving it a -9999999px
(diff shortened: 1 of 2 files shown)


src/lib/axes/canvas_text_bbox_calculator.ts
```diff
@@ -13,15 +14,21 @@ export class CanvasTextBBoxCalculator implements BBoxCalculator {
     this.context = this.offscreenCanvas.getContext('2d');
     this.attachedRoot = rootElement || document.documentElement;
     this.attachedRoot.appendChild(this.offscreenCanvas);
+    this.fontScale = 100;
```
markov00: Isn't that too much? with a font size of 10 and a monospace, we will have 1000px height and 1000px * n.chars width. What do you think? I don't know the impact in performance for that.
Maybe it's enough have a the base font size of 100 and than reduce it to the requested font size:
```
this.context.font = `100px ${fontFamily}`;
const measure = this.context.measureText(text);
const width =  100 / measure.width * fontSize;
```

src/lib/axes/canvas_text_bbox_calculator.ts
```diff
@@ -5,6 +5,7 @@ export class CanvasTextBBoxCalculator implements BBoxCalculator {
   private attachedRoot: HTMLElement;
   private offscreenCanvas: HTMLCanvasElement;
   private context: CanvasRenderingContext2D | null;
+  private fontScale: number;
   // TODO specify styles for text
   // TODO specify how to hide the svg from the current dom view
   // like moving it a -9999999px
```
markov00: nit: since we are here, we can remove this old commits
Can you open an issue for that: put the offscreen canvas completely offscreen?

emmacunningham: Ah that's what those comments meant!  Yes, will open an issue and remove these :)

src/lib/axes/canvas_text_bbox_calculator.ts
emmacunningham: Yeah I probably went a little overboard there with the scaling factor :)  Adjusted to always use `100px` and then scale back down based on the actual font size 1f1a05edbaaa01c3fd7d9b606e9e309e862637f9

emmacunningham: Will also play around a bit to test the performance of various scaling factors.

src/lib/axes/canvas_text_bbox_calculator.ts
emmacunningham: Comments removed (1f1a05edbaaa01c3fd7d9b606e9e309e862637f9) and issue created https://github.com/elastic/elastic-charts/issues/43

PR #44, opened 2019-02-06
feat: add dark theme

### LINKED ISSUE #35: Add dark mode switch button
Prepare a story or a global knob to switch EUI darkmode on/off

markov00:
This PR introduce the a better theming on charts. Now the theme has a clear and flatter structure.

A new theme is added: `DARK_THEME`.

A switch to test the DARK theme is added to storybook

feb-06-2019 11-16-43

fix #35 

**BREAKING CHANGES:**

The `Theme` interface was refactored completely with a flatter structure.

### DIFF
--- a/src/components/react_canvas/area_geometries.tsx
+++ b/src/components/react_canvas/area_geometries.tsx
@@ -4,8 +4,13 @@ import React from 'react';
 import { Circle, Group, Path } from 'react-konva';
 import { animated, Spring } from 'react-spring/konva';
 import { LegendItem } from '../../lib/series/legend';
-import { AreaGeometry, GeometryValue, getGeometryStyle, PointGeometry } from '../../lib/series/rendering';
-import { AreaSeriesStyle } from '../../lib/themes/theme';
+import {
+  AreaGeometry,
+  GeometryValue,
+  getGeometryStyle,
+  PointGeometry,
+} from '../../lib/series/rendering';
+import { AreaSeriesStyle, SharedGeometryStyle } from '../../lib/themes/theme';
 import { ElementClickListener, TooltipData } from '../../state/chart_state';
 
 interface AreaGeometriesDataProps {
@@ -13,6 +18,7 @@ interface AreaGeometriesDataProps {
   areas: AreaGeometry[];
   num?: number;
   style: AreaSeriesStyle;
+  sharedStyle: SharedGeometryStyle;
   onElementClick?: ElementClickListener;
   onElementOver: ((tooltip: TooltipData) => void) & IAction;
   onElementOut: (() => void) & IAction;
@@ -24,7 +30,7 @@ interface AreaGeometriesDataState {
 export class AreaGeometries extends React.PureComponent<
   AreaGeometriesDataProps,
   AreaGeometriesDataState
-  > {
+> {
   static defaultProps: Partial<AreaGeometriesDataProps> = {
     animated: false,
     num: 1,
@@ -86,31 +92,31 @@ export class AreaGeometries extends React.PureComponent<
       [] as JSX.Element[],
     );
   }
-  private renderPoints = (points: PointGeometry[], i: number): JSX.Element[] => {
-    const { style } = this.props;
+  private renderPoints = (areaPoints: PointGeometry[], i: number): JSX.Element[] => {
+    const { radius, stroke, strokeWidth } = this.props.style.point;
     const { overPoint } = this.state;
 
-    return points.map((point, index) => {
-      const { x, y, color, value, transform } = point;
+    return areaPoints.map((areaPoint, index) => {
+      const { x, y, color, value, transform } = areaPoint;
       return (
         <Group key={`point-${i}-${index}`}>
           <Circle
             x={transform.x + x}
             y={y}
-            radius={style.dataPointsRadius * 2.5}
+            radius={radius * 2.5}
             onClick={this.onElementClick(value)}
-            onMouseOver={this.onOverPoint(point)}
+            onMouseOver={this.onOverPoint(areaPoint)}
             onMouseLeave={this.onOutPoint}
             fill={'gray'}
-            opacity={overPoint === point ? 0.3 : 0}
+            opacity={overPoint === areaPoint ? 0.3 : 0}
           />
           <Circle
             x={transform.x + x}
             y={y}
-            radius={style.dataPointsRadius}
+            radius={radius}
             strokeWidth={0}
             fill={color}
-            opacity={overPoint === point ? 0.5 : 0}
+            opacity={overPoint === areaPoint ? 0.5 : 0}
             strokeHitEnabled={false}
             listening={false}
             perfectDrawEnabled={true}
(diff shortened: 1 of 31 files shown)


src/lib/themes/theme.ts
```diff
@@ -17,10 +18,15 @@ export interface AxisConfig {
...
   tickFontFamily: string;
   tickFontStyle: string;
+  tickLabelColor: string;
+  tickStrokeWidth: number;
   titleFontSize: number;
   titleFontFamily: string;
   titleFontStyle: string;
   titlePadding: number;
+  titleColor: string;
+  lineStrokeColor: string;
```
emmacunningham: What part of the axis does this affect?

src/lib/themes/theme.ts
```diff
@@ -17,10 +18,15 @@ export interface AxisConfig {
...
+  tickStrokeColor: string;
```
emmacunningham: Not necessarily advocating for this change here (maybe this can be a separate enhancement PR), but wondering if we might want to think about breaking up the AxisConfig into the configs for the title and labels.  Maybe something like:

```
interface AxisConfig {
  tickLabel: TickLabelConfig;
  axisTitle: TitleLabelConfig;
}
```

and since there is a significant amount of overlap between the configs for tick labels and titles, these configs could be and intersection with some base interface for the shared config properties.  We may even be able to organize the configs in such a way that for passing props into the rendered components, we could pass the whole config in as opposed to how we have it right now, needing to first access the specific properties from the config and then create a new props object.

src/lib/themes/theme.ts
markov00: is the axis line, please can you suggest me a better naming?

src/lib/themes/theme.ts
markov00: yes you are right. I was thinking the same when writing this code.

src/lib/themes/theme.ts
```diff
@@ -91,133 +97,129 @@ export const DEFAULT_GRID_LINE_CONFIG: GridLineConfig = {
...
+      tickFontStyle: 'normal',
+      tickLabelColor: darkMode ? 'white' : 'gray',
+      tickStrokeColor: darkMode ? 'white' : 'gray',
+      tickStrokeWidth: 1,
+      titleFontSize: 12,
+      titleFontStyle: 'bold',
+      titleFontFamily: `'Open Sans', Helvetica, Arial, sans-serif`,
+      titlePadding: 5,
+      titleColor: darkMode ? 'white' : 'gray',
+      lineStrokeColor: darkMode ? 'white' : 'gray',
```
snide: I'm just randomly jumping around here, but if possible I would try to build your theming variables into a single document, similar to how EUI handles this. So rather than checking for `darkMode` on each prop, you would have a single file with ALL tokenization for light theme, and then a copy document for dark mode. Then, just import the variables and they use the same nameschema in your document.

That way when you need a new theme you copy the file and replace the vars. You don't want to hand touch everything and put boolean checks everywhere.

Essentially, don't assume dark and light mode but _dozens_ of themes.

src/lib/themes/theme.ts
emmacunningham: ah, yes that makes sense.  since there could be a few different lines associated with the axis (axis line, tick line, grid line, etc.), perhaps this could be `axisLineStrokeColor`?

though also, i think this becomes more clear once we have clearer separation of the different parts of axis as discussed in [this comment](https://github.com/elastic/elastic-charts/pull/44#discussion_r254378877), so i'm not sure this needs to be changed (i was just asking because i saw it defined but had seen `lineStrokeColor` on other configs and wanted to make sure i understood what the difference between all of them is).

src/lib/themes/theme.ts
markov00: You are absolutely right. I will review this to work in that way

snide: Yep. And just so you know we're starting to run intot he same problem in Kibana too. So we're gonna try and think up something there as well. https://github.com/elastic/kibana/pull/30341#discussion_r254539068

cc @chandlerprall

src/lib/themes/theme.ts
markov00: resolved on https://github.com/elastic/elastic-charts/pull/44/commits/6fc53b7572c254db01452b1b6faa70f4e51a2dd4

src/lib/themes/theme.ts
markov00: @snide @chandlerprall I've refactored the theme here: https://github.com/elastic/elastic-charts/pull/44/commits/5b7445f5678b1920a827ff3f11cac043dcd69ba4
I've added a more flat structure on the theme object, splitted into various files each theme. Now we are no more linked to darkMode, now you can:
- add a new Theme and export it
- extend an existing one
- use a completely new custom theme

markov00: We have just one more place to handle the dark vs light mode: https://github.com/elastic/elastic-charts/commit/5b7445f5678b1920a827ff3f11cac043dcd69ba4#diff-9650d2817e08427313c3e1f5d20218e6R49
I will update it to be darkmode agnostic

markov00: I've removed the darkMode props from settings here https://github.com/elastic/elastic-charts/pull/44/commits/ca797f760a29ac53b043812e16ae23a88d0f9d48. Now we are theme agnostic

src/components/react_canvas/area_geometries.tsx
```diff
@@ -116,12 +116,12 @@ export class AreaGeometries extends React.PureComponent<
...
             y={y}
-            radius={style.dataPointsRadius}
+            radius={style.point.radius}
             onMouseOver={this.onOverPoint(point)}
             onMouseLeave={this.onOutPoint}
             fill={'transparent'}
-            stroke={style.dataPointsStroke}
-            strokeWidth={style.dataPointsStrokeWidth}
+            stroke={style.point.stroke}
+            strokeWidth={style.point.strokeWidth}
```
emmacunningham: For this and the other above Circle, could we just spread `style.point`?

src/components/react_canvas/line_geometries.tsx
```diff
@@ -114,12 +114,12 @@ export class LineGeometries extends React.PureComponent<
...
             y={y}
-            radius={style.dataPointsRadius}
+            radius={style.point.radius}
             onMouseOver={this.onOverPoint(point)}
             onMouseLeave={this.onOutPoint}
             fill={'transparent'}
-            stroke={style.dataPointsStroke}
-            strokeWidth={style.dataPointsStrokeWidth}
+            stroke={style.point.stroke}
+            strokeWidth={style.point.strokeWidth}
```
emmacunningham: Similar to other comment, could these be spread?

src/lib/themes/theme.ts
```diff
@@ -1,26 +1,38 @@
...
-  tickFontFamily: string;
-  tickFontStyle: string;
-  titleFontSize: number;
-  titleFontFamily: string;
-  titleFontStyle: string;
-  titlePadding: number;
+  axisTitleStyle: TextStyle;
+  axisLineStyle: StrokeStyle;
+  tickLabelStyle: TextStyle;
+  tickLineStyle: StrokeStyle;
```
emmacunningham: 👏

PR #54, opened 2019-02-13
style(legend): align k7 design

### LINKED ISSUE #34: Align Legend design to K7 Design
With this PR: https://github.com/elastic/kibana/pull/29896 the icon to collapse the legend is now always on the same place (bottom left) and disappear when moving the mouse focus out of the chart panel.
We need to align the current implementation with that style

markov00:
This PR align the legend design to the K7 design with the toggle button is always on the bottom left of the chart.

fix #34

### DIFF
--- a/src/components/legend.tsx
+++ b/src/components/legend.tsx
@@ -40,8 +40,10 @@ class LegendComponent extends React.Component<ReactiveChartProps> {
     const legendClasses = classNames(
       'elasticChartsLegend',
       `elasticChartsLegend--${legendPosition}`,
-      legendCollapsed.get() && 'elasticChartsLegend--collapsed',
-      debug && 'elasticChartsLegend--debug',
+      {
+        'elasticChartsLegend--collapsed': legendCollapsed.get(),
+        'elasticChartsLegend--debug': debug,
+      },
     );
     let paddingStyle;
     if (isVertical(legendPosition)) {
--- a/src/components/legend.tsx
+++ b/src/components/legend.tsx
@@ -1,37 +1,13 @@
-import {
-  EuiButtonIcon,
-  EuiFlexGroup,
-  EuiFlexItem,
-  EuiIcon,
-  EuiText,
-} from '@elastic/eui';
+import { EuiFlexGroup, EuiFlexItem, EuiIcon, EuiText } from '@elastic/eui';
 import classNames from 'classnames';
 import { inject, observer } from 'mobx-react';
 import React from 'react';
+import { isVertical } from '../lib/axes/axis_utils';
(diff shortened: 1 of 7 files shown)


src/components/legend.tsx
```diff
@@ -50,44 +25,48 @@ class LegendComponent extends React.Component<ReactiveChartProps> {
...
-      debug && 'euiChartLegend--debug',
-    );
-
-    const legendCollapser = classNames(
-      'euiChartLegendCollapser',
-      `euiChartLegendCollapser--${legendPosition}`,
+      'elasticChartsLegend',
+      `elasticChartsLegend--${legendPosition}`,
+      legendCollapsed.get() && 'elasticChartsLegend--collapsed',
+      debug && 'elasticChartsLegend--debug',
```
emmacunningham: Could use object for conditional classes?

```
{
'elasticChartsLegend--collapsed': legendCollapsed.get(),
'elasticChartsLegend--debug': debug,
}
```

src/components/tooltips.tsx
```diff
@@ -58,7 +58,10 @@ class TooltipsComponent extends React.Component<ReactiveChartProps> {
   ) => {
     return (
       <div
-        className={classNames('euiChartTooltip', showTooltip ? null : 'euiChartTooltip--hidden')}
+        className={classNames(
+          'elasticChartsTooltip',
+          showTooltip ? null : 'elasticChartsTooltip--hidden',
```
emmacunningham: similar to the suggestion in [this comment](https://github.com/elastic/elastic-charts/pull/54/files#r256561655),  could use object for the conditional class?

emmacunningham:
Not sure if it is intentional, but it looks like the top position of the legend toggle changes slightly on hover?  This is most noticeable when the toggle has focus w/ the gray background as there appears to be whitespace between the toggle and the edge of the container:

legend_tooltip_hover_focus

markov00:
> Not sure if it is intentional, but it looks like the top position of the legend toggle changes slightly on hover? This is most noticeable when the toggle has focus w/ the gray background as there appears to be whitespace between the toggle and the edge of the container:
> 
> legend_tooltip_hover_focus

@emmacunningham this is the default behaviour of `EuIButtonIcon`, @cchaos what do you think? do I need to remove the animation? on dashboard we don't have this because we are not using EUIButton directly.

cchaos:
I dont think it's detrimental to the user, so I'd opt for leaving it as is.

emmacunningham:
👍 sounds good!  agree it's not detrimental to the user.  was just curious as i'm new to these things :)

PR #74, opened 2019-02-20
test(chart_state): fill in missing coverage

emmacunningham:
## Summary

This PR adds test coverage for the chart_state functions, bringing test coverage up to nearly 100% (there is 100% coverage of functions, but still a few branches and statements not covered).  The remaining missing coverage items reveal opportunities for refactoring the `computeChart` function to be a more pure function, so they are not addressed here but are left with the hope that our idea to refactor this will also lend itself to making these remaining states easier to test (see https://github.com/elastic/elastic-charts/issues/75).

### Checklist

### DIFF
--- a/src/state/chart_state.ts
+++ b/src/state/chart_state.ts
@@ -201,7 +201,7 @@ export class ChartStore {
     this.highlightedLegendItemIndex.set(legendItemIndex);
     if (this.onLegendItemOverListener) {
       const currentLegendItem = this.highlightedLegendItem.get();
-      const listenerData = currentLegendItem ? currentLegendItem.value : null;
+      const listenerData = currentLegendItem!.value;
       this.onLegendItemOverListener(listenerData);
     }
   });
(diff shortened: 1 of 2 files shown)


src/state/chart_state.ts
```diff
@@ -201,7 +201,7 @@ export class ChartStore {
     this.highlightedLegendItemIndex.set(legendItemIndex);
     if (this.onLegendItemOverListener) {
       const currentLegendItem = this.highlightedLegendItem.get();
-      const listenerData = currentLegendItem ? currentLegendItem.value : null;
+      const listenerData = currentLegendItem!.value;
```
markov00: So this means that `currentLegendItem` in this situation is never `null`? I don't think it's legit here to say so.
What happen if we call `onLegendItemOver` with an index that is not in the range of `legendItems`?
On our current code that will never happen, but one more check is always a nice to have.

emmacunningham: Added check for valid index number and sets it to null if out of range; updated the test to check for these cases.  429d51a

PR #95, opened 2019-03-08
feat(series): set custom series colors through spec prop

### LINKED ISSUE #39: Add external custom color from spec props
Specify a color for each data series on the chart specs

emmacunningham:
## Summary

close #39 

This PR adds the ability to specify a custom series colors through a prop on the series spec.

Each series can now receive an optional prop `customSeriesColor: Map<DataSeriesColorValues, string>`, where the key is used to identify the series and the string is a representation of the color.  Under the `Stylings` section of Storyboard, there is an example of this in action.

custom_series_colors

The color picker in the legend is also still usable to customize a series color; on color picker update, the `customSeriesColor` prop for the corresponding series is updated.

### Checklist

### DIFF
--- a/src/lib/series/series.test.ts
+++ b/src/lib/series/series.test.ts
@@ -265,7 +265,7 @@ describe('Series', () => {
 
     const emptyCustomColors = new Map();
 
-    const defaultColorMap = getSeriesColorMap(seriesColors, chartColors, emptyCustomColors, specs);
+    const defaultColorMap = getSeriesColorMap(seriesColors, chartColors, emptyCustomColors);
     const expectedDefaultColorMap = new Map();
     expectedDefaultColorMap.set('spec1', 'elastic_charts_c1');
     expect(defaultColorMap).toEqual(expectedDefaultColorMap);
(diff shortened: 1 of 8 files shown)


emmacunningham:
@markov00 for serializing the `Map<DataSeriesColorsValues, string>` prop: there are a few options I've thought about.  The current implementation has the advantage that it makes it easier for the downstream user to define than need to know more specifics about the internals of the library.  But if we want to have a JSON schema for the spec, then the user could define an Array of:

```
{
  dataSeriesColorValues: DataSeriesColorValues;
  color: string
}
```

This still allows the user to define in terms of DataSeriesColorValues (instead of needing to know more about the chart internals and forming the string series key themselves) while providing a bit of type safety to how they are defining these so that we at least get both a specId and colorValues to uniquely identify the series.  We would need to then change the way that updating the props works; because currently it is a map, we can just set the customSeriesColors map using the key when there is an update.  Using this new data structure, we'd need to iterate through the array for the spec and then find if a config with a matching DataSeriesColorValues exists. 

Another option would be to have this prop be an object (one advantage is that the updating logic would remain about the same, unlike the option above):

```
{
  [seriesKey: string]: string;
}
```

where the seriesKey property is the stringified form of the DataSeriesColorValues and the value is string representing color.  The drawback to this is that the user would need to be able to form the seriesKey, though we do have a function that they could call on DataSeriesColorValues to generate this string.

PR #107, opened 2019-03-19
test(utils/domain): add tests for domain computations

emmacunningham:
## Summary

Fills in missing coverage for this file, bringing test coverage up from 52.63% to 100% 😄 

### Checklist

### DIFF
--- a/src/lib/utils/domain.test.ts
+++ b/src/lib/utils/domain.test.ts
@@ -1,5 +1,5 @@
 import { AccessorFn } from './accessor';
-import { computeOrdinalDataDomain, computeContinuousDataDomain, computeStackedContinuousDomain } from './domain';
+import { computeContinuousDataDomain, computeOrdinalDataDomain, computeStackedContinuousDomain } from './domain';
 
 describe('utils/domain', () => {
   test('should compute ordinal data domain: sort & remove nulls', () => {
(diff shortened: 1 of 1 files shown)


src/lib/utils/domain.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    const yAccessor: AccessorFn = (datum: any) => datum.x;
+    const xAccessor: AccessorFn = (datum: any) => datum.y;
+    const scaleToExtent = true;
+
+    const stackedDataDomain = computeStackedContinuousDomain(
+      data,
+      xAccessor,
+      yAccessor,
+      scaleToExtent,
+    );
```
markov00: I was a bit shocked when I see that the domain was: [6, 20].... than I realized the the xAccessor is the y value and the yAccessor is the x value.
To avoid this ambiguity, could you please invert the xy accessors to their relative x and y keys and update also the data, like that:
```
    const data = [{ x: 'a', y: 12 }, { x: 'b', y: 6  }, { x: 'a', y: 8 }];
    const yAccessor: AccessorFn = (datum: any) => datum.y;
    const xAccessor: AccessorFn = (datum: any) => datum.x;
```

it's a bit more easier to read and think of

src/lib/utils/domain.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    const yAccessor: AccessorFn = (datum: any) => datum.x;
+    const xAccessor: AccessorFn = (datum: any) => datum.y;
+
+    const stackedDataDomain = computeStackedContinuousDomain(
+      data,
+      xAccessor,
+      yAccessor,
+    );
+
+    const expectedStackedDomain = [0, 20];
```
markov00: The same as the previous comment

src/lib/utils/domain.test.ts
emmacunningham: 👍 thank you for catching this!  have updated with this change in commit 80231c01e8384ce6a734cdfa462e6af8556518b9

src/lib/utils/domain.test.ts
emmacunningham: see https://github.com/elastic/elastic-charts/pull/107/commits/80231c01e8384ce6a734cdfa462e6af8556518b9

PR #114, opened 2019-03-22
fix(rendering): fix rendering values <= 0 on log scale

### LINKED ISSUE #63: Display orphan data points on line/area series
Some series may have orphan data points, points that are between two null datums like:
```
[
   ...,
   {x: 1, y: null},
   {x: 2, y: 10},
   {x: 3, y: null},
   ...,
]
```
When a series has orphan data points like the example above, we need to find a way to display the single point value.

Fitting functions (#55) can fill the gaps with specific values (like adding zeros or an average value), but the developer or user can decide to keep `null` values.

If the style of the line/area already shows single circles as data points we don't need to add anything to that rendering.

If the style configuration hides the data points we need to force displaying these points when orphans.

We also have to configure the line/area path generation `defined` function to correctly specify when a line/area is not defined: https://github.com/d3/d3-shape#line_defined https://github.com/d3/d3-shape#area_defined

**Kibana Cross Issues**

**Checklist**

### LINKED ISSUE #112: LineSeries throw warnings when using ScaleType.Log
**Describe the bug**
While setting a prop `yScaleType={ScaleType.Log}` to `LineSeries` component, get console warnings:

image

Here is an ES dataset: [data.txt](https://github.com/elastic/elastic-charts/files/2991696/data.txt) 

**To Reproduce**
The code looks like:
```
this.props.series.map(series => (
            <LineSeries
              key={series.id}
              id={getSpecId(series.label)}
              xScaleType={ScaleType.Time}
              yScaleType={ScaleType.Log}
              xAccessor={0}
              yAccessors={[1]}
              data={series.data}
              yScaleToDataExtent={false}
            />
          ))
```
**Expected behavior**
There should not be any warning

 - OS: Windows 10
 - Browser Chrome  v 72.0.3626.121
 - Elastic Charts: 3.2.0

markov00:
## Summary

Instead of throwing errors or warnings, if the user is using a log scale with a dataset with 0
values, we will hide them on the rendering of line and area charts. 

Screenshot 2019-03-22 at 19 38 28

This fix also the defined area of a line/area chart: if the y value is null, the area/line is depicted with an empty space between the previous non null value and the following one.

fix #112, fix #63

### Checklist


src/lib/series/rendering.ts
```diff
@@ -263,10 +283,17 @@ export function renderArea(
   areaGeometry: AreaGeometry;
   indexedGeometries: Map<any, IndexedGeometry[]>;
 } {
+  const isLogScale = yScale.type === ScaleType.Log;
```
emmacunningham: Since this logic is repeated in a few places, could create a helper function `isLogScale(type: ScaleType)` to ensure consistency.

markov00: resolved in https://github.com/elastic/elastic-charts/pull/114/commits/790d5f7f9e8143789e682d5a7d839b0f3ec6ed74
currently passing a Scale object instead of a type as it's done for the `isContinuousScale` or `isOrdinalScale`

PR #115, opened 2019-03-22
fix(chart): fix duplicated keys for chart elements

markov00:
## Summary

This PR will fix errors caused by duplicated react element keys in line and area chart.
It also move 3 global properties used on Konva elements to a global object.

### Checklist

### DIFF
--- a/src/components/react_canvas/globals.ts
+++ b/src/components/react_canvas/globals.ts
@@ -1,0 +1,7 @@
+import { ShapeConfig } from 'konva';
+
+export const GlobalKonvaElementProps: ShapeConfig = {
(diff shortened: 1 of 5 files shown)


src/components/react_canvas/globals.ts
```diff
@@ -0,0 +1,7 @@
+import { ShapeConfig } from 'konva';
+
+export const GlobalKonvaElementProps: ShapeConfig = {
+  strokeHitEnabled: false,
+  perfectDrawEnabled: false,
+  listening: false,
+};
```
emmacunningham: 👏

Issue #111, opened 2019-03-21
Axis domain optional min/max support


sulemanof:
**Is your feature request related to a problem? Please describe.**
There is no way to pass a domain prop to the `Axis` component as an empty object or an object with partial values (only `max` or only `min` value). 

**Describe the solution you'd like**
I think it should support the next object structures:
- `const domain = { min: undefined || null, max: undefined || null}` and pretty the same `const domain = {}` - in that case it should behave as without the domain prop;
- `const domain = { min: 10 }` or `const domain = { max: 10 }` - support of partial domain values if a user wants to use only one of them

**Additional context**
In that case I would never think about if `min` or `max` is defined and it would simplify the development
![image](https://user-images.githubusercontent.com/31325372/54743210-9109cf80-4bd4-11e9-9f8a-a96c2ab87b7e.png)

- I suppose the charts should show a message that user pass invalid max-min range instead of throwing an error which crash an app
![image](https://user-images.githubusercontent.com/31325372/54743760-1e99ef00-4bd6-11e9-8550-65489b89d5dd.png)
![image](https://user-images.githubusercontent.com/31325372/54743777-31acbf00-4bd6-11e9-9164-59e360640c0f.png)

**Kibana Cross Issues**
n/a

**Checklist**

- [x] this request is checked against already exist requests
- ~[ ] every related Kibana issue is listed under `Kibana Cross Issues` list~
- ~[ ] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present~

markov00:
Making min and max optionals is a nice enhancement. @emmacunningham could you take a look into this?
@sulemanof For the error throw related question: we can have two way to handle this:
- let's the developer catch that error and notify the user with the appropriate message
- display a message instead of the chart with that error.

Can happen that the final user is not the one that have changed the domain min/max on the chart: think about the case where you are integrating the library and you always precompute the domain min and max to adjust them a bit, if you have an error on your code, we can display a default message but the user doesn't know how to deal with it, we should than add a way to integrate custom/translated messages for each error message available.

I think we need to discuss this a bit more to get a correct understanding on how to deal with errors

markov00:
@sulemanof On the other way, we don't want to support null, undefined values or empty object checks. The typescript definition is rigid on the type of data to be passed and it already checks if the value is valid or not at compile time. 

We can change the `DomainRange` interface to something like this
```
// copied from stackoverflow :)
[...]

PR #116, opened 2019-03-22
feat: allow partial custom domain

emmacunningham:
## Summary

re #111 

This PR adds the enhancement to defining partially bounded custom domains.  Where `DomainRange` was previously typed as `{ min: number, max: number }`, it now is typed as such:

```
export type DomainRange = LowerBoundedDomain | UpperBoundedDomain | CompleteBoundedDomain;

export interface LowerBoundedDomain {
  min: number;
}

export interface UpperBoundedDomain {
  max: number;
}

export interface CompleteBoundedDomain {
  min: number;
  max: number;
}
```

This allows us to flexibly extend the interfaces as needed and also enforce that at least one of `max` and `min` are defined (while providing the downstream user with feedback on the type of their config to block `{}` as a valid prop).  [The other benefit to this type is that internally, we can refer to the individual types of bounded domains and use them in a typeguard; see the updated `mergeDomainsByGroupId` implementation.  Additionally, we can use the typeguards when merging X and Y domains.]

For a partial custom domain, we also check its validity (for example, a LowerBoundedDomain should not have a min greater than the max of the computed domain).  Currently this throws, but per #117, once there is an agreement on error handling, this may change.

### Checklist

### DIFF
--- a/src/lib/axes/axis_utils.test.ts
+++ b/src/lib/axes/axis_utils.test.ts
@@ -1034,6 +1034,134 @@ describe('Axis computational utils', () => {
     );
   });
 
+  test('should merge axis domains by group id: partial upper bounded prevDomain with complete domain', () => {
+    const groupId = getGroupId('group_1');
+    const domainRange1 = {
+      max: 9,
+    };
+
+    const domainRange2 = {
+      min: 0,
+      max: 7,
+    };
+
+    verticalAxisSpec.domain = domainRange1;
+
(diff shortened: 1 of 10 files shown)


emmacunningham:
@markov00 using the typeguards instead of casting as `Partial<CompleteBoundedDomain>`, see fbbe316058c9020f69a16b1c6b5b53e933a53986

PR #118, opened 2019-03-24
fix(chart_state): maintain series visibility state on props update

### LINKED ISSUE #113: Series visibility state is lost when a chart props changed
**Describe the bug**

The visibility state of a series is lost when sending updated props to a chart

image

It is also reproducible in the [storybook](https://elastic.github.io/elastic-charts/?selectedKind=Stylings&selectedStory=margins%20and%20paddings&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs)

series_visibility_issue

**To Reproduce**
1. Press the eye button on a series -> the series should become invisible
2. Change any prop in `KNOBS` tab
3. the series become visible again

**Expected behavior**
The visibility of the series should stay in that state until press the button again or the chart  unmount.

**Screenshots**
It becomes useless in kibana TSVB because it updates on every click:

series_visibility_issue_in_tsvb

 - OS: Windows 10
 - Browser Chrome 73
 - Elastic Charts: 3.2.0

**Errors in browser console**
-

**Kibana Cross Issues**
Add any [Kibana](https://github.com/elastic/kibana) related issues here.

**Checklist**

emmacunningham:
## Summary

fix #113 

This PR fixes the issue where the visibility of the data series would reset on every state change.

Previously, `resetSelectedDataSeries` was being called in `componentDidUpdate` on `SpecsParser`.  Instead of keeping track of `selectedDataSeries` (and assuming `isVisible: false` as the default), we now assume visibility as the default and keep track of `deselectedDataSeries`.  This has the advantage that we do not need to reset any kind of visibility and visibility state can be better preserved as specs are being updated & added.

- splitAccessors change
visibility_split_change

- yAccessors change
visibility_y_change

- changes in split accessor values in data
visibility_data_change

- removing a series that is the only selected visible series (this is the same as the functionality in current kibana)
visibility_remove_data

### Checklist

### DIFF
--- a/src/state/utils.ts
+++ b/src/state/utils.ts
@@ -26,7 +26,6 @@ import { isEqualSeriesKey } from '../lib/series/series_utils';
 import {
   AreaSeriesSpec,
   AxisSpec,
-  BarSeriesSpec,
   BasicSeriesSpec,
   DomainRange,
   LineSeriesSpec,
@@ -51,7 +50,7 @@ export interface BrushExtent {
   maxY: number;
 }
 
-export function findSelectedDataSeries(
+export function findDataSeriesByColorValues(
   series: DataSeriesColorsValues[] | null,
   value: DataSeriesColorsValues,
 ): number {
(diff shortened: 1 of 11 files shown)


src/state/utils.ts
```diff
@@ -63,6 +64,30 @@ export function findSelectedDataSeries(
...
+  prevSeriesColors: Map<string, DataSeriesColorsValues>,
+  prevSeriesSpecs: Map<SpecId, BasicSeriesSpec>,
+  prevSelectedDataSeries: DataSeriesColorsValues[] | null,
+  nextSeriesSpec: BasicSeriesSpec | LineSeriesSpec | AreaSeriesSpec | BarSeriesSpec,
+
+): boolean {
+  const nextSeriesSpecs = new Map([...prevSeriesSpecs]);
+  nextSeriesSpecs.set(nextSeriesSpec.id, nextSeriesSpec);
+
+  const nextSeriesColors = getSplittedSeries(nextSeriesSpecs, prevSelectedDataSeries).seriesColors;
```
markov00: using `getSplittedSeries` is quite expensive, specially on charts with an high number of elements/series.
It's an  heavy function because it internally use `splitSeries` that cycle all the datums for each series and all the x values after that.
Can't we find a better way to detect that change? checking only if we are adding a new spec or if it's already there and differ from the existing one. 
In case you can create a test case to to check if that decrease or not the performance when having 200 data points with 20/30 series (that doesn't mean to have a readable chart but we can stress test it a bit)

emmacunningham: Instead of storing the selected (visible) data series, we store the deselected data series.  This removes the need to reset in response to certain conditions because by default all series will be visible and we only keep track of what is deselected (not visible).  This also better preserves visibility state when making changes to the spec.

stories/interactions.tsx
```diff
@@ -177,9 +177,57 @@ storiesOf('Interactions', module)
...
+      min: number('yDomain min', 0, {}, notSpecChange),
+      max: number('yDomain max', 10, {}, notSpecChange),
+    };
+
+    const scaleTypeOptions: { [key: string]: ScaleType.Linear | ScaleType.Log } = {
+      linear: ScaleType.Linear,
+      log: ScaleType.Log,
+    };
+
+    const xScaleType = select('xScaleType', scaleTypeOptions, ScaleType.Linear, specChange);
```
markov00: I think we have a bug and we allow also log scale for x scale, that's not actually possible.
could you please correct this on the story, we can open a PR for fixing this on the props.

emmacunningham: 👍 For xDomain it should only be Time, Linear, and Ordinal, right?

markov00: yep

emmacunningham: Removed the option to change xScaleType on this story (it doesn't really make sense to switch it for an ordinal or time scale type with this dataset and the point that changing scale types doesn't change the visibility state is already demonstrated by yScaleType changes) in this commit https://github.com/elastic/elastic-charts/pull/118/commits/fe7b307a01224210aee7e7011cf68ce14dc12d29

Created issue #122 to address the bug.

PR #126, opened 2019-03-27
feat(annotations): render line annotations via LineAnnotation spec

### LINKED ISSUE #104: Add annotation line, tooltip, and marker
Add ability to add annotation to a chart with a line (vertical or horizontal), tooltip, and marker.

User should be able to define an x or y value for the line, the marker, and the text to appear within the tooltip.

We can also expose a helper utility to the user to be able to generate the annotation configs for something like "get all the X positions for these data points matching these values".

This is one of three annotation-related tasks (see #81).  

**Additional context**
Current feature in Kibana:
kibana_annotation

**Kibana Cross Issues**

**Checklist**

emmacunningham:
## Summary

close #104 

This PR introduces a LineAnnotation spec so the user can specify what annotations on a chart they want.

Since this is the first of the Annotation features, this also introduces a bit of scaffolding for the annotations overall.  This implementation allows a user to define a set of annotations by defining one of three specs (LineAnnotation, RectangleAnnotation, and TextAnnotation).

Within a LineAnnotation spec a user may define:

```
  /** The id of the annotation */
  annotationId: AnnotationId;
  /** Annotation type: line, rectangle, text */
  /** The ID of the axis group, generated via getGroupId method */
  groupId: GroupId; // defaults to __global__; needed for yDomain position
  /** Annotation domain type: XDomain or YDomain */
  domainType: AnnotationDomainType;
  /** Data values defined with value, details, and header */
  dataValues: AnnotationDatum[];
  /** Custom line styles */
  lineStyle?: Partial<AnnotationLineStyle>;
  /** Custom marker */
  marker?: JSX.Element;
```

`dataValues` are defined in terms of data relevant to a scale within the chart. If the `domainType` is xDomain, then the annotation points will be scaled relative to the xDomain scale.  If the `domainType` is yDomain, then the annotation points will be scaled relative to a yDomain scale; we get the relevant scale by lookup up against `this.yScales` with the `groupId` defined on the spec.

x_continuous_annotation
ordinal_annotation
time_series_annotation

### Checklist


src/state/annotation_utils.ts
```diff
@@ -0,0 +1,521 @@
...
+export function computeXDomainLineAnnotationDimensions(
+  dataValues: AnnotationDatum[],
+  xScale: Scale,
+  chartRotation: Rotation,
+  lineOverflow: number,
+  axisPosition: Position,
+  chartDimensions: Dimensions,
+  lineColor: string,
+  marker?: JSX.Element,
+) {
```
markov00: add return type

src/state/annotation_utils.ts
```diff
@@ -0,0 +1,521 @@
...
+): AnnotationLineProps[] | null {
+  const { domainType, dataValues, marker } = annotationSpec;
+
+  // TODO : make line overflow configurable via prop
+  const lineOverflow = DEFAULT_LINE_OVERFLOW;
+
+  const lineColor = getAnnotationLineColor(annotationSpec.lineStyle);
+
+  switch (domainType) {
+    case AnnotationDomainType.XDomain: {
```
markov00: it's the first time I see someone using the braces on a `case` :)

src/state/annotation_utils.ts
```diff
@@ -0,0 +1,521 @@
...
+): Map<AnnotationId, AnnotationDimensions> {
+  const annotationDimensions = new Map<AnnotationId, AnnotationDimensions>();
+
+  annotations.forEach((annotationSpec: AnnotationSpec, annotationId: AnnotationId) => {
+    switch (annotationSpec.annotationType) {
+      case AnnotationType.Line:
+        const { groupId, domainType } = annotationSpec;
+        const annotationAxis = getAnnotationAxis(axesSpecs, groupId, domainType);
+
+        const axisPosition = annotationAxis || Position.Bottom;
```
markov00: shouldn't we need to skip computing the dimension if we don't have an annotation axis? specifying as a default Position.Bottom doesn't create a possibly wrong annotation?

src/state/annotation_utils.ts
```diff
@@ -0,0 +1,521 @@
...
+    : cursorPosition.y >= startY && cursorPosition.y <= endY;
+  return isCursorWithinXBounds && isCursorWithinYBounds;
+}
+
+export function getAnnotationLineStrokeWidth(lineStyle?: Partial<AnnotationLineStyle>): number {
+  if (lineStyle && lineStyle.line && (lineStyle.line.strokeWidth !== null)) {
+    return lineStyle.line.strokeWidth;
+  }
+  return DEFAULT_ANNOTATION_LINE_STYLE.line.strokeWidth;
+}
```
markov00: We are using `Partial<AnnotationLineStyle>` because the spec props is `lineStyle?: Partial<AnnotationLineStyle>;`
We are also merging with the default style in the render method (`canvas/reactive_chart`).
Can we merge instead before saving the lineannotationspec into the chartstore so we don't need this and the next function to get the `lineStokeWidth` and the `lineColor` ?

markov00: We can do the same also for the chart theme in the future

src/state/annotation_utils.ts
```diff
@@ -0,0 +1,521 @@
...
+  return dataValues.map((datum: AnnotationDatum): AnnotationLineProps => {
+    const { dataValue } = datum;
+    const details = {
+      detailsText: datum.details,
+      headerText: datum.header || dataValue.toString(),
+    };
+
+    // TODO: make offset dependent on annotationSpec.alignment (left, center, right)
+    const offset = xScale.bandwidth / 2;
+    const xDomainPosition = xScale.scale(dataValue) + offset;
```
markov00: if `dataValue` is not part of the x domain it will return `NaN` (try to change in the `basic xDomain ordinal` story the annotations value to `a,,d` where the second is an `''` string)
I think we can check the NaN here and skip the annotation line in that case.

src/state/annotation_utils.ts
```diff
@@ -0,0 +1,521 @@
...
+  const markerDimensions = { width: 16, height: 18 }; // todo : get from spec
+
+  return dataValues.map((datum: AnnotationDatum): AnnotationLineProps => {
+    const { dataValue } = datum;
+    const details = {
+      detailsText: datum.details,
+      headerText: datum.header || dataValue.toString(),
+    };
+
+    const yDomainPosition = yScale.scale(dataValue);
```
markov00: As for xScale, if `dataValue` is not part of the y domain it will return a value larger or lower than the range (try to change in the `basic yDomain` story the `generateAnnotationData([7.2]);`

I'm not sure if we can limit currently the scale value for continuous scales to limit the value in the domain (if I'm not wrong we are using this extra domain feature to compute cluster positions), but you can check the limit here and in case we are out of the range we can skip the annotation

src/state/annotation_utils.ts
markov00: Same happens for the continuous scale as for the yScale annotation

src/state/annotation_utils.ts
emmacunningham: Yes, thanks for catching!  Meant to leave a note/todo to early return here in case there is no corresponding axis.  Also will make a note to make sure we mention this in our documentation that an annotation must have a corresponding axis (and if the user doesn't want to show the axis, they can use the `hide` attribute)

src/components/annotation_tooltips.tsx
```diff
@@ -0,0 +1,83 @@
...
+        markers.push(markerElement);
+      });
+    });
+
+    return markers;
+  }
+
+  render() {
+    return (
+      <div>
```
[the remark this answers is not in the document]
emmacunningham: 👍 did not know about Fragments, cool!  Addressed in ecef817389abaa60c2cde3271aa247c5307d6fc0

src/state/annotation_utils.ts
emmacunningham: Addressed in 46c20b0006acafd2d45541fc4ffa83f1430e4ce5

src/state/annotation_utils.ts
emmacunningham: 👍 am now merging the styles on `addAnnotationSpec` (called both on mount and update)

addressed in 46c20b0006acafd2d45541fc4ffa83f1430e4ce5

src/lib/series/specs.ts
```diff
@@ -147,3 +147,42 @@ export enum Position {
...
+export enum AnnotationType {
+  Line = 'line',
+  Rectangle = 'rectangle',
+  Text = 'text',
+}
+
+export enum AnnotationDomainType {
+  XDomain = 'xDomain',
+  YDomain = 'yDomain',
+}
```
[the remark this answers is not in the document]
emmacunningham: Removed the two enums (AnnotationDomainType and AnnotationType) and instead export an object with the possible values that the user can use.

See b4f085cfed22b944784766cf8e7bf4cd9650d55b and d68416a7a18647e12ce97d0220c92f8220e140d4

src/state/annotation_utils.ts
```diff
@@ -0,0 +1,521 @@
...
+
+export interface AnnotationMarker {
+  icon: JSX.Element;
+  transform: string;
+  color: string;
+}
+
+export type AnnotationLinePosition = [number, number, number, number];
+export interface AnnotationLineProps {
+  position: AnnotationLinePosition; // Or AnnotationRectanglePosition or AnnotationTextPosition
```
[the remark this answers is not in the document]
emmacunningham: Ah yes, that was an old comment (have editied & moved it to the `AnnotationDimensions`)

src/state/annotation_utils.ts
```diff
@@ -0,0 +1,521 @@
...
+export function computeYDomainLineAnnotationDimensions(
+  dataValues: AnnotationDatum[],
+  yScale: Scale,
+  chartRotation: Rotation,
+  lineOverflow: number,
+  axisPosition: Position,
+  chartDimensions: Dimensions,
+  lineColor: string,
+  marker?: JSX.Element,
+) {
```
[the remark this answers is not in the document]
emmacunningham: added here 8e17bdf9a79fe1a9ff534e7fcaf9d5c67e722fca

src/state/annotation_utils.ts
emmacunningham: added here 8e17bdf9a79fe1a9ff534e7fcaf9d5c67e722fca

src/state/annotation_utils.ts
emmacunningham: See 15f0c54f60a2bef252e59b391375f0fedbf18f8d

src/state/annotation_utils.ts
emmacunningham: See 15f0c54f60a2bef252e59b391375f0fedbf18f8d

src/components/react_canvas/annotation.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+    return <Line key={`tick-${i}`} {...lineProps} />;
+  }
+
+  private renderAnnotation = () => {
+    const { chartDimensions, lines } = this.props;
+
+    return (
+      <Group x={chartDimensions.left} y={chartDimensions.top}>
+        <Group key="grid-lines">{lines.map(this.renderAnnotationLine)}</Group>
```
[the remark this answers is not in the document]
emmacunningham: Removed the outer group here 60ca86538188905b357972f6705977a049467939

src/state/annotation_utils.ts
emmacunningham: updated the stories so it is easier to add/remove values to test this 

y_range

emmacunningham: also in this commit we only check for the data value being out of the domain if the scale is not ordinal 99bab2805a74eb72157f5ab2a02b0a99a2b1226c

emmacunningham:
Made `LineAnnotationSpec.style` partial and optional here https://github.com/elastic/elastic-charts/pull/126/commits/0e588eb6820e289ebba1868d5961720b5c047213 (and added story for styling https://github.com/elastic/elastic-charts/pull/126/commits/ff6b70f94ad9e804fda7c344e73b7755efc9b18e)

We explicitly check for the empty string now to prevent the error when `d3.scale('')` returns 0. (https://github.com/elastic/elastic-charts/pull/126/commits/47406ae0398d1c0dab126fc38624b9b84a6cf932 and https://github.com/elastic/elastic-charts/pull/126/commits/1a2e08fd76f03bd4a1d183ba45f820709ec5fae8)

empty_string

Issue #121, opened 2019-03-26
Cleanup defaults


markov00:
The idea is to have a cleaner chart defaults parameters to minimize the code that a developer needs to wrote to build a basic chart.

Ideally this is the list of tasks thats needs to be implemented:
- [x] defaults to `canvas` as renderer
- [x] provides a default right-bottom axes configuration. If an axis component is added to the JSX, the existing default axis configuration will be removed.
- [x] default to legend on right shown or hidden
- [x] change `animateData` props with a clear name see (https://github.com/elastic/elastic-charts/pull/129#discussion_r270021088)
- [x] default `TooltipType`:
    - `None` can be used on sparklines
    - `None` is used for chart suggestion previews on new viz editor
    - `Follow` is best candidate when having more than 4 series (limiting the tooltip height)
- [x] `yScaleToDataExtent={false}` used in most charts as default
- [x] reenable animation props #129
- [x] convert Chart `size` into `width` and `height` that accept strings
- ~[ ] Make the curve type default to the smooth, curved version (CURVE_CATMULL_ROM)~
...

PR #129, opened 2019-03-27
fix(animation): re-enabled animateData prop to disable animation

markov00:
## Summary

The `animateData` prop was not used. This PR reenable this functionality and add a max thresholds of bars and lines/area points before disabling automatically the animation.

### Checklist

### DIFF
--- a/src/state/chart_state.ts
+++ b/src/state/chart_state.ts
@@ -69,6 +69,7 @@ import {
   findDataSeriesByColorValues,
   getAxesSpecForSpecId,
   getUpdatedCustomSeriesColors,
+  isChartAnimatable,
   isLineAreaOnlyChart,
   Transform,
   updateDeselectedDataSeries,
@@ -93,7 +94,6 @@ export type ElementClickListener = (values: GeometryValue[]) => void;
 export type ElementOverListener = (values: GeometryValue[]) => void;
 export type BrushEndListener = (min: number, max: number) => void;
 export type LegendItemListener = (dataSeriesIdentifiers: DataSeriesColorsValues | null) => void;
-// const MAX_ANIMATABLE_GLYPHS = 500;
 
 export class ChartStore {
   debug = false;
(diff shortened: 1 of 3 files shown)


src/state/utils.ts
```diff
@@ -481,3 +535,18 @@ export function isLineAreaOnlyChart(specs: Map<SpecId, BasicSeriesSpec>) {
...
+
+export function isChartAnimatable(
+  geometriesCounts: GeometrisCounts,
+  animationEnabled: boolean,
+): boolean {
+  if (!animationEnabled) {
+    return false;
+  }
+  const { bars, linePoints, areasPoints } = geometriesCounts;
+  if (bars > MAX_ANIMATABLE_BARS || linePoints + areasPoints > MAX_ANIMATABLE_LINES_AREA_POINTS) {
```
emmacunningham: Instead of the if/else statement, could we just return the boolean condition?

```
const isBarsAnimatable = bars <= MAX_ANIMATABLE_BARS;
const isLinesAndAreasAnimatable = linePoints + areasPoints <= MAX_ANIMATABLE_LINES_AREA_POINTS;
return isBarsAnimatable && isLinesAndAreasAnimatable;
```

src/state/chart_state.ts
```diff
@@ -789,13 +792,8 @@ export class ChartStore {
...
     this.axesVisibleTicks = axisTicksPositions.axisVisibleTicks;
     this.axesGridLinesPositions = axisTicksPositions.axisGridLinesPositions;
-    // if (glyphsCount > MAX_ANIMATABLE_GLYPHS) {
-    //   this.canDataBeAnimated = false;
-    // } else {
-    //   this.canDataBeAnimated = this.animateData;
-    // }
-    this.canDataBeAnimated = true;
 
+    this.canDataBeAnimated = isChartAnimatable(seriesGeometries.geometriesCounts, this.animateData);
```
emmacunningham: Perhaps we can rename `this.animateData` to something that indicates that it is a boolean setting?  `this.isAnimationEnabled`

markov00: Yes absolutely, I prefer to address this into https://github.com/elastic/elastic-charts/issues/121
so we can clean up a bit the interface with one single breaking change instead of a lot of small ones

src/state/utils.ts
markov00: solved in 7bbbaaa

src/state/utils.ts
```diff
@@ -50,6 +53,15 @@ export interface BrushExtent {
   maxY: number;
 }
 
+export interface GeometrisCounts {
```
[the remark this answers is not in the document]
markov00: solved in 7bbbaaa

PR #151, opened 2019-04-05
fix(timescale): consider timezone on axis ticks

### LINKED ISSUE #130: Enable ScaleType.Time independently of timezone
**Describe the bug**
The current implementation requires the data to be in UTC to work correctly.
We need to digest any type of date independently of the timezone.

**Errors in browser console**
If applicable, add any console error.

**Kibana Cross Issues**

- https://github.com/elastic/kibana/issues/32252

**Checklist**

markov00:
## Summary

The `d3-scaleTime` and `d3-scaleUtc` functions, used to compute axis ticks, compute a discrete and nice rounded number of ticks on a time scale. This rounding (the start of a day, an hour, a month, a year depending on the data interval) is applied to UTC or local timezones, meaning that we cannot display a nicely rounded tick if we want to display data in a timezone different from the local or utc one. 

This PR includes a new optional prop to each series `timeZone` that can be used to configure this behaviour (default to utc).
To enable a nice rounded ticks on d3-scaleUtc on any timezone we followed this approach:

your current status:
- my local timezone is UTC+1
- Kibana, or what else, is configured to use UTC-6 as timezone.
- I want to visualize my data on that timezone, that means: if my dataset is composed by 3 days from 00:00:00 of the 1st january to 00:00:00 of 3rd of january in UTC-6 I want visualize 3 data points, one on the left edge, one in the middle and the last in the right edge of my chart.
- I also want to visualize the relative ticks on the bottom axis whenever possible, on the same timezone of my data, means I want to visualize at least 3 ticks each one at the midnight of the 3 days.
- my data looks like the following:
```
[
  ['2014-01-01T00:00:00.000-06:00', 1388556000000, 6206],
  ['2015-01-01T00:00:00.000-06:00', 1420092000000, 5674],
  ['2016-01-01T00:00:00.000-06:00', 1451628000000, 4148],
  ['2017-01-01T00:00:00.000-06:00', 1483250400000, 6206],
  ['2018-01-01T00:00:00.000-06:00', 1514786400000, 3698],
]
```

the d3 `scaleUtc` scale is implemented to display nicely rounded axis ticks, rounded to hours, days, months days, depending on the domain extent.
This means that if my domain is `[1388556000000,1514786400000]` `scaleUtc` will find the nice time based on UTC timezone, that means:
- **t0** => `2014-01-01T00:00:00.000-06:00` in milliseconds is `1388556000000`
- `1388556000000` in UTC is `2014-01-01T06:00:00.000Z`
- `scaleUtc` will compute the ticks and will round that nice value to `2014-04-01T00:00:00.000Z`
- this means that rounded time is - 6 hours from the starting point **t0** `1388556000000` and graphically it will represented as the following
```
            t0
             |
--------------------------
  |
t0-6
```
no matter what type of formatting you are applying to the text label, the positioning will be always  shifted by -6 hours

To fix this, the idea is to apply a shift only on the axis scale. We want to shift the nice rounded value to the right place:
1) shift the domain back by the same utc offset:
   from `[1388556000000,1514786400000]` to `[1388556000000 ,1514786400000]`

WIP

 

fix #130

Display data with UTC-6 using UTC-6 as display timezone (the screenshot local timezone is UTC+1):

Screenshot 2019-04-05 at 12 13 15

Display UTC timezoned data using UTC as display timezone
Screenshot 2019-04-05 at 12 13 26

Display UTC timezoned data using local  timezone (UTC+1) as display timezone (data is shifted to +1 but the ticks are nicely kept to UTC+1 rounded days

Screenshot 2019-04-05 at 12 13 37

### Checklist

### DIFF
--- a/src/lib/utils/scales/scale_continuous.ts
+++ b/src/lib/utils/scales/scale_continuous.ts
@@ -164,7 +164,7 @@ export function isLogarithmicScale(scale: Scale) {
 }
 
 function invertValue(
-  minDomain: number,
+  domainMin: number,
   invertedValue: number,
   minInterval: number,
   stepType?: StepType,
@@ -177,7 +177,7 @@ function invertValue(
         return linearStepBefore(invertedValue, minInterval);
       case StepType.Step:
       default:
-        return linearStep(minDomain, invertedValue, minInterval);
+        return linearStep(domainMin, invertedValue, minInterval);
     }
   }
   return invertedValue;
@@ -197,13 +197,14 @@ export function linearStepAfter(invertedValue: number, minInterval: number): num
  * Return an inverted value that is valid from the half point before and half point
  * after the value. |----****|*****----|
  * till the end of the interval.
+ * @param domainMin the domain's minimum value
  * @param invertedValue the inverted value
  * @param minInterval the data minimum interval grether than 0
  */
-export function linearStep(minDomain: number, invertedValue: number, minInterval: number): number {
-  const diff = (invertedValue - minDomain) / minInterval;
+export function linearStep(domainMin: number, invertedValue: number, minInterval: number): number {
+  const diff = (invertedValue - domainMin) / minInterval;
   const base = diff - Math.floor(diff) > 0.5 ? 1 : 0;
-  return minDomain + Math.floor(diff) * minInterval + minInterval * base;
+  return domainMin + Math.floor(diff) * minInterval + minInterval * base;
 }
 
 /**
--- a/src/lib/utils/scales/scale_continuous.ts
+++ b/src/lib/utils/scales/scale_continuous.ts
@@ -1,4 +1,4 @@
-import { scaleLinear, scaleLog, scaleSqrt, scaleTime } from 'd3-scale';
+import { scaleLinear, scaleLog, scaleSqrt, scaleUtc } from 'd3-scale';
 import { DateTime } from 'luxon';
 import { ScaleContinuousType, ScaleType } from './scales';
 import { Scale } from './scales';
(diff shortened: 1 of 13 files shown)


src/lib/utils/scales/scale_continuous.ts
```diff
@@ -179,10 +200,10 @@ export function linearStepAfter(invertedValue: number, minInterval: number): num
  * @param invertedValue the inverted value
  * @param minInterval the data minimum interval grether than 0
  */
-export function linearStep(invertedValue: number, minInterval: number): number {
-  const diff = invertedValue / minInterval;
+export function linearStep(minDomain: number, invertedValue: number, minInterval: number): number {
```
emmacunningham: if I'm following correctly, `minDomain` represents the domain's minimum value instead of the minimum possible domain.  so I wonder if it this parameter could be `domainMin` as opposed to `minDomain` (which, especially since another parameter is called `minInterval`, makes it seem like this is the minimum domain amount instead of the domain mininum).  I guess the point I'm trying to get at is changing the name to reflect that in `minInterval` we are describing the distance between numbers, while the other parameter currently called `minDomain` is representing the minimum bound of a range of numbers.

maybe also we could add this param to the `@param` docs on the function to clarify what this value is (and that it comes from the computed domain).

src/lib/utils/scales/scale_continuous.ts
```diff
@@ -147,7 +163,12 @@ export function isLogarithmicScale(scale: Scale) {
   return scale.type === ScaleType.Log;
 }
 
-function invertValue(invertedValue: number, minInterval: number, stepType?: StepType) {
+function invertValue(
+  minDomain: number,
```
emmacunningham: similar to the comment on `minDomain` parameter in `linearStep`, a suggestion to comment what this parameter represents (and possibly change the name).

Issue #144, opened 2019-04-02
Add Band area chart


markov00:
**Is your feature request related to a problem? Please describe.**
Add a band area chart like in the following screenshot.
TSVB currently has this function when using percentile aggregation in band mode.
![6a00e554d9fb998833014e87c5dded970d](https://user-images.githubusercontent.com/1421091/55410259-d5498800-5563-11e9-8c60-ecbbd9ddb7c3.png)

**Describe the solution you'd like**
The area chart specification can be extended to add a y0 accessor to specify the minimum value of the area chart if not zero. This will be used to compute the band.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

**Kibana Cross Issues**
n/a

**Checklist**

- [x] this request is checked against already exist requests
- [x] every related Kibana issue is listed under `Kibana Cross Issues` list
- [x] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

PR #157, opened 2019-04-08
feat: add band area chart

markov00:
## Summary

close https://github.com/elastic/elastic-charts/issues/144

This PR adds a `y0Accessors` prop for each `SeriesSpec`: this property can be used to describe a dataset with bands ( with a min and max value for each data point).
It's mainly used for with `AreaSeries` to descibe a bounded area chart, bounded on the y axis. The  `y0` value, is usually a lower value than the `y1`, it will be rendered perpendicularly to the `y1`  point. On a standard area chart, the `y0` value usually correspond to `0` or to the previous value in the stack. In a band-area chart, the `y0` can be any value below the `y1` value.

It can also be applied to bar charts, with some limitations as described below. On a barchart, the `y0` value is the bottom edge of the bar.

On a stacked bar/area chart a series with a `y0Accessor` will be stacked on top of the below series stacking the below series `y1` value with the band series `y0` value. This means that if the below series at a point X has an height of `10`, and the bound series has the following values: `y0: 2, y1: 5` the resulting chart will push the bound series to the top starting using the following values: `y0:12, y1: 15`

Screenshot 2019-04-09 at 23 52 35

The `IndexedGeometry` object is also refactored. To limit amount of redundant code/memory object we used the same `Geometry` elements used on the highlight/hover process.
The structure of `IndexedGeometry` is slightly changed, to avoid duplicate values and to simplify tooltip formatting when rendering either y1 and y0 values.  

A candlestick-like barchart can be also configured, but the y0 value is not currently displayed on the tooltip.

Screenshot 2019-04-10 at 00 27 24

After few investigation on tests, I've currently removed from the `indexedGeometries` map all geometry with `null` values on `y1`. This leads to a minor problem: we are not showing any tooltip on that interval, the tooltip of the crosshair is never shown.

To Dos:
- ~~[ ] bring back animation on area charts~~ see https://github.com/elastic/elastic-charts/issues/161
- [x] add test on tooltip with stacked bars, check the right value for each bar
- ~~[ ] enable if possible the visualization of the y0 value for bars~~ see https://github.com/elastic/elastic-charts/issues/162

### Checklist


markov00:
@emmacunningham thanks for the review. I will address those problems in 3 separated PRs: one for testing negative/mixed values, one for cleaning the DataSeriesDatum types and one to add either values of the band to the legend.

PR #167, opened 2019-04-11
fix: remove old specs with changed ids

### LINKED ISSUE #158: Old series spec configuration remains when spec id is changed
**Describe the bug**
When a SeriesSpec id is changed, the chart updates with the new spec configurations but the older spec configurations also exist as a separate copy.  This is most noticeable when switching data sources; if you switch data sources and also update the spec id, you will see both data sets in the chart.

This is because on `componentDidUpdate`, we update the series specs by adding the updated spec to `chartStore.seriesSpecs` with the specId; if no specId changes are involved, the chart will update correctly.  But when the specId changes too (see the code example below), then the original spec does not get updated.  A first pass basic fix would be to check on update if the specId has changed; if it has, remove the previous spec from map before adding the updated spec with the new specId.

**To Reproduce**
1. Add a SeriesSpec w/ a specId (e.g. 'foo')
2. Toggle data source & also on toggle, change the specId (e.g. 'bar')
3. See that the old data source is still in the chart along with the new one.

Example spec:

```
const MyChart = () => {
  const [isUpdated, update] = useState(false);
  return (
    <>
      <Chart renderer="canvas" className={'story-chart'}>
        <LineSeries
          id={getSpecId(isUpdated ? 'lines2' : 'lines')}
          xScaleType={ScaleType.Time}
          yScaleType={ScaleType.Linear}
          xAccessor={0}
          yAccessors={[1]}
          data={
            isUpdated
              ? KIBANA_METRICS.metrics.kibana_os_load[0].data.map(([ts, val]) => [ts, val / 4])
              : KIBANA_METRICS.metrics.kibana_os_load[0].data
          }
          yScaleToDataExtent={false}
        />
      </Chart>
      <button onClick={() => update(true)}>Update</button>
    </>
  );
};
```

**Expected behavior**
Updated spec should replace previous spec; in this screenshot below, there should only be one line.

**Screenshots**

series_switch

 - OS: Mac 10.14.1 (will affect all OS)
 - Browser: Firefox 65.0.1 (will affect all browsers)
 - Elastic Charts: v3.8.0

**Additional context**

**Errors in browser console**
N/A

**Kibana Cross Issues**
N/A

**Checklist**

emmacunningham:
## Summary

fix #158 

This PR implements the fix for the issue of SpecComponent updates not clearing the previous spec version when the update involves a specId change.  Now, we check if the specId has changed on update, and if so, remove the spec by the previous specId from the chartStore.

Before:

broken_bar_spec

broken_line_spec

broken_area_spec

After:

fixed_bar_spec

fixed_line_spec

fixed_area_spec

### Checklist

~- [ ] Unit tests were updated or added to match the most common scenarios~

### DIFF
--- a/src/specs/area_series.tsx
+++ b/src/specs/area_series.tsx
@@ -22,9 +22,12 @@ export class AreaSeriesSpecComponent extends PureComponent<AreaSpecProps> {
     const { chartStore, children, ...config } = this.props;
     chartStore!.addSeriesSpec({ ...config });
   }
-  componentDidUpdate() {
+  componentDidUpdate(prevProps: AreaSpecProps) {
     const { chartStore, children, ...config } = this.props;
     chartStore!.addSeriesSpec({ ...config });
+    if (prevProps.id !== this.props.id) {
+      chartStore!.removeSeriesSpec(prevProps.id);
+    }
   }
   componentWillUnmount() {
     const { chartStore, id } = this.props;
--- a/src/specs/bar_series.tsx
+++ b/src/specs/bar_series.tsx
@@ -22,9 +22,12 @@ export class BarSeriesSpecComponent extends PureComponent<BarSpecProps> {
     const { chartStore, children, ...config } = this.props;
     chartStore!.addSeriesSpec({ ...config });
   }
-  componentDidUpdate() {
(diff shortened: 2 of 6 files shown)


markov00:
jenkins test this!

PR #170, opened 2019-04-12
feat: allow individual series styling

### LINKED ISSUE #138: Possibility to style each series
Currently I can style only the whole chart via `theme` prop in `<Settings />` component as described in the [storybook](https://elastic.github.io/elastic-charts/?selectedKind=Stylings&selectedStory=theme%2Fstyle&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybook%2Fstories%2Fstories-panel).
But the TSVB currently has a possibility to style each series separately:
image

I think each series should have its own style option, which should have `fill`, `line width` and `point size` fields inside:
e.x. `<LineSeries theme={ { fill: 0.5, lineWidth: 2, pointSize: 3} }`

**Additional context**

**Kibana Cross Issues**

**Checklist**

emmacunningham:
## Summary

close #138 

This PR introduces the ability to customize the styles for a series independently from the main chart theme (which is shared across all series).  If both a custom individual series style and custom chart theme are defined, precedence will be given to the custom series style (though the chart theme can still be used to change styles for the series which do not have their own custom series styles).

Currently, this supports the same set of customizable styling options as the main chart theme, but can be extended to cover additional styles currently not implemented as customizable.

The following examples can be found under `Styling` in storybook:

### Bars
currently supported customizable options:
- border stroke width
- border stroke color
- border visibility
- bars opacity (currrently not configurable from chart theme, but seems something that TSVB allows as a series-specific customization)

bar_style

### Lines
currently supported customizable options:
- line stroke width
- line visibility
- line opacity 
- point visibility
- point radius
- point opacity

line_style

### Areas
currently supported customizable options:
- line stroke width
- line visibility
- line opacity 
- point visibility
- point radius
- point opacity
- area visibility
- area opacity

area_style

### Checklist

### DIFF
--- a/src/lib/series/rendering.ts
+++ b/src/lib/series/rendering.ts
@@ -1,6 +1,14 @@
 import { area, line } from 'd3-shape';
 import { mutableIndexedGeometryMapUpsert } from '../../state/utils';
-import { CustomBarSeriesStyle, LineSeriesStyle, LineStyle, PointStyle, SharedGeometryStyle } from '../themes/theme';
+import {
+  AreaSeriesStyle,
+  AreaStyle,
+  CustomBarSeriesStyle,
+  LineSeriesStyle,
+  LineStyle,
+  PointStyle,
+  SharedGeometryStyle,
+} from '../themes/theme';
 import { SpecId } from '../utils/ids';
 import { isLogarithmicScale } from '../utils/scales/scale_continuous';
 import { Scale, ScaleType } from '../utils/scales/scales';
@@ -71,6 +79,8 @@ export interface AreaGeometry {
     y: number;
   };
   geometryId: GeometryId;
+  seriesAreaStyle?: AreaStyle;
+  seriesAreaLineStyle?: LineStyle;
 }
 
 export function isPointGeometry(ig: IndexedGeometry): ig is PointGeometry {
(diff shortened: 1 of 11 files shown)


src/lib/themes/theme.ts
```diff
@@ -90,16 +90,24 @@ export interface Theme {
 export interface BarSeriesStyle {
   border: StrokeStyle & Visible;
 }
+
+export type CustomBarSeriesStyle = BarSeriesStyle & Partial<Opacity>;
```
emmacunningham: This was created because BarSeriesStyle currently does not have a way to control opacity of the bar elements itself outside of the hover-effects; if we want to change this and add opacity as a property directly on the BarSeriesStyle interface, I can get rid of this extra type.

src/lib/series/rendering.ts
```diff
@@ -38,6 +46,7 @@ export interface PointGeometry {
   };
   geometryId: GeometryId;
   value: GeometryValue;
+  seriesPointStyle?: PointStyle;
```
markov00: Since we don't have individual style for each point, I think we can move this to the AreaGeometry and LineGeometry. Will avoid having this style copied to all data point and we can avoid checking for visibility on each datapoint on a line (we can just check it in the first series cycle. What do you think?

emmacunningham: moved `seriesPointStyle` to `AreaGeometry` and `LineGeometry`.  also moved the series-related style computation to when we iterate over each line/area (before this was happening for every point, even though they will all have the same styles (strokeWidth, radius, opacity) per line/area).  

see https://github.com/elastic/elastic-charts/pull/170/commits/f4658070979a99b0d9329a7b87df215e23fed212.

PR #173, opened 2019-04-16
fix: cleanup example prop default values

markov00:
## Summary

Because we started having some users of the library , it's better to start cleaning up a bit the default and the examples to avoid propagate unnecessary styles to new users.
This PR can partially solve the issues described here: https://github.com/elastic/elastic-charts/issues/121

This PR cleans up a bit the current examples:
- configure canvas as default renderer and remove the renderer props from examples
- as `yScaleToDataExtent` is default to false, I've removed the unneccessary prop from example
- removed the `basic` scale type from code. It was never used and can lead to possible misunderstandings
- updated the overview docs to reflect the current specs interfaces

### Checklist

- ~~[ ] Unit tests were updated or added to match the most common scenarios~~


markov00:
jenkins test this

markov00:
jenkins, test this

PR #180, opened 2019-04-20
feat(rect_annotation): add RectAnnotation type

### LINKED ISSUE #105: Add rectangular annotation & tooltip
Add ability to add annotation to a chart with a rectangle and tooltip.

User should be able to define an coordinates for the rectangle & the text to display for the tooltip.  User should also be able to define the style for the annotation (fill, strokeWidth, color, etc.) and default to base styles for undefined styles.

This is one of three annotation-related tasks (see #81).  

**Additional context**
Current feature in Kibana:
kibana_annotation

**Kibana Cross Issues**

**Checklist**

emmacunningham:
## Summary

close #105 

This PR introduces the functionality to add a `RectAnnotation` spec which will draw a rectangular annotation in a chart.  It also refactors the Annotation specs such that both `LineAnnotationSpec` and `RectAnnotationSpec` extend the same `BaseAnnotationSpec` for shared properties:

`BaseAnnotationSpec` specifies:
```
export interface BaseAnnotationSpec {
  /** The id of the annotation */
  annotationId: AnnotationId;
  /** Annotation type: line, rectangle, text */
  /** The ID of the axis group, generated via getGroupId method
   * @default __global__
   */
  groupId: GroupId; // defaults to __global__; needed for yDomain position
  /** Data values defined with coordinates and details */
  dataValues: AnnotationDatum[];
  /** Custom annotation style */
  style?: Partial<AnnotationStyle>;
  /** Toggles tooltip annotation visibility */
  hideTooltips?: boolean;
  /** z-index of the annotation relative to other elements in the chart
   * @default 0
   */
  zIndex?: number;
}
```

The `RectAnnotationSpec` is defined as follows:
```
export interface RectAnnotationSpec {
  annotationType: 'rectangle';
  /** Custom rendering function for tooltip */
  renderTooltip?: (
      position: { transform: string; top: number; left: number; }, 
      details?: string
    ) => JSX.Element;
  /** Data values defined with coordinates and details */
  dataValues: RectAnnotationDatum[];
  /** Custom annotation style */
  style?: Partial<RectAnnotationStyle>;
}
```

where each annotation dataValue (`RectAnnotationDatum`) has the shape:
```
export interface RectAnnotationDatum {
  coordinates: {
    x0?: any;
    x1?: any;
    y0?: any;
    y1?: any;
  };
  details?: string;
}
```

rect_anno_rotations

We validate & coerce the coordinate values such that if no valid coordinates are found, the annotation will be skipped & not render; so long as one of the coordinates is defined, we will draw the annotation by coercing the undefined coordinates to the chart extents such that:

| x0        | x1           | y0  |  y1  | render implications |
| ------------- |:-------------:|-----:|-----:|-----:|
| undefined    | undefined      |   undefined | undefined | no rendered annotation |
| **defined**    | undefined (coerced to x domain start)   |   undefined (coerced to y domain end) | undefined (coerced to y domain start) |  rect will be drawn from xDomain start to x1 with full chart height |
| undefined  (coerced to x domain end)   | **defined**      |   undefined  (coerced to y domain end) | undefined (coerced to y domain start) | rect will be drawn from x1 to x domain end with full chart height |
| undefined  (coerced to x domain end) | undefined (coerced to x domain start) |   **defined** | undefined (coerced to y domain start) | rect will be drawn from y domain start to y0 with full chart width |
| undefined (coerced to x domain end)  | undefined (coerced to x domain start) |   undefined  (coerced to y domain end) | **defined** | rect will be drawn from y1 to y domain end with full chart width |

defined_coord

The annotation may also be styled by passing a custom style, which allows customization of `stroke`, `strokeWidth`, `opacity`, and `fill`.

rect_anno_style

In the case where an annotation's hover tooltip is visible where there is also a chart element hover tooltip, precedence will go to the annotation tooltip unless the chart element hover has a specific element highlighted in which case the chart element tooltip will be visible instead.  (This matches the behavior of the tooltip on the annotations in Discover)

rect_anno_hover

Finally, we added a new prop `zIndex` which controls the relative z positioning of the annotations:

zIndex

### Checklist


emmacunningham:
jenkins test this

src/lib/series/specs.ts
```diff
@@ -229,5 +230,40 @@ export interface LineAnnotationSpec {
...
+  /** The id of the annotation */
+  annotationId: AnnotationId;
+  /** Annotation type: line, rectangle, text */
+  annotationType: AnnotationType;
+  /** The ID of the axis group, generated via getGroupId method
+   * @default __global__
+   */
+  groupId: GroupId; // defaults to __global__; needed for yDomain position
+  /** Custom marker */
+  marker?: JSX.Element;
```
markov00: I think we have to differentiate this marker from the one used on a LineAnnotation.
For Lines the markers are annotations on the chart, ideally we could only show those markers instead of lines and tooltips.
For RectangleAnnotations we are adding this marker to the tooltip. Maybe a different name, or at least a better description on the comment can be useful.
In that case we can thing of adding a prop like: 'tooltip?: JSX.Element` that is the content of the tooltip that takes two props that takes a `RectAnnotationDatum` as one passed prop, if the user submit the tooltip JSX element, we will render that, if not we fallback to a simple one.

src/lib/series/specs.ts
```diff
@@ -229,5 +230,40 @@ export interface LineAnnotationSpec {
...
+    y2: any;
+  };
+  details?: string;
+}
+
+export interface RectAnnotationSpec {
+  /** The id of the annotation */
+  annotationId: AnnotationId;
+  /** Annotation type: line, rectangle, text */
+  annotationType: AnnotationType;
```
markov00: Can't we create a generic interface for Annotations and have the type hardcoded for each spec? I personally don't understand why I've to specify an annotation type rectangle if I'm using a RectAnnotationSpec.

src/components/react_canvas/reactive_chart.tsx
```diff
@@ -181,19 +183,31 @@ class Chart extends React.Component<ReactiveChartProps, ReactiveChartState> {
         return;
       }
 
-      // We merge custom style w/ the default on addAnnotationSpec, so this is guaranteed
-      // to be complete by the time we get to rendering
-      const lineStyle = spec.style as AnnotationLineStyle;
+      if (isLineAnnotation(spec)) {
+        const lineStyle = spec.style as LineAnnotationStyle;
```
markov00: I think if we use a generic Annotation interface with generics (as specified below), we can have a cleaner code here

src/components/annotation_tooltips.tsx
```diff
@@ -12,31 +12,76 @@ interface AnnotationTooltipProps {
 class AnnotationTooltipComponent extends React.Component<AnnotationTooltipProps> {
   static displayName = 'AnnotationTooltip';
 
-  renderTooltip() {
-    const annotationTooltipState = this.props.chartStore!.annotationTooltipState.get();
-    if (!annotationTooltipState || !annotationTooltipState.isVisible) {
-      return <div className="elasticChartsAnnotation__tooltip elasticChartsAnnotation__tooltip--hidden" />;
-    }
+  renderLineAnnotationTooltip(tooltipState: AnnotationTooltipState) {
```
markov00: can we move these `renderLineAnnotationTooltip` and the `renderRectAnnotationTooltip` to to external components that accept style + the other props? we can use that as the default implementation, but the user can in the future add its own tooltip as part of the spec (see below)

src/lib/series/specs.ts
```diff
@@ -229,5 +230,40 @@ export interface LineAnnotationSpec {
...
 }
 
-// TODO: RectangleAnnotationSpec & TextAnnotationSpec
-export type AnnotationSpec = LineAnnotationSpec;
+export interface RectAnnotationDatum {
+  coordinates: {
+    x1: any;
+    x2: any;
+    y1: any;
+    y2: any;
```
markov00: A major remark here: it's super nice to have y1 and y2 here, but the user has to compute externally always the min and the max value of the dataset if we want to display a rect annotation that cover the whole domain.
Can we leave all them as optionals? if omitted we are specifying that we want to cover the whole extent to top or bottom or both.
Maybe the same can be done also for x1 and x2, it can simplify a bit the coding and the user doesn't always have to put there the initial x value or the final one

Another thing to notice: on code we are using always y0 and y1, where y0 is the bottom and y1 is the top value

emmacunningham: Renamed the coordinates to conform to zero-index naming: https://github.com/elastic/elastic-charts/pull/180/commits/9caa5bd35672c4a1abbf07683ce4d13a35c18f0b

And made all coordinates optional; so long as one value is defined, the other values will be defaulted to the domain extents: https://github.com/elastic/elastic-charts/pull/180/commits/cc4bb74b94d084c0b10fec965c0ffe642262980f

Below, we toggle which one coordinate is defined and the remaining coordinate values are set to the extent where if only x0 is defined, the rect is drawn from the x domain start to x0, if only x1 is defined, the rect is drawn from the x1 to the x domain end, if only y0 is defined, the rect is drawn from the y domain start to y0, and if only y1 is defined, the rect is drawn from y1 to the y domain end.
defined_coord

src/lib/series/specs.ts
emmacunningham: Have replaced `marker` with `renderTooltip` which allows the user to define a custom rendering. https://github.com/elastic/elastic-charts/pull/180/commits/60073b691dbe18a5c8ccfee461aae9de967468f1 (and also the following commit which refactors to move the tooltip rendering into ist own component https://github.com/elastic/elastic-charts/pull/180/commits/0dccb13e4f1807fd3623496bff99153708ea391c).

src/components/annotation_tooltips.tsx
emmacunningham: Moved rendering functions into own components: https://github.com/elastic/elastic-charts/pull/180/commits/0dccb13e4f1807fd3623496bff99153708ea391c

src/lib/series/specs.ts
emmacunningham: Have added a BaseAnnotationSpec which the annotation specs extend and hardcode the annotation type (was already not required to be supplied by the consumer because it was being defaulted in the respective spec components but now the specific type is also hardcoded).  https://github.com/elastic/elastic-charts/pull/180/commits/aecb912c10c4b29d0a068392dd752789da5d9f92

src/components/react_canvas/reactive_chart.tsx
emmacunningham: Do you mean the casting to the annotation style types here?  This is necessary to accomodate partial custom style definitions that the consumer might define and pass in as a spec prop (`spec.style` is typed as `Partial<LineAnnotationStyle>`); because we merge the custom styles with the full default styles on `addSpec`, `spec.style` will actually always be a fully defined `LineAnnotationStyle`, so this casting just reflects that.

markov00:
@emmacunningham I think we should also export on the index.ts the `RectAnnotationStyle` and the `LineAnnotationStyle`

emmacunningham:
@markov00 Added the `LineAnnotationSpec` & `RectAnnotationSpec` exports https://github.com/elastic/elastic-charts/pull/180/commits/2bd9451d8199214089d3d6fc162eedcfe1361e6d

emmacunningham:
@markov00 Added the zIndex prop to BaseAnnotationSpec, which will now control the ordering of how the elements.  Because of how react-konva doesn't handle zIndex (and instead relies on rendering order for this), this required a change in the implementation of the annotations; they now are rendered in the same layer as the chart elements.  I will leave the remaining work of ordering the chart elements to a separate PR as that will require grouping the chart elements by series & rendering the groups per zIndex on the series.

zIndex

emmacunningham:
jenkins test this

src/specs/rect_annotation.tsx
```diff
@@ -10,6 +10,7 @@ export class RectAnnotationSpecComponent extends PureComponent<RectAnnotationPro
   static defaultProps: Partial<RectAnnotationProps> = {
     groupId: getGroupId('__global__'),
     annotationType: 'rectangle',
+    zIndex: 0,
```
markov00: I think it's better to move the rect annotation on the background as default with a zIndex of -1 or -10 (so that on our current Kibana use cases we don't have to specify the zIndex for each RectAnnotation)

src/specs/line_annotation.tsx
```diff
@@ -14,6 +14,7 @@ export class LineAnnotationSpecComponent extends PureComponent<LineAnnotationPro
     style: DEFAULT_ANNOTATION_LINE_STYLE,
     hideLines: false,
     hideTooltips: false,
+    zIndex: 0,
```
markov00: We should already specify a default zIndex of 1 or 10 for lines so that our current use-cases are already satisfied bu our defaults

src/lib/series/specs.ts
```diff
@@ -255,6 +255,10 @@ export interface BaseAnnotationSpec {
   style?: Partial<AnnotationStyle>;
   /** Toggles tooltip annotation visibility */
   hideTooltips?: boolean;
+  /** z-index of the annotation relative to other elements in the chart
+   * @default 0
```
markov00: I think we should specify the default in a different way, as per my comment below we should use two different defaults, one for Line and one for Rect

src/specs/rect_annotation.tsx
emmacunningham: 👍 added here https://github.com/elastic/elastic-charts/pull/180/commits/d9aa483362e096dcf5866223599d8045de7c6f0c

src/specs/line_annotation.tsx
emmacunningham: added default here https://github.com/elastic/elastic-charts/pull/180/commits/d9aa483362e096dcf5866223599d8045de7c6f0c

src/lib/series/specs.ts
emmacunningham: have added a comment to the individual specs for RectAnnotation and LineAnnotation the different default values https://github.com/elastic/elastic-charts/pull/180/commits/d9aa483362e096dcf5866223599d8045de7c6f0c

PR #182, opened 2019-04-22
feat: add option to toggle value labels on bar charts

### LINKED ISSUE #67: Value labels option for charts without hovering
Often raised by users,  users use dashboard/visualization in kiosk node displaying it on monitor, they can't tell the value of the chart without hovering on the area of the chart, this issue often also raised in the context of reporting, when you export a report  isn't possible to hover and know the value of the chart. 
It will be helpful if we had an option to display the chart with the values labels (similar to pie)
image

emmacunningham:
## Summary

close #67 

This PR adds the option to toggle value labels on chart elements.  This does not handle edge cases (long display values overflowing their bars, etc.) and just implements these text labels directly on the BarGeometries.  Future iterations of this feature may handle edge cases as well as implementing the text labels as `TextAnnotation`s.

This PR adds a `showValueLabels` prop to the SeriesSpec such that it toggles the appearance of the value lables for a series.  `displayValue` has been added to the SeriesStyle such that it can be customized either from the theme or the series' custom style.

value_label

One can defined the prop `alternatingDisplayValue=true` so that the value labels alternate every other element.

If an element is not tall enough or the value label would overflow the top edge of the chart, we re-position the value label for that bar only (in the first case, we position the label above the bar; in the second, we position the label within the bar).

value_label

### Checklist


markov00:
We can maybe have a small subset of checking cases, like:
- move the text above the bar top edge if the y1 value is lower than the font size or below the bar top edge if the value is near the top of the chart
- add a simple feature to toggle text visibility in even/odd manner, a simplified method to avoid text overlapping.
I think we don't have to implement nothing more that that for this first relase

emmacunningham:
@markov00 Added the toggling between alternating display values and repositioning for the edge cases (when bar height is too small or text appearing above bar):

value_label

emmacunningham:
The implementation has been rewrittten to support the additional feedback/requests.  Previously, we were drawing the labels directly in the bar chart elements, but that was causing issues where the text would be covered up by any neighboring bars due to react-konva z-indexing based on render order.  Instead, now we create the labels on their own Layer which prevents the labels from getting covered up.  When we add the TextAnnotation feature, this layer can be re-used to feed values into the TextAnnotation layer.

Values will be positioned relative to chartRotation:

value_rotations

`offsetX` and `offsetY` were added to the spec to allow the user to directly manipulate these offsets without having to rely on padding (would like to address padding in a separate PR as there are various edge cases to think through for this).

value_offsets

There is another prop that the user can set to have the value labels contained within the bar element itself (by default values will overflow): 

value_contained

Also there is another prop which can be used to control if clipped values (overflowing the edges of the chart) should be hidden or visible (would like to address in a separate PR how to handle these values besides just hiding them):

value_hide_clipped

Something else (that I would like to address in a followup PR) is how to handle automatically adjusted values: currently if a bar height is not tall enough to hold the display value, we move the label to above the bar.  This causes some confusion when also applying an offset; should the label also move (as it currently does)?  If so, then the offset amount will be greater than for the other labels which fit inside of the bar.  One strategy might be to only apply the offset once it exceeds the automatically adjusted text's already offset position.

A select knob has been added to the story to change the data volume from small, medium, and high:

value_data

emmacunningham:
fyi just checked how the labels look for stacked & split series (this is added to stories also):

value_split_stacked

PR #200, opened 2019-04-30
fix(scales): use bisect to handle invertWithStep

### LINKED ISSUE #183: Tooltips on hover disappear with discontinuous data
With this data:

`const data = [[1555819200000, 507741], [1555862400000, 505757], [1555905600000, 502]]`

and this JSX for the chart:
```
        <Chart size={[800, 300]} renderer="canvas">
          <Settings
            theme={getChartTheme()}
            xDomain={{min: 1553861780116, max: 1556021780116}}
          />
          <Axis
            id={getAxisId('bottom')}
            position={Position.Bottom}
            showOverlappingTicks={true}
            tickFormat={dateFormatter}
          />
          <Axis id={getAxisId('left')} title={aggLabel} position={Position.Left} />
            return (
              <LineSeries
                key={key}
                id={getSpecId(key)}
                xScaleType={ScaleType.Time}
                yScaleType={ScaleType.Linear}
                data={data}
                xAccessor={0}
                yAccessors={[1]}
                timeZone="America/New_York"
              />
            );
          <LineSeries
            id={getSpecId('threshold')}
            xScaleType={ScaleType.Time}
            yScaleType={ScaleType.Linear}
            data={[[domain.min, watch.threshold], [domain.max, watch.threshold]]}
            xAccessor={0}
            yAccessors={[1]}
            timeZone="America/New_York"
            yScaleToDataExtent={true}
            customSeriesColors={thresholdCustomSeriesColors}
          />
        </Chart>
```
tooltips do not appear on hover.  If the domain is not wider than the data, the tooltips appear, if not then they don't.

### LINKED ISSUE #195: elasticChartsCrosshair is not positioned correctly for some intervals
**Describe the bug**
elasticChartsCrosshair is not positioned correctly for some intervals

In my example: timeFrame: 1week, interval: 12h

**Screenshots**
chart_issue_1

markov00:
## Summary

close #195
close #183 

We actually relay on the fact that data is evenly distributed among our x axis. We also rely on the `minInterval` value (the minimum interval between two consecutive x values on your dataset) to determine the current interval between x values.
Those wrong assumptions leads to issues when interacting with a chart: the interaction invert the pixel coordinate of the mouse to a scale value: the `invertWithStep` function to get the exact `x` value of the dataset for a bar chart with linear x axis, using the minInterval and the bandwidth to compute it. If we have odd placed x values or a domain that is larger than the current max x value, the tooltip and the highlighter are misaligned and wrongly displayed as in the closing issues #195 #183

Instead of relying on the minInterval, I've refactored the code to use a `bisect` function on `invertWithStep`, to get back the right x value from the chart.

I've also pixel adjusted few tests, that use the previous `invertWithStep` function.

### Checklist


src/state/crosshair_utils.ts
```diff
@@ -76,7 +77,7 @@ export function getCursorBandPosition(
   }
   const isHorizontalRotated = isHorizontalRotation(chartRotation);
   const snappedPosition = getSnapPosition(
-    xScale.invertWithStep(isHorizontalRotated ? x : y),
+    xScale.invertWithStep(isHorizontalRotated ? x : y, data),
```
emmacunningham: Since `invertWithStep` may return `undefined` for continuous scales, do we need to update the type signature for `getSnapPosition`?  Currently `getSnapPosition` defines `value: string | number` for its first parameter, but if the scale is a continuous scale, it's possible that the return value of `invertWithStep` is `undefined`.  (I think we already handle if the scaled value is undefined, so the implementation doesn't need to change but rather just the type signature to more accurately reflect the possible values)  Further, should we tighten the return type of `invertWithStep` in the `Scale` interface?  Currently it is defined as `(value: number, data: any[]) => any` but `getSnapPosition` expects `string | number`.

markov00: As in its `Scale` parent class, `invertWithScale` return type is `any` because at runtime we can be never sure that the `data` object passed to the `invertWithScale` function contains only numbers. I've add few more checks on d42bf42  for `getCursorBandPosition` and for `invertWithStep` to avoid problems with undefined/null values. We should however check all the scales and the data types to have a correct handling of null/undefined values

emmacunningham: 👍 we could decode runtime values with something like `io-ts` but for now manually validating the data types within these functions works

PR #208, opened 2019-05-20
fix(css): remove dependency on EUI components and use only EUI styles

markov00:
## Summary

This PR update the EUI dependency and clean its use inside the library. The following thing are changed:
- each css style is prefixed with `ech` instead of the long `elasticCharts`
- the new css is fully compatible with IE11, Chrome and Firefox
- the **legend toggle** is currently commented out ( we can decide to reintroduce it or leave that to the external container)
- the **color picker** and the contextual menu on the legend are temporarily disabled. They are actually part of EUI, so we can have add them as external prop components in a later version (only vislib is currently use this function)
- the style for the `eye` open and closed on the legend is different from the EUI one: there is no animation on that button, but only a static change.

*Before*
Screenshot 2019-05-20 at 23 04 00

*After*
Screenshot 2019-05-20 at 23 04 10

- text are a bit more condensed and the width of the legend is changed from 160px to 200px.

*Before*
Screenshot 2019-05-20 at 23 11 06

*After* 
Screenshot 2019-05-20 at 23 10 55

- fixed the alignment of latest element on bottom/top legend

*Before*
Screenshot 2019-05-20 at 23 01 32

*After*
Screenshot 2019-05-20 at 23 01 44

### Checklist


stories/bar_chart.tsx
```diff
@@ -857,12 +860,13 @@ storiesOf('Bar Chart', module)
           tickFormat={(d) => Number(d).toFixed(2)}
         />
 
-        <BarSeries
+        <AreaSeries
```
emmacunningham: Should this be `BarSeries` since this is the bar chart story group?

src/components/icons/assets/alert.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { Props } from '../icon';
+
+export function AlertIcon(extraProps: Props) {
+  return (
+    <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" {...extraProps}>
+      <path
+        fill-rule="evenodd"
```
emmacunningham:
```suggestion
        fillRule="evenodd"
```

This error appears when using the dash-delimited version:
```Warning: Invalid DOM property `fill-rule`. Did you mean `fillRule`?```

**Comment by cchaos** [comment]
I am struggling to test this in Kibana. I have built it with `yarn build` then linked it over in Kibana. I see both the `theme_light.css` and `theme_dark.css` in node_modules. But when I add those theme files to https://github.com/elastic/kibana/blob/cc7a6ec2c11b51bbbc64a319b29750232a5dc175/src/legacy/ui/ui_render/ui_render_mixin.js#L119-L125

like so:

```js
darkMode ?
              [
                `${basePath}/node_modules/@elastic/eui/dist/eui_theme_dark.css`,
                `${basePath}/node_modules/@kbn/ui-framework/dist/kui_dark.css`,
                `${basePath}/node_modules/@elastic/charts/dist/theme_dark.css`,
              ] : [
                `${basePath}/node_modules/@elastic/eui/dist/eui_theme_light.css`,
                `${basePath}/node_modules/@kbn/ui-framework/dist/kui_light.css`,
                `${basePath}/node_modules/@elastic/charts/dist/theme_light.css`,
              ]
```

I get 404 errors and it completely borks Kibana, getting the "Kibana didn't load properly" error.

I can import the `theme_light.scss` file into https://github.com/elastic/kibana/blob/master/src/legacy/ui/public/styles/_styling_constants.scss via:

```scss
@import '@elastic/charts/src/theme_light';
```

But of course, that duplicates these styles numerous times and I'm pretty sure it doesn't swap light for dark.

Any thoughts on how to appropriately handle the themes and importing?

markov00:
@cchaos I think you should also add the following:
```
  server.exposeStaticDir('/node_modules/@elastic/charts/dist/{path*}', fromRoot('node_modules/@elastic/charts/dist'));

```

at line 59 of `ui_render_mixin` https://github.com/elastic/kibana/blob/cc7a6ec2c11b51bbbc64a319b29750232a5dc175/src/legacy/ui/ui_render/ui_render_mixin.js#L59

I've just tried and it works

src/components/_tooltip.scss
```diff
@@ -1,18 +1,20 @@
-.elasticChartsTooltip {
-  @include euiBottomShadow($color: $euiColorFullShade);
-  @include euiFontSizeXS;
+.echTooltip {
```
cchaos: For tooltips, can you actually duplicate this mixin: https://github.com/elastic/eui/blob/master/src/components/tool_tip/_mixins.scss

Currently the shadow does not work in dark mode
Screen Shot 2019-05-21 at 14 30 24 PM

src/style/globals/variables/_typography.scss
```diff
@@ -0,0 +1,57 @@
...
+// EX: A proper line-height for text is 1.5 times the font-size.
+//     If our base font size (echFontSize) is 16, our baseline is 8 (16*1.5 / 3). To ensure the
+//     text stays on the baseline, we pass a multiplier to calculate a line-height in rems.
+@mixin lineHeightFromBaseline($multiplier: 3) {
+  line-height: convertToRem(($echFontSize/2) * $multiplier);
+}
+
+// Families
+
+$echFontFamily: 'Inter UI', -apple-system, BlinkMacSystemFont, 'Segoe UI', Helvetica, Arial,
```
cchaos: We shouldn't be importing 'Inter UI' in this base library. We should probably just stick to system fonts. In fact, I think there's a lot of unnecessary invisibles brought in here from EUI.

src/components/_tooltip.scss
markov00: Sure, working on that right now.

src/style/globals/variables/_typography.scss
markov00: @cchaos what are the system fonts i can use? `-apple-system, BlinkMacSystemFont, Helvetica, Arial, sans-serif` ?
I also need the right one to use on the canvas text elements, what I can use there?

snide: Removing the `Inter UI` bit is all you need. The ones that follow are "system fonts". Apple, then MS, then Linux, then Helvetica.

markov00:
Hey @cchaos are you still working on that? Should we start to wrap up and merge this PR?
If you are working on that, please could you list here what are your current task so I can help you on those?

cchaos:
@markov00 Yes, I'm actively working on this. I think I'm just about done, but have I'm double-checking it in EUI right now and will need to do so in Kibana as well.

cchaos:
@markov00 PR4U: https://github.com/markov00/elastic-charts/pull/2

markov00:
hey @cchaos I've merged your PR and rebased everything. 
What do you think about creating two different css output, one with reset and one without.
So that the user just need to import one or the other css, and not two separate files.
I'm also wondering if we need to apply the reset only under the ech container, and not to the whole html. I don't want to mess with the existing reset, but maybe there is something I'm not considering.

markov00:
@cchaos @snide I've also minor concern on the readability of the highlighted value on the tooltip.

This is the current implementation:
Jun-05-2019 15-52-54


This is the Caroline changes:
Jun-05-2019 15-46-51

From my point of view, adding changing the font-weight with a higher value create few disadvantages:
- the highlighted rows is poorly distinguishable from others (your eyes need to catch a small difference in a font size, than in a background color change) specially if you enter quickly into the chart.
- changing the font-weight to a higher value increase also the tooltip width creating a strange effect that grow and shrink the tooltip when highlighting a bar.

cchaos:
@markov00 I actually agree. I wanted to come back and revisit that but had forgotten. I will work on those states.

cchaos:
Pushed a fix for the tooltip:

cchaos:
> I'm also wondering if we need to apply the reset only under the ech container, and not to the whole html. I don't want to mess with the existing reset, but maybe there is something I'm not considering.

The problem with that is that it will override any base styles declared outside of the .echContainer. 

## Example

Reset file includes:

```css
a { color: blue }
```

Consumer's theme is:

```css
a { color: red }
```

But then including the charts styles with the reset would make it:

```css
.echContainer a { color: blue }
```

completely overriding the consumer's theme.

package.json
```diff
@@ -91,7 +90,6 @@
     "husky": "^1.3.1",
     "jest": "^24.1.0",
     "jest-environment-jsdom-fourteen": "^0.1.0",
-    "moment": "^2.24.0",
```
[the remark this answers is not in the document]
cchaos: EUI is only being consumed for it's SASS variables so there's no need fo this peer dependency.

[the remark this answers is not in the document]
nickofthyme: @cchaos cool thanks!

cchaos:
@markov00 @snide My latest commit does as @markov00 had asked and compiles one file with reset and theme, and then a theme only file.

snide:
@cchaos OK. Looks good. Thanks for updating the docs too.

cchaos:
> I've also played a bit with the dark mode, and seems that in darkmode the tooltip is fully black.

This is how the tooltip works in EUI. Not backgrounds should actually be fully black. Our default "white" background actually goes to a "dark gray" in dark mode:

Screen Shot 2019-06-06 at 13 39 14 PM

I had also changed that fully black background **to** our default background color here: https://github.com/elastic/elastic-charts/pull/208/files#diff-6cc5987d83a2bd4599131f08c70199cf

Screen Shot 2019-06-06 at 13 44 03 PM

You can see there is a slight variation in color. But that's how we work with them currently in EUI and would like to stick with the same method here.

cchaos:
Pushed more fixes to legend toggle.

markov00:
@cchaos I'm ok with the changes. If you think we are done, I will fix the squash commit and merge it

markov00:
@cchaos I can do that if you want, let me fix conflict, merge and I will create the PR in Kibana

PR #213, opened 2019-05-21
fix(specs):add custom style inside each specific series style

### LINKED ISSUE #188: The custom styling props should be series specific
**Describe the bug**
The current implementation of `BasicSeriesSpec` incorporate the custom styles for each different series.
```
export type BasicSeriesSpec = SeriesSpec & SeriesAccessors & SeriesScales & {
  barSeriesStyle?: CustomBarSeriesStyle;
  lineSeriesStyle?: LineSeriesStyle;
  areaSeriesStyle?: AreaSeriesStyle;
};
```

**Expected behavior**
Each custom style should be declared inside it's specific series style like

```
export type BarSeriesSpec = BasicSeriesSpec & {
  /** @default bar */
  seriesType: 'bar';
  barSeriesStyle?: CustomBarSeriesStyle;
};

export type LineSeriesSpec = BasicSeriesSpec & {
  /** @default line */
  seriesType: 'line';
  curve?: CurveType;
  lineSeriesStyle?: LineSeriesStyle;
};
```

rshen91:
## Summary

Fixes #188 
In specs.ts, BasicSeriesSpec included the styles for LineSeriesSpec, AreaSeriesSpec, and BarSeriesSpec. This PR removes barSeriesStyle, lineSeriesStyle, and areaSeriesStyle from BasicSeries style and includes the SeriesStyle per each chart type. 

Included in specs.ts are the type guards for utils.ts 

```
export function isBarSeriesSpec(spec: BasicSeriesSpec): spec is BarSeriesSpec {
  return spec.seriesType === 'bar';
}

export function isLineSeriesSpec(spec: BasicSeriesSpec): spec is LineSeriesSpec {
  return spec.seriesType === 'line';
}

export function isAreaSeriesSpec(spec: BasicSeriesSpec): spec is AreaSeriesSpec {
  return spec.seriesType === 'area';
}
```

### Checklist

~~-  Any consumer-facing exports were added to `src/index.ts` (and stories only import from `../src` except for test data & storybook)~~
~~-  Proper documentation or storybook story was added for features that require explanation or tutorials~~

~~- [ ] Unit tests were updated or added to match the most common scenarios~~

### DIFF
--- a/src/lib/series/specs.ts
+++ b/src/lib/series/specs.ts
@@ -122,6 +122,7 @@ export type BasicSeriesSpec = SeriesSpec & SeriesAccessors & SeriesScales & {
 export type BarSeriesSpec = BasicSeriesSpec & {
   /** @default bar */
   seriesType: 'bar';
+  barSeriesStyle?: CustomBarSeriesStyle;
 };
 
 /**
@@ -131,6 +132,7 @@ export type LineSeriesSpec = BasicSeriesSpec & {
   /** @default line */
   seriesType: 'line';
   curve?: CurveType;
+  lineSeriesStyle?: LineSeriesStyle;
 };
 
 /**
@@ -141,6 +143,7 @@ export type AreaSeriesSpec = BasicSeriesSpec & {
   seriesType: 'area';
   /** The type of interpolator to be used to interpolate values between points */
   curve?: CurveType;
+  areaSeriesStyle?: AreaSeriesStyle;
 };
 
 /**
--- a/src/lib/series/specs.ts
+++ b/src/lib/series/specs.ts
@@ -110,11 +110,7 @@ export interface SeriesScales {
   yScaleToDataExtent: boolean;
 }
 
-export type BasicSeriesSpec = SeriesSpec & SeriesAccessors & SeriesScales & {
-  barSeriesStyle?: CustomBarSeriesStyle;
-  lineSeriesStyle?: LineSeriesStyle;
-  areaSeriesStyle?: AreaSeriesStyle;
-};
+export type BasicSeriesSpec = SeriesSpec & SeriesAccessors & SeriesScales;
 
 /**
  * This spec describe the dataset configuration used to display a bar series.
@@ -298,4 +294,16 @@ export function isLineAnnotation(spec: AnnotationSpec): spec is LineAnnotationSp
 
 export function isRectAnnotation(spec: AnnotationSpec): spec is RectAnnotationSpec {
   return spec.annotationType === AnnotationTypes.Rectangle;
+}
+
+export function isBarSeriesSpec(spec: BasicSeriesSpec): spec is BarSeriesSpec {
+  return spec.seriesType === 'bar';
+}
+
+export function isLineSeriesSpec(spec: BasicSeriesSpec): spec is LineSeriesSpec {
+  return spec.seriesType === 'line';
+}
+
+export function isAreaSeriesSpec(spec: BasicSeriesSpec): spec is AreaSeriesSpec {
+  return spec.seriesType === 'area';
 }
--- a/src/state/utils.ts
+++ b/src/state/utils.ts
@@ -27,6 +27,9 @@ import {
   AxisSpec,
   BasicSeriesSpec,
   DomainRange,
+  isAreaSeriesSpec,
+  isBarSeriesSpec,
+  isLineSeriesSpec,
   LineSeriesSpec,
   Rotation,
 } from '../lib/series/specs';
@@ -343,8 +346,8 @@ export function renderGeometries(
       continue;
     }
     const color = seriesColorsMap.get(ds.seriesColorKey) || defaultColor;
-    switch (spec.seriesType) {
-      case 'bar':
+
+    if (isBarSeriesSpec(spec)) {
         const shift = isStacked ? indexOffset : indexOffset + i;
 
         // TODO: we can handle style merging here and not pass that off to the component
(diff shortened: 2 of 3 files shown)


src/lib/series/specs.ts
```diff
@@ -141,6 +143,7 @@ export type AreaSeriesSpec = BasicSeriesSpec & {
   seriesType: 'area';
   /** The type of interpolator to be used to interpolate values between points */
   curve?: CurveType;
+  areaSeriesStyle?: AreaSeriesStyle;
 };
```
emmacunningham: These should also be removed from BasicSeriesSpec so that they are defined only for the specific spec type.

rshen91: Ah! Following - added commit ef1bbfd

src/lib/series/specs.ts
```diff
@@ -110,39 +110,41 @@ export interface SeriesScales {
...
+export type  AreaSeriesSpec = SeriesSpec &  {
   /** @default area */
   seriesType: 'area';
   /** The type of interpolator to be used to interpolate values between points */
   curve?: CurveType;
+  areaSeriesStyle?: AreaSeriesStyle;
 };
 
+// export a joined type what series spec are available by different fnctions
+export type AnySeriesSpec = BarSeriesSpec | LineSeriesSpec | AreaSeriesSpec;
```
emmacunningham: We want to avoid doing this.  The purpose of this fix is to make each custom series style specific only to the relevant series spec.

src/lib/series/rendering.areas.test.ts
```diff
@@ -4,13 +4,13 @@ import { ScaleType } from '../utils/scales/scales';
...
 import { computeXScale, computeYScales } from './scales';
-import { AreaSeriesSpec } from './specs';
+import { AreaSeriesSpec, BasicSeriesSpec } from './specs';
 const SPEC_ID = getSpecId('spec_1');
 const GROUP_ID = getGroupId('group_1');
 
 describe('Rendering points - areas', () => {
   describe('Single series area chart - ordinal', () => {
-    const pointSeriesSpec: AreaSeriesSpec = {
+    const pointSeriesSpec: AreaSeriesSpec | BasicSeriesSpec = {
```
emmacunningham: Here and other places where you're casting the type as either the specific series type or the basic series: 

We should keep the casted type to just the specific series.

src/state/utils.ts
```diff
@@ -440,7 +441,7 @@ export function renderGeometries(
   };
 }
 
-export function getSpecById(seriesSpecs: Map<SpecId, BasicSeriesSpec>, specId: SpecId) {
+export function getSpecById(seriesSpecs: Map<SpecId, BasicSeriesSpec & AnySeriesSpec>, specId: SpecId) {
```
emmacunningham: Per my comment earlier, we don't want to introduce this additional type.

I think you've introduced this to get around the issues where the specific series styles are not defined on the BasicSeriesSpec (expected once we remove them), but this is not the way to go about resolving that problem as this introduce an additional generic type and more complications to the type system.  Instead we want to leave the specific style properties specific to the relevant spec type, that is:

```
export type BasicSeriesSpec = SeriesSpec & SeriesAccessors & SeriesScales;

/**
 * This spec describe the dataset configuration used to display a bar series.
 */
export type BarSeriesSpec = BasicSeriesSpec & {
  /** @default bar */
  seriesType: 'bar';
  barSeriesStyle?: CustomBarSeriesStyle;
};

/**
 * This spec describe the dataset configuration used to display a line series.
 */
export type LineSeriesSpec = BasicSeriesSpec & {
  /** @default line */
  seriesType: 'line';
  curve?: CurveType;
  lineSeriesStyle?: LineSeriesStyle;
};

/**
 * This spec describe the dataset configuration used to display an area series.
 */
export type AreaSeriesSpec = BasicSeriesSpec & {
  /** @default area */
  seriesType: 'area';
  /** The type of interpolator to be used to interpolate values between points */
  curve?: CurveType;
  areaSeriesStyle?: AreaSeriesStyle;
};
```

Then, in `renderGeometries` in `utils`, you'll need to correct the type errors by casting each series spec to the more specific type (they come in as `BasicSeriesSpec`s but once we know what the `seriesType` is, we can reliably cast them to the more specific types).

The quick fix is to just cast them inline; in `renderGeometries` there's the switch statement on `spec.seriesType`, to cast, say, the line spec, you could do:

```
case 'line': 
  const lineSpec = spec as LineSeriesSpec;
  const lineSeriesStyle = lineSpec.lineSeriesStyle; // This will now be defined because lineSpec has been cast as a LineSeriesSpec which does have this property unlike BasicSeriesSpec which doesnt
```

A slightly more elegant way to handle this which doesn't require that you explicitly cast within the `renderGeometries` function would be to use a type guard (you'd also then swap out the `switch/case` for an `if/else`) [see the user defined type guards section [here](https://www.typescriptlang.org/docs/handbook/advanced-types.html); there are also examples of this for the AnnotationSpec types and the way that the differently bounded domain types in `axis_utils` are handled]

markov00: Another suggestion, instead of casting the spec to a specific type with `spec as LineSeriesSpec;` you can change the `switch` to a series of if and add few utilities functions that helps TS identify your general type like:
```
function isLineSeriesSpec(spec: BasicSeriesSpec): spec is LineSeriesSpec {
    return spec.seriesType === 'line';
}
```
This will also helps in case we want to change the seriesType value in the future or the BasicSeriesSpec in general, we can just change the helper functions and not all places where we are testing `seriesType` attribute to find the specific type.

rshen91: Awesome! - addressed in commit a1f62ec

src/lib/series/rendering.areas.test.ts
rshen91: I reverted a few commits and then squashed them in 8c9653d

src/lib/series/specs.ts
rshen91: I reverted a couple commits and then squashed them in 8c9653d

Issue #94, opened 2019-03-07
Add tick label padding to axis spec


emmacunningham:
**Is your feature request related to a problem? Please describe.**
We add a small amount of default padding (1px) to axis tick labels to help with discrepancies in browser canvas text measurement.  Users might want to themselves be able to adjust the amount of padding for a tick label, which is what is proposed here.

**Describe the solution you'd like**
Add tick label padding to the axis spec and use this to adjust the computed bounding box dimensions for a tick label.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

**Kibana Cross Issues**
n/a

**Checklist**

- [x] this request is checked against already exist requests
- [x] every related Kibana issue is listed under `Kibana Cross Issues` list
- [x] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

PR #217, opened 2019-05-22
feat(axis): add tickLabelPadding prop 

rshen91:
## Summary

Closes https://github.com/elastic/elastic-charts/issues/94
This is a duplicate of the closed PR https://github.com/elastic/elastic-charts/pull/176 (closed in error)

Allows users to specify a tick label padding for charts. In storybook, when debug is checked, the user can add a specified tick label padding number to the bottom axis and rotations with ordinal axis.
In storybook this can be seen in the following charts:
- Axis basic
- Axis tick label rotation

Screen Shot 2019-05-21 at 10 48 32 AM

This functionality is also included as a theme for the bottom tick label padding: 

Screen Shot 2019-05-21 at 10 49 29 AM

In the Stylings stories, there is a tickLabelPadding both prop and theme story included to show both the axis spec and theme manipulating tickLabelPadding.
PR217

### Checklist


rshen91:
Copying from earlier [PR](https://github.com/elastic/elastic-charts/pull/176) 
> Have you tested to see what happens if there is an axis-specific padding configuration versus if it's on the theme itself? May be useful to have a story showing the difference (see the custom series styling stories for an example).

markov00:
hey @rshen91 is good to review  or still a WIP?

rshen91:
> hey @rshen91 is good to review or still a WIP?

Sorry about that @markov00! Please feel free to review when you have a chance

stories/styling.tsx
```diff
@@ -661,4 +661,49 @@ storiesOf('Stylings', module)
...
+          fill: color('tickFill', '#333', 'Tick Label'),
+          fontSize: range('tickFontSize', 0, 40, 10, 'Tick Label'),
+          fontFamily: `'Open Sans', Helvetica, Arial, sans-serif`,
+          fontStyle: 'normal',
+          padding: number('Tick Label Padding Theme', 1, {}, 'Tick Label'),
+        },
+      },
+    };
+    const customTheme = mergeWithDefaultTheme(theme, LIGHT_THEME);
+    const tickLabelPadding = number('Tick Label Padding Prop', 0);
```
emmacunningham: currently, the way this is set up the theme padding will never have an effect because the AxisSpec configuration will take precedence.  if you remove the knobbed value from one of the axis specs, then you should be able to see the theme padding affecting one axis and the axis spec padding affecting the other.

src/lib/series/specs.ts
```diff
@@ -177,6 +177,8 @@ export interface AxisSpec {
   title?: string;
   /** If specified, it constrains the domain for these values */
   domain?: DomainRange;
+  /** Handles the user supplied padding for axis debug */
```
emmacunningham: Can we update this comment to `Specifies the amount of padding on the tick label bounding box`?  This property doesn't really have anything to do with the `debug` setting other than that's how we're able to confirm its being applied correctly.

src/lib/series/rendering.ts
```diff
@@ -234,7 +237,8 @@ export function renderBars(
       (barGeometries.length % 2 === 0 ? formattedDisplayValue : undefined)
       : formattedDisplayValue;
 
-    const computedDisplayValueWidth = bboxCalculator.compute(displayValueText || '', fontSize, fontFamily).getOrElse({
+    const computedDisplayValueWidth = bboxCalculator.compute(displayValueText || '', padding, fontSize, fontFamily)
+                                                    .getOrElse({
```
emmacunningham: this is a kind of weird indentation given the following lines.  can we align the lines below so it's clearer that they belong to this?

src/lib/axes/canvas_text_bbox_calculator.ts
```diff
@@ -17,7 +17,7 @@ export class CanvasTextBBoxCalculator implements BBoxCalculator {
     this.attachedRoot.appendChild(this.offscreenCanvas);
     this.scaledFontSize = scaledFontSize;
   }
-  compute(text: string, fontSize = 16, fontFamily = 'Arial', padding: number = 1): Option<BBox> {
+  compute(text: string, padding: number, fontSize = 16, fontFamily = 'Arial'): Option<BBox> {
```
emmacunningham: @markov00 What do you think about doing some validation to prevent negative padding and if the supplied padding argumen is less than 1, set the padding to 1 instead?  And we could also surface this in our configuration warnings once we have those.  We are already accounting for padding in the computation of the label and if the user specifies a negative padding, then the text box would not be wide enough for the label.

src/lib/series/rendering.ts
rshen91: Good catch - included in commit 929b0ac thanks!

stories/axis.tsx
```diff
@@ -91,10 +95,11 @@ storiesOf('Axis', module)
           showOverlappingTicks={true}
           tickLabelRotation={number('bottom axis tick label rotation', 0, {
             range: true,
-            min: -90,
+            min: -900,
```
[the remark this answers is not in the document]
rshen91: Sorry about that! Updated in commit 929b0ac

src/components/react_canvas/axis.tsx
```diff
@@ -29,7 +29,12 @@ export class Axis extends React.PureComponent<AxisProps> {
     return this.renderAxis();
   }
   renderTickLabel = (tick: AxisTick, i: number) => {
-    const { padding, ...labelStyle } = this.props.chartTheme.axes.tickLabelStyle;
+// padding is already being computed through bbox_calculator using tickLabelPadding, set padding to 0 to avoid conflict
```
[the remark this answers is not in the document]
rshen91: Let me know what you think about the longer comment in commit 929b0ac for this, thanks!

src/lib/series/specs.ts
rshen91: Thanks! Addressed in commit 0f39133

stories/styling.tsx
rshen91: Oh! Addressed in commit 0f39133

src/lib/axes/canvas_text_bbox_calculator.ts
markov00: Yes I'm ok to add some validation here to prevent negative padding.

src/lib/themes/light_theme.ts
```diff
@@ -91,7 +91,7 @@ export const LIGHT_THEME: Theme = {
       fontFamily: `'Open Sans', Helvetica, Arial, sans-serif`,
       fontStyle: 'normal',
       fill: 'gray',
-      padding: 0,
+      padding: 1,
```
markov00: You missed to update also the dark theme with the same padding

rshen91: Thank you! Added commit edb1f4b to address

src/lib/axes/canvas_text_bbox_calculator.ts
rshen91: Added commit edb1f4b to address

markov00:
jenkins test this please

markov00:
@emmacunningham what do you think if we extend this PR to support custom axis styles?
because what is the meaning of having this padding as a separate spec property?
What if we extend this to have something that accept custom style for axis like:

```
<Axis
  id={getAxisId('bottom')}	        
  position={Position.Bottom}	        
  title={'Bottom axis'}	          
  showOverlappingTicks={true}
  style={{
    tickLabelStyle: {
      padding: 10
    }
  }}
/>	 
```

src/lib/axes/canvas_text_bbox_calculator.ts
```diff
@@ -17,11 +17,16 @@ export class CanvasTextBBoxCalculator implements BBoxCalculator {
...
     if (!this.context) {
       return none;
     }
 
+    // Avoid having negative padding that can obscure text
```
nickofthyme: I'd say generally comments as simple as this can be left for tests. Meaning you should create test cases that validate this line of code. That way if someone removes these lines of code, the tests break.

Not saying you can't leave comments, but do so for more complex code that requires explanation.

rshen91: Ok! I think I wrote one that gets to the issue in commit `1507d9c`. Let me know what you think thanks!

src/lib/axes/canvas_text_bbox_calculator.ts
rshen91: test added in 1507d9c to address

src/lib/axes/canvas_text_bbox_calculator.ts
nickofthyme: FYI Github will auto link and format the commit hash when you just put the hash or short hash like this 1507d9c but when you wrap the hash in backticks, or anything else for that matter, it doesn't create the link like this `1507d9c`. Here is a link to all the url fun https://help.github.com/en/articles/autolinked-references-and-urls

You can also paste code snippets by selecting code lines, use shift+click for multiple lines, and select `copy permalink`

image

Then you will get something like this...
https://github.com/elastic/elastic-charts/blob/f2b90dfbf8bb47387d3078118af9ae9bf44a563b/package.json#L14-L16

> Just FYI - It took me a while to realize how many helpful tools GH has for our use.

src/lib/axes/axis_utils.test.ts
```diff
@@ -1272,4 +1272,15 @@ describe('Axis computational utils', () => {
     expect(isBounded(lowerBounded)).toBe(true);
     expect(isBounded(upperBounded)).toBe(true);
   });
+  test('should not allow negative padding', () => {
```
nickofthyme: Nice 🎉

src/lib/axes/canvas_text_bbox_calculator.ts
rshen91: Just to confirm - should we set the padding to 1 when it's specified by the user at 0? Or should the validation only catch negative values? Thanks!

emmacunningham:
@rshen91 let's chat on zoom about how to do this.

> @emmacunningham @rshen91 what do you think if we extend this PR to support custom axis styles?
> because what is the meaning of having only the padding as a separate spec property here?
> What if we extend this to have something that accept custom style for axis like:
> 
> ```
> <Axis
>   id={getAxisId('bottom')}	        
>   position={Position.Bottom}	        
>   title={'Bottom axis'}	          
>   showOverlappingTicks={true}
>   style={{
>     tickLabelStyle: {
>       padding: 10
>     }
>   }}
> />	 
> ```

src/lib/axes/axis_utils.ts
```diff
@@ -70,7 +70,10 @@ export function computeAxisTicksDimensions(
     throw new Error(`Cannot compute scale for axis spec ${axisSpec.id}`);
   }
 
-  const tickLabelPadding = axisSpec.tickLabelPadding || axisConfig.tickLabelStyle.padding;
+  const tickLabelPadding =
+    axisSpec.style && axisSpec.style.tickLabelPadding
+      ? axisSpec.style.tickLabelPadding
+      : axisConfig.tickLabelStyle.padding;
```
emmacunningham: Can we rename this parameter name to `themeAxisConfig` to make it clearer where this is coming from?  I think eventually we'll want to think about how to better merge spec-specific and theme-general styles but for now, would be nice to update the parameter name.

src/lib/axes/canvas_text_bbox_calculator.ts
rshen91: improved comment in commit d8dd8f4

src/lib/axes/axis_utils.ts
rshen91: Thank you for helping me with this! The commit with the changes in naming and improving axis_utils test is addressed in commit d8dd8f4

src/lib/axes/axis_utils.ts
```diff
@@ -82,6 +87,13 @@ export function computeAxisTicksDimensions(
   };
 }
 
+export function getAxisTickLabelPadding(axisConfigTickLabelPadding: number, axisSpecStyle?: AxisStyle): number {
+  if (axisSpecStyle && axisSpecStyle.tickLabelPadding) {
```
markov00: I think you should check if `tickLabelPadding` is not undefined here, because checked like this can lead to the following:
a user wrongly specify a tickLabelPadding  to 0 (meaning I want the minimum padding for my labels) this `if` result than in a false statement and we then use the configured axis tick label padding, that can be a larger number.

src/lib/series/specs.ts
```diff
@@ -218,10 +218,17 @@ export interface AxisSpec {
...
+  /** Object to hold custom styling */
+  style?: AxisStyle;
 }
 
 export type TickFormatter = (value: any) => string;
 
+export interface AxisStyle {
+  /** Specifies the amount of padding on the tick label bounding box */
+  tickLabelPadding?: number;
+}
```
markov00: What if we start using a similar structure as the AxisConfiguration?
```
export interface AxisStyle {
  tickLabelStyle: {
    padding: number;
  }
}
```
This will allows us to slowly add more custom style on each single axis

src/lib/axes/axis_utils.ts
rshen91: Ah ok I think I'm following - basically reverse the conditionals? Let me know what you think about commit 83fcc7d or if I misunderstood. Thanks!

src/lib/axes/canvas_text_bbox_calculator.ts
rshen91: Clarified in zoom with Emma about this 👌

src/lib/axes/axis_utils.ts
markov00: I think it's more or less the same because `!axisSpecStyle.tickLabelPadding` is `true` if `tickLabelPadding` is 0.
So you should check it like that:
```
  if (axisSpecStyle && axisSpecStyle.tickLabelPadding !== undefined) {
```

rshen91: Oh!! Ok thank you for clarifying! Commit 725177f should include this improvement

src/lib/series/specs.ts
rshen91: Opened issue Refactor AxisStyle object to include custom styles for AxisSpec #260 for this change and also to address migrating other props to this object. Thank you!

Issue #102, opened 2019-03-19
Hiding a single series from the legend results in a wrong visualization


markov00:
**Describe the bug**
When you have a single series chart with legend, clicking on the eye to hide the series results in a wrongly represented chart as shown in screenshot

**To Reproduce**
1. Go to `Bar Chart with axis and legend` story
2. Click on  the eye on the legend

**Expected behavior**
We can add a clear message displaying something like: no data to show as in the EuiSeriesChart or we can disable any hide/show eye if we have one single series on the chart.

**Screenshots**
<img width="472" alt="Screenshot 2019-03-19 at 16 02 11" src="https://user-images.githubusercontent.com/1421091/54616737-6e56aa00-4a60-11e9-99f4-2f735996d9a0.png">

**Version (please complete the following information):**
 - OS: Mac OS 10.14.3
 - Browser: chrome
 - Elastic Charts: 3.1.1/master

**Additional context**
n/a

**Errors in browser console**

```
Konva warning: "points" attribute has non numeric element Infinity. Make sure that all elements are numbers.
Konva warning: Infinity is a not valid value for "x" attribute. The value should be a number.
```

**Kibana Cross Issues**
n/a

**Checklist**
- [x] every related Kibana issue is listed under `Kibana Cross Issues` list
- [x] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

PR #219, opened 2019-05-28
feat: add isChartEmpty

rshen91:
## Summary

Closes https://github.com/elastic/elastic-charts/issues/102

Instead of rendering the chart, this PR will show a message to the end consumer telling them they have no data to display

PR219

### Checklist

- [X] This was checked for cross-browser compatibility, [including a check against IE11]
PR102


src/state/chart_state.ts
```diff
@@ -806,8 +812,10 @@ export class ChartStore {
       this.deselectedDataSeries,
     );
 
-    // tslint:disable-next-line:no-console
-    // console.log({ legendItems: this.legendItems });
+    this.isChartEmpty = true;
+    this.legendItems.forEach((val, key) => {
+      this.determineEmptyChart(val);
+    });
```
emmacunningham: We can break early on the first legend item that has `isSeriesVisible` set to true if we use the `for...of` syntax.

```
  for (const [, legendItem] of this.legendItems) {
    if (legendItem.isSeriesVisible) {
        // set something to true
        break;
    }
  }
```

Further, can we move this entire function to `utils` and just assign `this.isChartEmpty` to the return value of that function call?  Doing so eliminates the need for a `let` and allows you to just return early at the first item with `isSeriesVisible === true`.   

That is, in `utils`:

```
export function isAllSeriesDeselected(legendItems: Map<string, LegendItem>): boolean {
  for (const [, legendItem] of this.legendItems) {
    if (legendItem.isSeriesVisible) {
        return false;
    }
  }
  return true;
}
```

And then back in `chart_state` you can delete the part where you iterate over `this.legendItems` and: 

```
this.isChartEmpty = isAllSeriesDeselected(this.legendItems);
```

src/state/chart_state.ts
```diff
@@ -806,8 +807,7 @@ export class ChartStore {
       this.deselectedDataSeries,
     );
 
-    // tslint:disable-next-line:no-console
-    // console.log({ legendItems: this.legendItems });
```
emmacunningham: Can we remove these comments?

src/lib/series/specs.ts
```diff
@@ -14,7 +14,7 @@ import { DataSeriesColorsValues } from './series';
 
 export type Datum = any;
 export type Rotation = 0 | 90 | -90 | 180;
-export type Rendering = 'canvas' | 'svg';
+export type Rendering = 'canvas' | 'svg' | 'empty';
```
emmacunningham: Can we remove `'empty'`?

src/components/react_canvas/reactive_chart.tsx
```diff
@@ -332,9 +332,16 @@ class Chart extends React.Component<ReactiveChartProps, ReactiveChartState> {
...
     } = this.props.chartStore!;
-
     // disable clippings when debugging
+    if (isChartEmpty) {
+      return (
+        <div>
+          There is no data to display
+        </div>
+      );
+    }
```
emmacunningham: Can we style this a bit here?  We can match what they have currently in EUI (but please note that we shouldn't pull any EUI components in but rather copy/paste any components we may need).  I think for now it's ok to just have the below without the icon (with the text centered within the containing element).

image

emmacunningham: Probably should move the comment about clippings being disabled to the more relevant section below.

src/lib/series/legend.ts
```diff
@@ -26,19 +26,20 @@ export function computeLegend(
   deselectedDataSeries?: DataSeriesColorsValues[] | null,
 ): Map<string, LegendItem> {
   const legendItems: Map<string, LegendItem> = new Map();
-
+  // export let isChartNotEmpty: number = 0;
```
emmacunningham: This isn't used anymore, can you remove?

src/lib/series/legend.ts
```diff
@@ -26,19 +26,20 @@ export function computeLegend(
...
   const sortedSeriesColors = getSortedDataSeriesColorsValuesMap(seriesColor);
+  // tslint:disable-next-line:prefer-const
+  // let isChartNotEmptyResult: number = 0;
```
emmacunningham: Can you remove these lines that aren't used?

src/lib/series/legend.ts
```diff
@@ -26,19 +26,20 @@ export function computeLegend(
...
 
   sortedSeriesColors.forEach((series, key) => {
     const spec = specs.get(series.specId);
-
     const color = seriesColorMap.get(key) || defaultColor;
     const hasSingleSeries = seriesColor.size === 1;
     const label = getSeriesColorLabel(series.colorValues, hasSingleSeries, spec);
+    // isChartEmpty(seriesColor, isChartNotEmptyResult, deselectedDataSeries);
```
emmacunningham: Can you remove these lines that aren't used?

src/lib/series/legend.ts
```diff
@@ -26,19 +26,20 @@ export function computeLegend(
...
     const isSeriesVisible = deselectedDataSeries
-      ? findDataSeriesByColorValues(deselectedDataSeries, series) < 0
-      : true;
-
+    ? findDataSeriesByColorValues(deselectedDataSeries, series) < 0
+    : true;
```
emmacunningham: Can you match the previous indentation?  If your editor does some weird format on save, we can look into it but the original indentation should be preserved.

src/lib/series/rendering.test.ts
```diff
@@ -99,7 +99,6 @@ describe('Rendering utils', () => {
     // no highlighted elements
     expect(defaultStyle).toEqual({ opacity: 1 });
 
-
```
emmacunningham: Can we remove this deleted line?  There are no other changes in this file so it would be nice to not include this in the PR.

src/state/utils.test.ts
```diff
@@ -810,4 +812,24 @@ describe('Chart State utils', () => {
...
+    });
+    legendItems1.set('specId:{bars},colors:{b}', {
+      key: 'specId:{bars},colors:{b}',
+      color: '#2B70F7',
+      label: 'b',
+      value: { specId: getSpecId('bars'), colorValues: ['b'], lastValue: 2 },
+      displayValue: { raw: 2, formatted: '2.00'},
+      isSeriesVisible: false,
+    });
+    expect(isAllSeriesDeselected(legendItems1)).toBe(true);
```
emmacunningham: Can you also test that the function returns false as expected?

src/state/chart_state.ts
rshen91: Ok I think these might be removed already

src/state/utils.test.ts
rshen91: Good point - added in commit bb29ec7

src/lib/series/legend.ts
rshen91: Yikes - I think this commit should address this bb29ec7

src/lib/series/legend.ts
rshen91: Yep - thanks for catching it  commit bb29ec7 should have this covered

src/lib/series/legend.ts
rshen91: Fixed in commit bb29ec7 thanks!

src/components/react_canvas/reactive_chart.tsx
rshen91: I think I'm following - attempted styling in commit bb29ec7

src/lib/series/specs.ts
rshen91: Good catch! Sorry about that - fixed in bb29ec7

src/lib/series/legend.ts
rshen91: I think these should be gone in commit bb29ec7. Please let me know if I'm missing something. thanks!

src/components/react_canvas/reactive_chart.tsx
emmacunningham: Ah, sorry if my previous comment was unclear.  I meant that the styling should be handled by CSS (in an external `.scss` file).  It might be easiest to make this its own file to avoid conflicts with the removal of EUI in https://github.com/elastic/elastic-charts/pull/208.

Inline styles in general should be avoided unless they are based on computed values.

Per the convention mentioned in #208, all classes in elastic-charts should be namespaced with `.ech`.

And we should not be using any `.eui` namespaced classes as they belong to EUI and we do  not use any EUI classes or components in elastic-charts.

So, to move forward, what you would want to do:
- remove the inline style on the `span`
- create a new `.scss` file (it can be a sibling to the ones already in `/components`).  we can call it `unavailable_chart.scss`
- you'll need to add the new file in `_chart.scss`
- write classes with the `.ech` namespace using BEM convention (which you can read about in the `SASS best practices` section in the [EUI docs](https://elastic.github.io/eui/#/guidelines/sass)).

In the updated version, you're using a `span` followed by a `p`; the top probably doesn't need to be a span.  Instead, I'd recommend: 

```
<div>
  <h3>Chart not available</h3>
  <h4>Missing data</h4>
</div>
```

We can add basic vertical and horizontal centering for the child elements of the parent div.

Finally, it occurs to me that we probably need to have that since there's text here, this will raise accessibility and localization issues.  Will open that up in a separate ticket as the styling and logic won't change, but we may need to expose some props at least for localization.

emmacunningham: Oh one another thing, can we change the text a little bit?  `Chart not available` is a bit misleading; it's more like in this case there's no data to show because no series are selected.

src/components/legend.tsx
```diff
@@ -70,7 +70,7 @@ class LegendComponent extends React.Component<ReactiveChartProps> {
             className="elasticChartsLegendListContainer"
             responsive={false}
           >
-            {[...legendItems.values()].map((item) => {
+            {[...legendItems.values()].map((item, key) => {
```
nickofthyme: is there any `key` in the `item` we can use rather than the index?

src/components/react_canvas/reactive_chart.tsx
```diff
@@ -332,8 +332,17 @@ class Chart extends React.Component<ReactiveChartProps, ReactiveChartState> {
       chartRotation,
       chartTransform,
       setCursorPosition,
+      isChartEmpty,
     } = this.props.chartStore!;
 
+    if (isChartEmpty) {
+      return (
+        <div className="euiEmptyPrompt euiSeriesChartContainer__emptyPrompt">
+          {<span className="euiTitle euiTitle--large" style={{paddingBottom: '1.2rem'}}>Chart not available</span>}
```
nickofthyme: I don't think you need the `{}` around the template code

```suggestion
          <span className="euiTitle euiTitle--large" style={{paddingBottom: '1.2rem'}}>Chart not available</span>
```

src/components/react_canvas/reactive_chart.tsx
```diff
@@ -332,8 +332,17 @@ class Chart extends React.Component<ReactiveChartProps, ReactiveChartState> {
...
+          {<p className="euiTextColor euiTextColor--subdued">There is no data to display</p>}
```
nickofthyme: Same here...
```suggestion
          <p className="euiTextColor euiTextColor--subdued">There is no data to display</p>
```

src/lib/series/rendering.test.ts
nickofthyme: We can enforce this and clean up existing with an eslint rule in the future.

src/components/react_canvas/reactive_chart.tsx
rshen91: Thank you! I added commit ae929db

src/components/react_canvas/reactive_chart.tsx
rshen91: Thanks! I've revised this a bit in commit ae929db to avoid bringing in EUI styling in this part. Let me know what you think

src/components/react_canvas/reactive_chart.tsx
rshen91: I think the changes in commit ae929db should resolve this - thanks!

src/components/react_canvas/reactive_chart.tsx
```diff
@@ -337,9 +337,8 @@ class Chart extends React.Component<ReactiveChartProps, ReactiveChartState> {
 
     if (isChartEmpty) {
       return (
-        <div className="euiEmptyPrompt euiSeriesChartContainer__emptyPrompt">
-          {<span className="euiTitle euiTitle--large" style={{paddingBottom: '1.2rem'}}>Chart not available</span>}
-          {<p className="euiTextColor euiTextColor--subdued">There is no data to display</p>}
+        <div className="elasticChartsUnavailable">
+          {<h3>No series selected to display</h3>}
```
nickofthyme: Ya same thing applies here. The beauty of jsx (or tsx) is that it is basically html but allows you to add js within the `{}`.

In this case this part here...

```html
<h3>No series selected to display</h3>
```

is all html so you don't need to wrap the html in curly braces.

You only need to use the curly braces when you are using javascript in jsx for example showing a component based on a boolean...

```jsx
<div>
  { this.showTitle : <TitleComponent /> ? null }
</div>
```

nickofthyme:
```suggestion
          <h3>No series selected to display</h3>
```

rshen91: Ah! Thanks! Changed in commit dd019cb

src/components/legend.tsx
rshen91: Thanks for jumping on a zoom to explain - I hope commit dd019cb will resolve this 💪

src/components/_unavailable_chart.scss
```diff
@@ -0,0 +1,4 @@
+.elasticChartsUnavailable {
```
nickofthyme: One last thing...classnames are generally dash-case

markov00:
jenkins test this please

src/components/_unavailable_chart.scss
emmacunningham: we use the `.ech` namespace for classes.  see https://github.com/elastic/elastic-charts/pull/208 for more detail on the conventions we're using for styling.

src/components/_unavailable_chart.scss
```diff
@@ -0,0 +1,4 @@
...
+  text-align: center;
+  padding: 150px 0;
```
emmacunningham: is the padding here for the vertical centering?  i would recommend finding a different way to center the text vertically here (flexbox should work pretty well).  

one reason to avoid using a fixed amount of padding here is that this assumes the height of the chart will be at least 300px + the text height, which is not an assumption we can make, especially in the cases where the user is creating a sparklines chart.

src/components/react_canvas/reactive_chart.tsx
```diff
@@ -349,8 +349,16 @@ class Chart extends React.Component<ReactiveChartProps, ReactiveChartState> {
       chartRotation,
       chartTransform,
       setCursorPosition,
+      isChartEmpty,
     } = this.props.chartStore!;
 
+    if (isChartEmpty) {
+      return (
+        <div className="elasticChartsUnavailable">
```
emmacunningham: see above about namespace convention for class names

src/components/_unavailable_chart.scss
rshen91: Ok! I think I'm following to get the text to the middle of the chart I did use the em css units instead of px in commit 0a4552d

src/components/_unavailable_chart.scss
```diff
@@ -0,0 +1,7 @@
+.echUnavailable {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  padding-top: 10em;
```
markov00: You can get rid of that. If the chart is too small, the padding will push the title on the bottom.
To use correctly flex you should add a size to the div. It can be done just adding the width and height to 100%
```
    width: 100%;
    height: 100%;
```

src/components/_unavailable_chart.scss
```diff
@@ -0,0 +1,7 @@
...
+}
```
markov00: I think we should also specify the h3 style using EUI

with 
```css
.echUnavailable {
   h3 {
     @include euiText;
    }
}
```
@cchaos  what do you think?

src/components/react_canvas/reactive_chart.tsx
```diff
@@ -349,8 +349,16 @@ class Chart extends React.Component<ReactiveChartProps, ReactiveChartState> {
       chartRotation,
       chartTransform,
       setCursorPosition,
+      isChartEmpty,
     } = this.props.chartStore!;
 
+    if (isChartEmpty) {
+      return (
+        <div className="echUnavailable">
+          <h3>No series selected to display</h3>
```
cchaos: Try not to use actual header level tags here. For accessibility, it's vital that the hierarchy of h1-h6 is correct within the DOM order and being that this is a plugin library, you can't really be sure where this component will fall within the DOM order. Best to just leave this is a `<p>` tag.

src/components/react_canvas/reactive_chart.tsx
```diff
@@ -349,8 +349,16 @@ class Chart extends React.Component<ReactiveChartProps, ReactiveChartState> {
       chartRotation,
       chartTransform,
       setCursorPosition,
+      isChartEmpty,
     } = this.props.chartStore!;
 
+    if (isChartEmpty) {
+      return (
+        <div className="echUnavailable">
```
cchaos: You'll want to be more specific with your class naming so that you don't accidentally create too generic of a class that it's not easily recognizable as to where the class is applied and/or that it doesn't get overridden. So for this component, you'd likely start with the exported class name and append `unavailable` as a child style like:

```suggestion
        <div className="echReactiveChart__unavailable">
```

src/components/react_canvas/reactive_chart.tsx
cchaos: I would just make this sentence more generic like:
```suggestion
          <p>No data to display</p>
```

src/components/_unavailable_chart.scss
cchaos: Actually, I would just do:
 ```scss
{
  @include euiFontSizeXS;
  color: $euiColorDarkShade;
}
```

This will match the dashboard example's font size and color.

File: src/components/react_canvas/reactive_chart.tsx (hunk 20 above)
Comment (rshen91) [reply src/components/react_canvas/reactive_chart.tsx]: Hi @cchaos thanks for the feedback! Let me know what you think about the changes in commit 6719484 

For instance, I wasn't positive about adding the chart icon over or just keeping it to the text message. I'm open to what you think might make the most sense. Thanks!

Comment (cchaos) [reply src/components/react_canvas/reactive_chart.tsx]: Agreed, it doesn't need any icon (at least not at the moment).

File: src/components/_unavailable_chart.scss [hunk 22]
```diff
@@ -0,0 +1,13 @@
+.echReactiveChart_unavailable {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 100%;
+  height: 100%;
+}
+
+.echReactiveChart_unavailable p {
```
cchaos: You'll also want to stay away from generic selectors like `p` here (in case that element ever changes in the future, you won't also have to change the CSS). It can just be removed here completely because the paragraph will inherit from it's parent.

rshen91: Thank you! I think I improved this in commit 64c1844. Let me know what you think!

emmacunningham: For clarification, the commit where the generic selector was removed is here: https://github.com/elastic/elastic-charts/pull/219/commits/7c3ab4f002c4a8739e49e7ccc36fff33bb5f52c3

The commit linked to above is the first pass at handling the cases where there's a legend visible; the refactored version, which may help with readability, is here https://github.com/elastic/elastic-charts/pull/219/commits/64c18445620ec0bc34b7ebd61d59daf12871eb12

src/components/react_canvas/reactive_chart.tsx
```diff
@@ -348,9 +348,35 @@ class Chart extends React.Component<ReactiveChartProps, ReactiveChartState> {
...
+      const { verticalWidth, horizontalHeight } = chartTheme.legend;
+
+      const marginStyle =
+        legendPosition && isVertical(legendPosition)
+          ? legendPosition === Position.Right
+            ? { marginLeft: -verticalWidth }
+            : { marginLeft: verticalWidth }
+          : legendPosition === Position.Top
+          ? { marginTop: horizontalHeight }
+          : { marginTop: -horizontalHeight };
```
cchaos: There is a bit of an issue with the margin approach. Margins don't actually shrink the contents of children. They will still grow to be 100% of the width of it's parent, it just gets pushed in one direction. So you could have a small chart where the legend takes up most of the space and get this message to be pushed out of the container:

Screen Shot 2019-07-09 at 11 07 16 AM

What you'll actually want to do is add this value as **padding** to the `.echReactiveChart_unavailable` element to contain the contents within:

Screen Shot 2019-07-09 at 11 12 18 AM

rshen91: Thank you @cchaos for explaining - I have added commit 0209ed3 to affect the padding vs the margins. Let me know if I'm misunderstanding anything. Thank you!

PR #220, opened 2019-05-29
fix(crosshair): adjust band position for rotation

emmacunningham:
## Summary

This PR introduces a fix so that the band highlighter position appears correctly positioned when chartRotation of -90 or 180 is defined.  Previously, these flipped rotations were not being accounted for in computing the position, causing the bands to appear on the wrong side of the chart.  (This is most noticeable with series which have discontinuous data.)

Previous bug:
band_rotation_bug

After implementing fix:
band_fix

### Checklist

### DIFF
--- a/src/state/crosshair_utils.linear_snap.test.ts
+++ b/src/state/crosshair_utils.linear_snap.test.ts
@@ -454,7 +454,7 @@ describe('Crosshair utils linear scale', () => {
           1,
         );
         expect(bandPosition).toEqual({
-          left: 119,
+          left: -1,
           top: 0,
           height: 100,
           width: 1,
@@ -472,7 +472,7 @@ describe('Crosshair utils linear scale', () => {
           1,
         );
         expect(bandPosition).toEqual({
-          left: 119,
+          left: -1,
           top: 0,
           height: 100,
           width: 1,
@@ -490,7 +490,7 @@ describe('Crosshair utils linear scale', () => {
           1,
         );
         expect(bandPosition).toEqual({
-          left: 79,
+          left: 39,
           top: 0,
           height: 100,
           width: 1,
@@ -508,7 +508,7 @@ describe('Crosshair utils linear scale', () => {
           1,
         );
         expect(bandPosition).toEqual({
-          left: 29,
+          left: 89,
           top: 0,
           height: 100,
           width: 1,
@@ -851,7 +851,7 @@ describe('Crosshair utils linear scale', () => {
         );
         expect(bandPosition).toEqual({
           left: 0,
-          top: 99,
+          top: -1,
           height: 1,
           width: 120,
         });
@@ -869,7 +869,7 @@ describe('Crosshair utils linear scale', () => {
         );
         expect(bandPosition).toEqual({
           left: 0,
-          top: 99,
+          top: -1,
           height: 1,
           width: 120,
         });
@@ -887,7 +887,7 @@ describe('Crosshair utils linear scale', () => {
         );
         expect(bandPosition).toEqual({
           left: 0,
-          top: 59,
+          top: 39,
           height: 1,
           width: 120,
         });
@@ -905,7 +905,7 @@ describe('Crosshair utils linear scale', () => {
         );
         expect(bandPosition).toEqual({
           left: 0,
-          top: 9,
+          top: 89,
           height: 1,
           width: 120,
         });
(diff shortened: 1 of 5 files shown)


src/state/crosshair_utils.ts
```diff
@@ -76,7 +77,22 @@ export function getCursorBandPosition(
...
   const isHorizontalRotated = isHorizontalRotation(chartRotation);
-  const invertedValue = xScale.invertWithStep(isHorizontalRotated ? x : y, data);
+  const xAxisCursorPosition = getValidXPosition(
+    x,
+    y,
+    chartRotation,
+    chartDimensions,
+  );
+
+  const yAxisCursorPosition = getValidXPosition(
```
markov00: Isn't this equal to the xAxisCursorPosition? seems that you are calling two time `getValidXPosition` on that line and on line 80.

src/state/crosshair_utils.ts
```diff
@@ -76,7 +77,22 @@ export function getCursorBandPosition(
...
+      x,
+      y,
+      chartRotation,
+      chartDimensions,
+    );
+
+  const invertedValue = xScale.invertWithStep(isHorizontalRotated ? xAxisCursorPosition : yAxisCursorPosition, data);
+
```
markov00: Of my previous comment is correct you should remove the `isHorizontalRotated` check here and use the only xAxisCursorPosition.

stories/bar_chart.tsx
```diff
@@ -238,8 +238,43 @@ storiesOf('Bar Chart', module)
...
+      0,
+    );
+
+    const theme = {
+      ...LIGHT_THEME,
+      chartMargins: {
+        left: 30,
+        right: 0,
+        top: 30,
+        bottom: 0,
```
markov00: why are you setting right and bottom margin to 0? on a first look, on storybook seems a bug, but I then look at the code and found those at 0. Are you going to check anything in particular?

stories/bar_chart.tsx
```diff
@@ -484,9 +519,44 @@ storiesOf('Bar Chart', module)
...
```
markov00: Same comment as before

src/state/crosshair_utils.linear_snap.test.ts
```diff
@@ -454,7 +454,7 @@ describe('Crosshair utils linear scale', () => {
           1,
         );
         expect(bandPosition).toEqual({
-          left: 0,
+          left: -1,
```
markov00: Are we sure that those changes in test values are fine? I think we should not have negative values on that computation, can you change those for the root cause? I've also seen a -21 value, is it correct?

src/state/crosshair_utils.linear_snap.test.ts
```diff
@@ -472,7 +472,7 @@ describe('Crosshair utils linear scale', () => {
...
```
nickofthyme: @emmacunningham I see now why you want property based testing so badly!

stories/bar_chart.tsx
emmacunningham: Was initially using these values to help with debugging but they're no longer need.  Have removed bee17510a49ac965e445434cd64595ac2d4c30ac.

stories/bar_chart.tsx
emmacunningham: Was initially using these values to help with debugging but they're no longer need.  Have removed bee17510a49ac965e445434cd64595ac2d4c30ac.

src/state/crosshair_utils.ts
emmacunningham: Have simplified since this is called in setCursorPosition in chart_state where the adjusted x & y are already computed and using those as the coordinate values, which also addresses this f13bded93801b8c53c1e44c70b490da4b3493f91

src/state/crosshair_utils.ts
emmacunningham: See f13bded93801b8c53c1e44c70b490da4b3493f91

src/state/crosshair_utils.linear_snap.test.ts
emmacunningham: I've changed the logic so that the cursor position values coming into this function are the normalized cursor position values, which helps with some reason around the expected values.  After the refactor, there was still one test that was failing with a negative value due to setting `DEFAULT_SNAP_POSITION_BAND` to 1 when we offset by the band position in the test; now the offset is only applied iff the scale has a bandwidth greater than 0.

Crazybus:
jenkins test this please

Issue #109, opened 2019-03-20
Remove debounce on resize for first render


markov00:
**Describe the bug**
There is a 200ms delay before you can get anything on the screen.
It's because the `chart-resizer`, that compute the chart parent size, has a `debounce` set to 200ms for any resize event. 

**To Reproduce**
1. Go to any chart
2. See the small delay between the page load and the rendering

**Expected behavior**
Rendering happens at the same exact time of rendering other contents

**Screenshots**
n/a

**Version (please complete the following information):**
 - OS: mac osx
 - Browser: chrome
 - Elastic Charts: master

**Additional context**
n/a

**Errors in browser console**
n/a

**Kibana Cross Issues**
n/a

**Checklist**
- [x] every related Kibana issue is listed under `Kibana Cross Issues` list
- [x] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

nickofthyme:
Closed by #229

Issue #75, opened 2019-02-21
Refactor computeChart function in chart state to be more functional and not recompute on every change


emmacunningham:
**Is your feature request related to a problem? Please describe.**
Our current `computeChart` function is massive and produces many side effects, which makes it hard to write tests for.  Further, any time `computeChart` is called, _all_ computations are run again even if certain values have remained the same and may not need to be re-computed.

**Describe the solution you'd like**
We can consider reducing the side effects produced by this function and reduce the size of the function (moving logic and computations into additional functions which can be themselves tested).  We can also consider ways to address partial (and not full) re-computing (memoization, transducers to chain transformations, etc.).

**Describe alternatives you've considered**

**Additional context**
n/a

**Kibana Cross Issues**
Add any [Kibana](https://github.com/elastic/kibana) related issues/feature requests here.

n/a

**Checklist**

- [x] this request is checked against already exist requests
- [x] every related Kibana issue is listed under `Kibana Cross Issues` list
- [x] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

PR #229, opened 2019-06-05
fix(chart_resizer): debounce resize set to leading

nickofthyme:
## Summary

Fix for `ChartResizer` initial render delay. A 200ms delay was applied to all initial renders.

Set debounce to `leading` to prevent initial load delay of 200ms.

Issue #109

### Before
Screen Recording 2019-06-05 at 05 14 PM

### After
Screen Recording 2019-06-05 at 05 17 PM

### Checklist

- [ ] ~~Any consumer-facing exports were added to `src/index.ts` (and stories only import from `../src` except for test data & storybook)~~
- [ ] ~~Proper documentation or storybook story was added for features that require explanation or tutorials~~

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,7 +129,8 @@
     "react-dom": "^16.8.3",
     "react-konva": "16.8.3",
     "react-spring": "^8.0.8",
-    "resize-observer-polyfill": "^1.5.1"
+    "resize-observer-polyfill": "^1.5.1",
+    "ts-debounce": "^1.0.0"
   },
   "peerDependencies": {
     "@elastic/datemath": "^5.0.2",
--- a/src/components/chart_resizer.tsx
+++ b/src/components/chart_resizer.tsx
@@ -14,7 +14,7 @@ class Resizer extends React.Component<ResizerProps> {
   constructor(props: ResizerProps) {
     super(props);
     this.containerRef = React.createRef();
-    this.ro = new ResizeObserver(debounce(this.onResize, 200));
+    this.ro = new ResizeObserver(debounce(this.onResize, 200, { leading: true }));
   }
 
   componentDidMount() {
--- a/src/components/chart_resizer.tsx
+++ b/src/components/chart_resizer.tsx
@@ -1,20 +1,29 @@
-import debounce from 'lodash/debounce';
 import { inject, observer } from 'mobx-react';
 import React, { RefObject } from 'react';
 import ResizeObserver from 'resize-observer-polyfill';
+import { debounce } from 'ts-debounce';
 import { ChartStore } from '../state/chart_state';
 
 interface ResizerProps {
   chartStore?: ChartStore;
 }
-class Resizer extends React.Component<ResizerProps> {
+interface ResizerState {
+  initialResizeComplete: boolean;
+}
+class Resizer extends React.Component<ResizerProps, ResizerState> {
+  state = {
+    initialResizeComplete: false,
+  };
+
   private containerRef: RefObject<HTMLDivElement>;
   private ro: ResizeObserver;
+  private onResizeDebounced: (entries: ResizeObserverEntry[]) => void;
 
   constructor(props: ResizerProps) {
     super(props);
     this.containerRef = React.createRef();
-    this.ro = new ResizeObserver(debounce(this.onResize, 200, { leading: true }));
+    this.onResizeDebounced = debounce(this.onResize, 200);
+    this.ro = new ResizeObserver(this.handleResize);
   }
 
   componentDidMount() {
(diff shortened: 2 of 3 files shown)


Crazybus:
jenkins test this please

nickofthyme:
@markov00 Ya I see your point. I was thinking this solution was too good to be true. I'll try your suggestion.

nickofthyme:
@markov00 Your suggestion worked! The implementation doesn't seem so elegant with having a state variable to track the first render and a separate function to handle the resize. If you have any other ideas of how to implement it better let me know. I think if this comes up more we may want to write our own debounce function to have this option.

## Demo
Screen Recording 2019-06-06 at 10 14 AM

For some reason, resizing the browser window _sometimes_, not always, call the `computeChart` method twice still. Any ideas? Seems to just be my computer struggling to keep up.

Also to your point about calling `computeChart` twice. I noticed that on initial load or refresh that method is called twice. See console traces below 👇. But only the second call, the one from `onResize`, renders the chart to the page. Is that expected?

<details><summary>First call to `computeChart` looks to be from specs_parser</summary>
<p>

```bash
push../src/state/chart_state.ts.ChartStore.computeChart	@	chart_state.ts:757
push../src/specs/specs_parser.tsx.SpecsSpecRootComponent.componentDidMount	@	specs_parser.tsx:17
commitLifeCycles	@	react-dom.development.js:17334
commitAllLifeCycles	@	react-dom.development.js:18736
callCallback	@	react-dom.development.js:149
invokeGuardedCallbackDev	@	react-dom.development.js:199
invokeGuardedCallback	@	react-dom.development.js:256
commitRoot	@	react-dom.development.js:18948
(anonymous)	@	react-dom.development.js:20418
unstable_runWithPriority	@	scheduler.development.js:255
completeRoot	@	react-dom.development.js:20417
performWorkOnRoot	@	react-dom.development.js:20346
performWork	@	react-dom.development.js:20254
performSyncWork	@	react-dom.development.js:20228
requestWork	@	react-dom.development.js:20097
scheduleWork	@	react-dom.development.js:19911
scheduleRootUpdate	@	react-dom.development.js:20572
updateContainerAtExpirationTime	@	react-dom.development.js:20600
updateContainer	@	react-dom.development.js:20657
push../node_modules/react-dom/cjs/react-dom.development.js.ReactRoot.render	@	react-dom.development.js:20953
(anonymous)	@	react-dom.development.js:21090
unbatchedUpdates	@	react-dom.development.js:20459
legacyRenderSubtreeIntoContainer	@	react-dom.development.js:21086
render	@	react-dom.development.js:21155
render	@	render.js:51
renderMain	@	render.js:89
renderMain	@	start.js:208
renderUI	@	start.js:228
dispatch	@	redux.js:220
_renderMain	@	config_api.js:98
render	@	config_api.js:42
(anonymous)	@	config_api.js:71
(anonymous)	@	config.ts:48
./.storybook/config.ts	@	config.ts:48
__webpack_require__	@	bootstrap:782
fn	@	bootstrap:150
0	@	styling.tsx:673
__webpack_require__	@	bootstrap:782
checkDeferredModules	@	bootstrap:45
webpackJsonpCallback	@	bootstrap:32
(anonymous)	@	main.2dbf464….bundle.js:1
```

</p>
</details>

<details><summary>Second call to `computeChart` looks to be from chart_resizer</summary>
<p>

```bash
chart_state.ts:757 computeChart
push../src/state/chart_state.ts.ChartStore.computeChart	@	chart_state.ts:757
push../src/state/chart_state.ts.ChartStore.updateParentDimensions	@	chart_state.ts:716
(anonymous)	@	chart_resizer.tsx:30
Resizer._this.onResize	@	chart_resizer.tsx:29
invokeFunc	@	debounce.js:95
trailingEdge	@	debounce.js:144
timerExpired	@	debounce.js:132
setTimeout (async)		
leadingEdge	@	debounce.js:103
debounced	@	debounce.js:172
```

</p>
</details>

markov00:
@nickofthyme I'm also not sure that this will be the right way to solve. Specially I'm not sure if we really need to use the react state for that usecase. I think we should simply use a private class variable to store that flag. This will avoid a re-rendering the component because of the state change.

Yes, you are right about the chart component that is called few times at the beginning. We have to clean that behaviour once we are in a good shape with the bug/feature requests. For the moment that method is called few times, but the real computation happens only when we have all the required paramenters to proceed (parent width, height, all the series etc). We will track this issue here: https://github.com/elastic/elastic-charts/issues/75

nickofthyme:
@markov00 Ya true not sure why I threw that in state. I replaced the state variable with a local private variable. Is there another way you thought of doing it entirely? Or is this approach ok?

markov00:
@nickofthyme I think it's good to merge. I've took a look at the codecov and we actually don't have any test on that component so we can go on and merge right now.
Please squash before merge, clean the squash message and description to be aligned with the `conventionalcommit` standard.

Issue #197, opened 2019-04-25
Cannot set the custom formatting function for Tooltip header / Anotation header


alexwizp:
**Describe the bug**
Currently we can specify the formatting function only for ticks in the axis.

We need to somehow format the title for the tool-tip.
![image](https://user-images.githubusercontent.com/20072247/56728360-46dbb700-675b-11e9-85fd-324f4b503196.png)

For Annotation we can do it by overriding header property like in example below
![image](https://user-images.githubusercontent.com/20072247/56728143-cfa62300-675a-11e9-9f9d-c24869da27b2.png)

But I prefer to have common property like e.g. `headerFormat` for both cases

markov00:
@emmacunningham @alexwizp does the merged PR https://github.com/elastic/elastic-charts/pull/233 fix this issue?

nickofthyme:
@alexwizp Is this still needed for the tsvp PR?

markov00:
@alexwizp can this request be closed?

Issue #191, opened 2019-04-24
Use elastic-charts in Kibana Discover


emmacunningham:
**Is your feature request related to a problem? Please describe.**
Additional features needed in elastic-charts:
- [x] `DONE` BarSeriesSpec needs to have a "histogram mode" for positioning ticks correctly (PR #218)
![image](https://user-images.githubusercontent.com/452850/58044656-2eaa4e00-7af5-11e9-8820-0727a71b5bb0.png)
  - [x] ticks should be aligned to the starting edge of a bar and bars cannot be clustered (only stacked)
  - [x] xDomain annotation positions should be aligned correctly (currently they snap to a band edge)
- [x] `DONE` customizable tooltip content (PR #233)
  - [x] need to be able to add x-axis title to chart tooltip (formatter for tooltip header)
  - [x] need to be able to add message about partial bucket to chart tooltip (additional information iff match for filter value)
- [x] `DONE` in histogram mode, need to extend line annotation (PR #236)
- [x] `DONE` check bucket sizing in histogram mode, at some time intervals, buckets appear much larger (PR #240)
- [x] `DONE` ticks in histogram mode need to be extended correctly (PR #251 )
- [x] `DONE` single datum tooltip bug (PR #259)
- [x] `DONE` show `now` line annotation in single datum yearly view (PR #265)

On the Kibana side: 
- [x] `IN PROGRESS` Use elastic-charts component (https://github.com/elastic/kibana/pull/43788)
  - [x] Partial bucket rect annotations 
  - [x] Partial bucket custom tooltip header
  - [x] Line annotation for current time
  - [x] Axis ticks line up correctly for each time interval
  - [x] Buckets render correctly for each time interval
    - [x] check all ranges at auto interval
    - [x] check all ranges at manually set intervals
  - [x] On brush, fetch new data
  - [x] styling (we will use the elastic-charts theme defaults but should check both dark & light modes)
  - [x] fix tests [note updating with elastic-charts did not break any tests]
  - [x] line annotation for `now` renders correctly
    - [x] within chart when now is within time range
    - [x] adjust position to make line visible when at domain edge
  - [x] localize custom text
  - [x] rect annotation tooltip should be formatted correctly
  - [x] y-axis formatter
  - [x] bug when hovering out of tooltip

**Describe the solution you'd like**
Use elastic-charts for the charts in Kibana Discover.

**Describe alternatives you've considered**

**Additional context**

**Kibana Cross Issues**
TODO

**Checklist**

[...]

PR #233, opened 2019-06-11
feat(tooltip): add custom headerFormatter

emmacunningham:
## Summary
addresses #197 & #191 

This PR introduces a SettingsSpec prop `headerFormatter` which the user can use to define a custom header formatter for the tooltip headers.

`headerFormatter` has the type `(data: TooltipValue) => JSX.Element | string` so that a user can either specify a simple value formatting function or a more complex one should they need to return a JSX element to be able to add additional text and/or styling that is not provided by the internal chart styles.  

We use this feature in Discover charts to show a custom message for the first and last bars in a series when they contain partial buckets:

Original Discover chart:
image

Version using elastic-charts with custom `headerFormatter`:
image

Notable implementation details:
- `emptyFormatter` was changed from always returning a string to returning the raw value; this was necessary because in Discover, we do a check to see if a value warrants the special header or not 
- if `headerFormatter` is defined, `emptyFormatter` is used in place of the tick formatter from the relevant domain axis.  This means that if `headerFormatter` is defined, the user must explicitly format the values themselves (in the Discover implementation, we re-use the axis formatter if the condition to display the special header is not met)
- in the tooltip rendering component `<p>` was changed to `<div>` to allow for nested elements (Kibana's DOM nesting validator throws a warning for nested elements in a `<p>`)
- *BREAKING CHANGE AND REFACTOR* With the addition of the header formatter prop, SettingsSpec  would take three tooltip-related props: `tooltipType`, `tooltipSnap`, and `tooltipHeaderFormatter`.  Instead of three different props all with the `tooltip` namespace, in this PR we create an interface for tooltip-related configuration to simplify things.  This is a breaking change, but as far as I can tell, no one is using `tooltipType` or `tooltipSnap` right now, so it shouldn't impact anyone to change this now.  Now, we define `TooltipProps` and in place of three individual props, expose `tooltipProps`:

```
interface TooltipProps {
  type?: TooltipType;
  snap?: boolean;
  headerForrmatter?: TooltipValueFormatter;
}

interface SettingSpecProps {
  chartStore?: ChartStore;
  theme?: Theme;
  rendering: Rendering;
  rotation: Rotation;
  animateData: boolean;
  showLegend: boolean;
  /** Props for customizing the chart tooltip on hover */
  tooltip?: TooltipType | TooltipProps;
  /** ... etc remaining props */
}
```

See discussion below for more on the new `tooltip` prop on `SettingSpecProps`.

### Checklist


src/lib/series/tooltip.ts
```diff
@@ -52,6 +52,6 @@ export function formatTooltip(
   };
 }
 
-function emptyFormatter(value: any): string {
-  return `${value}`;
+function emptyFormatter<T>(value: T): T {
```
nickofthyme: 👍

src/state/chart_state.test.ts
```diff
@@ -635,6 +635,36 @@ describe('Chart Store', () => {
...
+      tickFormat: (value: any) => `foo ${value}`,
+    };
+
+    store.addAxisSpec(axisSpec);
+    store.addSeriesSpec(spec);
+    store.tooltipType.set(TooltipType.Crosshairs);
+    store.tooltipHeaderFormatter = undefined;
+    store.computeChart();
+
+    // with no tooltipHeaderFormatter defined, should return value formatted using xAxis tickFormatter
```
nickofthyme: This seems like it should be a `test` in a `describe` rather than in a comment. Then all the setup could be in a `beforeEach` to run both assertions you are making below the commented lines? 

```ts
describe('can use a custom tooltip header formatter', () => {
  beforeEach(() => {
    const axisSpec: AxisSpec = {
      id: AXIS_ID,
      groupId: spec.groupId,
      hide: true,
      showOverlappingTicks: false,
      showOverlappingLabels: false,
      position: Position.Bottom,
      tickSize: 30,
      tickPadding: 10,
      tickFormat: (value: any) => `foo ${value}`,
    };
  
    store.addAxisSpec(axisSpec);
    store.addSeriesSpec(spec);
    store.tooltipType.set(TooltipType.Crosshairs);
    store.tooltipHeaderFormatter = undefined;
    store.computeChart();
  });

  test('with no tooltipHeaderFormatter defined, should return value formatted using xAxis tickFormatter', () => {
    store.setCursorPosition(10, 10);
    expect(store.tooltipData[0].value).toBe('foo 1');
  });

  test('with tooltipHeaderFormatter defined, should return value formatted', () => {
    store.tooltipHeaderFormatter = (value: TooltipValue) => `${value}`;
    store.setCursorPosition(10, 10);
    expect(store.tooltipData[0].value).toBe(1);
  });
});

```

Just a thought, maybe there are other implications to this test I don't see/know.

emmacunningham: 👍 Good catch and great suggestion!

emmacunningham:
@markov00 replies below inline:

> I've just take a look at the description and the code and I've few comments on that change:
> 
>     * what do you think of change the tooltip props name from `tooltipProps` to only `tooltip`? sounds much more simple and clean
> 
>     * what if the `tooltip` prop accept two types: the `TooltipType` and the `TooltipProps`? this can simplify a bit the code for consumers that want only to specify the type of tooltip.

Changed the name of `tooltipProps` on the `Settings` component to `tooltip` and allowed this prop to be either `TooltipType` or `TooltipProps`; added type guards to check which type the defined prop is and then using the guards to conditionally update depending on the type.  See commit https://github.com/elastic/elastic-charts/pull/233/commits/3efe360256bd8ccd8b9a2a4e6b06699cb7c90ca8 

>     * What are the use case where we want to change the full header (x value included) and what are the cases where we just want to add an additional test to the tooltip? I'm asking this because I just want to avoid consumers to go and implement their own header + add their own data formatter on the header if they want just to add a small text for specific cases.

The type annotation for the headerFormatter is `(data: TooltipValue) => JSX.Element | string` so this allows the user to configure a formatting function that just returns a string if they do not need to further style the inner component.  If they just want to format the string, they have the ability to do that and do not need to provide their own implementation of the header.

Another possibility could be to split up these configurations so that TooltipProps looks something like:

```
interface TooltipProps {
  type?: TooltipType;
  snap?: boolean;
  headerFormatter?: (dataValue: TooltipValue) => string;
  headerElement?: (dataValue: TooltipValue) => JSX.Element;
}
```

(We need the element to have access to the raw data value or else we won't be able to do conditional displays if they come in as formatted strings.)

In this case, then the user can:
- specify just the `headerFormatter` without `headerElement`, which would just render the string formatted per the defined function
- specify just the `headerElement` which may also include a custom formatter for the data value and allows them to customize the layout within the header
- specify both `headerFormatter` and `headerElement`, in which case, we could use the `headerFormatter` within the `headerElement` to  format the values and control how the formatted value is rendered.

>     * In the long run we can have the needs to:
>       
>       * enable custom header text (as in this PR)
>       * enable custom X value formatter
>       * enable custom Y values formatter
>       * enable general tooltip formatter (all values + header)
>         what is your ideas behind that future?

I focussed just on the `tooltipHeaderFormatter` here and created the more abstracted tooltip prop on `Settings` because the idea is that once we do implement these additional customization features, `headerFormatter` can get pulled into the larger strategy.  Eventually, we may have a tooltip interface that looks more like:

```
interface TooltipProps {
  type?: TooltipType;
  snap?: boolean;
  headerFormatter?: (xValue: TooltipValue) => JSX.Element | string; // or the broken up version above
  rowFormatter?: (yValue: TooltipValue) => JSX.Element | string; // or the broken up version above
  containerElement?: (header, yValues) => JSX.Element; 
};
```

Where, if `containerElement` is defined, the user is defining their own component for rendering the entire tooltip; we give them all of the values they need and they can decide how to structure & style.  If they define `headerFormatter` and/or `rowFormatter`, then we'd use those formatting functions within `containerElement` otherwise, we use the default formatted string.  

I think the implementation of the header customization doesn't commit us or block us from being able to address these other customizations, but also doesn't require that we tackle the entire question of what and how much the user should be able to customize these other things in order to get this one feature in which we know is needed for both TSVB and Discover replacements.

Issue #216, opened 2019-05-22
Style individual bars


markov00:
**Is your feature request related to a problem? Please describe.**
Some bar chart requires different colors for different bars. Is not a real good practice but in special case is permitted, specially when you want to visualize different categories with the same X axis. The current implementation only allows 1 color per series. Multiple, non stacked series are combined together but it will results in a clustered set of elements that can leave some empty space when an x value is not present on each series.

**Describe the solution you'd like**
I'd like to specify a color accessor or a color function for each bar series, that will be used to apply the color of a single or multiple bars.

- [ ] Group bars in any bar series given a property accessor or function accessor.
- [ ] Customize color of groups
- [ ] Groups should show in the legend and hide and show independently from the parent series.

**Describe alternatives you've considered**
There is currently a workaround that work only in specific situation:

```tsx
<Chart>
  <Settings debug={false} rotation={90} />
  <Axis id={getAxisId('categories')} title="" position={Position.Left} />
  <Axis id={getAxisId('count')} title="count" position={Position.Bottom} />
  <BarSeries
    id={getSpecId('a')}
    name={'Series A'}
    xScaleType={ScaleType.Ordinal}
    yScaleType={ScaleType.Linear}
    xAccessor="x"
    yAccessors={['y']}
    stackAccessors={['x']}
    data={[{ x: 'a', y: 1 }]}
  />
  <BarSeries
    id={getSpecId('b')}
    name={'Series B'}
    xScaleType={ScaleType.Ordinal}
    yScaleType={ScaleType.Linear}
    xAccessor="x"
    yAccessors={['y']}
    stackAccessors={['x']}
    data={[{ x: 'b', y: 5 }]}
  />
</Chart>
```
<img width="599" alt="Screenshot 2019-05-22 at 15 55 28" src="https://user-images.githubusercontent.com/1421091/58180290-221b2680-7caa-11e9-8105-7ccc065b4cf7.png">

**Additional context**

That type of chart is currently used in SIEM KPIs

<img width="284" alt="Screenshot 2019-05-22 at 21 55 01" src="https://user-images.githubusercontent.com/1421091/58180326-2fd0ac00-7caa-11e9-86c6-771c3e9d92c3.png">

**Kibana Cross Issues**
n/a

**Checklist**

- [x] this request is checked against already exist requests
- ~[ ] every related Kibana issue is listed under `Kibana Cross Issues` list~
[...]

PR #235, opened 2019-06-12
feat: style individual bars in a series

nickofthyme:
## Summary
Issue #216

Add ability to style(only color) individual bars in a series

Allow `colorAccessors` to be of type `AccessorFn`

https://github.com/elastic/elastic-charts/blob/79422c2d392e2912070b30d0efbbc36d58d3414e/src/lib/utils/accessor.ts#L3

### Checklist

### DIFF
--- a/src/lib/series/series.ts
+++ b/src/lib/series/series.ts
@@ -1,5 +1,5 @@
 import { ColorConfig } from '../themes/theme';
-import { Accessor } from '../utils/accessor';
+import { Accessor, AccessorFn, isAccessorFn } from '../utils/accessor';
 import { GroupId, SpecId } from '../utils/ids';
 import { splitSpecsByGroupId, YBasicSeriesSpec } from './domains/y_domain';
 import { formatNonStackedDataSeriesValues } from './nonstacked_series_utils';
@@ -92,7 +92,7 @@ export function splitSeries(
   splitSeriesLastValues: Map<string, any>;
 } {
   const { xAccessor, yAccessors, y0Accessors, splitSeriesAccessors = [] } = accessors;
-  const colorAccessors = accessors.colorAccessors ? accessors.colorAccessors : splitSeriesAccessors;
+  const colorAccessors = [...splitSeriesAccessors, ...accessors.colorAccessors || []];
   const isMultipleY = yAccessors && yAccessors.length > 1;
   const series = new Map<string, RawDataSeries>();
   const colorsValues = new Map<string, any[]>();
@@ -100,7 +100,6 @@ export function splitSeries(
   const splitSeriesLastValues = new Map<string, any>();
 
   data.forEach((datum) => {
-    const seriesKey = getAccessorsValues(datum, splitSeriesAccessors);
     if (isMultipleY) {
       yAccessors.forEach((accessor, index) => {
         const colorValues = getColorValues(datum, colorAccessors, accessor);
@@ -114,7 +113,7 @@ export function splitSeries(
         );
         splitSeriesLastValues.set(colorValuesKey, cleanedDatum.y1);
         xValues.add(cleanedDatum.x);
-        updateSeriesMap(series, [...seriesKey, accessor], cleanedDatum, specId, colorValuesKey);
+        updateSeriesMap(series, colorValues, cleanedDatum, specId, colorValuesKey);
       }, {});
     } else {
       const colorValues = getColorValues(datum, colorAccessors);
@@ -128,7 +127,7 @@ export function splitSeries(
       );
       splitSeriesLastValues.set(colorValuesKey, cleanedDatum.y1);
       xValues.add(cleanedDatum.x);
-      updateSeriesMap(series, [...seriesKey], cleanedDatum, specId, colorValuesKey);
+      updateSeriesMap(series, colorValues, cleanedDatum, specId, colorValuesKey);
     }
   }, {});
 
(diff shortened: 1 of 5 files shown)


nickofthyme:
@markov00 & @emmacunningham This has proven to be a little trickier than I initially thought but has given me good insight into the lifecycle of a chart.

Currently, the `colorAccessor` prop changes the legend to add a new series but does _NOT_ affect the chart itself. After exploring the bar chart rendering I was able to get the chart to render the `colorAccessor` series in a separate series that bifurcates the entire `DataSeries`.

```ts
const colorAccessors = [(d) => d.y1 >= 2 ? 'nick' : null];
```

## Before
image

## After
image

I wanted to get both of your thoughts on this before I do anything else. From what I see all the rendering functions are assuming `DataSeries` has only one color. In order to achieve this feature, I am thinking to create a series that has "buckets" for each `colorSeriesKey`.

I am not sure of the overall implications for that sort of change. One thing that will be a bit problematic is the legend interaction with this new "nested" series.

Let me know your thoughts. Thanks!

> Note: I am just looking at the bar chart for now as I think there are design implications for the area/line chart that need to be discussed.

src/lib/utils/accessor.ts
```diff
@@ -10,12 +10,18 @@ export type Accessor = AccessorString;
...
   throw new Error('Accessor must be a string or a function');
 }
+
+/**
+ * Returns if the accessor is an AccessorFn
+ * @param accessor the spec accessor
+ */
+export function isAccessorFn(accessor: Accessor | AccessorFn): boolean {
+  return typeof accessor === 'function';
+}
```
emmacunningham: If you specify the return type as a [type predicate](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types), then you won't need to do any explicitly casting to `AccessorFn` when the check passes (as is happening in `updateSeriesMap` where this function gets called).

src/lib/utils/domain.ts
```diff
@@ -13,13 +13,6 @@ export interface SpecDomain {
   isStacked?: boolean;
 }
 
-export interface ColorDomain {
-  accessors: Accessor[];
-  yAccessors?: Accessor[];
-  domain: string[];
-  scaleType: ScaleType;
-}
-
```
emmacunningham: 👏

src/lib/series/series.ts
```diff
@@ -169,25 +168,25 @@ function updateSeriesMap(
...
  */
-function getAccessorsValues(datum: Datum, accessors: Accessor[] = []): any[] {
+function getAccessorsValues(datum: Datum, accessors: Array<Accessor | AccessorFn> = []): any[] {
   return accessors
     .map((accessor) => {
-      return datum[accessor];
+      return isAccessorFn(accessor) ? (accessor as AccessorFn)(datum) : datum[accessor as Accessor];
     })
-    .filter((value) => value !== undefined);
+    .filter((value) => value !== undefined && value !== null);
```
emmacunningham: can be simplified to `value != null` since in javascript, `null == null` and `null == undefined`

markov00:
@nickofthyme I think we forgot a bit about the `colorAccessors` prop. I think it was originally configured to do something near what we want to do here, but we probably didn't take care of that on most of the code refactoring :(
I also don't see any stories around `colorAccessors` so I think it's fair to change a bit it's destiny. @emmacunningham what do you think? 
The initial idea, and also how it's implemented seems that we want to to use the `colorAccessors` a way to "split" series by a different set of properties. I think the initial idea was something like:
I've 1 `<BarSeries>` with a dataset that looks like `[x, y, group, colorGroup]` and a configuration that split the series in 3 `splitAccessors={['x', 'group']}`. I want to color 2 series with the same color using a third parameter like `colorAccessors={[`colorGroup']}`. 
This will be ideallty translated in a 3 series bar chart with two colors and only 2 items on the legend. But, now that I'm thinking about it seems a bit strange.
I think we can refactor the way the `colorAccessor` is used in the way we really want to use: have the possibility to specify a color or a color index in the used color palette depending on the data series value. 
Few things to notice:
- I'm not completely sure that `colorAccessors` is the right name for that prop, we are specifying a function that change the color of a bar depending on the bar value, it will be worth having a look at other libraries and how they call similar functions/props(eg Victory from Formidable, vx or react-vis)  
- this function should be called with the value, with the `specId` and the data series key (usually identified by a key composed by the values coming from the splitAccessors). We may need also the automatically assigned color of the bar
- we should extend that function to lines and areas in the future, where we can have a line colored with a gradient between points, from color A to color B

emmacunningham:
> I think we forgot a bit about the `colorAccessors` prop. I think it was originally configured to do something near what we want to do here, but we probably didn't take care of that on most of the code refactoring :(
> I also don't see any stories around `colorAccessors` so I think it's fair to change a bit it's destiny. @emmacunningham what do you think?

Yes, I agree!  (And I like the phrasing of "changing destiny" 😄 )

>     * I'm not completely sure that `colorAccessors` is the right name for that prop, we are specifying a function that change the color of a bar depending on the bar value, it will be worth having a look at other libraries and how they call similar functions/props(eg Victory from Formidable, vx or react-vis)

++ to thinking about a more accurate and descriptive name for this prop, especially now that we're altering destinies!

nickofthyme:
Closed in favor of #255

Issue #201, opened 2019-04-30
Allow partial changes to a theme


markov00:
**Is your feature request related to a problem? Please describe.**
I'd like to change the style of a single property of the theme without having to create the full object that specify the style.

**Describe the solution you'd like**
I'd like to have a recursive partial type for the Theme object so I can specify only a small portion of the theme.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

**Kibana Cross Issues**
n/a

**Checklist**

- [x] this request is checked against already exist requests
- [x] every related Kibana issue is listed under `Kibana Cross Issues` list
- [x] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

cchaos:
**Describe the solution you'd like**
Also, be able to use `mergeWithDefaultTheme` with partial objects.

PR #239, opened 2019-06-12
feat(theme): allow recursive partial theme

nickofthyme:
## Summary
Issue #201

Allow partial theme object for chart `Settings`

You can now specify a `PartialTheme` to pass to the `Settings` component where every field in the full `Theme` type is *recursively* optional.

### Before
If you wanted to change one property in a nested type lets say `left` margin we want to set to `0`. You would need to include the default/base values of `right`, `top` and `bottom` margins for those to carry over, otherwise, they will be removed!

```ts
const theme: PartialTheme = {
  chartMargins: {
    left: 0,
    right: 10,
    top: 10,
    bottom: 10,
  },
};
```

### After
You *ONLY* need to include the values you want, then pass that to the `Settings` component.
```ts
const theme: PartialTheme = {
  chartMargins: {
    left: 0,
  },
};
```

There is also no need to merge this `PartialTheme` theme using the `mergeWithDefaultTheme` function, although you can if you really want to 😏. This will happen automatically within the `Settings`. By default, it will **deep-merge** your `PartialTheme` with the `LIGHT_THEME` as a base. However, if you would like to set the base to be something else you can pass that as a prop.

```tsx
<Chart className="story-chart-dark">
  <Settings
    theme={customPartialTheme}
    baseThemeType={BaseThemeTypes.Dark}
  />
  ...
/>

```

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "scripts": {
     "cz": "git-cz",
     "build:clean": "rm -rf ./dist",
-    "build:ts": "tsc -p ./tsconfig.json",
+    "build:ts": "tsc -p ./tsconfig.lib.json",
     "build:sass": "node-sass src/theme_light.scss dist/theme_light.css --output-style compressed && node-sass src/theme_dark.scss dist/theme_dark.css --output-style compressed && node-sass src/theme_only_light.scss dist/theme_only_light.css --output-style compressed && node-sass src/theme_only_dark.scss dist/theme_only_dark.css --output-style compressed",
     "concat:sass": "node scripts/concat_sass.js",
     "build": "yarn build:clean && yarn build:ts && yarn build:sass && yarn concat:sass",
@@ -27,7 +27,7 @@
     "watch": "yarn test --watch",
     "semantic-release": "semantic-release",
     "typecheck:src": "yarn build:ts --noEmit",
-    "typecheck:all": "tsc -p ./tsconfig.all.json --noEmit",
+    "typecheck:all": "tsc -p ./tsconfig.json --noEmit",
     "playground": "cd .playground && webpack-dev-server"
   },
   "files": [
--- a/src/@types/global.d.ts
+++ b/src/@types/global.d.ts
@@ -1,7 +1,0 @@
-declare type RecursivePartial<T> = {
-  [P in keyof T]?: T[P] extends (infer U)[]
-    ? RecursivePartial<U>[]
-    : T[P] extends readonly (infer U)[] // eslint-disable-line prettier/prettier
-    ? readonly RecursivePartial<U>[]
-    : RecursivePartial<T[P]>
-};
--- a/src/@types/global.d.ts
+++ b/src/@types/global.d.ts
@@ -1,0 +1,7 @@
+declare type RecursivePartial<T> = {
+  [P in keyof T]?: T[P] extends (infer U)[]
+    ? RecursivePartial<U>[]
+    : T[P] extends readonly (infer U)[] // eslint-disable-line prettier/prettier
+    ? readonly RecursivePartial<U>[]
+    : RecursivePartial<T[P]>
+};
--- a/src/lib/themes/theme.ts
+++ b/src/lib/themes/theme.ts
@@ -1,5 +1,5 @@
 import { GeometryStyle } from '../series/rendering';
-import { mergePartial } from '../utils/commons';
+import { mergePartial, RecursivePartial } from '../utils/commons';
 import { Margins } from '../utils/dimensions';
 import { LIGHT_THEME } from './light_theme';
 
--- a/src/lib/themes/theme.ts
+++ b/src/lib/themes/theme.ts
@@ -100,10 +100,11 @@ export interface Theme {
 
 export type PartialTheme = RecursivePartial<Theme>;
 
-export enum BaseThemeType {
-  Light,
-  Dark,
-}
+export type BaseThemeType = 'light' | 'dark';
+export const BaseThemeTypes: Readonly<{ [key: string]: BaseThemeType }> = Object.freeze({
+  Light: 'light',
+  Dark: 'dark',
+});
 
 export type DisplayValueStyle = TextStyle & {
   offsetX?: number;
--- a/src/lib/themes/theme.ts
+++ b/src/lib/themes/theme.ts
@@ -1,4 +1,5 @@
 import { GeometryStyle } from '../series/rendering';
+import { mergePartial } from '../utils/commons';
 import { Margins } from '../utils/dimensions';
 import { LIGHT_THEME } from './light_theme';
 
@@ -97,6 +98,13 @@ export interface Theme {
   crosshair: CrosshairStyle;
 }
 
+export type PartialTheme = RecursivePartial<Theme>;
+
+export enum BaseThemeType {
+  Light,
+  Dark,
+}
+
 export type DisplayValueStyle = TextStyle & {
   offsetX?: number;
   offsetY?: number;
@@ -136,20 +144,6 @@ export interface LineAnnotationStyle {
 }
 
 export type RectAnnotationStyle = StrokeStyle & FillStyle & Opacity;
-
-export interface PartialTheme {
-  chartMargins?: Margins;
-  chartPaddings?: Margins;
-  lineSeriesStyle?: LineSeriesStyle;
-  areaSeriesStyle?: AreaSeriesStyle;
-  barSeriesStyle?: BarSeriesStyle;
-  sharedStyle?: SharedGeometryStyle;
-  axes?: Partial<AxisConfig>;
-  scales?: Partial<ScalesConfig>;
-  colors?: Partial<ColorConfig>;
-  legend?: Partial<LegendStyle>;
-  crosshair?: Partial<CrosshairStyle>;
-}
 
 export const DEFAULT_GRID_LINE_CONFIG: GridLineConfig = {
   stroke: 'red',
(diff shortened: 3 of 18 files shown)


src/specs/settings.tsx
```diff
@@ -42,6 +43,14 @@ interface SettingSpecProps {
   xDomain?: Domain | DomainRange;
 }
 
+function getTheme(theme?: Theme | PartialTheme): Theme {
+  if (theme) {
+    return mergeWithDefaultTheme(theme);
```
nickofthyme: Unable to differentiate between a `Theme` and a `PartialTheme` here so I have to merge every time with a deep merge which is not ideal.

I was thinking of adding a new prop `partialTheme` to explicitly designate the theme as Partial but thought that might be too verbose.

Thoughts?

package.json
```diff
@@ -118,6 +118,7 @@
     "d3-collection": "^1.0.7",
     "d3-scale": "^1.0.7",
     "d3-shape": "^1.3.4",
+    "deepmerge": "^3.2.0",
```
markov00: @nickofthyme I think we should use a different approach or a more secure library. This library suffer from prototype poisoning: https://medium.com/intrinsic/javascript-prototype-poisoning-vulnerabilities-in-the-wild-7bc15347c96

```js
const customTheme = JSON.parse(`{
  "areaStyle": {
    "__proto__": {
      "poisonedValue": 666
    }
  }
}`)
const defaultTheme = {
  areaStyle: {},
  lineStyle: {},
}

// vulnerable library
const deepmerge = require('deepmerge');
const merged = deepmerge(defaultTheme, customTheme);
console.log(`areaStyadminle: `,merged.areaStyle);
console.log(`has poisonedValue in prototype: `,merged.areaStyle.poisonedValue);



// secure library
const mergeDeep = require('merge-deep')
const merged2 = mergeDeep(defaultTheme, customTheme);
console.log(`areaStyadminle: `,merged2.areaStyle);
console.log(`has poisonedValue in prototype: `,merged2.areaStyle.poisonedValue);

```

In our case I don't think it's harmful, but adding a library like that can possibly cause a misuse of that function incurring into bad vulnerabilities

nickofthyme: That's interesting, never heard of that before. Very cool!

nickofthyme:
It appears that the `RecursivePartial` type is not showing/enforcing the nested types in vsCode. But the type is still enforced at compile time. I think it is caused by this [GH issue](https://github.com/Microsoft/TypeScript/issues/13195) ([fix](https://github.com/microsoft/TypeScript/pull/30796) 🚀). Also found this [SO post](https://stackoverflow.com/questions/54489817/typescript-partialt-type-without-undefined).

```ts
type PartialTheme = RecursivePartial<Theme>
// Creates this 👇 
type PartialTheme = {
    chartMargins?: RecursivePartial<Margins> | undefined;
    chartPaddings?: RecursivePartial<Margins> | undefined;
    lineSeriesStyle?: RecursivePartial<LineSeriesStyle> | undefined;
    ... 7 more ...;
    crosshair?: RecursivePartial<...> | undefined;
}
```

For example, this `PartialTheme` object does not know that `barSeriesStyle` is of type `BarSeriesStyle` and therefore does not type check the `BAD` value _ONLY_ in vsCode.

```ts
const customPartialTheme: PartialTheme = {
  barSeriesStyle: {
    border: {
      visible: true,
      BAD: true,
    },
  },
};
```

But will error in the compiler.

```bash
ERROR in /Users/nicholaspartridge/Documents/repos/elastic-charts/stories/styling.tsx
./stories/styling.tsx
[tsl] ERROR in /Users/nicholaspartridge/Documents/repos/elastic-charts/stories/styling.tsx(402,11)
      TS2322: Type '{ border: { stroke: string; visible: true; BAD: boolean; }; }' is not assignable to type 'RecursivePartial<BarSeriesStyle>'.
  Types of property 'border' are incompatible.
    Type '{ stroke: string; visible: true; BAD: boolean; }' is not assignable to type 'RecursivePartial<StrokeStyle & Visible>'.
      Object literal may only specify known properties, and 'BAD' does not exist in type 'RecursivePartial<StrokeStyle & Visible>'.
```

I think this is acceptable until MS makes the fix in typescript.

yarn.lock
```diff
@@ -4951,7 +4951,7 @@ deep-is@~0.1.3:
   resolved "https://registry.yarnpkg.com/deep-is/-/deep-is-0.1.3.tgz#b369d6fb5dbc13eecf524f91b070feedc357cf34"
   integrity sha1-s2nW+128E+7PUk+RsHD+7cNXzzQ=
 
-deepmerge@3.2.0:
+deepmerge@3.2.0, deepmerge@^3.2.0:
```
nickofthyme: @markov00 it looks like we are already using this library through some dependency. Not sure if that matters.

src/@types/global.d.ts
```diff
@@ -0,0 +1,7 @@
+declare type RecursivePartial<T> = {
+  [P in keyof T]?: T[P] extends (infer U)[]
+    ? RecursivePartial<U>[]
+    : T[P] extends readonly (infer U)[] // eslint-disable-line prettier/prettier
+    ? readonly RecursivePartial<U>[]
+    : RecursivePartial<T[P]>
+};
```
markov00: I think we don't need a declaration file for that type, you can add this into a normal `.ts` and I've seen that also solves the issue with VSCode

markov00: I think you can move that definition to `src/lib/utils/commons.ts`

src/lib/themes/theme.ts
```diff
@@ -97,6 +98,13 @@ export interface Theme {
   crosshair: CrosshairStyle;
 }
 
+export type PartialTheme = RecursivePartial<Theme>;
+
+export enum BaseThemeType {
+  Light,
+  Dark,
+}
```
markov00: Together with @emmacunningham we are trying to get rid of `enums` to move toward babel in the next future (as it's done in Kibana).
You can use the Emma way to create enums in ts:
```
export const HistogramModeAlignments = Object.freeze({
  Start: 'start' as HistogramModeAlignment,
  Center: 'center' as HistogramModeAlignment,
  End: 'end' as HistogramModeAlignment,
});

export type HistogramModeAlignment = 'start' | 'center' | 'end';
```

yarn.lock
markov00: yep, it's semantic-release via octokit... important thing is that we don't use it directly

tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "declaration": true,
     "outDir": "./dist/",
     "noImplicitAny": true,
-    "noUnusedLocals": true,
+    "noUnusedLocals": false,
```
markov00: why are you disabling this option?
If it's to enable an easier prototyping while code i think it can be better to disable on the general tsconfig and enable it on the build process.

nickofthyme: Sorry I was going to ask about this before adding but left it in by accident. The reason I removed it is because this can, and is, being enforced by the linter an by removing it we avoiding needless compile errors. I have noticed when developing I will use temporary variables or comment functions then realize the function was the only thing using a defined variable and `tsc` complains about the variable being unused so I have to go comment out that variable. This happens even more when I write units.

Any thoughts?

src/@types/global.d.ts
nickofthyme: Is there a reason you don't like to use global custom types? I just put it here to avoid having to import the type everywhere.

src/lib/themes/theme.ts
nickofthyme: OK cool what is the benefit of this approach?

src/@types/global.d.ts
markov00: I'm not against that, but I've found that moving the type to a .ts file solves the VSCode problem on detecting the `BAD: true` error.

src/lib/themes/theme.ts
markov00: Babel doesn't currenty support TS `const enums` https://github.com/babel/babel/issues/8741 
or at least from the latest news I know :) We are ideally following the way Kibana is doing compilation etc, so we will move to babel some time soon.
If you know other nice way to build enums in a more JS way feel free to go with that direction

tsconfig.json
markov00: Yes exactly, I've noticed that and I also wanted to reduce these errors during development.
It will be great if we still enforce this at build time (on the pre-commit hook and on the regular build) but disable that option during normal development time.
I think it can be achieved overriding that option on the `.lib` tsconfig, but not sure

src/lib/themes/theme.ts
nickofthyme: No that makes sense, I was just curious about the motivation. I think this is a great way of doing it.

tsconfig.json
nickofthyme: Ok great idea, that should work.

src/@types/global.d.ts
nickofthyme: Oh I see. It looks like some of the tsconfig changes fixed this. I am now able to see the type errors in VSCode.

image

If you are fine with the global type other than that I will keep it.

nickofthyme: @markov00 I couldn't find a way to include the global type in the output. I think the global types are more useful in an application build rather than a library. I'm sure there is a way but wasn't worth the hassle. I added the type to the common utils as you suggested.

tsconfig.json
nickofthyme: https://github.com/elastic/elastic-charts/pull/239/files#diff-d1a6ed5fa5c14bb4581808f36d005956R3

src/specs/settings.tsx
nickofthyme: Only run when settings props are updated.

PR #251, opened 2019-07-01
fix(ticks): fill in additional ticks for histogram

emmacunningham:
## Summary
This PR introduces a fix for the issue where, in histogram mode, the intermediary ticks for a time scale were not being rendered.

The issue stemmed from how the ticks were being extended in histogram mode.  Previously, we were just extending the ticks by 1 tick adding the scale's minInterval to the last computed tick.  However, in cases where the interval between ticks was smaller than the scale's minInterval, this would mean that while the last tick would be correct, there would be missing intermediary ticks between the last computed tick and the final added tick.  Now, we computed the distance between the ticks and add additional ticks with that interval until reaching the extended histogram domain end.

Before fix (note the missing ticks for the last bar):
custom_min_interval

After fix:
intermediary_ticks

And additional test has been added to simulate the situation visible in the Storybook example and have also cleaned up the tests a bit to be easier to follow.

### Checklist

### DIFF
--- a/src/lib/axes/axis_utils.test.ts
+++ b/src/lib/axes/axis_utils.test.ts
@@ -205,36 +205,73 @@ describe('Axis computational utils', () => {
     expect(xScale).toBeDefined();
   });
 
-  test('should compute available ticks', () => {
-    const scale = getScaleForAxisSpec(verticalAxisSpec, xDomain, [yDomain], 0, 0, 100, 0);
-    const axisPositions = getAvailableTicks(verticalAxisSpec, scale!, 0, false);
-    const expectedAxisPositions = [
-      { label: '0', position: 100, value: 0 },
-      { label: '0.1', position: 90, value: 0.1 },
-      { label: '0.2', position: 80, value: 0.2 },
-      { label: '0.3', position: 70, value: 0.3 },
-      { label: '0.4', position: 60, value: 0.4 },
-      { label: '0.5', position: 50, value: 0.5 },
-      { label: '0.6', position: 40, value: 0.6 },
-      { label: '0.7', position: 30, value: 0.7 },
-      { label: '0.8', position: 20, value: 0.8 },
-      { label: '0.9', position: 10, value: 0.9 },
-      { label: '1', position: 0, value: 1 },
-    ];
-    expect(axisPositions).toEqual(expectedAxisPositions);
+  describe('getAvailableTicks', () => {
+    test('should compute to end of domain when histogram mode not enabled', () => {
+      const enableHistogramMode = false;
+      const scale = getScaleForAxisSpec(verticalAxisSpec, xDomain, [yDomain], 0, 0, 100, 0);
+      const axisPositions = getAvailableTicks(verticalAxisSpec, scale!, 0, enableHistogramMode);
+      const expectedAxisPositions = [
+        { label: '0', position: 100, value: 0 },
+        { label: '0.1', position: 90, value: 0.1 },
+        { label: '0.2', position: 80, value: 0.2 },
+        { label: '0.3', position: 70, value: 0.3 },
+        { label: '0.4', position: 60, value: 0.4 },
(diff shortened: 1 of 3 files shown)


src/lib/axes/axis_utils.ts
```diff
@@ -372,8 +372,14 @@ export function getAvailableTicks(
   const ticks = scale.ticks();
 
   if (enableHistogramMode && scale.bandwidth > 0) {
-    const finalTick = ticks[ticks.length - 1] + scale.minInterval;
-    ticks.push(finalTick);
+    const lastComputedTick = ticks[ticks.length - 1];
+    const penultimateComputedTick = ticks[ticks.length - 2];
```
markov00: If we have only one data value, that will be `undefined`, you should check this against an example like the following:
```
.add('single histogram bar chart', () => {
    return (
      <Chart className={'story-chart'}>
        <Axis
          id={getAxisId('bottom')}
          title={'timestamp per 1 minute'}
          position={Position.Bottom}
          showOverlappingTicks={true}
        />
        <Axis
          id={getAxisId('left')}
          title={KIBANA_METRICS.metrics.kibana_os_load[0].metric.title}
          position={Position.Left}
        />
        <HistogramBarSeries
          id={getSpecId('bars')}
          xScaleType={ScaleType.Linear}
          yScaleType={ScaleType.Linear}
          xAccessor={0}
          yAccessors={[1]}
          data={KIBANA_METRICS.metrics.kibana_os_load[0].data.slice(0, 1)}
        />
      </Chart>
    );
  })
```

emmacunningham: See https://github.com/elastic/elastic-charts/pull/251/commits/739c6c2a5609ab3c98be0972e681fdb9dda4f2b9

Note that in order for the additional tickLabel to be visible here, the user must specify a custom `minInterval`.  The reason for that is with a single value scale, the default computed minInterval will be `1` which will render the incorrect label for the tick.  If the user doesn't specify a custom `minInterval`, the chart will still render but without the additional final tick.

image

src/lib/axes/axis_utils.test.ts
```diff
@@ -205,36 +205,90 @@ describe('Axis computational utils', () => {
...
+      const xBandDomain: XDomain = {
+        type: 'xDomain',
+        scaleType: ScaleType.Linear,
+        domain: [0, 100],
+        isBandScale: true,
+        minInterval: 10,
+      };
+      const xScale = getScaleForAxisSpec(horizontalAxisSpec, xBandDomain, [yDomain], 1, 0, 100, 0);
+      const histogramAxisPositions = getAvailableTicks(horizontalAxisSpec, xScale!, 1, enableHistogramMode);
+      const histogramTickLabels = histogramAxisPositions.map((tick: AxisTick) => tick.label);
```
nickofthyme: I think giving the type for HOF is a little redundant given the array should have the type, also destructuring. Just a thought, no big deal.

```suggestion
      const histogramTickLabels = histogramAxisPositions.map(({ label }) => label);
```

emmacunningham: I added the destructuring but kept the type annotation https://github.com/elastic/elastic-charts/pull/251/commits/5d338111bddbde46a38178c670f3b1d06912e350

This is maybe something we will want to discuss more as a whole team, but in these kinds of cases, I prefer to see the type (even though it is not strictly necessary) because it helps me directly see the type of thing I'm working with without having to reference the array being mapped over; the explicitness isn't just for checking purposes but also reasoning purposes.  Yes, it takes like milliseconds to check the type of the array and then infer the type of the function parameter, but making the type explicit frees up my working memory so that I don't have to do that inference myself.  It's a small marginal benefit to be sure, but I'm not sure there's any disadvantage to having the type annotation here, so I'd take the small marginal benefit when I can.

Maybe this is something we can discuss in one of our upcoming syncs with @markov00?

PR #252, opened 2019-07-02
refactor(legend): remove visibility button

markov00:
## Summary

This PR removes the visibility button. The visibility change handling is applied on the legend item title.

**BREAKING CHANGE**: the `onLegendItemClick` click handler is no longer applied when clicking on the title. Instead a simple visibility change is applied.

Before:
Screenshot 2019-07-02 at 10 13 31

After:
Screenshot 2019-07-02 at 10 12 48

cc @cchaos 
I've some doubts on this because removes two important features right now:
- without the eye/eyeclosed icons we don't have a graphic representation that specify that a series is hidden. Is a ~~strikethrough~~ enough or is much more difficult to read?
- There are few legend item interaction that we need to support:
    - Hide/show series (currently the only interaction on TSVB, but not an interaction on vizlib)
    - click to filter (on vizlib clicking on the legend item is used to create and apply a global filter)

Before merging I'd like to check with you @cchaos these things.

### Checklist

### DIFF
--- a/src/components/legend/legend_item.tsx
+++ b/src/components/legend/legend_item.tsx
@@ -52,8 +52,11 @@ class LegendItemComponent extends React.Component<LegendItemProps, LegendItemSta
     const isSelected = legendItemKey === this.props.chartStore!.selectedLegendItemKey.get();
     const hasDisplayValue = this.props.chartStore!.showLegendDisplayValue.get();
     const hasTitleClickListener = Boolean(this.props.chartStore!.onLegendItemClickListener);
+    const itemClasses = classNames('echLegendItem', {
+      'echLegendItem-isHidden': !isSeriesVisible,
+    });
     return (
-      <div className="echLegendItem" onMouseEnter={onMouseEnter} onMouseLeave={onMouseLeave}>
+      <div className={itemClasses} onMouseEnter={onMouseEnter} onMouseLeave={onMouseLeave}>
         {this.renderColor(this.toggleColorPicker, color, isSeriesVisible)}
         {this.renderTitle(label, onTitleClick, hasTitleClickListener, isSelected, hasDisplayValue)}
         {this.renderDisplayValue(displayValue, showLegendDisplayValue, isSeriesVisible)}
--- a/src/components/legend/legend_item.tsx
+++ b/src/components/legend/legend_item.tsx
@@ -54,20 +54,24 @@ class LegendItemComponent extends React.Component<LegendItemProps, LegendItemSta
     const hasTitleClickListener = Boolean(this.props.chartStore!.onLegendItemClickListener);
     return (
       <div className="echLegendItem" onMouseEnter={onMouseEnter} onMouseLeave={onMouseLeave}>
-        {this.renderColor(this.toggleColorPicker, color)}
+        {this.renderColor(this.toggleColorPicker, color, isSeriesVisible)}
         {this.renderTitle(label, onTitleClick, hasTitleClickListener, isSelected, hasDisplayValue)}
         {this.renderDisplayValue(displayValue, showLegendDisplayValue, isSeriesVisible)}
       </div>
     );
   }
-  renderColor(colorClickAction: () => void, color: string | undefined) {
+  renderColor(colorClickAction: () => void, color?: string, isSeriesVisible: boolean = true) {
     if (!color) {
       return null;
     }
-    // TODO add color picler
+    // TODO add color picker
+    const iconType = isSeriesVisible ? 'dot' : 'eyeClosed';
+    const iconColor = isSeriesVisible ? color : undefined;
+    const ariaLabel = isSeriesVisible ? 'series color' : 'series hidden';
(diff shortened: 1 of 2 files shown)


cchaos:
> without the eye/eyeclosed icons we don't have a graphic representation that specify that a series is hidden

What we do in Maps is replace the legend color indicator with the eyeClosed icon. We can do the same here though we may need to replace the icon with a smaller version since it's a bit overbearing compared to the dot.



> There are few legend item interaction that we need to support:
> * Hide/show series (currently the only interaction on TSVB, but not an interaction on vizlib)

Is this not the current interaction by clicking on the legend item?

> * click to filter (on vizlib clicking on the legend item is used to create and apply a global filter)

Yes back when that functionality was actually in elastic-charts I had spoken with @emmacunningham about the UI for this and had proposed that any "extra" functionality that the consumer needs to add per legend item would be housed in an overflow context menu. Essentially:

cchaos:
Also I'm not sure if this was introduced in this branch or not but I saw a bug in the **Legend / hide legend items by series** example when clicking to hide the series:

markov00:
> > 
> > * Hide/show series (currently the only interaction on TSVB, but not an interaction on vizlib)
> 
> Is this not the current interaction by clicking on the legend item?
> 
Clicking on vislib chart just only open the contextual menu with the color picker and the `+` `-` buttons to filter the element, where in tsvb clicking doesn't add any filter, just hide/show the series. 

So the click will just hide/show the series without adding any filter. On a different PR we can add the contextual menu back again, so you can add any type of action on the selected legend item.

markov00:
@cchaos I've updated the code here with the icon when hiding the series
Screenshot 2019-07-03 at 19 39 16

src/components/legend/legend_item.tsx
```diff
@@ -46,29 +46,32 @@ class LegendItemComponent extends React.Component<LegendItemProps, LegendItemSta
...
 
-    const onTitleClick = this.onLegendTitleClick(legendItemKey);
+    const onTitleClick = this.onVisibilityClick(legendItemKey);
 
     const showLegendDisplayValue = this.props.chartStore!.showLegendDisplayValue.get();
     const isSelected = legendItemKey === this.props.chartStore!.selectedLegendItemKey.get();
     const hasDisplayValue = this.props.chartStore!.showLegendDisplayValue.get();
     const hasTitleClickListener = Boolean(this.props.chartStore!.onLegendItemClickListener);
     return (
       <div className="echLegendItem" onMouseEnter={onMouseEnter} onMouseLeave={onMouseLeave}>
```
cchaos: Can you append a class of `echLegendItem-isHidden` if it's hidden and then in the CSS do
`.echLegendItem-isHidden { color: $euiColorDarkShade }`

src/components/legend/legend_item.tsx
```diff
@@ -46,29 +46,32 @@ class LegendItemComponent extends React.Component<LegendItemProps, LegendItemSta
...
+  renderColor(colorClickAction: () => void, color?: string, isSeriesVisible: boolean = true) {
     if (!color) {
       return null;
     }
-    // TODO add color picler
+    // TODO add color picker
+    const iconType = isSeriesVisible ? 'dot' : 'eyeClosed';
+    const iconColor = isSeriesVisible ? color : undefined;
+    const ariaLabel = isSeriesVisible ? 'series color' : 'series hidden';
+    const viewBox = isSeriesVisible ? undefined : '-3 -3 22 22';
```
cchaos: Interesting way of sizing down the SVG, does this work in all browsers?

src/components/legend/legend_item.tsx
```diff
@@ -46,29 +46,32 @@ class LegendItemComponent extends React.Component<LegendItemProps, LegendItemSta
...
-  renderColor(colorClickAction: () => void, color: string | undefined) {
+  renderColor(colorClickAction: () => void, color?: string, isSeriesVisible: boolean = true) {
     if (!color) {
       return null;
     }
-    // TODO add color picler
+    // TODO add color picker
+    const iconType = isSeriesVisible ? 'dot' : 'eyeClosed';
+    const iconColor = isSeriesVisible ? color : undefined;
+    const ariaLabel = isSeriesVisible ? 'series color' : 'series hidden';
```
cchaos: We have learned that `aria-label` attributes directly on `svg`'s don't work with screen readers. You should apply this to the wrapping div. I would also add this as the `title` attribute to the wrapping div.

markov00: done in e40a42d5f9c42b93535010e0004fc497668921f5

src/components/legend/legend_item.tsx
markov00: done in 42adb3b3e66c5db0f1148df855ec8047b92a8213

src/components/legend/legend_item.tsx
markov00: yes, or at least it works correctly on Chrome, Firefox and IE 11. `viewBox` is on the svg specs supported since 1.1

Issue #237, opened 2019-06-12
Move `gridLineStyle` to be part of Theme


cchaos:
**Is your feature request related to a problem? Please describe.**
At the moment, you have to declare your grid styles completely separate from the theme, which also makes dark vs light mode more complicated.

**Describe the solution you'd like**
Move `gridLineStyle` into the full Theme json object.

**Checklist**

- [x] this request is checked against already exist requests
- ~[ ] every related Kibana issue is listed under `Kibana Cross Issues` list~
- ~[ ] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present~

PR #257, opened 2019-07-02
feat(theme): add gridLineStyle to AxisConfig

rshen91:
## Summary
Fixes https://github.com/elastic/elastic-charts/issues/237

This PR will simplify the process of changing styles by moving gridLineStyle into the Theme interface through the AxisConfig interface. 

Light theme and dark theme are changed to include gridLineStyle.

BREAKING CHANGE: Added `GridLineStyle` to `Theme` (`theme.gridLineStyle.horizontal` and `theme.gridLineStyle.vertical`)

### Checklist


elasticcla:
Hi @rshen91, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/elastic-charts/pull/257.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

rshen91:
Zoomed with Emma about this PR: 

- Need to define a theme fallback if axisSpec isn't specified by the user but can keep axisSpec in reactiveChart working as it was previously
- AxisSpec should take precedence over the theme but if a prop is defined, use that instead
- Define a helper function, in reactive_chart, that will handle the merge of the theme and the prop specified 
- Add a story to show the different input similar to the styling story called custom series style bars - apply bar style (bar 1 series) defaults to the theme if the customization isn't specified

src/lib/themes/dark_theme.ts
```diff
@@ -87,6 +87,12 @@ export const DARK_THEME: Theme = {
       stroke: 'white',
       strokeWidth: 1,
     },
+    gridLineStyle: {
+      stroke: 'black',
+      strokeWidth: 1,
+      opacity: 0,
+      dash: [0, 0],
+    },
```
markov00: What if we add also the `visible:true/false` so on the theme we don't have to work on the opacity to show or hide them?
I would also propose to set some nice default, like a darkgray or lightgray stroke for the dark theme  grid and the same/similar for the light theme

src/components/react_canvas/grid.tsx
```diff
@@ -7,7 +7,7 @@ import { Dimensions } from '../../lib/utils/dimensions';
 interface GridProps {
   chartDimensions: Dimensions;
   debug: boolean;
-  gridLineStyle: GridLineConfig | undefined;
+  gridLineStyle?: GridLineConfig;
```
markov00: As this will come from the theme, it will always exist, so you can remove the optional parameter

cchaos:
I'm sorry to throw a wrench in this, but can I also request that there are 2 grid line styles in the theme? One for horizontal and one for vertical. Here you can see that the EUI theme uses 2 different styles for each:

rshen91:
Zoom with Emma
- Add visible to the theme
- Horizontal and Vertical specification for gridLineStyle 
- Add a new story in grid.tsx to address this new functionality (specify part of the axisSpec.gridLineStyle and make sure the theme generates the other properties for the gridLineStyle) 
-- Full editing feature for all parts of the theme gridLineStyle and a knob (a slider) for one part of the axisSpec (the others will be undefined by default)

src/components/react_canvas/grid.tsx
rshen91: Thank you! I think commit 79c2d11 addresses this

src/lib/themes/dark_theme.ts
rshen91: added visible here 3e55aec

cchaos:
Any movement on this one? I'd **really** like to have it! ❤️

nickofthyme:
@markov00 I think all the comments have been addressed.

PR #258, opened 2019-07-03
feat(style): allow fill and stroke overrides

markov00:
## Summary

This PR allows the computed series color to be overridden by the theme or the series specific style.

**BREAKING CHANGE**: `LineStyle`, `AreaStyle` and `BarSeriesStyle` types differs on the optional values.
`stroke` and `fill` on the theme or specific series style now override the computed series color.

The a style overrides with the following flows:
- if a style is declared at the Spec level (on a Bar/Area/Line component) it will take precedence over all the other style (any `stroke` or `fill` prop will override the computed series color for that specific style element). If not go to next
- if a `stroke` or `fill` is present on the chart `theme` prop for a Bar/Area/Line than it will use that color for the point (stroke or fill) line (stroke) bar(stroke, fill, borderStroke)
- if no `stroke` or `fill` is configured on theme or series specific style than we will use the one computed or assigned with `customSeriesColors` prop

Screenshot 2019-07-03 at 15 24 09
Screenshot 2019-07-03 at 15 24 17
Screenshot 2019-07-03 at 15 24 23

**notes**
- I've refactored the `build...props` used to compute the props for rendering lines/areas/bars/points, cleaning a but the function semantics.
- The payload of each individual `bar` rendered geometry is currently increased, but I will find a better way to specify that for bars in the future.
- I've cleaned the defaults of the two themes
- I've cleaned the stories for custom styles to align it with that new feature

### Checklist

### DIFF
--- a/src/components/react_canvas/bar_geometries.tsx
+++ b/src/components/react_canvas/bar_geometries.tsx
@@ -76,7 +76,7 @@ export class BarGeometries extends React.PureComponent<BarGeometriesDataProps, B
                   geometryStyle,
                 );
 
-                return <animated.Rect key={key} {...barProps} />;
+                return <animated.Rect {...barProps} key={key} />;
               }}
             </Spring>
           </Group>
@@ -94,7 +94,7 @@ export class BarGeometries extends React.PureComponent<BarGeometriesDataProps, B
         );
         return (
           <React.Fragment key={index}>
-            <Rect key={key} {...barProps} />
+            <Rect {...barProps} key={key} />
           </React.Fragment>
         );
       }
--- a/src/components/react_canvas/bar_geometries.tsx
+++ b/src/components/react_canvas/bar_geometries.tsx
@@ -4,13 +4,12 @@ import { Group, Rect } from 'react-konva';
 import { animated, Spring } from 'react-spring/renderprops-konva.cjs';
 import { LegendItem } from '../../lib/series/legend';
 import { BarGeometry, getGeometryStyle } from '../../lib/series/rendering';
-import { BarSeriesStyle, SharedGeometryStyle } from '../../lib/themes/theme';
-import { buildBarProps } from './utils/rendering_props_utils';
+import { SharedGeometryStyle } from '../../lib/themes/theme';
+import { buildBarRenderProps } from './utils/rendering_props_utils';
 
 interface BarGeometriesDataProps {
   animated?: boolean;
   bars: BarGeometry[];
-  style: BarSeriesStyle;
   sharedStyle: SharedGeometryStyle;
   highlightedLegendItem: LegendItem | null;
 }
@@ -40,11 +39,9 @@ export class BarGeometries extends React.PureComponent<BarGeometriesDataProps, B
 
   private renderBarGeoms = (bars: BarGeometry[]): JSX.Element[] => {
     const { overBar } = this.state;
-    const { style, sharedStyle } = this.props;
+    const { sharedStyle } = this.props;
     return bars.map((bar, index) => {
       const { x, y, width, height, color, seriesStyle } = bar;
-      const border = seriesStyle ? seriesStyle.border : style.border;
-      const customOpacity = seriesStyle ? seriesStyle.opacity : undefined;
 
       // Properties to determine if we need to highlight individual bars depending on hover state
       const hasGeometryHover = overBar != null;
@@ -58,49 +55,46 @@ export class BarGeometries extends React.PureComponent<BarGeometriesDataProps, B
         bar.geometryId,
         this.props.highlightedLegendItem,
         sharedStyle,
-        customOpacity,
+        seriesStyle.rect.opacity,
         individualHighlight,
       );
+      const key = `bar-${index}`;
 
       if (this.props.animated) {
         return (
           <Group key={index}>
             <Spring native from={{ y: y + height, height: 0 }} to={{ y, height }}>
               {(props: { y: number; height: number }) => {
-                const barProps = buildBarProps({
-                  index,
+                const barProps = buildBarRenderProps(
                   x,
-                  y: props.y,
+                  props.y,
                   width,
-                  height: props.height,
-                  fill: color,
-                  stroke: border.stroke,
-                  strokeWidth: border.strokeWidth,
-                  borderEnabled: border.visible,
+                  props.height,
+                  color,
+                  seriesStyle.rect,
+                  seriesStyle.rectBorder,
                   geometryStyle,
-                });
+                );
 
-                return <animated.Rect {...barProps} />;
+                return <animated.Rect key={key} {...barProps} />;
               }}
             </Spring>
           </Group>
         );
       } else {
-        const barProps = buildBarProps({
-          index,
+        const barProps = buildBarRenderProps(
           x,
           y,
           width,
           height,
-          fill: color,
-          stroke: border.stroke,
-          strokeWidth: border.strokeWidth,
-          borderEnabled: border.visible,
+          color,
+          seriesStyle.rect,
+          seriesStyle.rectBorder,
           geometryStyle,
-        });
+        );
         return (
           <React.Fragment key={index}>
-            <Rect {...barProps} />
+            <Rect key={key} {...barProps} />
           </React.Fragment>
         );
       }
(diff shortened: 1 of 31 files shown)


src/components/react_canvas/bar_geometries.tsx
```diff
@@ -58,49 +55,46 @@ export class BarGeometries extends React.PureComponent<BarGeometriesDataProps, B
...
+          color,
+          seriesStyle.rect,
+          seriesStyle.rectBorder,
           geometryStyle,
-        });
+        );
         return (
           <React.Fragment key={index}>
-            <Rect {...barProps} />
+            <Rect key={key} {...barProps} />
```
nickofthyme: Do you want to allow `barProps.key` to potentially override the `key` prop?
```suggestion
            <Rect {...barProps} key={key}/>
```

src/components/react_canvas/line_geometries.tsx
```diff
@@ -78,89 +66,56 @@ export class LineGeometries extends React.PureComponent<LineGeometriesDataProps,
...
-                  lineIndex,
-                  pointIndex,
-                  x,
-                  y,
-                  color,
-                  pointStyleProps,
-                });
-                return <animated.Circle {...pointProps} />;
+                const pointProps = buildPointRenderProps(x, y, pointStyleProps);
+                return <animated.Circle key={key} {...pointProps} />;
```
nickofthyme: Same question?

```suggestion
                return <animated.Circle {...pointProps} key={key} />;
```

src/components/react_canvas/bar_geometries.tsx
markov00: nice catch!

src/components/react_canvas/line_geometries.tsx
markov00: nice catch!

markov00: solved in 111a465 (with a check on all the keys props)

src/components/react_canvas/bar_geometries.tsx
markov00: solved in 111a465 (with a check on all the keys props)

src/lib/themes/theme.ts
```diff
@@ -107,36 +107,80 @@ export const BaseThemeTypes: Readonly<{ [key: string]: BaseThemeType }> = Object
...
+  /** a static stroke color if defined, if not it will use the color of the series */
+  stroke?: string;
+  /** the stroke width of the rect border */
+  strokeWidth: number;
+}
+export interface BarSeriesStyle {
+  rect: RectStyle;
+  rectBorder: RectBorderStyle;
+  displayValue: DisplayValueStyle;
+}
```
emmacunningham: For some of the properties above, we have defined interfaces for them already.  Should we use them where possible?  For example, we have an `Opacity` interface that could be referenced instead of duplicating it manually within `RectStyle`.  (The same for `FillStyle`, `StrokeStyle`, `Visible` and `StrokeDashArray`.)  If we aren't going to reuse those interfaces, should we delete them?  Looks like `Radius` was already deleted in favor of defining it inside the interface.

stories/bar_chart.tsx
```diff
@@ -119,7 +118,12 @@ storiesOf('Bar Chart', module)
       0,
     );
 
-    const theme = mergeWithDefaultTheme(barStyle, LIGHT_THEME);
+    const theme = mergeWithDefaultTheme(
+      {
+        barSeriesStyle: barStyle,
```
emmacunningham: can this be simplified to `barSeriesStyle: displayValueStyle`?  (and then can remove lines 105-6 above)

src/lib/themes/theme.ts
markov00: @emmacunningham here my difficulty was the following: the `StrokeStyle` is specified as 
```
export interface StrokeStyle {
  /** The stroke color in hex, rgba, hsl */
  stroke: string;
  /** The stroke width in pixel */
  strokeWidth: number;
}
```
but I want to to have the stroke optional and the strokeWidth required. Is there a way to achieve that? I can change the `StrokeStyle` type but we maybe need a different configuration in the future. Splitting the `StrokeStyle` into two types is useless and we can just go removing all the Stroke/Opacity interfaces.
What do you think?

stories/bar_chart.tsx
markov00: fixed in 274cf29

src/lib/themes/theme.ts
emmacunningham: I believe you can do something like this to extend `StrokeStyle` such that `stroke` becomes optional and `strokeWidth` remains required.

First, we need a way to keep one property as required and make the rest optional:

```
export type RequireOneAndMakeRestOptional<T, K extends keyof T> = { [X in Exclude<keyof T, K>]?: T[X] } & { [P in K]-?: T[P] };
```

and then `RectBorderStyle` for example could be:

```
export type RectBorderStyle = RequireOneAndMakeRestOptional<StrokeStyle, 'strokeWidth'> & Visible;
```

I'm not sure this is that much better since it requires other contributors to understand what `RequireOneAndMakeRestOptional` is doing to understand the type annotations where it's being used.  I think that the way you have it now is more easy to follow and read.

Having the interfaces for these common styling properties I think is useful because it makes it easier for other contributors to compose new interfaces that are consistent (so we don't accidentally type `stroke` as a `number` sometimes, for example).  In particulary, I think `StrokeDashArray` is especially useful because it is not obvious what this should be (in fact, we could probaably make this more explicit as `[number, number]`) for someone not familiar with the API to create a dashed stroke.

stories/bar_chart.tsx
```diff
@@ -119,7 +100,23 @@ storiesOf('Bar Chart', module)
       0,
     );
 
-    const theme = mergeWithDefaultTheme(barStyle, LIGHT_THEME);
+    const theme = mergeWithDefaultTheme(
```
nickofthyme: No longer need to `mergeWithDefaultTheme`

stories/styling.tsx
```diff
@@ -468,48 +478,36 @@ storiesOf('Stylings', module)
...
+      },
+      rect: {
+        fill: color('rect fill', '#22C61A', 'Bar 1 Style'),
+        opacity: range('rect opacity', 0, 1, 0.3, 'Bar 1 Style', 0.1),
+      },
+    };
 
-    const chartTheme = {
-      ...LIGHT_THEME,
+    const theme = mergeWithDefaultTheme({
```
nickofthyme: No longer need to `mergeWithDefaultTheme`

stories/styling.tsx
```diff
@@ -519,51 +517,37 @@ storiesOf('Stylings', module)
...
-    const lineSeriesStyle2 = useOnlyChartTheme ? undefined : generateLineSeriesStyleKnobs('lines2');
+    const applyLineStyles = boolean('apply line series style', true, 'Chart Global Theme');
+    const lineSeriesStyle1 = generateLineSeriesStyleKnobs('Line 1 style', 'line1', 'lime', 'green', 4, 10, 6);
+    const lineSeriesStyle2 = generateLineSeriesStyleKnobs('Line 2 style', 'line2', 'blue', 'violet', 2, 5, 4);
 
-    const chartTheme = {
-      ...LIGHT_THEME,
-      lineSeriesStyle: generateLineSeriesStyleKnobs('chartTheme'),
-    };
+    const chartTheme = mergeWithDefaultTheme({
```
nickofthyme: No longer need to `mergeWithDefaultTheme`

stories/styling.tsx
```diff
@@ -617,19 +596,28 @@ storiesOf('Stylings', module)
...
   })
   .add('custom series styles: area', () => {
-    const chartTheme = {
-      ...LIGHT_THEME,
-      areaSeriesStyle: generateAreaSeriesStyleKnobs('chartTheme'),
-    };
+    const applyLineStyles = boolean('apply line series style', true, 'Chart Global Theme');
 
-    const useOnlyChartTheme = boolean('ignore series style (use only chart theme)', false, 'chartTheme');
+    const chartTheme = mergeWithDefaultTheme({
```
nickofthyme: No longer need to `mergeWithDefaultTheme`

src/lib/themes/theme.ts
markov00: I will leave this as it is for the moment, and we can find a clear type later. You are completely right about the fact that a common styling type is required, I will like to open a PR with that in the next future https://github.com/elastic/elastic-charts/issues/264

PR #259, opened 2019-07-03
fix: position tooltip within chart with single value xScale

### LINKED ISSUE #254: Single datum chart on element hover tooltip overflows the chart
**Describe the bug**
When we only have a single datum chart, the tooltip that appears on hover is positioned either to the left or the right of the element, pushing it outside of the chart.

**To Reproduce**
1. create single datum chart
2. hover over chart element
3. see tooltip overflowing chart

**Expected behavior**
Tooltip should be visible within the chart.  For single datum charts, we probably want the tooltip to follow the cursor within the element.

**Screenshots**
single_datum_tooltip

 - OS: all
 - Browser all
 - Elastic Charts: 7.0.1

**Additional context**

**Errors in browser console**
none

**Kibana Cross Issues**
n/a

**Checklist**

emmacunningham:
## Summary

fix #254 

This PR introduces a fix for the tooltip positioning on a chart with a single value xScale.  

I'm not entirely sure where the best place to position the tooltip, so this presents the minimal effort version, where the tooltip will be positioned inside of the chart aligned along the chart's edge (left if horizontal rotation, top if vertical rotation):

single_value

Implementation-wise, we've added a method `isSingleValue` to the `Scale` classes so that we can check if a scale is a single value scale (so that we can apply the different positioning in that case).  The `customDomain` knob was added to the stories to demonstrate that, when the chart only has one datapoint but has a multiple value scale, the tooltips are positioned as usual (since there is room to display them as normal).

### Checklist

### DIFF
--- a/stories/bar_chart.tsx
+++ b/stories/bar_chart.tsx
@@ -874,19 +874,16 @@ storiesOf('Bar Chart', module)
       0,
     );
 
-    const theme = mergeWithDefaultTheme(
-      {
-        scales: {
-          barsPadding: number('bars padding', 0.25, {
-            range: true,
-            min: 0,
-            max: 1,
-            step: 0.1,
-          }),
-        },
+    const theme = {
+      scales: {
+        barsPadding: number('bars padding', 0.25, {
+          range: true,
+          min: 0,
+          max: 1,
+          step: 0.1,
+        }),
       },
-      LIGHT_THEME,
-    );
+    };
 
     return (
       <Chart className={'story-chart'}>
@@ -926,19 +923,16 @@ storiesOf('Bar Chart', module)
       0,
     );
 
-    const theme = mergeWithDefaultTheme(
-      {
-        scales: {
-          barsPadding: number('bars padding', 0.25, {
-            range: true,
-            min: 0,
-            max: 1,
-            step: 0.1,
-          }),
-        },
+    const theme = {
+      scales: {
+        barsPadding: number('bars padding', 0.25, {
+          range: true,
+          min: 0,
+          max: 1,
+          step: 0.1,
+        }),
       },
-      LIGHT_THEME,
-    );
+    };
 
     return (
       <Chart className={'story-chart'}>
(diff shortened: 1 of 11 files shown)


markov00:
hey @emmacunningham just a suggestion, what if in the "single value" mode we switch between the "crosshair/vertical" mode and we manually enable the "follow" mode?

stories/bar_chart.tsx
```diff
@@ -855,9 +855,42 @@ storiesOf('Bar Chart', module)
...
+      {
+        '0 deg': 0,
+        '90 deg': 90,
+        '-90 deg': -90,
+        '180 deg': 180,
+      },
+      0,
+    );
+
+    const theme = mergeWithDefaultTheme(
```
nickofthyme: No need to use `mergeWithDefaultTheme` anymore. Will merge with `LIGHT_THEME` by default.

stories/bar_chart.tsx
```diff
@@ -872,11 +905,63 @@ storiesOf('Bar Chart', module)
...
```
nickofthyme: Same here

stories/bar_chart.tsx
emmacunningham: https://github.com/elastic/elastic-charts/pull/259/commits/c2c13bd9e1e2aa97daa11f6eade317e4d80325ce

stories/bar_chart.tsx
emmacunningham: https://github.com/elastic/elastic-charts/pull/259/commits/c2c13bd9e1e2aa97daa11f6eade317e4d80325ce

emmacunningham:
> hey @emmacunningham just a suggestion, what if in the "single value" mode we switch between the "crosshair/vertical" mode and we manually enable the "follow" mode?

@markov00 yes, good idea.  have implemented that logic here: https://github.com/elastic/elastic-charts/pull/259/commits/121a8fb8cf96bb633f3fdb0d041044462de5f475

PR #261, opened 2019-07-05
fix(last_value): compute last value for non stacked series

markov00:
## Summary

Fix an error introduced in #250 where the last value on the legend was undefined for non stacked series.

When using the stack percentage mode, the data values are computed on a percentage scale. As we want to display the percent value and not the original value when displaying a tooltip, I've changed the place where the lastValue value is captured: from the original dataset to the formatted one.

The commit fix the the fact that the last value was computed only for stacked series.

### Checklist

### DIFF
--- a/src/lib/series/series.ts
+++ b/src/lib/series/series.ts
@@ -89,7 +89,6 @@ export function splitSeries(
   rawDataSeries: RawDataSeries[];
   colorsValues: Map<string, any[]>;
   xValues: Set<any>;
-  splitSeriesLastValues: Map<string, any>;
 } {
   const { xAccessor, yAccessors, y0Accessors, splitSeriesAccessors = [] } = accessors;
   const colorAccessors = accessors.colorAccessors ? accessors.colorAccessors : splitSeriesAccessors;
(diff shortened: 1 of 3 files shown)


markov00:
> Looks like something weird is going on when the series is clustered but also `stackAsPercentage={true}`:

Are you referring to the legend showing 600 % and 200 % ? I will update the storybook for that, the wrong thing here is that I'm multipling by 100 the value on the `tickFormatter` either for the stacked as percentage mode and for non stacked one

emmacunningham:
> > Looks like something weird is going on when the series is clustered but also `stackAsPercentage={true}`:
> 
> Are you referring to the legend showing 600 % and 200 % ? I will update the storybook for that, the wrong thing here is that I'm multipling by 100 the value on the `tickFormatter` either for the stacked as percentage mode and for non stacked one

Yes, the legend display value is what I was referring to.  I wasn't exactly sure how we want to handle this since `clustered` + `stackAsPercentage` isn't an ideal configuration, but I see now how that's just a Storybook configuration thing that can be easily solved :D

PR #271, opened 2019-07-24
feat(bar_chart): color overrides

nickofthyme:
## Summary

Allow user to override colors of given datum based on accessor function.

```ts
interface RawDataSeriesDatum {
  x: number | string;
  y1: number | null;
  y0?: number | null;
  datum?: any;
}
interface GeometryId {
  specId: SpecId;
  seriesKey: any[];
}
type Color = string;
type StyleOverride = RecursivePartial<BarSeriesStyle> | Color | null;
type StyleAccessor = (datum: RawDataSeriesDatum, geometryId: GeometryId) => StyleOverride;
```

BREAKING CHANGE:  Replace ColorAccessor prop and functionality with StyleAccessor

Screen Recording 2019-07-24 at 02 02 PM

#216

### Checklist


nickofthyme:
I didn't label these changes as breaking because I don't think anyone is using the `ColorAccessors` prop of the `SeriesSpec`. But technically, I am removing `ColorAccessors` in place of `ColorAccessor`.

src/chart_types/xy_chart/rendering/rendering.ts
```diff
@@ -113,6 +115,33 @@ export function mutableIndexedGeometryMapUpsert(
   }
 }
 
+export function getStyleOverrides(
+  styleAccessor: StyleAccessor | null = null,
```
markov00: Since styleAccessor is defined as optional on the spec, I will prefer to have the same here as it will never be null.
I know that you are checking it with `styleAccessor &&` but I think it's better to keep the same typing and don't mix them other types.
You can easily move the argument as the last in the list and set it as optional.

src/chart_types/xy_chart/rendering/rendering.ts
```diff
@@ -113,6 +115,33 @@ export function mutableIndexedGeometryMapUpsert(
...
+  datum: DataSeriesDatum,
+  geometryId: GeometryId,
+  seriesStyle: BarSeriesStyle,
+): BarSeriesStyle {
+  const styleOverride = styleAccessor && styleAccessor(datum, geometryId);
+
+  if (styleOverride) {
```
markov00: Can we apply the return early practice here: https://github.com/elastic/kibana/blob/master/style_guides/js_style_guide.md#returnthrow-early-from-functions

src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -97,7 +101,7 @@ export interface SeriesAccessors {
   /** An array of fields thats indicates the stack membership */
   stackAccessors?: Accessor[];
   /** An optional array of field name thats indicates the stack membership */
```
markov00: can you update also the docs here?

src/chart_types/xy_chart/rendering/rendering.ts
nickofthyme: Ya I good point I just wanted to avoid reordering the args but that is fair

src/chart_types/xy_chart/rendering/rendering.ts
nickofthyme: The `else` statement should actually return the result of `mergePartial` directly, I'll fix that.

But were you referring to doing something like this 👇?

```ts
if (!styleOverride) {
  return seriesStyle;
}

// other conditions
```

markov00: yes

src/chart_types/xy_chart/utils/specs.ts
nickofthyme: see d5ae861

src/chart_types/xy_chart/rendering/rendering.ts
nickofthyme: see d5ae861

src/chart_types/xy_chart/rendering/rendering.ts
nickofthyme: see d5ae861

nickofthyme:
@markov00 Could you review this again when you have time?

PR #274, opened 2019-07-30
fix(tooltip): fix overflow for long series names

### LINKED ISSUE #270: Tooltip overflowing on long series names
**Describe the bug**
The tooltip overflows if the series names get too long.

**Expected behavior**
Should either wrap or expand - probably wrap. Current Visualize has the same problem, this is currently fixed in the PR https://github.com/elastic/kibana/pull/41703

**Screenshots**
Screenshot 2019-07-24 at 10 53 34

**Checklist**

nickofthyme:
## Summary

Allow long series names in tooltip to wrap to next line

fix #270

### Before
Screen Shot 2019-07-30 at 12 51 08 PM

### After
Screen Shot 2019-07-30 at 1 57 12 PM

### Checklist

### DIFF
--- a/src/components/_tooltip.scss
+++ b/src/components/_tooltip.scss
@@ -10,13 +10,13 @@
   table {
     border-collapse: collapse;
     border-spacing: 0;
+    table-layout: fixed;
+    width: 100%;
+
     td,
     th {
       padding: 3px;
     }
-  }
-  table {
-    width: 100%;
   }
 }
 
@@ -25,10 +25,13 @@
   padding: $euiSizeXS ($euiSizeXS * 2);
 }
 
-.echTooltip__label {
-  @include euiTextTruncate;
-  // Border indicates series color
+.echTooltip__cell {
+  display: flex;
   border-left: $euiSizeXS solid transparent;
+}
+
+.echTooltip__label {
+  flex: 1;
 }
 
 .echTooltip__value {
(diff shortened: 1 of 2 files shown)


src/components/_tooltip.scss
```diff
@@ -10,35 +10,38 @@
   table {
     border-collapse: collapse;
     border-spacing: 0;
+    table-layout: fixed;
```
nickofthyme: NOTE: `<table>` does __NOT__ respect relative widths without this being `table-layout: fixed`.

nickofthyme:
Tested for long URLs and Long descriptions with word breaks...

image

markov00:
@nickofthyme changing the tooltip from table to list makes every line on the tooltip independent resulting in different width of the value column:

**from (master)**
Screenshot 2019-08-02 at 15 51 02

**to**
Screenshot 2019-08-02 at 15 50 04

I just want to ask @cchaos what she think about that. For sure is more compact and nice but I'd like a design opinion on that

cchaos:
I don't have a strong opinion either way, however if there's space to show more of the series label then it's nice to use it.

nickofthyme:
@markov00 Good catch! As long as @cchaos is cool with it.

I noticed that the legend does this too so I was trying to resemble that UI.
image

PR #281, opened 2019-08-02
feat: new chart API

markov00:
### Todos

- [x] reenable legend values
- [x] reenable external cursor events/share cursor position
- [x] ~check  deselectedDataSeries: DataSeriesColorsValues[];~
- [x] `isCursorOnChart`
- [x] fix brush end

## Breaking Changes

- `SpecId`,`AxisId`, `AnnotationId` types are down-casted to a `string` type. The `getSpecId`, `getAxisId`m `getAnnotationId` methods still exist and but return just the same passed string until deprecated in a future version.
- The spec ids, previously `id`, `axisId`,`annotationId` etc are now aligned to use the same prop name: `id`
- The chart rendering status `data-ech-render-complete` and `data-ech-render-count` is no more at the root level of the `echChart` div, but on its child element: `echChartStatus` 

This PR refactor the current MobX implementation of elastic-chart in favour of redux and a better API to include new chart types.

### GlobalChartState
The new redux state is configured as the following:
```ts
export interface GlobalChartState {
  // an unique ID for each chart used by re-reselect to memoize selector per chart
  chartId: string;
  // true when all all the specs are parsed ad stored into the specs object
  specsInitialized: boolean;
  // true if the chart is rendered on dom
  chartRendered: boolean;
  // incremental count of the chart rendering
  chartRenderedCount: number;
  // the map of parsed specs
  specs: SpecList;
  // the chart type depending on the used specs
  chartType: ChartType | null;
  // a chart-type-dependant class that is used to render and share chart-type dependant functions
  internalChartState: InternalChartState | null;
  // the dimensions of the parent container, including the legend
  parentDimensions: Dimensions;
  // the state of the interactions
  interactions: InteractionsState;
  // external event state
  externalEvents: ExternalEventsState;
}
```

That global state maintain mainly the `specs` map (parsed from the `<SpecParser />` component, similarly to the previous MobX implementation. Each spec is now stored into a plain JS object where each key is the id of the `Spec`.
A `Spec` has now few more properties
```ts
export interface Spec {
  /** unique Spec identifier */
  id: string;
  /** Chart type define the type of chart that use this spec */
  chartType: ChartType;
  /** The type of spec, can be series, axis, annotation, settings etc*/
  specType: string;
}
```

### New Chart types

Each `ChartType` needs one or more `Spec` and needs to register its own `InternalChartState` class into the `initInternalChartState` function of `src/state/chart_state.ts`.
The `InternalChartState` class follow this interface:

```ts
export interface InternalChartState {
  // the chart type
  chartType: ChartType;
  // returns a JSX element with the chart rendered (lenged excluded)
  chartRenderer(containerRef: BackwardRef, forwardStageRef: RefObject<Stage>): JSX.Element | null;
  // true if the brush is available for this chart type
  isBrushAvailable(globalState: GlobalChartState): boolean;
  // true if the chart is empty (no data displayed)
  isChartEmpty(globalState: GlobalChartState): boolean;
  // return the list of legend items
  getLegendItems(globalState: GlobalChartState): Map<string, LegendItem>;
  // return the list of values for each legend item
  getLegendItemsValues(globalState: GlobalChartState): Map<string, TooltipLegendValue>;
  // return the CSS pointer cursor depending on the internal chart state
  getPointerCursor(globalState: GlobalChartState): string;
}
```

close https://github.com/elastic/elastic-charts/issues/75

### Checklist


monfera:
First of all, great changes overall toward data flow simplification! 

There are a couple of places where some more uncoupling could be done, here's the main one: likely because of what chart types have been covered so far, `Chart` (`src/components/chart.tsx`) now depends on and builds an `xy_chart` (ie. stuff under `src/chart_types/xy_chart`) even when instantiating something like
```
<Chart>...
  <Pie...>
</Chart>
```
where `Pie` is in a separate directory under `/chart_types/` (even appears to work as long as structural typing is observed but won't use the code under `/chart_types/circline`)
  
 As `Chart` is a quite general name, and likely meant to be a top-level container for all things in `elastic-charts` from the API viewpoint, this coupling would ideally give way to a plurality of projection types incl. non-Cartesian ones. An alternative is to rename it eg. `CartesianChart`, and there may be other ways already thought out, or we all could brainstorm

monfera:
The other uncoupling question/suggestion is about whether certain computations could be separated from views that they are often associated with. For example, `get_cursor_band.ts` or `get_tooltip_values_highlighted_geoms.ts` seems to depend on axes, which are basically, things to render only. So maybe it could be split to two parts: the computational part would have the responsibility of returning which value, band, point etc. the cursor is over; and things like the Axis, Crosshair and Tooltip would merely be responsible for rendering. Ie. no sideways communication. I see the attraction, and how a coupling can evolve historically, eg. we need a crosshair and need to compute where the cursor is in data space, and then later we add tooltips, which can reuse some values already computed for the crosshair. It'd be neat to separate because eg. some charts may need tooltips that don't need crosshair.

Another example is `get_tooltip_values_highlighted_geoms.ts` and `tooltip.ts`. Here it's coupled, and data flow is happening sideways, because the tooltip takes formatting spec defaults from the axes (maybe other things happen too or there's a deeper explanation). Perhaps the default formatters could belong to the projection (each scale, eg. `xScale` and `yScale`) rather than the axes, or at least it'd be neat to reconsider data transfers between sibling components like tooltip vs axis. 

I'm not necessarily suggesting that instead of an `<Axis...>` tag, that currently specifies a `tickFormatter`, we mandate library users to use some newfangled, and for many users, hard to grasp `<Projection...>` tag under `<Chart>`, even though it'd make our technical work easier. But maybe it's possible to make it so that the `<Tooltip>` component need not depend on `<Axis>` because from the tooltip viewpoint, the `tickFormat` comes from "upward". For a hypothetical example, if later on, someone decides that the default `tickFormat` needs to come from somewhere else, eg. `<Projection>` or `<Formatters>` or `<Config>` or whatever, the tooltip code shouldn't need to change as where the spec or default comes _from_ should be outside the concern of the `Tooltip` component itself.  Btw. I might be off base as I'm new to the code, or there may be obstacles to such separation; or maybe such changes have already been planned, as it's a WIP PR

monfera:
Likely a known issue: apparently randomly, storybook examples fail:
```
Cannot read property 'legendPosition' of undefined
TypeError: Cannot read property 'legendPosition' of undefined
    at http://localhost:9001/main.08acef345097e16e8e69.bundle.js:10032:57
```
then it may work on another try. Could be some race condition

markov00:
> There are a couple of places where some more uncoupling could be done, here's the main one: likely because of what chart types have been covered so far, `Chart` (`src/components/chart.tsx`) now depends on and builds an `xy_chart` (ie. stuff under `src/chart_types/xy_chart`) even when instantiating something like

You are right. as this is a WIP PR there still some refactoring that I need to go through before making it really available for review. Right not `Chart` depends on `xy_chart` for some event actions that needs to be moved into the XY chart process itself:
```
 this.chartStore.subscribe(() => {
      const state = this.chartStore.getState();
      const settings = getSettingsSpecSelector(state);
      if (this.state.legendPosition !== settings.legendPosition) {
        this.setState({
          legendPosition: settings.legendPosition,
        });
      }
      if (!isInitialized(state) || state.chartType !== ChartTypes.XYAxis) {
        return;
      }
      onElementOverCaller(state);
      onElementOutCaller(state);
      onElementClickCaller(state);
    });
```
IMHO I prefer to keep the main root component as `Chart` because I don't want to provide N multiple container classes that need to be changed depending on the chart type. When I will move the dependency into the XY chart there will no need to rename it in any other way.

monfera:
Definitely not news to you but I was curious, all MobX dependencies can now be removed  🎉
```
index b97abbc..5eabf14 100644
--- a/package.json
+++ b/package.json
@@ -139,8 +139,6 @@
     "fp-ts": "^1.14.2",
     "konva": "^2.6.0",
     "luxon": "^1.11.3",
-    "mobx": "^4.9.2",
-    "mobx-react": "^5.4.3",
     "newtype-ts": "^0.2.4",
```

monfera:
Hover, click events on legends don't show up in Storybook not sure if it's to do with the code change or something that changed in relation to Storybook (I suspect the latter due to test cases passing)
image

monfera:
Brush select also doesn't report events in Storyboard

monfera:
Crosshairs events don't get reported in either [here](https://elastic.github.io/elastic-charts/?path=/story/interactions--crosshair-with-time-axis) or in this branch, so maybe it's not meant to or regressed earlier

monfera:
Render change action is unreported as well in Storybook

yarn.lock
```diff
@@ -2,52 +2,25 @@
 # yarn lockfile v1
```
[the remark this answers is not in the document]
markov00: it's committed, maybe I didn't understood your statement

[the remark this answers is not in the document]
markov00: ok I've seen what you mean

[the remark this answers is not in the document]
markov00: updated on cf1a93b

markov00:
> Hover, click events on legends don't show up in Storybook not sure if it's to do with the code change or something that changed in relation to Storybook (I suspect the latter due to test cases passing)
> image

@monfera There was few missing things on this. I've fixed it on 50c6997
I've re-enabled the listeners + add the click event (not available in the current mobx version) and added few tests

.playground/webpack.config.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
         exclude: /node_modules/,
         options: {
           configFile: 'tsconfig.json',
+          transpileOnly: true,
```
monfera: This makes an amazing speed difference to Playground, great find! 🎉

package.json
```diff
@@ -153,10 +155,14 @@
     "mobx-react": "^5.4.3",
```
monfera: Please `yarn remove mobx mobx-react` 😄

src/chart_types/xy_chart/state/selectors/compute_brush_extent.ts
```diff
@@ -0,0 +1,13 @@
...
+import { getSettingsSpecSelector } from '../../../../state/selectors/get_settings_specs';
+import { computeChartTransformSelector } from './compute_chart_transform';
+
+export const computeBrushExtentSelector = createCachedSelector(
+  [computeChartDimensionsSelector, computeChartTransformSelector, getSettingsSpecSelector],
+  (chartDimensions, chartTransform, settingsSpecs): BrushExtent => {
+    // console.log('--- 8 computeBrushExtentSelector ---');
+    return computeBrushExtent(chartDimensions.chartDimensions, settingsSpecs.rotation, chartTransform);
+  },
+)((state) => state.chartId);
```
monfera: For consideration for a future, non-time-critical PR: `reselect` is great and one of its benefits is that it uses a lifting pattern (it lifts an ordinary function that works on regular values to a level where it'll work with the input of other selectors), though its syntax hides it. `reselect` has various alternatives so it'd be useful to eventually abstract it out and not necessarily use its syntax. The constituents could be broken up a bit more too. An example (not sure if I got all parts right):

```
const chartId = (state) => state.chartId;

const brushExtentSelector = select(
  (chartDimensions, chartTransform, settingsSpecs): BrushExtent =>
    computeBrushExtent(chartDimensions.chartDimensions, settingsSpecs.rotation, chartTransform),
)(computeChartDimensionsSelector, computeChartTransformSelector, getSettingsSpecSelector);

export const computeBrushExtentSelector = brushExtentSelector(chartId);
```

or with less verbose names:

```
const extent = select(
  (dimensions, transform, settings): BrushExtent =>
    brushExtent(dimensions.chartDimensions, settings.rotation, transform),
)(dimensions, transform, settings);

export const brushExtent = extent(chartId);
```
```

**Comment by markov00** [comment]
> Brush select also doesn't report events in Storybook (events get reported on the time charts though)

Fixed on c18ba7e
I've also fixed the brush extents on bar charts

File: src/chart_types/xy_chart/state/selectors/compute_brush_extent.ts (hunk 4 above)
Comment (monfera) [reply src/chart_types/xy_chart/state/selectors/compute_brush_extent.ts]: ... my phrasing was unclear, names that tell what the function is about, preferably with no or very few, widely used abbreviations, are a good thing and this can make function names a bit long. I think it's an OK price to pay as the code is more self-documenting, a little bit closer to literate programming, and it makes the code more accessible. It's the little pre-and postfix words like compute, get, selector etc. that can be eliminated (my example is guilty of some 😄)

Also, in my last example, the `getExtent` is not only noisily named; it also looks like a shallow wrapper over the Real Deal, which is `brushExtent()`. Sometimes such shallow (un)wrappers are useful, but too many of them may just add noise. So it's possible that a better example from me would've been a very slight refactor of `brushExtent` (=`computeBrushExtent`) such that we can do
```
const chartId = (state) => state.chartId;

const extent = select(brushExtent)(dimensions, transform, settings);

export const brushExtent = extent(chartId);
```
or, if the rotation really needs to be unwrapped,
```
const chartId = (state) => state.chartId;

const rotation = (transform) => transform.rotation;
const chartDimensions = (dimensions) => dimensions.chartDimensions;

const extent = select(brushExtent)(
  select(chartDimensions)(dimensions),
  select(rotation)(transform), 
  settings
);

export const brushExtent = extent(chartId);
```
Or some variation, eg. with `const rotation = select(transform => transform.rotation);`

Also, I forgot to add that other tools that do very similar lifting are Flyd, most.js, crosslink etc. so the data piping logic can be tweaked (eg. performance optimized) with no change to the userland code.

File: src/chart_types/xy_chart/state/selectors/compute_brush_extent.ts [hunk 5]
```diff
@@ -0,0 +1,13 @@
+import createCachedSelector from 're-reselect';
+import { computeBrushExtent, BrushExtent } from '../utils';
+import { computeChartDimensionsSelector } from './compute_chart_dimensions';
+import { getSettingsSpecSelector } from '../../../../state/selectors/get_settings_specs';
+import { computeChartTransformSelector } from './compute_chart_transform';
+
+export const computeBrushExtentSelector = createCachedSelector(
```
monfera: Maybe I'm looking at it the wrong way, but don't immediately see where `computeBrushExtentSelector` is used

monfera: Maybe I'm looking at it the wrong way, but don't immediately see where `computeBrushExtentSelector` is used. Also, some other selectors eg. `isBrushingEnabledSelector`, `isCrosshairVisibleSelector`, `isLegendVisibleSelector`, `isTooltipEnabledSelector` and some non-selectors seem to be unused, eg.  `createOnChartRenderedCalled`, `buildArcRenderProps`, `isSettingsSpec` and tens of others. Based on their names and them being nested deep in code structures, they don't look like part of an API, nor are there tests. There's also a possibility that some unused exports might not show up as unused because they're covered with a test case

package.json
```diff
@@ -153,10 +155,14 @@
...
     "newtype-ts": "^0.2.4",
     "prop-types": "^15.7.2",
+    "re-reselect": "^3.4.0",
```
monfera: `reselect` caches the value for the last call only (by default, based on a shallow equality check).  

I haven't immediately seen a default cache size limitation in the `re-reselect` doc, is there something in the doc about a default limit? Question 3 [in the FAQ])https://github.com/toomuchdesign/re-reselect#faq) says that if we want to limit the number of cache entries, a `cacheObject` is needed, eg. `LruObjectCache`, but there don't appear to be any reference to these. It could mean that there's a memory leak, with memory growing over use. I haven't confirmed it and depending on selector update types and frequencies, it may or may not matter, but setting a limit may be good insurance

src/chart_types/xy_chart/state/selectors/compute_brush_extent.ts
markov00: There can be for sure some unused selector here and there, I will check and remove them

package.json
markov00: Good catch, it's using `FlatObjectCache` that can be "unlimited" depending on key.
The key used is the `chartId` that can be passed on the component or generated on every new chart instantiation. Using the `LruObjectCache` can be the right approach. What could be the cacheSize? we need to think to keep at least the number of charts in a dashboard so something around 10 to 50 visualizations?

monfera: One use for the the earlier suggestion of abstracting around the selectors with eg. `select`- instead of importing it directly from `re-reselect` in a bunch of files -  is that we could try what `re-reselect` buys us right now. For example, does performance deteriorate if we switch back to `reselect`.  

- are we utilizing the cache, given that the [doc](https://github.com/toomuchdesign/re-reselect#re-reselect-solution) says "[`cacheKey` is] by default a string or number but it can be anything depending on the chosen cache strategy (see cache objects docs)" while our selectors tend to use a good number of objects?
- if we are utilizing the cache, are we realizing the benefits of `re-reselect`, ie. does it lead to noticeably or measurably better performance?

On the offchance we're not utilizing the cache or it doesn't bring us benefits, we might want to temporarily revert to basic `reselect` until further measures. Should be a change of a few lines after abstracting out the selector.

One concern about abstracting out the selector could be that in the future, if/when we need it, we might want to use various caching strategies, ie. a variety of selectors could be needed. And it's a possible need in the future, eg. use a different strategy when the selector can be cached by primitive key vs. shallow object vs. deep object (eg. with expensive computations, it may be worth doing a deep equality check on a small object). But nothing stops us from having one `select` function that takes an optional second argument as the selector strategy, eg. 
```
import { select } from `ourSelectorAbstraction`;
...
const hoveredDataPoint = select(xyToDataPoint, {
      selector: `re-reselect`, 
      keys: `primitives`, 
      strategy: `LRU`, 
      limit: 100
  })(x, y);
```
or, to cut down on noise, via abstracting out common strategies, and using pre-configured shorthand literals, eg.
```
const hoveredDataPoint = select(xyToDataPoint, primitiveLruCache)(x, y);
```

src/chart_types/xy_chart/state/selectors/compute_brush_extent.ts
markov00: I've cleaned the code a bit, removing most of the unused functions and some exports in e585909

package.json
markov00: done :D

.playground/webpack.config.js
markov00: 👍

markov00:
> Render change action is unreported as well in Storybook

fixed on 0d93c57

markov00:
> Crosshairs events don't get reported in either [here](https://elastic.github.io/elastic-charts/?path=/story/interactions--crosshair-with-time-axis) or in this branch, so maybe it's not meant to or regressed earlier

We don't have events from the crosshair itself, but there are events for what we call cursor position (https://elastic.github.io/elastic-charts/?path=/story/interactions--cursor-update-action)

src/chart_types/xy_chart/state/selectors/compute_brush_extent.ts
monfera: Via discussion, resolved as done by someone in a subsequent PR

package.json
monfera: Resolved as @markov00 making a subsequent issue or PR for eg. a LRU cache with a limit; also, my initial concern was mitigated as I thought the proliferation of cached items will be pretty bad, but Marco explained about a key `re-reselect` point

monfera: Again, subsequent PR by any one of us

src/chart_types/xy_chart/state/selectors/compute_brush_extent.ts
monfera: There's still about a dozen; resolution is that someone (@nickofthyme?) would remove or comment out unused functions

src/chart_types/xy_chart/renderer/canvas/axis.tsx
```diff
@@ -227,4 +214,108 @@ export class Axis extends React.PureComponent<AxisProps> {
...
+          debug={debug}
+          chartDimensions={chartDimensions}
+        />,
+      );
+    });
+    return axesComponents;
+  }
+}
+
+const mapDispatchToProps = () => ({});
```
rshen91: This is more of a question than stating that this needs to be fixed as it doesn't seem linked to any major performance issues. I was seeing in a couple files, `mapDispatchToProps` that don't seem to be used/needed. One example is here in axis, and the others I found are in the following: bar_values.tsx, grid.tsx, annotation_tooltip.tsx, brush.tsx, crosshair.tsx, highlighter.tsx, tooltip.tsx, and chart_status.tsx. Do these need to be included in the `connect()` at all? Or is it included for a reason? Like the redux structure to be in place incase we would want to build on it later? https://stackoverflow.com/questions/41670146/how-does-connect-work-without-mapdispatchtoprops

markov00: fix in ae398b5

src/components/chart.tsx
```diff
@@ -30,70 +38,70 @@ interface ChartProps {
...
   private chartStageRef: React.RefObject<Stage>;
   constructor(props: any) {
     super(props);
     this.chartContainerRef = createRef();
     this.chartStageRef = createRef();
-    this.chartSpecStore = new ChartStore(props.id);
+
+    const storeReducer = chartStoreReducer(uuid.v4());
+    if (process.env.NODE_ENV !== 'production') {
+      this.chartStore = createStore(storeReducer, devToolsEnhancer({ trace: true }));
```
nickofthyme: 👍

src/components/chart.tsx
```diff
@@ -30,70 +38,70 @@ interface ChartProps {
...
+      legendPosition: Position.Right,
     };
 
-    this.chartSpecStore.chartInitialized.observe(({ newValue, oldValue }) => {
-      if (newValue !== oldValue) {
+    const onElementClickCaller = createOnElementClickCaller();
+    const onElementOverCaller = createOnElementOverCaller();
+    const onElementOutCaller = createOnElementOutCaller();
+    const onBrushEndCaller = createOnBrushEndCaller();
+    const onPointerMoveCaller = createOnPointerMoveCaller();
```
nickofthyme: I think this is the best way to do this

src/components/chart.tsx
```diff
@@ -150,34 +158,21 @@ export class Chart extends React.Component<ChartProps, ChartState> {
...
           <div className="echContainer">
-            <ChartResizer />
-            <Crosshair />
-            {// TODO reenable when SVG rendered is aligned with canvas one
-            renderer === 'svg' && <ChartContainer forwardRef={this.chartStageRef} />}
-            {renderer === 'canvas' && <ChartContainer forwardRef={this.chartStageRef} />}
-            <Tooltips getChartContainerRef={this.getChartContainerRef} />
-            <AnnotationTooltip getChartContainerRef={this.getChartContainerRef} />
-            <Highlighter />
+            <ChartContainer getChartContainerRef={this.getChartContainerRef} forwardStageRef={this.chartStageRef} />
```
nickofthyme: Nice!

src/components/chart_status.tsx
```diff
@@ -0,0 +1,39 @@
...
+  rendered: boolean;
+  renderedCount: number;
+  onRenderChange?: RenderChangeListener;
+}
+class ChartStatusComponent extends React.Component<ChartStatusStateProps> {
+  componentDidMount() {
+    const { onRenderChange, rendered } = this.props;
+    if (onRenderChange) {
+      onRenderChange(rendered);
+    }
```
nickofthyme: Could be helper function on component

src/components/icons/assets/dot.tsx
```diff
@@ -1,15 +1,17 @@
...
-    <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" {...extraProps}>
-      <defs>
-        <circle id="dot-a" cx="8" cy="8" r="4" />
-      </defs>
-      <g>
-        <use xlinkHref="#dot-a" />
-      </g>
-    </svg>
-  );
+export class DotIcon extends React.PureComponent<Props> {
```
nickofthyme: Is this for a reason? You could just use `memo` with the FC

src/components/icons/icon.tsx
```diff
@@ -31,25 +31,27 @@ export interface IconProps {
 
 export type Props = Omit<SVGAttributes<SVGElement>, 'color' | 'type'> & IconProps;
 
-export const Icon = (props: Props) => {
-  const { type, color, className, tabIndex, ...rest } = props;
-  let optionalCustomStyles = null;
+export class Icon extends React.PureComponent<Props> {
```
nickofthyme: Is this for a reason? You could just use `memo` with the FC

src/specs/settings.tsx
```diff
@@ -99,135 +93,48 @@ export interface SettingSpecProps {
...
-  if (onLegendItemMinusClick) {
-    chartStore.setOnLegendItemMinusClickListener(onLegendItemMinusClick);
-  }
-  if (onCursorUpdate) {
-    chartStore.setOnCursorUpdateListener(onCursorUpdate);
-  }
-  if (onRenderChange) {
-    chartStore.setOnRenderChangeListener(onRenderChange);
-  }
-}
```
nickofthyme: Nice!!

src/specs/specs_parser.tsx
```diff
@@ -1,26 +1,37 @@
...
-import { ChartStore } from '../chart_types/xy_chart/store/chart_state';
+import React, { useEffect } from 'react';
+import { bindActionCreators, Dispatch } from 'redux';
+import { connect } from 'react-redux';
+import { specParsed, specUnmounted } from '../state/actions/specs';
 
-export interface SpecProps {
-  chartStore?: ChartStore; // FIX
-}
+export const SpecsParserComponent: React.FunctionComponent<{}> = (props) => {
```
nickofthyme: I think the type here is not necessary 

```suggestion
export const SpecsParserComponent = (props: DispatchProps) => {
```

src/state/selectors/get_deselected_data_series.ts
```diff
@@ -0,0 +1,16 @@
+// import createCachedSelector from 're-reselect';
```
nickofthyme: ?

src/state/selectors/get_legend_size.ts
```diff
@@ -0,0 +1,96 @@
...
+  },
+)(getChartIdSelector);
+
+const MARKER_WIDTH = 16;
+// const MARKER_HEIGHT = 16;
+const MARKER_LEFT_MARGIN = 4;
+const VALUE_LEFT_MARGIN = 4;
+const VERTICAL_PADDING = 4;
+
+export const getLegendSizeSelector = createCachedSelector(
```
nickofthyme: Nice!

src/state/spec_factory.ts
```diff
@@ -0,0 +1,61 @@
...
+export function specComponentFactory<U extends Spec, D extends keyof U>(
+  defaultProps: Pick<U, D | 'chartType' | 'specType'>,
+) {
+  const spec = (props: U & DispatchProps) => {
+    const prevId = usePrevious(props.id);
+    const { removeSpec, upsertSpec, ...spec } = props;
+    useEffect(() => {
+      // console.log('updating specs');
+      if (prevId && prevId !== props.id) {
+        // console.log('remove existing specs, ', prevId);
```
nickofthyme:
```suggestion
```

src/state/spec_factory.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export function specComponentFactory<U extends Spec, D extends keyof U>(
+  defaultProps: Pick<U, D | 'chartType' | 'specType'>,
+) {
+  const spec = (props: U & DispatchProps) => {
+    const prevId = usePrevious(props.id);
+    const { removeSpec, upsertSpec, ...spec } = props;
+    useEffect(() => {
+      // console.log('updating specs');
```
nickofthyme:
```suggestion
```

src/state/spec_factory.ts
```diff
@@ -0,0 +1,61 @@
...
+      if (prevId && prevId !== props.id) {
+        // console.log('remove existing specs, ', prevId);
+        removeSpec(prevId);
+      }
+      upsertSpec(spec);
+    });
+    useEffect(
+      () => () => {
+        // console.log('remove spec on unmount', props.id);
```
nickofthyme:
```suggestion
```

src/utils/bbox/dom_text_bbox_calculator.ts
```diff
@@ -0,0 +1,39 @@
+import { Option, some } from 'fp-ts/lib/Option';
```
nickofthyme: What is `fp-ts` doing?

src/utils/ids.ts
```diff
@@ -1,25 +1,17 @@
...
-const annotationIdIso = iso<AnnotationId>();
-
-export function getGroupId(id: string): GroupId {
-  return groupIdIso.wrap(id);
+export function getGroupId(id: string): string {
+  return id;
 }
-export function getAxisId(id: string): AxisId {
-  return axisIdIso.wrap(id);
+export function getAxisId(id: string): string {
```
nickofthyme: Feels so good!

integration/page_objects/common.ts
```diff
@@ -21,7 +21,7 @@ class CommonPage {
 
     return `${baseUrl}?${query}${query ? '&' : ''}knob-debug=false`;
   }
-  async getBoundingClientRect(selector = '.echChart[data-ech-render-complete=true]') {
+  async getBoundingClientRect(selector = '.echChart') {
```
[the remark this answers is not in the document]
markov00: No this is because I've moved the `data render` attributes into a child element as written on the breaking changes notes in the PR description. The reason was to separate the two element to avoid unnecessary re-rendering

src/chart_types/index.ts
```diff
@@ -0,0 +1,5 @@
+export const ChartTypes = Object.freeze({
+  Global: 'global' as 'global',
+  Pie: 'pie' as 'pie',
+  XYAxis: 'xy_axis' as 'xy_axis',
+});
```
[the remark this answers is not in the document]
markov00: fixed in 5fc93cff

src/chart_types/xy_chart/legend/legend.ts
```diff
@@ -40,20 +40,31 @@ export function getPostfix(spec: BasicSeriesSpec): Postfixes {
...
+  axesSpecs: AxisSpec[],
+  deselectedDataSeries: DataSeriesColorsValues[] = [],
 ): Map<string, LegendItem> {
   const legendItems: Map<string, LegendItem> = new Map();
   const sortedSeriesColors = getSortedDataSeriesColorsValuesMap(seriesColor);
 
   sortedSeriesColors.forEach((series, key) => {
     const { banded, specId, lastValue, colorValues } = series;
-    const spec = specs.get(specId);
+    const spec = getSpecsById<BasicSeriesSpec>(specs, specId);
```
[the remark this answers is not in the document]
markov00: I'm not much a fan of classes, but we can create for sure a set of utilities for that!

src/chart_types/xy_chart/renderer/canvas/axis.tsx
```diff
@@ -9,13 +9,24 @@ import {
...
+import { Dimensions } from '../../../../utils/dimensions';
+import { connect } from 'react-redux';
+import { GlobalChartState } from '../../../../state/chart_state';
+import { computeAxisVisibleTicksSelector } from '../../state/selectors/compute_axis_visible_ticks';
+import { getAxisSpecsSelector } from '../../state/selectors/get_specs';
+import { AxisId } from '../../../../utils/ids';
+import { computeAxisTicksDimensionsSelector } from '../../state/selectors/compute_axis_ticks_dimensions';
+import { getChartThemeSelector } from '../../../../state/selectors/get_chart_theme';
+import { computeChartDimensionsSelector } from '../../state/selectors/compute_chart_dimensions';
+import { LIGHT_THEME } from '../../../../utils/themes/light_theme';
```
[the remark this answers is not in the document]
markov00: Absolutely, I've added aliases at the beginning but it creates a bit of issues sometimes when moving a lot of files and I always have to restart TS Server on vscode for that

src/chart_types/xy_chart/renderer/canvas/reactive_chart.tsx
```diff
@@ -0,0 +1,390 @@
...
+
+    return [
+      {
+        element,
+        zIndex: 0,
+      },
+    ];
+  };
+
+  // getAxes = (): AxisProps[] => {
```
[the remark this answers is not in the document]
markov00: fix in a2540567

src/chart_types/xy_chart/renderer/canvas/index.tsx
```diff
@@ -0,0 +1,25 @@
+// renderBrushTool = () => {
```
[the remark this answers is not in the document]
markov00: removed in a2540567

src/chart_types/xy_chart/renderer/canvas/reactive_chart.tsx
```diff
@@ -0,0 +1,390 @@
...
+}
+interface ReactiveChartDispatchProps {
+  onChartRendered: typeof onChartRendered;
+}
+interface ReactiveChartOwnProps {
+  forwardStageRef: RefObject<Stage>;
+}
+interface ReactiveChartElementIndex {
+  element: JSX.Element;
+  zIndex: number;
```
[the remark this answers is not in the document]
markov00: It's mostly used to change the zIndex of the annotation in respect of the geometries. We can also apply it to the series if we want

src/chart_types/xy_chart/renderer/canvas/line_annotation.tsx
```diff
@@ -1,19 +1,18 @@
...
 }
 
 export class LineAnnotation extends React.PureComponent<LineAnnotationProps> {
   render() {
-    return this.renderAnnotation();
+    const { lines } = this.props;
+
+    return <Group key={'line_annotations'}>{lines.map(this.renderAnnotationLine)}</Group>;
   }
   private renderAnnotationLine = (lineConfig: AnnotationLineProps, i: number) => {
```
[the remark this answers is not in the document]
markov00: renamed in a697b928

src/chart_types/xy_chart/renderer/canvas/line_annotation.tsx
```diff
@@ -1,19 +1,18 @@
...
   lines: AnnotationLineProps[];
   lineStyle: LineAnnotationStyle;
 }
 
 export class LineAnnotation extends React.PureComponent<LineAnnotationProps> {
   render() {
-    return this.renderAnnotation();
+    const { lines } = this.props;
+
+    return <Group key={'line_annotations'}>{lines.map(this.renderAnnotationLine)}</Group>;
```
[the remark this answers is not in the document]
markov00: removed in a697b928 (together with the same in rect_annotations)

src/chart_types/xy_chart/renderer/canvas/reactive_chart.tsx
```diff
@@ -0,0 +1,390 @@
...
+import { PointGeometry, BarGeometry, AreaGeometry, LineGeometry } from '../../../../utils/geometry';
+import { LegendItem } from '../../../../chart_types/xy_chart/legend/legend';
+import { getSettingsSpecSelector } from '../../../../state/selectors/get_settings_specs';
+import { computeChartDimensionsSelector } from '../../state/selectors/compute_chart_dimensions';
+import { getChartContainerDimensionsSelector } from '../../../../state/selectors/get_chart_container_dimensions';
+import { Clippings } from './bar_values_utils';
+
+interface ReactiveChartStateProps {
+  initialized: boolean;
+  geometries: {
```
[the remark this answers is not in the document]
markov00: fix on 95276b69

src/chart_types/xy_chart/renderer/canvas/rect_annotation.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+    const rectProps = {
+      ...this.props.rectStyle,
+      x,
+      y,
+      width,
+      height,
+    };
+
+    return <Rect {...rectProps} key={`rect-annotation-${i}`} />;
```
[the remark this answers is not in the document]
markov00: what do you mean by directly? on props at time like `x={x} y={y} ` etc?

src/chart_types/xy_chart/renderer/dom/tooltips.tsx
```diff
@@ -67,22 +81,21 @@ class TooltipsComponent extends React.Component<TooltipProps> {
...
     const chartContainerRef = getChartContainerRef();
-    if (chartContainerRef.current === null || !isVisible) {
+    let tooltip;
+    if (chartContainerRef.current === null || !isTooltipVisible) {
       return null;
     } else {
       tooltip = (
         <div className="echTooltip" ref={this.tooltipRef}>
-          <div className="echTooltip__header">{this.renderHeader(tooltipData[0], tooltipHeaderFormatter)}</div>
+          <div className="echTooltip__header">{this.renderHeader(tooltipValues[0], tooltipHeaderFormatter)}</div>
```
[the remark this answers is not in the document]
markov00: I've opened an issue to track this: https://github.com/elastic/elastic-charts/issues/460

src/chart_types/xy_chart/rendering/rendering.areas.test.ts
```diff
@@ -1,19 +1,21 @@
-/* eslint @typescript-eslint/no-object-literal-type-assertion: off */
-import { computeSeriesDomains } from '../store/utils';
-import { getGroupId, getSpecId, SpecId } from '../../../utils/ids';
+/* eslint-disable @typescript-eslint/no-object-literal-type-assertion */
```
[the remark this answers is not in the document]
markov00: fix in 555db69c I've removed the eslint disable line without needs to use forceTypes in tests

src/chart_types/xy_chart/state/chart_state.tsx
```diff
@@ -0,0 +1,49 @@
...
+      <React.Fragment>
+        <Crosshair />
+        <ReactiveChart forwardStageRef={forwardStageRef} />
+        <Tooltips getChartContainerRef={containerRef} />
+        <AnnotationTooltip getChartContainerRef={containerRef} />
+        <Highlighter />
+        <BrushTool />
+      </React.Fragment>
+    );
+  }
```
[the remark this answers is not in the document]
markov00: your comment is much more like black magic ?

src/chart_types/xy_chart/specs/rect_annotation.tsx
```diff
@@ -1,33 +1,24 @@
...
-import { getGroupId } from '../../../utils/ids';
-import { SpecProps } from '../../../specs/specs_parser';
+import { specComponentFactory, getConnect } from '../../../state/spec_factory';
+import { DEFAULT_ANNOTATION_RECT_STYLE } from '../../../utils/themes/theme';
+import { ChartTypes } from '../../index';
 
-type RectAnnotationProps = SpecProps & RectAnnotationSpec;
+const defaultProps = {
+  chartType: ChartTypes.XYAxis,
+  specType: 'annotation' as 'annotation',
```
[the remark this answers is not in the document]
markov00: fix on 56992787

src/chart_types/xy_chart/specs/rect_annotation.tsx
```diff
@@ -1,33 +1,24 @@
...
+  groupId: DEFAULT_GLOBAL_ID,
+  annotationType: 'rectangle' as 'rectangle',
```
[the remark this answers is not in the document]
markov00: fix on 56992787

src/chart_types/xy_chart/specs/histogram_bar_series.tsx
```diff
@@ -1,42 +1,36 @@
...
 import { ScaleType } from '../../../utils/scales/scales';
-import { SpecProps } from '../../../specs/specs_parser';
+import { specComponentFactory, getConnect } from '../../../state/spec_factory';
+import { ChartTypes } from '../../../chart_types';
 
-type HistogramBarSpecProps = SpecProps & HistogramBarSeriesSpec;
+const defaultProps = {
+  chartType: ChartTypes.XYAxis,
+  specType: 'series' as 'series',
+  seriesType: 'bar' as 'bar',
```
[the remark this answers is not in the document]
markov00: fixed on 56992787

src/chart_types/xy_chart/state/chart_state.interactions.test.ts
```diff
@@ -0,0 +1,657 @@
...
+  });
+
+  test('store is correctly configured', () => {
+    // checking this to avoid broken tests due to nested describe and before
+    const seriesGeoms = computeSeriesGeometriesSelector(store.getState());
+    expect(seriesGeoms.scales.xScale).not.toBeUndefined();
+    expect(seriesGeoms.scales.yScales).not.toBeUndefined();
+  });
+
+  // test('set cursor from external source', () => {
```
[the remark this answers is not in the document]
markov00: fixed on 018883b9
I've skipped them but commented the inside code because it's related to the old implementation. I will take that in a future PR

src/chart_types/xy_chart/state/utils.ts
```diff
@@ -72,6 +64,34 @@ export interface GeometriesCounts {
...
 }
 
+export interface ComputedScales {
+  xScale: Scale;
+  yScales: Map<GroupId, Scale>;
+}
+
+export interface ComputedGeometries {
+  scales: ComputedScales;
+  geometries: {
```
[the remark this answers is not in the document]
markov00: fixed in 95276b696406b262fd6872ceffe0e4fd7a8add09

src/chart_types/xy_chart/state/selectors/merge_y_custom_domains.ts
```diff
@@ -0,0 +1,71 @@
...
+      };
+
+      if (isBounded(mergedDomain)) {
+        domainsByGroupId.set(groupId, mergedDomain);
+      }
+    } else {
+      domainsByGroupId.set(groupId, domain);
+    }
+  });
+  // console.log('--- 0 mergeYCustomDomains ---');
```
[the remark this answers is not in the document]
markov00: fixed in 76e7d346

src/chart_types/xy_chart/state/selectors/is_chart_animatable.ts
```diff
@@ -0,0 +1,17 @@
+import createCachedSelector from 're-reselect';
+import { getSettingsSpecSelector } from '../../../../state/selectors/get_settings_specs';
+import { computeSeriesGeometriesSelector } from './compute_series_geometries';
+import { getChartIdSelector } from '../../../../state/selectors/get_chart_id';
+// import { isChartAnimatable } from '../utils';
+
+export const isChartAnimatableSelector = createCachedSelector(
+  [computeSeriesGeometriesSelector, getSettingsSpecSelector],
+  () => {
+    // console.log('--- 13 isChartAnimatableSelector ---');
```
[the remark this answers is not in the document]
markov00: fixed in 76e7d346

src/chart_types/xy_chart/state/selectors/get_series_color_map.ts
```diff
@@ -0,0 +1,22 @@
...
+  [getSeriesSpecsSelector, computeSeriesDomainsSelector, getChartThemeSelector],
+  (seriesSpecs, seriesDomainsAndData, chartTheme): Map<string, string> => {
+    const updatedCustomSeriesColors = getUpdatedCustomSeriesColors(seriesSpecs);
+
+    const seriesColorMap = getSeriesColorMap(
+      seriesDomainsAndData.seriesColors,
+      chartTheme.colors,
+      updatedCustomSeriesColors,
+    );
+    // console.log('--- 2 computeSeriesDomainsSelector ---');
```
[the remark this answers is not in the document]
markov00: fixed in 76e7d346

src/chart_types/xy_chart/state/selectors/count_bars_in_cluster.ts
```diff
@@ -0,0 +1,15 @@
...
+import { countBarsInCluster } from '../../utils/scales';
+import { getChartIdSelector } from '../../../../state/selectors/get_chart_id';
+
+export const countBarsInClusterSelector = createCachedSelector(
+  [computeSeriesDomainsSelector],
+  (seriesDomainsAndData): number => {
+    const { formattedDataSeries } = seriesDomainsAndData;
+
+    const { totalBarsInCluster } = countBarsInCluster(formattedDataSeries.stacked, formattedDataSeries.nonStacked);
+    // console.log('--- 4 countBarsInCluster ---');
```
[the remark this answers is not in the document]
markov00: fixed in 76e7d346

src/chart_types/xy_chart/state/selectors/compute_series_geometries.ts
```diff
@@ -0,0 +1,48 @@
...
+    settingsSpec,
+    seriesSpecs,
+    seriesDomainsAndData,
+    seriesColorMap,
+    chartTheme,
+    chartDimensions,
+    axesSpecs,
+    isHistogramMode,
+  ): ComputedGeometries => {
+    // console.log('--- 9 computeSeriesGeometriesSelector ---');
```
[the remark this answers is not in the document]
markov00: fixed in 76e7d346

src/chart_types/xy_chart/state/selectors/compute_series_domains.ts
```diff
@@ -0,0 +1,24 @@
...
+import { SeriesDomainsAndData } from '../utils';
+import { GlobalChartState } from '../../../../state/chart_state';
+import { getChartIdSelector } from '../../../../state/selectors/get_chart_id';
+
+const getDeselectedSeriesSelector = (state: GlobalChartState) => state.interactions.deselectedDataSeries;
+
+export const computeSeriesDomainsSelector = createCachedSelector(
+  [getSeriesSpecsSelector, mergeYCustomDomainsByGroupIdSelector, getDeselectedSeriesSelector, getSettingsSpecSelector],
+  (seriesSpecs, customYDomainsByGroupId, deselectedDataSeries, settingsSpec): SeriesDomainsAndData => {
+    // console.log('--- 1 computeSeriesDomainsSelector ---', seriesSpecs);
```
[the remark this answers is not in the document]
markov00: fixed in 76e7d346

src/chart_types/xy_chart/state/selectors/compute_legend.ts
```diff
@@ -0,0 +1,40 @@
...
+  ],
+  (
+    seriesSpecs,
+    seriesDomainsAndData,
+    chartTheme,
+    seriesColorMap,
+    axesSpecs,
+    deselectedDataSeries,
+  ): Map<string, LegendItem> => {
+    // console.log('--- 3 computeLegend ---');
```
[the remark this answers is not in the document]
markov00: fixed in 76e7d346

src/chart_types/xy_chart/state/selectors/compute_chart_transform.ts
```diff
@@ -0,0 +1,13 @@
+import createCachedSelector from 're-reselect';
+import { computeChartTransform, Transform } from '../utils';
+import { computeChartDimensionsSelector } from './compute_chart_dimensions';
+import { getSettingsSpecSelector } from '../../../../state/selectors/get_settings_specs';
+import { getChartIdSelector } from '../../../../state/selectors/get_chart_id';
+
+export const computeChartTransformSelector = createCachedSelector(
+  [computeChartDimensionsSelector, getSettingsSpecSelector],
+  (chartDimensions, settingsSpecs): Transform => {
+    // console.log('--- 7 computeChartTransformSelector ---');
```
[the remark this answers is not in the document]
markov00: fixed in 76e7d346

src/chart_types/xy_chart/state/selectors/compute_axis_visible_ticks.ts
```diff
@@ -0,0 +1,60 @@
...
+    chartTheme,
+    settingsSpec,
+    axesSpecs,
+    axesTicksDimensions,
+    seriesDomainsAndData,
+    totalBarsInCluster,
+    isHistogramMode,
+    barsPadding,
+  ): AxisVisibleTicks => {
+    // console.log('--- 10 computeAxisVisibleTicksSelector ---');
```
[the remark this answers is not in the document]
markov00: fixed in 76e7d346

src/chart_types/xy_chart/state/selectors/compute_axis_ticks_dimensions.ts
```diff
@@ -0,0 +1,61 @@
...
+      );
+      if (
+        dimensions &&
+        (!settingsSpec.hideDuplicateAxes || !isDuplicateAxis(axisSpec, dimensions, axesTicksDimensions, axesSpecs))
+      ) {
+        axesTicksDimensions.set(id, dimensions);
+      }
+    });
+    bboxCalculator.destroy();
+    // console.log('--- 5 computeAxisTickDimensions ---');
```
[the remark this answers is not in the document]
markov00: fixed in 76e7d346

src/chart_types/xy_chart/state/selectors/compute_annotations.ts
```diff
@@ -0,0 +1,45 @@
...
+  (
+    annotationSpecs,
+    chartDimensions,
+    settingsSpec,
+    seriesGeometries,
+    axesSpecs,
+    totalBarsInCluster,
+    isHistogramMode,
+  ): Map<AnnotationId, AnnotationDimensions> => {
+    // console.log('--- 11 computeAnnotationDimensionsSelector ---', annotationSpecs);
```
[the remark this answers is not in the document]
markov00: fixed in 76e7d346

src/components/chart_status.tsx
markov00: fixed in 4c9c7a04

src/components/icons/assets/dot.tsx
markov00: Yes, but I've found that too late. I will take this in a future PR

src/components/icons/icon.tsx
markov00: Yes, but I've found that too late. I will take this in a future PR

src/state/spec_factory.ts
markov00: fixed in 76e7d34

src/state/spec_factory.ts
markov00: fixed in 76e7d34

src/state/spec_factory.ts
markov00: fixed in 76e7d34

src/utils/bbox/dom_text_bbox_calculator.ts
markov00: This is a nice functional programming library that provides some nice types. In this case returning an Option types mean that, on the result you can call `getOrElse` and specify a default instead of checking for null values or similar.
I've removed that here 14d7a946 because right now it's not really useful. I've moved the default return value (used by all the callers) inside the calculator function.

src/chart_types/xy_chart/annotations/line_annotation_tooltip.ts
```diff
@@ -0,0 +1,438 @@
...
+}
+
+export function getAnnotationLineTooltipXOffset(chartRotation: Rotation, axisPosition: Position): number {
+  let xOffset = 0;
+  const isChartHorizontalRotation = isHorizontalRotation(chartRotation);
+
+  if (isHorizontalAxis(axisPosition)) {
+    xOffset = isChartHorizontalRotation ? 50 : 0;
+  } else {
+    xOffset = isChartHorizontalRotation ? (axisPosition === Position.Right ? 100 : 0) : 50;
```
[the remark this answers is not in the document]
markov00: fixed in 9b0bb9fa

src/chart_types/xy_chart/annotations/rect_annotation_tooltip.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+  }
+  return {
+    isVisible: false,
+  };
+}
+
+export function isWithinRectBounds(
+  cursorPosition: Point,
+  { startX, endX, startY, endY }: { startX: number; endX: number; startY: number; endY: number },
```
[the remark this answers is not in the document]
markov00: fixed in 9b0bb9fa

src/chart_types/xy_chart/annotations/line_annotation_tooltip.ts
```diff
@@ -0,0 +1,438 @@
...
+
+/**
+ * Checks if the cursorPosition is within the line annotation marker
+ * @param cursorPosition the cursor position relative to the projected area
+ * @param marker the line annotation marker
+ */
+export function isWithinLineMarkerBounds(cursorPosition: Point, marker: AnnotationMarker): boolean {
+  const { top, left } = marker.position;
+  const { width, height } = marker.dimension;
+  const markerRect = { startX: left, startY: top, endX: left + width, endY: top + height };
```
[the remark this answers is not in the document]
markov00: fixed in 9b0bb9fa

src/chart_types/xy_chart/annotations/rect_annotation_tooltip.ts
```diff
@@ -0,0 +1,169 @@
...
+      };
+    }
+  }
+  return {
+    isVisible: false,
+  };
+}
+
+export function isWithinRectBounds(
+  cursorPosition: Point,
```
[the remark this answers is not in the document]
markov00: fixed in 9b0bb9fa

src/chart_types/xy_chart/renderer/canvas/rect_annotation.tsx
```diff
@@ -0,0 +1,29 @@
...
+  rects: AnnotationRectProps[];
+  rectStyle: RectAnnotationStyle;
+}
+
+export class RectAnnotation extends React.PureComponent<Props> {
+  render() {
+    const { rects } = this.props;
+    return <Group key={'rect_annotations'}>{rects.map(this.renderAnnotationRect)}</Group>;
+  }
+  private renderAnnotationRect = (props: AnnotationRectProps, i: number) => {
```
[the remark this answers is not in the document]
markov00: fix in 428151ae

src/specs/specs_parser.tsx
markov00: It's required to type the `children` property of props

src/state/selectors/get_deselected_data_series.ts
markov00: removed in 7d74a5a

markov00:
@nickofthyme I've fixed the Annotation tooltips here: 143276cc
I've left the `shift+click` for a future PR

src/chart_types/xy_chart/renderer/canvas/rect_annotation.tsx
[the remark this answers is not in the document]
nickofthyme: Yeah exactly. I just feel like defining the props that you then spread into the component is silly, when you can just set the props on the component itself. Then you will see what props are missing directly.

No need to fix now.

src/chart_types/xy_chart/rendering/rendering.areas.test.ts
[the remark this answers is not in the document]
nickofthyme: Yeah I think it wasn't the exact case here that I was referring to.

src/chart_types/xy_chart/specs/rect_annotation.tsx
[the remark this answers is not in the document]
nickofthyme: Love it!! : 😍

src/chart_types/xy_chart/state/chart_state.interactions.test.ts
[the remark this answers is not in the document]
nickofthyme: Oh I see, where there would be TS issues? Got it.

src/components/icons/assets/dot.tsx
nickofthyme: No worries!!

src/utils/bbox/dom_text_bbox_calculator.ts
nickofthyme: oh I see. Reminds me of the little scala I used.

nickofthyme:
YAAAAY!!!!

PR #290, opened 2019-08-06
fix(lines): line and point rendered on the same layer for each series

### LINKED ISSUE #287: Dots on line charts have sit above all series lines
**Describe the bug**
When there are multiple series of lines on one chart, all of the dots sit above all of the lines. This means that if a line crosses the dot of a different series, the dot will always be on top of the line.

**To Reproduce**
Create any multi-series line chart with data that allows the lines to cross.

**Expected behavior**
The dots **and** lines of each series should be grouped so that no line or dot can sit between the dot and line of another series.

**Screenshots**
Screen Shot 2019-08-05 at 18 33 09 PM

 - OS: Mac
 - Browser: Chrome
 - Elastic Charts: 8.1.5

markov00:
## Summary

This commit change the rendering order of the line series. Now we render: for each series the line
and each data point, instead of rendering first all the lines and than all the points

fix #287

**before**
Screenshot 2019-08-06 at 11 35 06

**after**
Screenshot 2019-08-06 at 11 35 21

### Checklist

### DIFF
--- a/src/components/react_canvas/line_geometries.tsx
+++ b/src/components/react_canvas/line_geometries.tsx
@@ -38,26 +38,9 @@ export class LineGeometries extends React.PureComponent<LineGeometriesDataProps,
     return (
       <Group ref={this.barSeriesRef} key={'bar_series'}>
         {this.renderLineGeoms()}
-        {this.renderLinePoints()}
       </Group>
     );
   }
-
-  private renderLinePoints = (): JSX.Element[] => {
-    const { lines } = this.props;
-    return lines.reduce(
-      (acc, glyph, i) => {
-        const { points, seriesPointStyle, color } = glyph;
-
-        if (!seriesPointStyle.visible) {
-          return acc;
-        }
-        const pointStyleProps = buildPointStyleProps(color, seriesPointStyle);
-        return [...acc, ...this.renderPoints(points, i, pointStyleProps)];
-      },
-      [] as JSX.Element[],
-    );
-  };
 
   private renderPoints = (
     linePoints: PointGeometry[],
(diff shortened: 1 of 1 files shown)


src/components/react_canvas/line_geometries.tsx
```diff
@@ -90,35 +73,47 @@ export class LineGeometries extends React.PureComponent<LineGeometriesDataProps,
...
-    });
-
-    return lineElements;
+      return acc;
+    }, []);
   };
+
+  getLineToRender(glyph: LineGeometry, sharedStyle: SharedGeometryStyle, index: number) {
+    const { line, color, transform, geometryId, seriesLineStyle } = glyph;
+    const key = `line-${index}`;
```
nickofthyme: Can we use something other than the index for the key?

src/components/react_canvas/line_geometries.tsx
```diff
@@ -90,35 +73,47 @@ export class LineGeometries extends React.PureComponent<LineGeometriesDataProps,
...
+    const customOpacity = seriesLineStyle ? seriesLineStyle.opacity : undefined;
+    const geometryStyle = getGeometryStyle(geometryId, this.props.highlightedLegendItem, sharedStyle, customOpacity);
+
+    if (this.props.animated) {
+      return (
+        <Group key={index} x={transform.x}>
```
nickofthyme: Also here?

src/components/react_canvas/line_geometries.tsx
```diff
@@ -90,35 +73,47 @@ export class LineGeometries extends React.PureComponent<LineGeometriesDataProps,
...
+          <Spring native reset from={{ opacity: 0 }} to={{ opacity: 1 }}>
+            {() => {
```
nickofthyme: Can we extract this function so it doesn't create a new function on each rerender? I don't think it will be that different on performance but best to keep up with best practices.

src/components/react_canvas/line_geometries.tsx
markov00: fixed on https://github.com/elastic/elastic-charts/pull/290/commits/5d712a590948d6463b5cae6b73be3d494c5dc59d

src/components/react_canvas/line_geometries.tsx
markov00: fixed on https://github.com/elastic/elastic-charts/pull/290/commits/5d712a590948d6463b5cae6b73be3d494c5dc59d

src/components/react_canvas/line_geometries.tsx
markov00: removed the animation in https://github.com/elastic/elastic-charts/pull/290/commits/5d712a590948d6463b5cae6b73be3d494c5dc59d

src/chart_types/xy_chart/rendering/rendering.ts
```diff
@@ -493,3 +496,7 @@ export function isPointOnGeometry(
   const { width, height } = indexedGeometry;
   return yCoordinate >= y && yCoordinate <= y + height && xCoordinate >= x && xCoordinate <= x + width;
 }
+
+export function getGeometryIdAsString(geometryId: GeometryId, prefix?: string, postfix?: string) {
```
nickofthyme: This is fine but what are your thoughts on going away from the `_____AsString` naming to `_____Key` because technically `GeometryIdAsString` would just be `GeometryIdAsString.toString()`. I think the key syntax makes the logic clearer to understand.

Fine to merge but just sharing my opinion.

src/components/react_canvas/line_geometries.tsx
```diff
@@ -44,76 +48,48 @@ export class LineGeometries extends React.PureComponent<LineGeometriesDataProps,
...
-                return <animated.Circle {...pointProps} key={key} />;
-              }}
-            </Spring>
-          </Group>,
-        );
-      } else {
-        const pointProps = buildPointRenderProps(transform.x + x, y, pointStyleProps);
-        linePointsElements.push(<Circle {...pointProps} key={key} />);
-      }
+      const key = `line-point-${lineKey}-${pointIndex}`;
```
nickofthyme: This value of `pointIndex` could possibly change right? Then making this key change and the element rerender...right 🤔?

src/components/react_canvas/line_geometries.tsx
```diff
@@ -44,76 +48,48 @@ export class LineGeometries extends React.PureComponent<LineGeometriesDataProps,
...
-            }}
-          </Spring>
-        </Group>
-      );
-    } else {
-      const lineProps = buildLineRenderProps(transform.x, line, color, seriesLineStyle, geometryStyle);
-      return <Path {...lineProps} key={key} />;
-    }
+    const lineProps = buildLineRenderProps(transform.x, line, color, seriesLineStyle, geometryStyle);
+    return <Path {...lineProps} key={key} />;
```
nickofthyme: 👍

src/components/react_canvas/area_geometries.tsx
```diff
@@ -38,110 +43,118 @@ export class AreaGeometries extends React.PureComponent<AreaGeometriesDataProps,
...
-      });
+  private renderPoints = (
+    areaPoints: PointGeometry[],
+    areaIndex: number,
+    pointStyleProps: PointStyleProps,
+    geometryId: GeometryId,
+  ): JSX.Element[] => {
+    return areaPoints.map((areaPoint, pointIndex) => {
+      const { x, y, transform } = areaPoint;
+      const key = getGeometryIdAsString(geometryId, `area-point-${areaIndex}-${pointIndex}-`);
```
nickofthyme: Same idea here re: `pointIndex`

src/components/react_canvas/area_geometries.tsx
```diff
@@ -38,110 +43,118 @@ export class AreaGeometries extends React.PureComponent<AreaGeometriesDataProps,
...
+    const geometryStyle = getGeometryStyle(geometryId, highlightedLegendItem, sharedStyle, seriesAreaLineStyle.opacity);
 
-      const geometryStyle = getGeometryStyle(
-        geometryId,
-        this.props.highlightedLegendItem,
-        sharedStyle,
-        seriesAreaLineStyle.opacity,
-      );
+    return lines.map((linePath, lineIndex) => {
+      const key = getGeometryIdAsString(geometryId, `area-line-${areaIndex}-${lineIndex}`);
```
nickofthyme: and here

src/components/react_canvas/line_geometries.tsx
markov00: yes, one point can be added, one removed. we don't have a real key for points. maybe only the data value, but this can still change

src/components/react_canvas/area_geometries.tsx
markov00: yea, unfortunately we don't have a static value that is maintained during an update.

src/components/react_canvas/area_geometries.tsx
markov00: same

src/chart_types/xy_chart/rendering/rendering.ts
markov00: not sure I understand your opinion: do you mean having a function like
```
getGeometryIdKey(geometryId: GeometryId, prefix?: string, postfix?: string):string 
```

Issue #245, opened 2019-06-19
Allow customization of sub series names


emmacunningham:
**Is your feature request related to a problem? Please describe.**
Currently the name for a sub series is automatically created by concatenating the split series values and yAccessors.  This leads to issues in both the legend and element hover tooltip when the split series values need to be formatted.

(in the example below, the UUIDs are defined as `yAccessors`, not `splitSeriesAccessors`)
![image](https://user-images.githubusercontent.com/452850/59775215-a8a13480-9265-11e9-8da1-f20f7a6733e2.png)

**Describe the solution you'd like**
This is related to the request in #241 though not completely the same as with that request, the user may want to do more than just the series name formatting.

For this enhancement, we could do something similar to how @nickofthyme  is extending the Accessor type for colorAccessors so that we can define either a string field accessor as we do now (in which case the series name would be the default concatenation) or a function which has `specId` and `colorValues` params that can be used to return a custom formatted string.

**Describe alternatives you've considered**
We already have the `name` prop on a series and could use that; the problem there is that for a split series, we may want different custom names per split series and so just using the one `name` prop may not be the best.

The other possibility is that instead of adding the function as a type on `splitSeriesAccessors` and `yAccessors`, we could add it to `name` instead and ensure that if `name` is defined, it always takes precedence over the default.

**Additional context**
related to #241 

**Kibana Cross Issues**
Add any [Kibana](https://github.com/elastic/kibana) related issues/feature requests here.

**Checklist**

- [ ] this request is checked against already exist requests
- [ ] every related Kibana issue is listed under `Kibana Cross Issues` list
- [ ] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

angorayc:
Hello,
I am excited about this feature, as we have similar request in SIEM.
When splitAccessor's got a single value, the legend label shows different from what we expected.
https://github.com/elastic/kibana/pull/48260#issuecomment-542651714
I'm looking forward to have the same behaviour for showing the legend label doesn't matter how many group we have in the dataset.
Thank you for taking this into consideration.

nickofthyme:
Hey @angorayc we needed to refactor some of the code in order to make this work in a useful way without going through the rigmarole of this 👇 

```tsx
const barCustomSeriesNames: CustomSeriesNamesMap = new Map();
const barDataSeries: DataSeriesColorsValues = {
  colorValues: ['cloudflare.com', 'direct-cdn', 'y2'],
[...]

PR #298, opened 2019-08-06
feat: allow custom series naming

nickofthyme:
## Summary

Allow user to set custom names for series based on `DataSeriesColorsValues`.

The names are applied similarly to how `customSeriesColors` are applied using `DataSeriesColorsValues` in a `Map` tied to a label `string` key.

> Changes applied to both the tooltip and legend logic.

```tsx
const barCustomSeriesNames: CustomSeriesNamesMap = new Map();
const barDataSeries: DataSeriesColorsValues = {
  colorValues: ['cloudflare.com', 'direct-cdn', 'y2'],
  specId: getSpecId('bars'),
};
barCustomSeriesNames.set(barDataSeries, 'MY CUSTOM SERIES');

return (
    {/* base config stuff */}

    <BarSeries
      id={getSpecId('bars')}
      xScaleType={ScaleType.Linear}
      yScaleType={ScaleType.Linear}
      xAccessor="x"
      yAccessors={['y1', 'y2']}
      splitSeriesAccessors={['g1', 'g2']}
      customSeriesNames={barCustomSeriesNames}
      data={TestDatasets.BARCHART_2Y2G}
    />
  </Chart>
)
```

### Result
Screen Shot 2019-08-06 at 1 21 50 PM

#245

### Checklist

### DIFF
--- a/src/chart_types/xy_chart/legend/legend.test.ts
+++ b/src/chart_types/xy_chart/legend/legend.test.ts
@@ -1,7 +1,7 @@
 import { AxisId, getAxisId, getGroupId, getSpecId, SpecId } from '../../../utils/ids';
 import { ScaleType } from '../../../utils/scales/scales';
 import { computeLegend, getSeriesColorLabel } from './legend';
-import { DataSeriesColorsValues } from '../utils/series';
+import { DataSeriesColorsValues, getColorValuesAsString } from '../utils/series';
 import { AxisSpec, BasicSeriesSpec, Position } from '../utils/specs';
 
 const colorValues1a = {
@@ -65,6 +65,7 @@ axesSpecs.set(axisSpec.id, axisSpec);
 describe('Legends', () => {
   const seriesColor = new Map<string, DataSeriesColorsValues>();
   const seriesColorMap = new Map<string, string>();
+  const seriesNameMap = new Map<string, string>();
   const specs = new Map<SpecId, BasicSeriesSpec>();
   specs.set(spec1.id, spec1);
   specs.set(spec2.id, spec2);
@@ -77,7 +78,7 @@ describe('Legends', () => {
   });
   it('compute legend for a single series', () => {
     seriesColor.set('colorSeries1a', colorValues1a);
-    const legend = computeLegend(seriesColor, seriesColorMap, specs, 'violet', axesSpecs);
+    const legend = computeLegend(seriesColor, seriesColorMap, seriesNameMap, specs, 'violet', axesSpecs);
     const expected = [
       {
         color: 'red',
@@ -94,7 +95,7 @@ describe('Legends', () => {
   it('compute legend for a single spec but with multiple series', () => {
     seriesColor.set('colorSeries1a', colorValues1a);
     seriesColor.set('colorSeries1b', colorValues1b);
-    const legend = computeLegend(seriesColor, seriesColorMap, specs, 'violet', axesSpecs);
+    const legend = computeLegend(seriesColor, seriesColorMap, seriesNameMap, specs, 'violet', axesSpecs);
     const expected = [
       {
         color: 'red',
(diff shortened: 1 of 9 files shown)


cchaos:
From an outsiders perspective it feel weird to be using `colorValues` to set a custom `name`.

Also, what if I wanted to change multiple series names?

nickofthyme:
Ya I agree, the issue I have is that we are using the `specId` and `colorValues` to differentiate series from each other.

The `colorValues` are actually accessor values, I think that is just bad naming on our part that has spread throughout the code.

The `DataSeriesColorsValues` interface is used more as a `DataSeriesIdentifier` in that it gives us all the values we need to classify a given series. We also use the `DataSeriesColorsValues` interface to know which series to deselect. But changing this naming as it would cause a breaking change to those who use `CustomSeriesColors`, which I want to avoid.

I'd like to cleanup how we are differentiating series but just not at the moment.

nickofthyme:
In regards to multiple series names: I realized that this approach would be very verbose but this is the same approach we are using for applying custom colors and that reasoning ties back in how we are differentiating/identifying the series.

cchaos:
Since this library is still in the early stages, there's not a **ton** of usages yet, and you're still doing quite a bit of breaking changes, I'd consider maybe taking a deeper stab at this naming and utility functions to drive away from the "Color" word usage to something more generic.
_Just my 2 cents_

**Multiple series**

I was more asking **how** is this done. Would it be like this?:

```js
const barCustomSeriesNames: CustomSeriesNamesMap = new Map();

const barDataSeries1: DataSeriesColorsValues = {
  colorValues: ['cloudflare.com', 'direct-cdn', 'y1'],
  specId: getSpecId('bars'),
};
barCustomSeriesNames.set(barDataSeries1, 'MY CUSTOM SERIES 1');

const barDataSeries2: DataSeriesColorsValues = {
  colorValues: ['cloudflare.com', 'direct-cdn', 'y2'],
  specId: getSpecId('bars'),
};
barCustomSeriesNames.set(barDataSeries2, 'MY CUSTOM SERIES 2');

```

markov00:
Hey @nickofthyme I think it's nice to clean this up a bit.
For historical reason, the `colorValues` name was used because of the `colorAccessors` idea: you can have a set of series, but you want to color them with a single color, maintaining them visually separated.  With that in mind, we kept the `colorValues` property to identify each created series.

We for sure can simplify this with something along those lines. Instead of have to build different maps for each custom property, we can have one single element that can be used to customize color and name, and other prop in the future.
We can completely remove the `specId` parameter that we can get from the spec. We can refactor the colorValues into: `splitSeriesValues` (an ordered array of value for each split series accessor) + and optional yAccessor string that identify the yAccessor when we are using multiple metrics.
That property can be a single object (overwrite everything for every series, used mainly with single series) or can be an array of this custom properties (used to override one or multiple data series)

```tsx
interface CustomSeriesProps{
  splitSeriesValues?: any[],
  yAccessor?: string,
  name?: string,
  color?: string,
}



// without splitted series and only one yAccessor
<BarSeries
  id={getSpecId('control3')}
  data={data}
  xAccessor={'x'}
  yAccessors={['y']}
  customProps={{
      color: 'blue',
  }}
/>


// with splitted series and multiple one yAccessor
<BarSeries
  id={getSpecId('control3')}
  data={data}
  xAccessor={'x'}
  yAccessors={['y']}
  splitSeriesAccessors={['g']}
  customProps={[
    {
      // this set custom color and name
      splitSeriesValues:['group1'],
      name: 'first series',
      color: 'blue',
    },
    {
      // this set custom name
      splitSeriesValues:['group2'],
      name: 'second series',
    },
     {
       // this set custom color
      splitSeriesValues:['group3'],
      color: 'red',
    },
  ]}
/>

// with splitted series and multiple yAccessors
<BarSeries
  id={getSpecId('control3')}
  data={data}
  xAccessor={'x'}
  yAccessors={['y1','y2']}
  splitSeriesAccessors={['g']}
  customProps={[
    {
      splitSeriesValues:['group1'],
      yAccessor: 'y1'
      name: 'first series - metric 1',
      color: 'blue',
    },
    {
      splitSeriesValues:['group2'],
      yAccessor: 'y1'
      name: 'second series - metric 2',
    },
  ]}
/>
```

nickofthyme:
close in place of #419

Issue #83, opened 2019-02-28
Sync crosshair/vertical cursor across charts


markov00:
**Is your feature request related to a problem? Please describe.**
Multiple charts can share the same vertical cursor when they have the same x axis.

**Describe the solution you'd like**
We need to add two props like the following to the chart:
- `onExternalCursorChange` that is populated with the value from an external cursor
- `onCursorChange` that accept a listener and is called when the cursor change it's position

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

**Kibana Cross Issues**
Add any [Kibana](https://github.com/elastic/kibana) related issues/feature requests here.

**Checklist**

- [x] this request is checked against already exist requests
- [ ] every related Kibana issue is listed under `Kibana Cross Issues` list
- [ ] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

nickofthyme:
@markov00 what are your thoughts on limiting this functionality to time series only?

I need to trigger the event with the actual `xValue` (i.e. time) rather than just the cursor `Dimensions` becuase charts could be different widths on the dashboard.

It's not that hard to parse the `xValue` for all charts but just curious if there is a use case to sync the cursors accross non-time-series charts.

markov00:
closed by #304

PR #304, opened 2019-08-12
feat: add cursor sync mechanism

nickofthyme:
## Summary

Add `CursorUpdateListener` as props on `Settings` component to allow `Chart` consumer to synchronize cursors across multiple `Chart`s by calling `dispatchExternalCursorEvent` on the `Chart` ref to update cursor value.

Issue #83 

### Demo
Screen Recording 2019-08-12 at 06 32 PM

### Checklist
- [x] Proper documentation or storybook story was added for features that require explanation or tutorials (*Not able to easily add story with 2 `Chart`s sharing state*)

### DIFF
--- a/src/components/chart.tsx
+++ b/src/components/chart.tsx
@@ -11,8 +11,8 @@ import { Highlighter } from './highlighter';
 import { Legend } from './legend/legend';
 import { LegendButton } from './legend/legend_button';
 import { ReactiveChart as ReactChart } from './react_canvas/reactive_chart';
-// import { ReactiveChart as SVGChart } from './svg/reactive_chart';
 import { Tooltips } from './tooltips';
+import { CursorEvent } from '../specs/settings';
 
 interface ChartProps {
   /** The type of rendered
@@ -34,6 +34,26 @@ export class Chart extends React.Component<ChartProps> {
     this.chartSpecStore = new ChartStore();
     this.legendId = htmlIdGenerator()('legend');
   }
+
+  dispatchExternalCursorEvent(event?: CursorEvent) {
+    this.chartSpecStore.setActiveChartId(event && event.chartId);
+    const isActiveChart = this.chartSpecStore.isActiveChart.get();
+
+    if (!event) {
+      if (!isActiveChart) {
+        this.chartSpecStore.setCursorPosition(-1, -1);
+      }
+    } else {
+      if (
+        !isActiveChart &&
+        this.chartSpecStore.xScale!.type === event.scale &&
+        (!event.unit || event.unit === this.chartSpecStore.xScale!.unit)
+      ) {
+        this.chartSpecStore.setCursorValue(event.value);
+      }
+    }
+  }
+
   render() {
     const { renderer, size, className } = this.props;
     let containerStyle: CSSProperties;
--- a/src/components/chart.tsx
+++ b/src/components/chart.tsx
@@ -47,7 +47,7 @@ export class Chart extends React.Component<ChartProps> {
       if (
         !isActiveChart &&
         this.chartSpecStore.xScale!.type === event.scale &&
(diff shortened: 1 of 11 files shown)


nickofthyme:
Not sure if the issue https://github.com/elastic/kibana/issues/14799 will be fixed by this, the sync still seems pretty slow. @markov00 any ideas?

nickofthyme:
@markov00 I updated the logic as we discussed. Much better performance 👍 

I also added back the `xScale.type` check and added an `xScale.unit` property for the future to check that the axis is the same unit.

## Demo
Screen Recording 2019-08-13 at 11 49 AM

src/components/chart.tsx
```diff
@@ -34,6 +34,26 @@ export class Chart extends React.Component<ChartProps> {
...
+
+    if (!event) {
+      if (!isActiveChart) {
+        this.chartSpecStore.setCursorPosition(-1, -1);
+      }
+    } else {
+      if (
+        !isActiveChart &&
+        this.chartSpecStore.xScale!.type === event.scale &&
+        (!event.unit || event.unit === this.chartSpecStore.xScale!.unit)
```
markov00: better to check if unit is `undefined` because `!''` is true also if `''` is a valid empty string.

markov00: There can be case where we can get an empty string that we may want to threat them as valid units

src/specs/settings.tsx
```diff
@@ -23,6 +26,17 @@ interface TooltipProps {
...
 }
 
+export interface CursorEvent {
+  chartId: string;
+  scale: ScaleTypes;
+  /**
+   * @todo
+   * unit for event (i.e. `time`, `feet`, `count`, etc.)
+   */
+  unit?: string;
```
markov00: FYI the concept of unit doesn't exist in Kibana already. There is the concept of field formatter type, that can, sometimes, correspond to a unit like time, percent, bytes, but can contain anything like numeric or custom formats.

src/specs/settings.tsx
```diff
@@ -23,6 +26,17 @@ interface TooltipProps {
   headerFormatter?: TooltipValueFormatter;
 }
 
+export interface CursorEvent {
```
markov00: You missed to export that type from the main index.ts

markov00: Can you also please add a bit more documentation for that interface?

src/specs/settings.tsx
```diff
@@ -23,6 +26,17 @@ interface TooltipProps {
...
+  chartId: string;
+  scale: ScaleTypes;
+  /**
+   * @todo
+   * unit for event (i.e. `time`, `feet`, `count`, etc.)
+   */
+  unit?: string;
+  value: number;
```
markov00: This one should be `number | string` according to the current `setCursorValue` implementation

src/components/chart.tsx
nickofthyme: Good catch

src/specs/settings.tsx
nickofthyme: Good to know. I just added this to go that way because using the scale type is ambiguous. 

Also, I think it would just be a localized thing, meaning: I want to sync these three charts in a single file only if their `unit` matches. The unit is local to that file, the rest of the app doesn't need to know about the unit, per se.

src/specs/settings.tsx
nickofthyme: 6baa552

src/specs/settings.tsx
nickofthyme: 6baa552

src/components/chart.tsx
nickofthyme: 6baa552

Issue #311, opened 2019-08-14
Dynamic legend content size


nickofthyme:
**Describe the bug**
When using elastic charts the legend should adjust based on content with a `max-width`/`max-height` for side/bottom legends.

**Screenshots**
![image](https://user-images.githubusercontent.com/19007109/63042579-f0bc8c00-be8f-11e9-9a85-86fa82cddb73.png)

See https://github.com/elastic/kibana/pull/33558#issuecomment-520847052

nickofthyme:
Closed by https://github.com/elastic/elastic-charts/pull/316

Issue #268, opened 2019-07-22
Get rid of the legend toggle


cchaos:
**Is your feature request related to a problem? Please describe.**
The visibility of the legend should be based on the consumer not dictated by the library.

**Describe the solution you'd like**
Remove this toggle:

<img src="https://d.pr/free/i/qq1Rhv+" width="50%" />

In favor of just providing a simple prop for legend visibility where then the consumer can hook up their own button or service to provide that visibility.

**Checklist**

- [x] this request is checked against already exist requests
- ~[ ] every related Kibana issue is listed under `Kibana Cross Issues` list~
- ~[ ] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present~

PR #316, opened 2019-08-15
feat: auto legend resize

nickofthyme:
## Summary

Issue #268 & #311 

- [x] Remove UI legend toggle
- [x] Allow legend content to determine size of legend.
- [x] Pull Legend out of Canvas space
- [x] Update styles for top/bottom legend to use [css grid](https://css-tricks.com/snippets/css/complete-guide-grid/)

The simple explanation of changes:
#### Changes to `Right` and `Left` legends
The old legend was statically positioned using a hardcoded width. This created a lot of useless white space for series with short labels and small values like this..............................👇👇👇👇👇👇 

image

With these changes, the `legendItems` are computed and rendered, then measured to determine the width. The width includes the formatted series label plus the formatted last value.

Image 2019-08-20 at 5 54 06 PM

Then a so-called "buffer" is added to this measured width value to account for the variability in value length (i.e. values ranging from `0` to `1000000000`). This buffer is a new property on the `Theme` type `theme.legend.spacingBuffer`. This should be increased for datasets with large variability.

> If the value plus label ever exceeds the allowable width, the label will be truncated as it does currently.

Image 2019-08-20 at 5 56 54 PM

This measured width is capped at a max value set by the `Theme` value `theme.legend.verticalWidth`.

#### Changes to `Top` and `Bottom` legends
The previously hardcoded `height` is now set to be a `max-height` to reduce white space.

This new `max-height` value is determined from the `Theme` value `theme.legend.horizontalHeight`.

CSS flex was replaced with CSS grid to better adjust to window sizing. Previously the widths were hardcoded and if too long would just wrap to a new row. like this 👇 

Screen Recording 2019-08-20 at 06 12 PM

With CSS grid and `repeat(auto-fill, [VALUE])` the columns grow and shrink to fill the empty space and wrap only when the `min-width` is not available like so 👇 

Screen Recording 2019-08-20 at 06 16 PM

> Note: The CSS grid changes do not apply to IE11 because IE11 😩. IE will still use css flex.

#### Empty State
As the legend is now a separate containing element, the chart will grow to fill only the available space, rather than before having to compute and position each accordingly. As a result, the empty state is now centered relative to the available chart element, see below 👇 

image

### Left/Right
Screen Recording 2019-08-20 at 06 23 PM

### Top/Bottom
Screen Recording 2019-08-20 at 06 24 PM

### Checklist


nickofthyme:
@markov00 The toggle removal was pretty trivial but sizing the legend was quite tricky and this is the best solution I could come up with.

I had to `observe` the `legendPosition` value to update the `Chart` clases. I pulled the legend out from the `echContainer` so that the legend can take up the room it needs then the container can take up the room left available for rendering the canvas elements. I don't really like that logic required in `chart.tsx` so I am open to ideas.

Another thing I had to change was that the `legendItems` are computed in the `computeChart` method but I need the legend items for the parent dimensions to be set. I'm basically breaking on the first time when the legend is initialized to rerun the computations.

Lastly the `Right` and `Left` legends now having a not-static width will cause the chart to twitch when the values in the legend start to change. A quick fix to this was just to set a `min-width` to prevent that, but it's only to a certain point. See gif below for what I'm referring to.

Screen Recording 2019-08-14 at 10 23 PM

markov00:
hey @nickofthyme I don't like much the fact that the chart twitch when moving the mouse over, nor I don't like much when, for `Top` and `Bottom` legend the legend twitch itself in the same way.
Also the wired connection between  store - container - legend to specify the flex nature of the chart container is a bit hacky IMHO.

What if we try a different way? so before we used to have a fixed width/height. What if we can compute that dimension? It doesn't have to be pixel perfect, but enough to nicely fit on the screen:
- we can get the longest data series name +
- we can also compute an hypothetical max value to be displayed on the legend (hypothetical because depends on the formatter, and a value of 1.000.000 can be shortest than 100.000 if we use unit formatter like 1M 100K)
- we use these two values to compute an average legend item fixed width. everything else should shrink in that max width.

This is not an optimal approach but can work.

A feature that we can include is also a user defined size for the legend.

src/specs/specs_parser.tsx
```diff
@@ -7,10 +7,6 @@ export interface SpecProps {
 }
 
 export class SpecsSpecRootComponent extends PureComponent<SpecProps> {
-  static getDerivedStateFromProps(props: SpecProps) {
-    props.chartStore!.specsInitialized.set(false);
```
nickofthyme: @markov00 this was causing the `specsInitialized` to be set to `false` just for top/bottom legends, which was breaking the visibility toggle. Is this necessary?

Screen Recording 2019-08-19 at 08 44 AM

markov00: Do you mean: is the visibility toggle necessary? if so yes, it's a feature that we need to keep on the library.

nickofthyme: No that is definitely necessary. I'm referring to the line of code. Is the `getDerivedStateFromProps` lifecycle method required here? Otherwise, I am removing it.

markov00: So it's strange that this `getDerivedStateFromProps` is triggered only for top/bottom legends. That should be used called on every mount and on every update. Right now it's only used for one variable reset on the chart_state but I'm pretty sure it's not really required.
I prefer to leave it there if possible and check what is the root cause of the missing visibility toggle.
Second thing: I've tested on a Right legend and I've seen this big space between the chart and the legend:
Screenshot 2019-08-19 at 18 22 34

nickofthyme: Ok I'll look deeper into why the `specParser` is rerendering and calling `getDerivedStateFromProps`

The merge created the gap issue, fixed that here 51041af

src/components/chart.tsx
```diff
@@ -57,33 +82,28 @@ export class Chart extends React.Component<ChartProps> {
...
+    const Horizontal = isHorizontal(this.state.legendPosition);
+    const chartClassNames = classNames('echChart', className, {
+      'echChart--column': Horizontal,
+    });
+
     return (
       <Provider chartStore={this.chartSpecStore}>
-        <Fragment>
+        <div style={containerStyle} className={chartClassNames}>
+          <Legend legendId={this.legendId} />
```
markov00: As you removed the legend button, I thin you can also remove entirely the `legendId`

src/chart_types/xy_chart/store/chart_state.ts
```diff
@@ -232,15 +233,9 @@ export class ChartStore {
   canDataBeAnimated = false;
 
   showLegend = observable.box(false);
-  legendCollapsed = observable.box(false);
-  legendPosition: Position | undefined;
+  legendPosition = observable.box<Position>(Position.Right);
```
markov00: ❤️

nickofthyme: lol I was tied of setting a default 🙄

src/components/chart.tsx
nickofthyme: Oh good idea, I wasn't sure what that was for.

nickofthyme:
@markov00 good catch, sorry I forgot about safari 🙃.

The problem was that we are setting a `max-height` on the `Legend` then the scrolling `LegendListContainer` had a height of `100%` that safari was not obeying. So putting the `max-height` on the scroll container fixed it, looks good in all browsers now.

### Safari
Screen Recording 2019-08-21 at 09 50 AM

src/components/chart.tsx
nickofthyme: ea01ba3

nickofthyme:
@markov00 the `flex: 1 1 auto` caused this to happen only in safari 👇 Where the `No data to display` text is at the top not the center.

Screen Recording 2019-08-21 at 10 15 AM

PR #317, opened 2019-08-16
feat: add chart size type overrides

### LINKED ISSUE #177: Chart `size` attribute type doesn't support "100%" as a valid size
**Describe the bug**

The `size` attribute definition only supports numbers BUT it should support both `string` and `number` so the user can define something like:

```jsx
<Chart size={['100%', 400]}>...</Chart>
```
This actual works the only issue is Typescript throws a linting error.

**To Reproduce**

1. Create a chart with a size set to `['100%', 400]`
2. Watch Typescript complain
3. Make sad trombone sound with mouth :(

**Expected behavior**

Shouldn't throw typescript error when defining width or height as `"100%"`

markov00:
## Summary

fix #177 

Add multiple override types for `<Chart>` `size` props.

The `size` parameters now accept `ChartSize`:
```ts
type ChartSize = number | string | ChartSizeArray | ChartSizeObject;
type ChartSizeArray = [number | string | undefined, number | string | undefined];
type ChartSizeObject = {
  width?: number | string;
  height?: number | string;
}

```
The default value for an `undefined` parameter is `100%`.

### Checklist

### DIFF
--- a/src/utils/chart_size.ts
+++ b/src/utils/chart_size.ts
@@ -1,0 +1,33 @@
+type ChartSizeArray = [number | string | undefined, number | string | undefined];
+interface ChartSizeObject {
+  width?: number | string;
+  height?: number | string;
+}
+export type ChartSize = number | string | ChartSizeArray | ChartSizeObject;
+
+function isSizeArray(size: ChartSize): size is ChartSizeArray {
+  return Array.isArray(size);
+}
+function isSizeObject(size: ChartSize): size is ChartSizeObject {
+  return !Array.isArray(size) && typeof size === 'object';
+}
+
+export function getChartSize(size: ChartSize) {
+  if (isSizeArray(size)) {
+    return {
+      width: size[0] === undefined ? '100%' : size[0],
+      height: size[1] === undefined ? '100%' : size[1],
+    };
+  }
+  if (isSizeObject(size)) {
+    return {
+      width: size.width === undefined ? '100%' : size.width,
+      height: size.height === undefined ? '100%' : size.height,
+    };
+  }
+  const sameSize = size === undefined ? '100%' : size;
+  return {
+    width: sameSize,
+    height: sameSize,
+  };
+}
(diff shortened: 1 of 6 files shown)


src/utils/chart_size.ts
```diff
@@ -0,0 +1,34 @@
+export type ChartSizeArray = [number | string | undefined, number | string | undefined];
+export interface ChartSizeObject {
+  width?: number | string;
+  height?: number | string;
+}
+
+export type ChartSize = number | string | ChartSizeArray | ChartSizeObject;
+
+function isSizeArray(size: ChartSize): size is ChartSizeArray {
```
nickofthyme: is this abstraction necessary? I think the `getChartSize` can just check if is array or object else.

PS: This overloading thing in TS is the one thing that makes me miss JS.

src/utils/chart_size.ts
```diff
@@ -0,0 +1,34 @@
...
+  return Array.isArray(size);
+}
+function isSizeObject(size: ChartSize): size is ChartSizeObject {
+  return !Array.isArray(size) && typeof size === 'object';
+}
+
+export function getChartSize(size: ChartSize) {
+  if (isSizeArray(size)) {
```
nickofthyme:
```suggestion
  if (Array.isArray(size)) {
```

src/utils/chart_size.ts
```diff
@@ -0,0 +1,34 @@
...
+    return {
+      width: size[0] === undefined ? '100%' : size[0],
+      height: size[1] === undefined ? '100%' : size[1],
+    };
+  }
+  if (isSizeObject(size)) {
```
nickofthyme:
```suggestion
  if (typeof size === 'object') {
```

markov00: done in 3330002

src/utils/chart_size.ts
markov00: done in 3330002

src/utils/chart_size.ts
markov00: you are right, there is no real needs for all that.
refactored in 3330002

PR #319, opened 2019-08-16
feat: customize number of axis ticks

markov00:
## Summary

This PR add the possibility to customize the number of ticks displayed on each axis.

A `ticks` parameter is added to the `Axis` component. The number specified is not always fully respected as the real number of displayed ticks because the scale function computes ticks that are uniformly spaced with human-readable values (such as multiples of powers of 10), and are guaranteed to be within the extent of the domain (see https://github.com/d3/d3-scale#continuous_ticks)

Aug-17-2019 00-11-16

### Checklist

### DIFF
--- a/src/chart_types/xy_chart/annotations/annotation_marker.test.tsx
+++ b/src/chart_types/xy_chart/annotations/annotation_marker.test.tsx
@@ -21,7 +21,11 @@ describe('annotation marker', () => {
   const maxRange = 100;
 
   const continuousData = [0, 10];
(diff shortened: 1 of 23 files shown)


markov00:
Hey @nickofthyme so we are using https://github.com/d3/d3-scale#continuous_ticks to generate human readable ticks on axis. This unfortunately doesn't allows us directly to specify to show min and max labels but we can add that manually on the code. We can do that my adding also the `nice` feature https://github.com/d3/d3-scale#continuous_nice that will round the domain to a nicer value and it generally shows the max value (the min is always displayed)

PR #320, opened 2019-08-16
fix: remove clippings from chart

### LINKED ISSUE #20: Avoid clipping tooltip markers on edges
Find a way to fully display tooltip markers near the edges.

screenshot 2019-01-31 at 22 32 13

Kibana Cross Issues:
* https://github.com/elastic/kibana/issues/13248

markov00:
## Summary

fix #20

This PR remove the clippings from the rendered geometries to allow full rendering elements on the edges.

It moves also the axis layer 1 level below the rendered geometries to clearly visualize the elements on top of the axis lines:

Screenshot 2019-08-17 at 00 58 34

cc @cchaos: FYI the following is an example with the axis layers 1 level above the chart geometries, what do you think it's better: showing the axis behind the chart elements or in front of them?:
Screenshot 2019-08-17 at 00 59 47

### Checklist

### DIFF
--- a/src/components/react_canvas/reactive_chart.tsx
+++ b/src/components/react_canvas/reactive_chart.tsx
@@ -380,15 +380,6 @@ class Chart extends React.Component<ReactiveChartProps, ReactiveChartState> {
         </div>
       );
     }
-    // disable clippings when debugging
-    const clippings = debug
-      ? {}
-      : {
-          clipX: 0,
-          clipY: 0,
-          clipWidth: [90, -90].includes(chartRotation) ? chartDimensions.height : chartDimensions.width,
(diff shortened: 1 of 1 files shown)


cchaos:
Thanks @markov00 

I definitely think the chart geometries should be on top. However, I am not super happy with the fact that the fills then overlap the axis lines.



The dot definitely needs to be above the axis lines but the dot, line, and fills should stay grouped together so they change in opacity together. So I'm not really sure what to do about that. 

At the very least, it is only an issue for areas, I think. How do bars fit near the axes?

markov00:
Let's keep in that way right now. We will find a good way to clip out the axis labels from the geometries later

PR #321, opened 2019-08-16
fix(legend): disable fade other if legend item is not visible

markov00:
## Summary

When hovering on a legend item that is currently hidden, don't fade the other series https://d.pr/free/i/DbbyWC

TODO

The current status correctly avoid hiding the other series when mouse over an hidden one:
Oct-29-2019 18-07-38

What is missing is the following: fade in all the series right after I clicked a series in the legend. Currently, if you click to hide a series, the rendered series are in a fade status. We need to have a consistent behaviour and, right after a click I just want to hide the clicked series and also I want to see all the other clearly

wrong behaviour
Oct-29-2019 18-11-35

### Checklist

### DIFF
--- a/src/chart_types/xy_chart/store/chart_state.test.ts
+++ b/src/chart_types/xy_chart/store/chart_state.test.ts
@@ -174,6 +174,33 @@ describe('Chart Store', () => {
     expect(outListener.mock.calls.length).toBe(1);
   });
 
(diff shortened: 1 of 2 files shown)


nickofthyme:
@markov00 Did you want to add anything more to this PR?

Issue #292, opened 2019-08-06
Change `baseThemeType` to `baseTheme`


cchaos:
**Is your feature request related to a problem? Please describe.**
I've created two EUI specific themes for light and dark which Kibana should use as a consumer of EUI. However, the new `baseThemeType` doesn't take into account a custom theme. Therefore it's still hard to combine a partial theme with a custom theme. Still have to use `mergeWithDefaultTheme`.

**Describe the solution you'd like**
Change `baseThemeType` to be `baseTheme` which accepts a custom theme object.

PR #333, opened 2019-08-20
feat(theme): base theme prop

nickofthyme:
## Summary

Allow `baseTheme` prop to set base theme.

#292

BREAKING CHANGE: remove `baseThemeType` prop on `Settings` Component and `BaseThemeTypes` type.

### Checklist


nickofthyme:
Yup I just think a breaking change would not be needed.

nickofthyme:
@markov00 what do you think of setting the default theme in `Settings` not the `chart_state`?

https://github.com/elastic/elastic-charts/blob/7d5540ce76bb2a764c7e03a41a0747c43dab39e1/src/chart_types/xy_chart/store/chart_state.ts#L152-L155

https://github.com/elastic/elastic-charts/blob/7d5540ce76bb2a764c7e03a41a0747c43dab39e1/src/specs/settings.tsx#L121

see 7d5540c

Issue #332, opened 2019-08-20
Tooltip style error IE11


nickofthyme:
**Describe the bug**
Tooltip style issue

**To Reproduce**
1. Go to any chart with a legend in IE11
2. Notice the misformed tooltip like below

**Expected behavior**
The tooltip should look the same in IE11 as it does in Chrome

**Screenshots**
![Screen Recording 2019-08-20 at 01 47 PM](https://user-images.githubusercontent.com/19007109/63375376-4510b180-c351-11e9-9888-458f09d065fc.gif)

**Version (please complete the following information):**
 - OS: Win7
 - Browser: IE11
 - Elastic Charts: `9.x`

**Checklist**
- [x] every related Kibana issue is listed under `Kibana Cross Issues` list
- [x] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

PR #334, opened 2019-08-20
fix(tooltip): ie11 flex sizing

nickofthyme:
## Summary
Fix tooltip flex positioning issue. #332

For some reason the default flex in Chrome/Firefox is different from IE.

```scss
// default in Crome
.class {
  flex: 1 1 auto;
}

// default in IE
.class {
  flex: 1 1 0;
}
```

In the future use `flex: 1 1 auto` instead of `flex: 1` for row elements that you want to _occupare_ the remaining space. This is not the same for column elements as safari will not _occupare_ the space.

Also added `playground:ie` script to start playground webpack server on local ip.

### Chrome
Screen Recording 2019-08-20 at 03 51 PM

### IE 11

Screen Recording 2019-08-20 at 03 53 PM

### Checklist


nickofthyme:
@markov00 I noticed that the error of the work break in IE11 still cuts words up. Is there an issue yet in EUI for that, relating to the `euiTextOverflowWrap` mixin?

image

nickofthyme:
@markov00 fix it for future

image

PR #335, opened 2019-08-21
fix(renderer): stroke opacity

nickofthyme:
## Summary

Fix stoke opacity on bar charts.

Add a property for `strokeOpacity` on `RectBorderStyle`. Fallback uses `opacity` on `RectStyle`.

If `strokeOpacity` is provided the `unhighlighted` stoke opacity is averaged between the provided value and the global `unhighlighted` opacity value.

This change requires splitting the fill and stroke into __TWO__ seperate `Rect` elements as the [Konva](https://konvajs.org/) API does not allow for stroke opacity but forces inheritance from the fill opacity.

Currently, the `perfectDrawEnabled` is set to `false` which causes the border to overlap the fill. These changes inset the border/stroke inside of the filled rectangle/bar. https://konvajs.org/docs/performance/Disable_Perfect_Draw.html

image

### Demo

Screen Recording 2019-08-21 at 03 02 PM

### Checklist

### DIFF
--- a/src/chart_types/xy_chart/rendering/rendering.ts
+++ b/src/chart_types/xy_chart/rendering/rendering.ts
@@ -34,6 +34,7 @@ export interface GeometryValue {
 /** Shared style properties for varies geometries */
 export interface GeometryStyle {
   opacity: number;
+  strokeOpacity?: number;
 }
 
 export type IndexedGeometry = PointGeometry | BarGeometry;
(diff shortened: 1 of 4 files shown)


nickofthyme:
@markov00 a strange consequence of this is when stacking bars the bar segments don't look as uniform.

Image 2019-08-21 at 3 24 51 PM

PR #356, opened 2019-08-27
feat: add prop to set debounce time, lower default

nickofthyme:
## Summary

Adds prop on `Settings` component to set the debounce resize timing.

### Checklist

### DIFF
--- a/src/chart_types/xy_chart/store/chart_state.ts
+++ b/src/chart_types/xy_chart/store/chart_state.ts
@@ -143,7 +143,7 @@ export class ChartStore {
     maxY: 0,
   };
 
-  resizeDebounce: number = 0;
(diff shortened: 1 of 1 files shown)


src/chart_types/xy_chart/store/chart_state.ts
```diff
@@ -143,6 +143,8 @@ export class ChartStore {
     maxY: 0,
   };
 
+  resizeDebounce: number = 0;
```
markov00: can we configure that to the same default of settings? it will not change anything by just to be aligned

markov00: yep I will leave it like it is right now and fix on redux

PR #361, opened 2019-08-28
Bump Node.js to 10.16.3

jbudz:
Companion to https://github.com/elastic/kibana/pull/43543

I'm not familiar with the publishing process if there's anyone that can help with that afterwords

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,7 +142,7 @@
     }
   },
   "engines": {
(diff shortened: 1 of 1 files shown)


markov00:
close in favour of https://github.com/elastic/elastic-charts/pull/363

PR #365, opened 2019-09-03
feat(chart_state): add render change event

nickofthyme:
## Summary

Add render change (`onRenderChange`) lifecycle hook to `Settings`.

`onRenderChange` is a listener to be called when chart render state changes. `isRendered` value is `true` when rendering is complete and `false` otherwise.

Required to implement visualization count for Discover and TSVB.

Screen Recording 2019-08-31 at 11 21 AM

http://localhost:9001/?path=/story/interactions--render-change-action

### Checklist

### DIFF
--- a/src/chart_types/xy_chart/store/chart_state.test.ts
+++ b/src/chart_types/xy_chart/store/chart_state.test.ts
@@ -323,6 +323,24 @@ describe('Chart Store', () => {
     expect(store.onRenderChangeListener).toEqual(listener);
   });
 
+  test('should observe chartInitialized value', () => {
+    const listener = jest.fn();
+    store.chartInitialized.set(false);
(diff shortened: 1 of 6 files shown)


nickofthyme:
Good point I added the type here but an explanation would be good

https://github.com/elastic/elastic-charts/blob/462b35acbb0d7abb994dff242082661489bc78f9/src/chart_types/xy_chart/store/chart_state.ts#L109

nickofthyme:
retest please

nickofthyme:
jenkins test this

nickofthyme:
@markov00 when you return, could you please look at this to see if the timing logic is correct?

I was thinking to put the trigger on the mount/update of the reactive_chart component but the timing was virtually identical to observing the `chartInitialized` value. Let me know what you think.

PR #370, opened 2019-09-10
feat(axis): option to hide duplicate axes

### LINKED ISSUE #368: Prop to coalesce duplicate Axes
Allow `hideDuplicateAxes ` prop to remove duplicate axes.

Axis should be hidden if _all_ of the following conditions are true:
- The first and last tick labels (with applied `tickFormat`) match
- The `length` of the ticks match
- The `titles` match, can be both `undefined`
- The `position` on the spec match (i.e. `top`, `right`, `bottom`, `left`)

`hideDuplicateAxes ` prop to remove duplicate axes.

Check axis domains manually

**Kibana Cross Issues**

**Checklist**

nickofthyme:
## Summary

Adds prop `hideDuplicateAxes` to settings to hide axes based on tick labels, position and title.

* Refactor axes render function.

Axis will be hidden if all of the following conditions are true:
- The first and last tick labels (with applied `tickFormat`) match
- The `length` of the ticks match
- The `titles` match, can be both `undefined`
- The `position` on the spec match

closes #368

Screen Recording 2019-09-10 at 05 37 PM

### Checklist

### DIFF
--- a/src/chart_types/xy_chart/store/chart_state.ts
+++ b/src/chart_types/xy_chart/store/chart_state.ts
@@ -134,7 +134,7 @@ export const isDuplicateAxis = (
     ) {
       const spec = specMap.get(axisId);
 
-      if (spec && spec.position === position && ((!title && !spec.title) || title === spec.title)) {
+      if (spec && spec.position === position && title === spec.title) {
         hasDuplicate = true;
       }
     }
--- a/src/chart_types/xy_chart/store/chart_state.ts
+++ b/src/chart_types/xy_chart/store/chart_state.ts
@@ -114,6 +114,35 @@ export type CursorUpdateListener = (event?: CursorEvent) => void;
 export type RenderChangeListener = (isRendered: boolean) => void;
 export type BasicListener = () => undefined | void;
 
+export const isDuplicateAxis = (
+  { position, title }: AxisSpec,
+  { tickLabels }: AxisTicksDimensions,
+  tickMap: Map<AxisId, AxisTicksDimensions>,
+  specMap: Map<AxisId, AxisSpec>,
+): boolean => {
+  const firstTickLabel = tickLabels[0];
+  const lastTickLabel = tickLabels.slice(-1)[0];
+
+  let hasDuplicate = false;
+  tickMap.forEach(({ tickLabels: axisTickLabels }, axisId) => {
+    if (
+      !hasDuplicate &&
+      axisTickLabels &&
+      tickLabels.length === axisTickLabels.length &&
+      firstTickLabel === axisTickLabels[0] &&
+      lastTickLabel === axisTickLabels.slice(-1)[0]
+    ) {
+      const spec = specMap.get(axisId);
+
+      if (spec && spec.position === position && ((!title && !spec.title) || title === spec.title)) {
+        hasDuplicate = true;
+      }
+    }
+  });
+
+  return hasDuplicate;
+};
+
 export class ChartStore {
   constructor(id?: string) {
     this.id = id || uuid.v4();
@@ -845,35 +874,6 @@ export class ChartStore {
     this.annotationSpecs.delete(annotationId);
   }
 
-  isDuplicateAxis(
-    { position, title }: AxisSpec,
-    { tickLabels }: AxisTicksDimensions,
-    tickMap: Map<AxisId, AxisTicksDimensions>,
-    specMap: Map<AxisId, AxisSpec>,
-  ): boolean {
-    const firstTickLabel = tickLabels[0];
-    const lastTickLabel = tickLabels.slice(-1)[0];
-
-    let hasDuplicate = false;
-    tickMap.forEach(({ tickLabels: axisTickLabels }, axisId) => {
-      if (
-        !hasDuplicate &&
-        axisTickLabels &&
-        tickLabels.length === axisTickLabels.length &&
-        firstTickLabel === axisTickLabels[0] &&
-        lastTickLabel === axisTickLabels.slice(-1)[0]
-      ) {
-        const spec = specMap.get(axisId);
-
-        if (spec && spec.position === position && ((!title && !spec.title) || title === spec.title)) {
-          hasDuplicate = true;
-        }
-      }
-    });
-
-    return hasDuplicate;
-  }
-
   computeChart() {
     this.chartInitialized.set(false);
     // compute only if parent dimensions are computed
@@ -959,8 +959,7 @@ export class ChartStore {
 
       if (
         dimensions &&
-        (!this.hideDuplicateAxes ||
-          !this.isDuplicateAxis(axisSpec, dimensions, this.axesTicksDimensions, this.axesSpecs))
+        (!this.hideDuplicateAxes || !isDuplicateAxis(axisSpec, dimensions, this.axesTicksDimensions, this.axesSpecs))
       ) {
         this.axesTicksDimensions.set(id, dimensions);
       }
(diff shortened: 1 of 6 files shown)


src/chart_types/xy_chart/store/chart_state.ts
```diff
@@ -844,6 +845,35 @@ export class ChartStore {
...
+    { tickLabels }: AxisTicksDimensions,
+    tickMap: Map<AxisId, AxisTicksDimensions>,
+    specMap: Map<AxisId, AxisSpec>,
+  ): boolean {
+    const firstTickLabel = tickLabels[0];
+    const lastTickLabel = tickLabels.slice(-1)[0];
+
+    let hasDuplicate = false;
+    tickMap.forEach(({ tickLabels: axisTickLabels }, axisId) => {
+      if (
```
monfera: Instead of a `let hasDuplicate` and `forEach`, let's use a `tickmap.some(...)`

src/chart_types/xy_chart/store/chart_state.ts
```diff
@@ -844,6 +845,35 @@ export class ChartStore {
     this.annotationSpecs.delete(annotationId);
   }
 
+  isDuplicateAxis(
```
monfera: On an initial look, this method isn't using `this`, it would be better to turn it into a regular function `const isDuplicateAxis = ():({}) = {}` completely outside anything that's to do with objects/classes. Even if such functions use `this` here and there, it'd be advantageous to represent logic as that; we can still pass `this.foo` or whatever upon invocation. This lets us lighten up our objects, make the logic reusable and refactorable (even into another file if needed), and help us reduce object orientation.

src/components/react_canvas/reactive_chart.tsx
```diff
@@ -151,40 +161,50 @@ class Chart extends React.Component<ReactiveChartProps, ReactiveChartState> {
...
-        />,
-      );
-    });
-    return axesComponents;
+  getAxes = (): AxisConfig[] => {
+    const { axesVisibleTicks, axesSpecs, axesTicksDimensions, axesPositions } = this.props.chartStore!;
+    const ids = [...axesVisibleTicks.keys()];
+
+    return ids.reduce(
+      (acc, id) => {
```
monfera: I'd usually retain `reduce` for slightly more complex cases. Here it's essentially a `ids.filter().map()` except that you want to avoid the recalculation of ticks, axisSpec, axisTicksDimensions and axisPosition. But the result already carries these, so we could use `ids.map(id => {... return {..., configured: ticks && axisSpec && axisTicksDimensions && axisPosition}}).filter(({configured}) => configured)` - not sure if _configured_ is the right name - or just make the `map` return `null` when a condition fails and then `.filter(identity)` where `const identity = d => d`

src/components/react_canvas/reactive_chart.tsx
```diff
@@ -153,40 +162,44 @@ class Chart extends React.Component<ReactiveChartProps, ReactiveChartState> {
...
+    const { axesVisibleTicks, axesSpecs, axesTicksDimensions, axesPositions } = this.props.chartStore!;
+    const ids = [...axesVisibleTicks.keys()];
+
+    return ids
+      .map((id) => ({
+        id,
+        ticks: axesVisibleTicks.get(id),
+        axisSpec: axesSpecs.get(id),
+        axisTicksDimensions: axesTicksDimensions.get(id),
+        axisPosition: axesPositions.get(id),
```
markov00: you can map this to something that respect the props of the `<Axis>` component like {key, ticks, axisSpec, axisTickDimensions, axisPosition }` so then you can just render it like
```
<Axis
   {...axis}
   chartTheme={chartTheme}
   debug={debug}
   chartDimensions={chartDimensions}
/>
```

src/chart_types/xy_chart/store/chart_state.ts
```diff
@@ -114,6 +114,35 @@ export type CursorUpdateListener = (event?: CursorEvent) => void;
...
+    if (
+      !hasDuplicate &&
+      axisTickLabels &&
+      tickLabels.length === axisTickLabels.length &&
+      firstTickLabel === axisTickLabels[0] &&
+      lastTickLabel === axisTickLabels.slice(-1)[0]
+    ) {
+      const spec = specMap.get(axisId);
+
+      if (spec && spec.position === position && ((!title && !spec.title) || title === spec.title)) {
```
markov00: can we simplify this using `spec && spec.position === position && title == spec.title`?

src/components/react_canvas/reactive_chart.tsx
nickofthyme: see 2474776

src/chart_types/xy_chart/store/chart_state.ts
nickofthyme: see 2474776

src/chart_types/xy_chart/store/chart_state.ts
```diff
@@ -134,7 +134,7 @@ export const isDuplicateAxis = (
     ) {
       const spec = specMap.get(axisId);
 
-      if (spec && spec.position === position && ((!title && !spec.title) || title === spec.title)) {
+      if (spec && spec.position === position && title === spec.title) {
```
markov00: it's a double equal here if we want to mark as duplicate if one title is null and the other undefined (that in our case we can threat as no value)
```suggestion
      if (spec && spec.position === position && title == spec.title) {
```

src/components/react_canvas/reactive_chart.tsx
```diff
@@ -188,17 +188,8 @@ class Chart extends React.Component<ReactiveChartProps, ReactiveChartState> {
...
-        axisSpec={axisSpec}
-        axisTicksDimensions={axisTicksDimensions}
-        axisPosition={axisPosition}
-        ticks={ticks}
-        chartTheme={chartTheme}
-        debug={debug}
-        chartDimensions={chartDimensions}
-      />
+    return axes.map(({ key, ...axisProps }) => (
+      <Axis {...axisProps} key={key} chartTheme={chartTheme} debug={debug} chartDimensions={chartDimensions} />
```
markov00: Have you extracted the key a bit more explicit on the code?

src/chart_types/xy_chart/store/chart_state.ts
nickofthyme: The value cannot be `null` based on the types correct?

Nonetheless, I also don't like double equals in javascript because it is loose equality, and generally a bad practice. It also can cause unwanted conversions producing inaccurate results. I'd like to enforce this throughout the code eventually.

I think the explicit equality adding twice the code to clearly indicate the comparison is worth it.

src/components/react_canvas/reactive_chart.tsx
nickofthyme: Sorry what do you mean by this?

I tried spreading the props with the key included and I get a linting error not adding the key to `map`d child. Is that what you mean?

markov00: yes, sorry :( 
strange that the ts linter doesn't know about that

nickofthyme: Ya seems like they can get the keys of the object to verify

image

src/chart_types/xy_chart/store/chart_state.ts
markov00: yep that value cannot be null, but I was thinking on the case the chart used in a JS world (people can do strange things there).
You are right about the loose equality, I use that only in rare case checking when checking for null and undefined, in that case you are right, we can get some strange results. So please ignore my comment here

markov00:
jenkins, test this please

src/chart_types/xy_chart/store/chart_state.ts
nickofthyme: To your point about people using this library in javaScript and not minding the types.

I 100% agree, but I think if we want to address that there are numerous places where we assume values to be provided/required. I think the `Theme` is a perfect example where we assume options are provided on the type were in js they may not.

PR #374, opened 2019-09-12
feat(axis): add visibility to tick style

### LINKED ISSUE #330: Remove tick lines
**Describe the bug**
The tick lines that are visible between the tick label and the grid adds lots of noise to the tick labels. Instead we should be relying on the grid to mark out the values across the whole chart so users can actually follow a line to the correct value.

**Expected behavior**
I tried removing them by setting `theme.tickLineStyle.strokeWidth = 0`. While this does make them disappear, there is still a lot of extra space between the tick label and the axis line leftover from where the tick line should be.

image

**Either,** 
A. just get rid of the idea of tick lines
B. make sure we can fully remove theme via the theme

 - Elastic Charts: 9.0.4

nickofthyme:
## Summary

Add visibility property to `tickLineStyle` in `Theme`.

BREAKING CHANGE: `theme.axes.tickLineStyle.visible` is now required (default base is `false`)

closes #330

Screen Recording 2019-09-12 at 01 20 PM

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+## [12.0.2](https://github.com/elastic/elastic-charts/compare/v12.0.1...v12.0.2) (2019-09-16)
+
+
+### Bug Fixes
+
+* **reactive_chart:** fix order of instantiation of onBruchEnd callback ([#376](https://github.com/elastic/elastic-charts/issues/376)) ([527d68d](https://github.com/elastic/elastic-charts/commit/527d68d)), closes [#360](https://github.com/elastic/elastic-charts/issues/360)
+
+## [12.0.1](https://github.com/elastic/elastic-charts/compare/v12.0.0...v12.0.1) (2019-09-12)
+
+
+### Bug Fixes
+
+* **theme:** fix grid position check ([#373](https://github.com/elastic/elastic-charts/issues/373)) ([af4805f](https://github.com/elastic/elastic-charts/commit/af4805f)), closes [#372](https://github.com/elastic/elastic-charts/issues/372)
+
 # [12.0.0](https://github.com/elastic/elastic-charts/compare/v11.2.0...v12.0.0) (2019-09-11)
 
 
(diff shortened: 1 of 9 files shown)


markov00:
Can we have the default visibility to true, and change the default to false when we are going to implement a better default style?

nickofthyme:
@markov00 I set the default to false as that is the style EUI is suggesting. Are you saying to set the default to true and then convert to false later?

markov00:
Yes, I would like to have a single PR that adjust the default theme to a better one. For example in this case hide the ticks but also reduce the tick size to 0 for example. Together with some better defaults of style for the axis colors, grid style, line annotations etc

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@elastic/charts",
   "description": "Elastic-Charts data visualization library",
-  "version": "12.0.2",
+  "version": "12.0.0",
```
markov00: we cannot go back

CHANGELOG.md
```diff
@@ -1,17 +1,3 @@
...
-
-* **reactive_chart:** fix order of instantiation of onBruchEnd callback ([#376](https://github.com/elastic/elastic-charts/issues/376)) ([527d68d](https://github.com/elastic/elastic-charts/commit/527d68d)), closes [#360](https://github.com/elastic/elastic-charts/issues/360)
-
-## [12.0.1](https://github.com/elastic/elastic-charts/compare/v12.0.0...v12.0.1) (2019-09-12)
-
-
-### Bug Fixes
-
-* **theme:** fix grid position check ([#373](https://github.com/elastic/elastic-charts/issues/373)) ([af4805f](https://github.com/elastic/elastic-charts/commit/af4805f)), closes [#372](https://github.com/elastic/elastic-charts/issues/372)
-
```
markov00: Why does changes?

package.json
nickofthyme: Sorry I didn't see that, must have been a bad merge. Fix in 6eda1fa

CHANGELOG.md
nickofthyme: Fix in 6eda1fa

markov00:
jenkins, test this please

PR #385, opened 2019-09-23
feat(style): point style overrides for area and line charts

nickofthyme:
#### BarSeries
Each bar `datum` of a **Bar** data series can be assigned a custom color or style with the `styleAccessor` prop.

The `styleAccessor` prop expects a callback function which will be called on _every_ `datum` in _every_ bar series with the signiture below. This callback should return a color `string` or a partial `BarSeriesStyle`, which will override any series-level styles for the respective `datum`. You are passed `geometryId` to identify the series the `datum` belongs to and the raw `datum` to derive conditions against.

```ts
type BarStyleAccessor = (
  datum: RawDataSeriesDatum,
  geometryId: GeometryId,
) => RecursivePartial<BarSeriesStyle> | Color | null;
```

#### LineSeries and AreaSeries points
Each point `datum` of a **Line** or **Area** data series can be assigned a custom color or style with the `pointStyleAccessor` prop.

The `pointStyleAccessor` prop expects a callback function which will be called on _every_ `datum` in _every_ line or area series with the signiture below. This callback should return a color `string` or a partial `PointStyle`, which will override any series-level styles for the respective `datum`. You are passed `geometryId` to identify the series the `datum` belongs to and the raw `datum` to derive conditions against.

```ts
type PointStyleAccessor = (
  datum: RawDataSeriesDatum,
  geometryId: GeometryId,
) => RecursivePartial<PointStyle> | Color | null;
```

## Demo

Screen Recording 2019-09-22 at 08 59 PM

### Checklist

### DIFF
--- a/wiki/overview.md
+++ b/wiki/overview.md
@@ -216,13 +216,26 @@ In the case of small multiples, each `SplittedSeries` computes its own x and y d
 
 ### Color/Style overrides
 
-Each `datum` of a **Bar Chart** data series can be assigned a custom color or style with the `styleAccessor` prop.
+#### BarSeries
+Each bar `datum` of a **Bar** data series can be assigned a custom color or style with the `styleAccessor` prop.
 
 The `styleAccessor` prop expects a callback function which will be called on _every_ `datum` in _every_ bar series with the signiture below. This callback should return a color `string` or a partial `BarSeriesStyle`, which will override any series-level styles for the respective `datum`. You are passed `geometryId` to identify the series the `datum` belongs to and the raw `datum` to derive conditions against.
 
 ```ts
-type StyleAccessor = (
+type BarStyleAccessor = (
   datum: RawDataSeriesDatum,
   geometryId: GeometryId,
 ) => RecursivePartial<BarSeriesStyle> | Color | null;
+```
+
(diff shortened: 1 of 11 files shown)


wiki/overview.md
```diff
@@ -216,13 +216,26 @@ In the case of small multiples, each `SplittedSeries` computes its own x and y d
...
+#### LineSeries and AreaSeries points
+Each point `datum` of a **Line** or **Area** data series can be assigned a custom color or style with the `pointStyleAccessor` prop.
+
+The `pointStyleAccessor` prop expects a callback function which will be called on _every_ `datum` in _every_ line or area series with the signiture below. This callback should return a color `string` or a partial `PointStyle`, which will override any series-level styles for the respective `datum`. You are passed `geometryId` to identify the series the `datum` belongs to and the raw `datum` to derive conditions against.
+
+```ts
+type PointStyleAccessor = (
+  datum: RawDataSeriesDatum,
+  geometryId: GeometryId,
+) => RecursivePartial<PointStyle> | Color | null;
```
markov00: Could you add a description of the `null` return, when it needs to be used?

nickofthyme: ef0c1ff

PR #389, opened 2019-09-23
feat(axis): add option for integer only axis ticks

### LINKED ISSUE #387: Ability to set axis to integer value
Allow prop on `Axis` (i.e. `integersOnly`) to remove non-integer ticks, tick labels and grid lines.

**Additional context**
Gridlines will also need to be removed pertaining to removed ticks.

**Kibana Cross Issues**
- https://github.com/elastic/kibana/issues/28614

**Checklist**

rshen91:
## Summary

Fixes #387

This PR should allow the ability for users to filter the ticks to be only integers vs float values:
PR387

### Checklist

### DIFF
--- a/src/chart_types/xy_chart/utils/axis_utils.test.ts
+++ b/src/chart_types/xy_chart/utils/axis_utils.test.ts
@@ -31,6 +31,7 @@ import {
   mergeYCustomDomainsByGroupId,
   isVerticalGrid,
   isHorizontalGrid,
+  getIntegerTicks,
 } from './axis_utils';
 import { CanvasTextBBoxCalculator } from '../../../utils/bbox/canvas_text_bbox_calculator';
 import { SvgTextBBoxCalculator } from '../../../utils/bbox/svg_text_bbox_calculator';
@@ -85,6 +86,7 @@ describe('Axis computational utils', () => {
       return `${value}`;
     },
     showGridLines: true,
+    integersOnly: false,
   };
 
   const horizontalAxisSpec: AxisSpec = {
@@ -99,6 +101,7 @@ describe('Axis computational utils', () => {
     tickFormat: (value: any) => {
       return `${value}`;
     },
+    integersOnly: false,
   };
 
   const verticalAxisSpecWTitle: AxisSpec = {
@@ -115,6 +118,7 @@ describe('Axis computational utils', () => {
       return `${value}`;
     },
     showGridLines: true,
+    integersOnly: false,
   };
 
   // const horizontalAxisSpecWTitle: AxisSpec = {
@@ -1362,5 +1366,15 @@ describe('Axis computational utils', () => {
     const axisConfigTickLabelPadding = 1;
 
     expect(getAxisTickLabelPadding(axisConfigTickLabelPadding, axisSpecStyle)).toEqual(2);
+  });
+  test('should expect only integers', () => {
+    const yScale = getScaleForAxisSpec(verticalAxisSpec, xDomain, [yDomain], 0, 0, 100, 0);
+    expect(yScale!.ticks()).toEqual([0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1]);
+    expect(getIntegerTicks(true, yScale!)).toEqual([0, 1]);
+  });
+  test('should not expect only integers', () => {
+    const yScale = getScaleForAxisSpec(verticalAxisSpec, xDomain, [yDomain], 0, 0, 100, 0);
+    expect(yScale!.ticks()).toEqual([0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1]);
+    expect(getIntegerTicks(false, yScale!)).toEqual([0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1]);
   });
 });
--- a/src/chart_types/xy_chart/utils/axis_utils.test.ts
+++ b/src/chart_types/xy_chart/utils/axis_utils.test.ts
@@ -32,6 +32,7 @@ import {
   isVerticalGrid,
   isHorizontalGrid,
   getTicks,
+  getTickLabel,
 } from './axis_utils';
 import { CanvasTextBBoxCalculator } from '../../../utils/bbox/canvas_text_bbox_calculator';
 import { SvgTextBBoxCalculator } from '../../../utils/bbox/svg_text_bbox_calculator';
@@ -1374,5 +1375,10 @@ describe('Axis computational utils', () => {
   test('should not expect only integers', () => {
     const yScale = getScaleForAxisSpec(verticalAxisSpec, xDomain, [yDomain], 0, 0, 100, 0);
     expect(getTicks(yScale!)).toEqual([0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1]);
+  });
+  test('expect tick labels to be whole integers', () => {
+    const tickFormat = (d: number) => Number(d).toFixed(2);
+    expect(getTickLabel(0.0, tickFormat, true)).toEqual('0');
+    expect(getTickLabel(0.0, tickFormat, false)).toEqual('0.00');
   });
 });
--- a/src/chart_types/xy_chart/utils/axis_utils.test.ts
+++ b/src/chart_types/xy_chart/utils/axis_utils.test.ts
@@ -1370,11 +1370,11 @@ describe('Axis computational utils', () => {
   test('should expect only integers', () => {
     const yScale = getScaleForAxisSpec(verticalAxisSpec, xDomain, [yDomain], 0, 0, 100, 0);
     expect(yScale!.ticks()).toEqual([0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1]);
-    expect(getIntegerTicks(true, yScale!)).toEqual([0, 1]);
+    expect(getIntegerTicks(yScale!, true)).toEqual([0, 1]);
   });
   test('should not expect only integers', () => {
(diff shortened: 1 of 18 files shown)


src/chart_types/xy_chart/utils/axis_utils.ts
```diff
@@ -400,7 +400,7 @@ export function getAvailableTicks(
   totalBarsInCluster: number,
   enableHistogramMode: boolean,
 ): AxisTick[] {
-  const ticks = scale.ticks();
+  const ticks = axisSpec.integersOnly ? scale.ticks().filter((item) => item % 1 === 0) : scale.ticks();
```
nickofthyme: since `ticks()` returns type `any[]` you may want to add `typeof item === 'number'` to the check
```suggestion
  const ticks = axisSpec.integersOnly ? scale.ticks().filter((item) => typeof item === 'number' && item % 1 === 0) : scale.ticks();
```

rshen91: This makes so much sense! I'm adding this in commit 000a4ac

rshen91:
@nickofthyme Something I'm wondering about with the current implementation is if in the specs.ts file

``` js 
integersOnly: boolean
``` 

makes sense or if it should be an optional property in the AxisSpec? I was leaning towards non-optional to avoid undefined issues but I can see it making sense either way.

src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -238,6 +238,8 @@ export interface AxisSpec {
   domain?: DomainRange;
   /** Object to hold custom styling */
   style?: AxisStyle;
+  /** Show only integar values **/
+  integersOnly: boolean;
```
rshen91: @nickofthyme  I was leaning towards non-optional to avoid undefined issues but I can see it making sense either way.

nickofthyme: Ya totally agree. Anything that should obviously be defaulted a certain way should not be required. Like the `showLegend` prop is `false` by default and if people had to set that for every `Chart` that would be a bit much.

Only things that prevent us from rendering the `Chart` like `data`, `yAccessor`, etc. should be required. At least that's how I think about it. There's probably a few exceptions to that rule but limited.

src/chart_types/xy_chart/utils/axis_utils.ts
```diff
@@ -393,14 +393,16 @@ export function getLeftAxisMinMaxRange(chartRotation: Rotation, height: number)
       return { minRange: 0, maxRange: height };
   }
 }
-
+export function getIntegerTicks(scale: Scale, hasInteger: boolean = false) {
```
nickofthyme: Sorry to nitpick here but I think this function should be `getTicks` because it returns both ticks and integer ticks. Then you can pass the `integersOnly` boolean as an argument.

```suggestion
export function getTicks(scale: Scale, integersOnly: boolean = false) {
```

src/chart_types/xy_chart/utils/axis_utils.test.ts
```diff
@@ -1363,4 +1367,14 @@ describe('Axis computational utils', () => {
 
     expect(getAxisTickLabelPadding(axisConfigTickLabelPadding, axisSpecStyle)).toEqual(2);
   });
+  test('should expect only integers', () => {
+    const yScale = getScaleForAxisSpec(verticalAxisSpec, xDomain, [yDomain], 0, 0, 100, 0);
+    expect(yScale!.ticks()).toEqual([0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1]);
```
nickofthyme: I would say this test is redundant, meaning that `ticks` is a method on the `scale` coming from d3. You're welcome to keep it if you disagree.

src/chart_types/xy_chart/utils/axis_utils.ts
```diff
@@ -393,14 +393,16 @@ export function getLeftAxisMinMaxRange(chartRotation: Rotation, height: number)
...
+  return hasInteger ? scale.ticks().filter((item) => typeof item === 'number' && item % 1 === 0) : scale.ticks();
```
nickofthyme: I think it might be logical to map all ticks to `.toFixed(0)` to actually make them an integer and not `1.00`. What do you think?

image

src/chart_types/xy_chart/utils/axis_utils.test.ts
rshen91: Yeah, that makes sense to remove it. Initially, my thought was to include that test to show the entire scale.ticks(), but there's the test right below it showing the full range of tick options so a new person can see that and figure out what the test is doing. 👌

src/chart_types/xy_chart/utils/axis_utils.ts
rshen91: I like this! I'm still working on better naming of variables etc so this is very appreciated. Thanks!

nickofthyme: I takes practice and there are sometimes not great options.

src/chart_types/xy_chart/utils/axis_utils.ts
rshen91: What you're saying makes sense. 
tickFormat in the grid story is hardcoded as ```(d) => Number(d).toFixed(2)``` and I'm working through how to conditionally apply .toFixed(0) when integersOnly is true

rshen91: Ok I think this is working in commit 607af9b

src/chart_types/xy_chart/utils/axis_utils.test.ts
rshen91: addressing in commit 607af9b

src/chart_types/xy_chart/utils/axis_utils.ts
rshen91: commit 607af9b for changes

nickofthyme:
I'm also thinking if this should come from the scale itself. Because if someone calls ticks on the scale they will not get the integer values. If you want to zoom and discuss lmk.

rshen91:
> I'm also thinking if this should come from the scale itself. Because if someone calls ticks on the scale they will not get the integer values. If you want to zoom and discuss lmk.

Commit a6896d2 added to address this refactor. I still have to play with how to properly implement `.toFixed(0)` and add some tests. But wanted to get this commit out since it seems to be working with scale_continuous now in the grid basic story 🎉

src/utils/scales/scale_continuous.ts
```diff
@@ -188,11 +192,18 @@ export class ScaleContinuous implements Scale {
           return this.domain[0] + i * this.minInterval;
         });
       } else {
-        this.tickValues = this.d3Scale.ticks(ticks);
+        this.tickValues = this.getTicks(ticks, integersOnly!);
```
markov00: This only applies to Linear/Log/Sqrt scales on the Y axis (the only axis what has minInterval === 0. I think it's enough for now (as the main use case for this property is to be used on the Y axis) but we should also take a look at applying the same principle also on the X scale.

The `minInterval` is always computed on the xScale as the minimum difference between consecutive values in the dataset, and it's never computed on the yScale(default to 0).

src/utils/scales/scale_continuous.ts
```diff
@@ -105,6 +105,8 @@ interface ScaleOptions {
   ticks: number;
   /** true if the scale was adjusted to fit one single value histogram */
   isSingleValueHistogram: boolean;
+  /** Show only integar values **/
```
markov00: typo: `integar` -> `integer`

rshen91: Ah! Good catch, thanks for this. I'm addressing in commit 9d00f81

src/utils/scales/scale_continuous.test.ts
```diff
@@ -388,4 +388,14 @@ describe('Scale Continuous', () => {
...
+  describe('ticks as integers or floats', () => {
+    test('should expect integers for tickValues', () => {
+      const domain: Domain = [0, 7];
+      const minRange = 0;
+      const maxRange = 100;
+      const scale = new ScaleContinuous({ type: ScaleType.Linear, domain, range: [minRange, maxRange] });
+      expect(scale.getTicks(10, true)).toEqual([0, 1, 2, 3, 4, 5, 6, 7]);
+      expect(scale.getTicks(10, false)).toEqual([0, 0.5, 1, 1.5, 2, 2.5, 3, 3.5, 4, 4.5, 5, 5.5, 6, 6.5, 7]);
+    });
+  });
```
nickofthyme: This is good but each `test` should test one thing to make tests easier to fix when they break. Lumping all varients into one `test` block is bad but lumping all into on `describe` with nested `test`s is great.

```ts
test('should return normal ticks', () => {
...
});

test('should return only integer ticks', () => {
...
});
```

src/utils/scales/scale_continuous.ts
```diff
@@ -188,11 +192,18 @@ export class ScaleContinuous implements Scale {
...
+        this.tickValues = this.getTicks(ticks, integersOnly!);
       }
     }
   }
-
+  getTicks(ticks: number, integersOnly: boolean) {
+    return integersOnly
+      ? this.d3Scale
+          .ticks(ticks)
+          .filter((item: number) => item % 1 === 0)
```
nickofthyme: is this safe without the `typeof === 'number'` check?

rshen91: Gosh I forgot this part T_T thanks! Adding in commit c5c0e66

src/utils/scales/scale_continuous.test.ts
rshen91: This makes sense! Addressing in commit c5c0e66

src/utils/scales/scale_continuous.ts
rshen91: Opened new issue for tracking this thanks!

PR #391, opened 2019-09-24
feat(tooltip): allow postfix banded area series

### LINKED ISSUE #162: Display y0 and y1 names on tooltip for band charts
A band area chart tooltip does't display two different names for the y0 and y1 value.
The same happens for a bar chart with y0 accessors, it will display only the y1 value.

nickofthyme:
## Summary

Allow user to set postfix string or accessor for upper and lower bound of banded series to distinguish between values.

Adds `y1AccessorFormat` and `y0AccessorFormat` props to `BasicSeriesSpec`. `y1AccessorFormat` defaults to `' - upper'` and `y0AccessorFormat` defaults to `' - lower'`

```ts
AccessorFormat = string | ((value: string) => string);
```

closes #162

see local storybook link http://localhost:9001/?path=/story/area-chart--band-area-chart

## Area banded series
Screen Recording 2019-09-24 at 12 17 PM

### Checklist


src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -89,6 +89,18 @@ export interface SeriesSpec {
...
+   *
+   * @default 'upper'
+   */
+  y0AccessorPostfix?: string;
+  /**
+   * Postfix for y1 accesor when using `y0Accessors`
+   *
+   * @default 'lower'
+   */
+  y1AccessorPostfix?: string;
```
markov00: Hey @nickofthyme 
can we change this to a postfix string OR a function that is invoked with the series name and return the user formatted series name like:
```
y0AccessorName={(seriesName) => { return `[min] ${seriesName}`; }}
```
because I think the position of the min/max fix depends on the user's language.
We can support both a string or a function?
In the case of the function, we can use it also on non-banded situations?

nickofthyme: Dangit! You're right, I should have thought of that.

nickofthyme: dca7589

src/chart_types/xy_chart/tooltip/tooltip.ts
```diff
@@ -25,30 +26,29 @@ export function getSeriesTooltipValues(tooltipValues: TooltipValue[], defaultVal
...
-    name = seriesKey.join(' - ');
+    displayName = seriesKey.join(' - ');
   } else {
-    name = spec.name || `${spec.id}`;
+    displayName = name || `${id}`;
+  }
+
+  if (banded) {
+    const formatter = accessor === 'y0' ? y0AccessorFormat : y1AccessorFormat;
+    displayName = getAccessorFormatLabel(formatter, displayName);
```
markov00: One last comment on that: as this can be used also in other cases, like when you want to change the name of a specific series, why not remove the if(banded) check and apply the accessorFormat if and only if there is an accessorFormat configured? can be only for y0 or only for y1

nickofthyme: I think that would be for naming the series itself. This is merly to append or prepend the given label.

The ability to name the series based on accessors etc. should not be in the PR.

src/chart_types/xy_chart/store/utils.ts
```diff
@@ -474,7 +474,7 @@ export function renderGeometries(
         color,
         (spec as LineSeriesSpec).curve || CurveType.LINEAR,
         ds.specId,
-        Boolean(spec.y0Accessors),
+        Boolean(spec.y0Accessors && spec.y0Accessors.length > 0),
```
markov00: good catch!

src/chart_types/xy_chart/tooltip/tooltip.ts
```diff
@@ -40,7 +40,7 @@ export function formatTooltip(
     displayName = name || `${id}`;
   }
 
-  if (banded) {
+  if (y0Accessors && y0Accessors.length > 0) {
```
markov00: nit: small utility function? 
as we are using this also on `src/chart_types/xy_chart/store/utils.ts`?
```
hasY0Accessors({y0Accessors}) {
    return y0Accessors && y0Accessors.length > 0
}
```

nickofthyme: 7304cb3

PR #395, opened 2019-09-27
fix(rendering): out-of-domain rendering of points/bars/lines/areas

### LINKED ISSUE #386: End Points on Line/Area are clipped when used with Bar charts
**Describe the bug**
When using a line/area series with a bar series. The range is set in a way to exclude the last point in the line series.

**To Reproduce**

```ts
// playground.tax
import React, { Fragment } from 'react';
import {
  Axis,
  Chart,
  getAxisId,
  getSpecId,
  Position,
  ScaleType,
  Settings,
  BarSeries,
  LineSeries,
  AreaSeries,
} from '../src';

  render() {
    return (
      <Fragment>
        <div className="chart">
            <Settings
              theme={{
                areaSeriesStyle: {
                  point: {
                    visible: true,
                  },
                },
              }}
            />
            <Axis
              id={getAxisId('bottom')}
              position={Position.Bottom}
              title={'Bottom axis'}
              showOverlappingTicks={true}
            />
            <Axis
              id={getAxisId('left')}
              title={'Left axis'}
              position={Position.Left}
              tickFormat={(d: any) => Number(d).toFixed(2)}
            />

            <BarSeries
              id={getSpecId('bar')}
              xScaleType={ScaleType.Linear}
              yScaleType={ScaleType.Linear}
              xAccessor="x"
              yAccessors={['y']}
              data={[{ x: 0, y: 2 }, { x: 1, y: 7 }, { x: 2, y: 3 }, { x: 3, y: 6 }]}
            />

            <LineSeries
              id={getSpecId('line')}
              xScaleType={ScaleType.Linear}
              yScaleType={ScaleType.Linear}
              xAccessor="x"
              yAccessors={['y']}
              data={[{ x: 0, y: 1 }, { x: 1, y: 6 }, { x: 2, y: 2 }, { x: 3, y: 5 }]}
            />

            <AreaSeries
              id={getSpecId('area')}
              xScaleType={ScaleType.Linear}
              yScaleType={ScaleType.Linear}
              xAccessor="x"
              yAccessors={['y']}
              data={[{ x: 0, y: 0.5 }, { x: 1, y: 4 }, { x: 2, y: 1 }, { x: 3, y: 4 }]}
            />
          </Chart>
        </div>
      </Fragment>
    );
  }
}
```

**Expected behavior**
Points on line and area series should clear the `range` check and render on the chart.

**Screenshots**
image

**Additional context**
The range check in the `renderPoints` function below returns false claiming the x value is outside of the range. Could be computing the Bar chart range incorrectly. A solution might be to compare with domain instead of the range.

https://github.com/elastic/elastic-charts/blob/e1b0cd607b8aeaa49f2c68f32eabcc085604be17/src/chart_types/xy_chart/rendering/rendering.ts#L170-L172

 - OS: any
 - Browser: chrome
 - Elastic Charts: `latest`

markov00:
## Summary
On a previous PR: https://github.com/elastic/elastic-charts/pull/355 we fixed the clipping of
elements on the chart edges. This introduced an unwanted bug where we where removing from the chart also points that are within the chart range itself. Instead of checking if a geometry position is
within the range of the scale, this PR checks for the value to be within the domain of the scale.

fix #386

On the following examples (the same that comes with the playground) the X domain is limited to a max value of 3.8. This will remove the bar that is at position 4, the point at position 4 and limit the line to the last valid value (3).

In contrast from the current(master) version where the bar at position 4 is displayed together with a non-delimited line that goes from 3 to "somewhere"
**from:**
Screenshot 2019-09-27 at 14 41 47

In this PR we remove everything that is after the point 3.8, removing also the line segment that connect the 3rd point with the "removed" 4th.

**to:**
Screenshot 2019-09-27 at 14 42 21

### Checklist

### DIFF
--- a/src/utils/scales/scale_continuous.ts
+++ b/src/utils/scales/scale_continuous.ts
@@ -269,6 +269,9 @@ export class ScaleContinuous implements Scale {
     const max = this.domain[this.domain.length - 1];
     return max === min;
   }
+  isValueInDomain(value: number) {
+    return value >= this.domain[0] && value <= this.domain[1];
+  }
 }
 
 export function isContinuousScale(scale: Scale): scale is ScaleContinuous {
--- a/.playground/playgroud.tsx
+++ b/.playground/playgroud.tsx
@@ -1,50 +1,75 @@
 import React, { Fragment } from 'react';
-import { Axis, Chart, getAxisId, getSpecId, Position, ScaleType, Settings, BarSeries } from '../src';
+import {
+  Axis,
+  Chart,
+  getAxisId,
+  getSpecId,
+  Position,
+  ScaleType,
+  Settings,
+  BarSeries,
+  LineSeries,
+  AreaSeries,
+} from '../src';
 
 export class Playground extends React.Component {
   render() {
     return (
       <Fragment>
         <div className="chart">
-          <Chart>
+          <Chart className="story-chart">
             <Settings
-              showLegend={true}
               theme={{
-                axes: {
-                  gridLineStyle: {
-                    horizontal: {
-                      stroke: 'red',
-                      strokeWidth: 0.5,
-                      opacity: 1,
-                      dash: [0, 0],
-                    },
-                    vertical: {
-                      stroke: 'blue',
-                      strokeWidth: 0.5,
-                      opacity: 1,
-                      dash: [4, 4],
-                    },
+                areaSeriesStyle: {
+                  point: {
+                    visible: true,
                   },
                 },
               }}
+              xDomain={{
+                max: 3.8,
+              }}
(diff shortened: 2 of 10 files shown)


src/utils/scales/scale_continuous.ts
```diff
@@ -269,6 +269,9 @@ export class ScaleContinuous implements Scale {
     const max = this.domain[this.domain.length - 1];
     return max === min;
   }
+  isValueInDomain(value: number) {
+    return value >= this.domain[0] && value <= this.domain[1];
```
nickofthyme: I love this check on the scale!

PR #402, opened 2019-10-03
refactor(typings): prepare for upgrade TS to 3.7

markov00:
## Summary

There are currently to main issue that needs to be resolved before upgrading TS to 3.7: 
- a duplicated `GeometryStyle` type, that I renamed to `GeometryStateStyle` (the one used on the `SharedGeometryStyle` renamed to `SharedGeometryStateStyle` to align better its name to its function )
- an issue with the missing `layerX` and `layerY` property on the MouseEvent after this PR merged into TS https://github.com/microsoft/TypeScript/pull/32578/files#diff-46fd87925e4552c166ec188712741c3fL10335-L10336 I've switched to use `offsetX` and `offsetY` that we are already using on the `mouseMove` event listener.

**BREAKING CHANGE:**
We have few exported styles, used in the Theme that are changed: `SharedGeometryStyle` to `SharedGeometryStateStyle` and `GeometryStyle` to `GeometryStateStyle`

**@reviewers** please double check locally on your browsers if the behaviour of the mouse and the brush is the same (I've tested it on chrome, safari, firefox and IE11 using various chart configuration in the playground).

ref issue in Kibana: https://github.com/elastic/kibana/pull/47188

### Checklist


src/chart_types/xy_chart/utils/series.ts
```diff
@@ -255,8 +255,8 @@ export function getRawDataSeries(
     areaSeries: 0,
   };
   const seriesSpecsCount = seriesSpecs.length;
-  let i;
-  for (i = 0; i < seriesSpecsCount; i++) {
+  let i = 0;
+  for (; i < seriesSpecsCount; i++) {
```
nickofthyme: what is the point of this? over `for (let i = 0;...`

src/components/react_canvas/reactive_chart.tsx
```diff
@@ -61,8 +61,8 @@ function limitPoint(value: number, min: number, max: number) {
 }
 function getPoint(event: MouseEvent, extent: BrushExtent): Point {
   const point = {
-    x: limitPoint(event.layerX, extent.minX, extent.maxX),
-    y: limitPoint(event.layerY, extent.minY, extent.maxY),
+    x: limitPoint(event.offsetX, extent.minX, extent.maxX),
+    y: limitPoint(event.offsetY, extent.minY, extent.maxY),
```
nickofthyme: Ya I tried to figure this out 😞 

It has 3 type defs, one of which is the right one but just can't resolve correctly.

src/chart_types/xy_chart/utils/series.ts
markov00: in our case is basically the same, but moving the variable initialization out of the for loop is like gain super speed powers: https://medium.com/kbdev/voyage-to-the-most-efficient-loop-in-nodejs-and-a-bit-js-5961d4524c2e

markov00: we should remove all the forEach/map/reduce loops if we want to speed up things

nickofthyme: Oh sweet, looks like a good read.

What is the performance gain? 2x? 3x?

markov00: 2.2x for big arrays

nickofthyme: what's _big_? `1k`? `100k`? `1M`?

monfera: I support `for` loops for any time the count of things can be unbound, can be done even if it's in the dozens, hundreds, thousands or more (depends on other things eg. is the loop already inside some other loop). Speedup due to this is indeed significant (I think it's always a low multiple, but there are constant factors so overall speedup on a 10 element array won't be mind-blowing), and especially compared to `map`, `filter` and most `reduce` calls, the memory allocation can be avoided.
 
We should exploit the limited scope of the `let` binding, making the index only accessible inside the loop, which is also the more idiomatic JavaScript, eg. 
```
for(let i = 0; i < seriesSpecsCount; i++) { ... }
```

monfera: Ah just see the actual Q (`let` before, or inside) - the article is from 2017 and the top part says it's fixed now. I think it's (well, in the case of `let`, was) OK to use such micro-optimizations (that lead to others wondering why) if performance profiling shows that part to be a tight spot; in such a case we might add a small comment with the url of the article or similar

monfera: Re `forEach`, `map` etc. vs `for`: the former notations are usually easier to grasp, so if I see a `for` somewhere, I assume that it's either because it's a foreseen or real performance  hotspot, or the structure of the problem doesn't lend itself well to functional equivalents. My feeling is that for our use cases, it's fine to keep functional forms around until they're shown to be a bottleneck. Ie. I wouldn't go ahead refactoring code for mass converting `map`s to `for`s; more might be lost in code clarity than what's gained in perf.

If we really need the performance somewhere, the row (object) organized data representation is already quite bad (CPU cache locality; memory requirements; garbage generation etc.) and the most optimal structure is often column-oriented, ie. instead of 
```
[
   { x: 3, y: 5 },
   { x: 4, y: 6 },
   ...
]
```
the column oriented, maybe even typed (for WebGL, or WebWorker transferable object*) alternative:
```
{ x: [3, 4, ...], y: [5, 6, ...] }
```
(*) Chrome and IIRC the other browsers automatically represent untyped arrays of homogeneous values as typed arrays. So using a typed array won't always lead to a better read performance. But making a typed array might be slightly cheaper, and it's more reliable for performance (and buys WebGL and WebWorker transferability).

So the upshot is that switching plain code, which makes and disuses large numbers of small objects like `{ x: 4, y: 6 }` won't be helped as much by `for` (vs. `map`) as code whose data structures are designed for high performance, memory locality etc. off the bat, in which case `map` wouldn't even work 😄 Sorry for stating what's mostly obvious for all of us but it's neat to chat about performance 😃

nickofthyme: Ya love the talk on performance. @markov00 and I were thinking if there was a way to convert all HOF prototypes to a more performant version using [sanic.js](https://github.com/kb-dev/sanic.js#api)

joshdover:
Can this be merged? TypeScript 3.7 has now been released and we'd like to get Kibana upgraded.

PR #405, opened 2019-10-07
ci: use linux docker nodes

nickofthyme:
## Summary

Tests have proven to fail on several `linux` nodes.

Switching to `docker` nodes

https://github.com/GoogleChrome/puppeteer/issues/807#issuecomment-330056692


nickofthyme:
use `docker` nodes instead.

PR #413, opened 2019-10-10
fix: handle null y0 values on y log scale rendering

markov00:
## Summary

The y0 value of each bar is `null` by default if we don't have a y0 value. With a log scale we were not check for that null value, creating an invalid height value for the bars. The same happens with other scale type.

The issue (two bars should be displayed):
Screenshot 2019-10-10 at 15 38 50

The fix:
Screenshot 2019-10-10 at 15 38 46

I've also fixed a small issue in a story with a wrong data set

### Checklist


src/chart_types/xy_chart/rendering/rendering.ts
```diff
@@ -405,7 +409,13 @@ export function renderLine(
 
   const pathGenerator = line<DataSeriesDatum>()
     .x(({ x }) => xScale.scale(x) - xScaleOffset)
-    .y(({ y1 }) => yScale.scale(y1))
+    .y(({ y1 }) => {
+      if (y1 !== null) {
+        return yScale.scale(y1);
+      }
+      // this should never happen thanks to the defined function
```
nickofthyme: 😄

PR #416, opened 2019-10-11
feat: fit functions for null y1 values

### LINKED ISSUE #388: Stacked area charts render with gaps when the X coordinates are not continuous
**Describe the bug**
Gaps appear in the rendering when you skip an X coordinate in your stacked area chart. It only appears in stacked areas, since stacked bars render correctly.

Screenshot 2019-09-23 14 08 34

**To Reproduce**
This is an example of discontinuous data:

```
      <AreaSeries
        splitSeriesAccessors={['split']}
        stackAccessors={['x']}
        id={getSpecId('y')}
        xAccessor={'x'}
        yAccessors={['y']}
        data={[
          {
            x: 5,
            y: 10,
            split: 'a',
          },
          {
            x: 5,
            y: 30,
            split: 'b',
          },
          {
            x: 10,
            y: 2,
            split: 'a',
          },
          // {
          //   x: 10,
          //   y: 3,
          //   split: 'b',
          // },
          {
            x: 20,
            y: 30,
            split: 'a',
          },
          {
            x: 20,
            y: 20,
            split: 'b',
          },
        ]}
        xScaleType={ScaleType.Linear}
        yScaleType={ScaleType.Linear}
      />
```

**Expected behavior**
With the data above, I would expect the chart library to insert an extra vertex at the missing X coordinates.

nickofthyme:
## Summary

Add fit functions below to fill `null` `y1` values (`y0` `null` values not yet supported)

Works for **ONLY** non-stacked series. Stacked series requires another approach having to account for the fitted value of lower series.

Name | Description  | Examples |
--------|------------------|---------------
None | Don't draw that value on the graph | `[2, null null, 8]`
Carry | Use the last non null value before that | `[2, 2, 2, 8]`
Nearest | Use the closest value (either before or after) that was non null | `[2, 2, 8, 8]`
Lookahead | Use the next non null value after that (opposite of *Carry*) | `[2, 8, 8, 8]`
Average | Use the average of the last and next non null value | `[2, 5, 5, 8]`
Linear | Linear interpolate between closest values | `[2, 4, 6, 8]`
Zero | Fills all `null` values with `0` | `[2, 0, 0, 8]`
Explicit | Specify an explicit value (`x`), that should be used instead | `[2, x, x, 8]`

> Note: All "fitted" (aka "filled") points are not rendered but only used to draw the line/area.

Related to fix #388

## Usage
within a series spec definition, define the `fit` type.

```tsx
import { Fit } from '@elastic/charts';

<AreaSeries
  fit={Fit.Nearest}
/>
```

when using the explicit type you must provide a value like so...

```tsx
<AreaSeries
  fit={{
    type: Fit.Explicit,
    value: 5,
  }}
/>
```

By default, indeterminate endvalues, values not able to be fitted, are set to `null`. To set a specific value rather than `null`, you can pass an `endValue` prop like so...

```tsx
<AreaSeries
  fit={{
    type: Fit.Average,
    endValue: 4,
  }}
/>
```

## Demos `[Non-stacked]`

Assuming the following data
```ts
const data = [
  { x: 0, y: null },
  { x: 1, y: 3 },
  { x: 2, y: 5 },
  { x: 3, y: null },
  { x: 4, y: 4 },
  { x: 5, y: null },
  { x: 6, y: 5 },
  { x: 7, y: 6 },
  { x: 8, y: null },
  { x: 9, y: null },
  { x: 10, y: null },
  { x: 11, y: 12 },
  { x: 12, y: null },
];
```

### `Fit.None` `[default]`
image

### `Fit.Carry`
image

### `Fit.Nearest`
image

### `Fit.Lookahead`
image

### `Fit.Average`
image

### `Fit.Linear`
image

### `Fit.Zero`
image

### `Fit.Explicit` with value set to `4`
image

### Cured Linear

Also Works with `curves`
image

> Note: If the explicit value is set outside of the xScale domain, the charts will not recompute the domain to include it. This will have to be fixed in a later PR.

### Using `endValue`

The `endValue` property allows you to set the end `null` point values to a specific value. This value is only used in the case where the fitting function does not have sufficient information to determine the fitted value. For example, using `Fit.Average` requires both the trailing and leading non-null boundaries, thus if the endpoints are both null the value can not be found for either.

```tsx
<AreaSeries
  fit={{
    type: Fit.Average,
    endValue: 4,
  }}
/>
```

image

### Checklist


src/chart_types/xy_chart/utils/series.ts
```diff
@@ -13,23 +13,23 @@ export interface FilledValues {
...
+export interface RawDataSeriesDatum<T = any> {
   /** the x value */
   x: number | string;
   /** the main y metric */
   y1: number | null;
   /** the optional y0 metric, used for bars or area with a lower bound */
   y0?: number | null;
   /** the datum */
-  datum?: any;
+  datum?: T;
```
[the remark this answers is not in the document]
nickofthyme: I'm not sure of the full ramifications of this change so I left the `any` default. I just think it would be nice to have the datum types accessible when possible.

src/chart_types/xy_chart/utils/fit_function.ts
```diff
@@ -0,0 +1,205 @@
...
+          nextNonNullDatum = value as FullDataSeriesDatum;
+          break;
+        }
+      }
+    }
+
+    const newValue =
+      current.y1 === null ? getValue(current, previousNonNullDatum, nextNonNullDatum, type, endValue) : current;
+
+    // TODO and end condition check to fill end values that are still null
```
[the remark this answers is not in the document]
nickofthyme: Nope that's done I'll remove the comment.

src/chart_types/xy_chart/utils/fit_function.ts
```diff
@@ -0,0 +1,205 @@
...
+        };
+      }
+    }
+  }
+
+  if (endValue === undefined) {
+    return current;
+  }
+
+  // No mtching fit - should only fall here on end conditions
```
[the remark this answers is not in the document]
nickofthyme: 26244212d84fdec35b95e7eb969fe1f754c352ec..f7452b844dbeed2174f61732dedf4627129086d5

src/chart_types/xy_chart/utils/fit_function.ts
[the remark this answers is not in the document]
nickofthyme: 26244212d84fdec35b95e7eb969fe1f754c352ec..f7452b844dbeed2174f61732dedf4627129086d5

nickofthyme:
Still need to add support for ordinal scale type

jest.config.js
```diff
@@ -2,7 +2,9 @@ module.exports = {
   roots: ['<rootDir>/src'],
   preset: 'ts-jest',
   testEnvironment: 'jest-environment-jsdom-fourteen',
-  setupFilesAfterEnv: ['<rootDir>/scripts/setup_enzyme.ts'],
+  setupFilesAfterEnv: ['jest-extended', '<rootDir>/scripts/setup_enzyme.ts', '<rootDir>/scripts/custom_matchers.ts'],
+  coveragePathIgnorePatterns: ['<rootDir>/src/mocks/', '<rootDir>/node_modules/'],
```
nickofthyme: @markov00 I don't know why the coverage went down to 78% but it does not appear to be because of my code changes. Maybe this is causing some issues. Any thoughts?

markov00: I think the current coverage only cover the modules/files imported on each test. Adding the `coveragePathIgnorePatterns` maybe count also the missing modules without a test file importing them

CHANGELOG.md
```diff
@@ -42,981 +42,840 @@
 
 ## [13.5.7](https://github.com/elastic/elastic-charts/compare/v13.5.6...v13.5.7) (2019-10-23)
 
-
 ### Bug Fixes
 
-* switch to momentjs to handle timezones ([#436](https://github.com/elastic/elastic-charts/issues/436)) ([a9f98c8](https://github.com/elastic/elastic-charts/commit/a9f98c8))
+- switch to momentjs to handle timezones ([#436](https://github.com/elastic/elastic-charts/issues/436)) ([a9f98c8](https://github.com/elastic/elastic-charts/commit/a9f98c8))
```
markov00: could you please checkout the CHANGELOG file from master? this was wrongly changed by prettier

src/mocks/series/series.ts
```diff
@@ -0,0 +1,181 @@
+import { shuffle } from 'lodash';
```
markov00: what if we avoid completely `lodash` and add a utils function to shuffle an array like:
```
function shuffle(a) {
    for (let i = a.length - 1; i > 0; i--) {
        const j = Math.floor(Math.random() * (i + 1));
        [a[i], a[j]] = [a[j], a[i]];
    }
    return a;
}
```
ref: https://stackoverflow.com/questions/6274339/how-can-i-shuffle-an-array

src/chart_types/xy_chart/rendering/rendering.ts
```diff
@@ -53,6 +54,13 @@ export interface GeometryStyle {
 
 export type IndexedGeometry = PointGeometry | BarGeometry;
 
+/**
+ * Array of **range** clippings [x1, x2] to be excluded during rendering
+ *
+ * Note: Must be scaled **range** values NOT domain
+ */
+export type ClippedRanges = [number, number][];
```
markov00: Is `x1` and `x2` in pixel coordinates? if so can you update the jsdocs?

src/chart_types/xy_chart/rendering/rendering.ts
```diff
@@ -33,7 +34,7 @@ export const AccessorType = Object.freeze({
   Y1: 'y1' as 'y1',
 });
 
-export type AccessorType = typeof AccessorType.Y0 | typeof AccessorType.Y1;
+export type AccessorType = $Values<typeof AccessorType>;
```
markov00: ❤️

src/chart_types/xy_chart/utils/series.ts
```diff
@@ -40,17 +40,21 @@ export interface DataSeriesDatum {
...
 
 export interface DataSeries {
   specId: SpecId;
   key: any[];
   seriesColorKey: string;
   data: DataSeriesDatum[];
+  /**
+   * Ranges of `[x0, x1]` pairs to clip from series
+   */
+  clippedRanges?: DataSeriesDatum[];
```
markov00: I think this is a typo because the `clippedRanges` are defined on the geometry type

src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -54,6 +55,90 @@ interface DomainMinInterval {
...
+   * [2, null, null, 8] => [2, 2, 8, 8]
+   * ```
+   */
+  Nearest: 'nearest' as 'nearest',
+  /**
+   * Average between the closest non-`null` values
+   *
+   * Example:
+   * ```js
+   * [2, null, null, 8] => [2, 5, 5, 8]
```
markov00: If the average is between the closest non-null value, wouldn't the fitted array be 
`[2, 5, 6.5, 8] `?

src/mocks/series/series.ts
nickofthyme: I like to have the functionality of `lodash` in development only, especially in tests.

Having `lodash` prevents us from making mistakes in the function logic and that way we don't have to reinvent the wheel and it doesn't bloat the final bundle size.

nickofthyme: If you don't like having `lodash` I can remove it.

src/chart_types/xy_chart/utils/specs.ts
nickofthyme: No `Average` should fill all points in between two non-null values with the average. see https://github.com/elastic/kibana/issues/17717

What you are describing is the `Linear` fit type

src/chart_types/xy_chart/utils/series.ts
nickofthyme: 4d990b2

src/chart_types/xy_chart/rendering/rendering.ts
nickofthyme: 4d990b2

CHANGELOG.md
nickofthyme: 4d990b2

src/chart_types/xy_chart/utils/specs.ts
markov00: ok thanks

PR #426, opened 2019-10-20
chore: add command to show unused exports

monfera:
## Summary

Invoke [ts-prune](https://github.com/nadeesha/ts-prune) with `yarn ts:prune` to output unused exports in the console. The library name and its description is a misnomer as it doesn't remove anything but if there's a short, easy to remember thing instead of `ts:prune` it can change, name suggestions are welcome!

### Checklist

- ~~[ ] Any consumer-facing exports were added to `src/index.ts` (and stories only import from `../src` except for test data & storybook)~~
- [ ] ~~This was checked for cross-browser compatibility, [including a check against IE11](https://github.com/elastic/kibana/blob/master/CONTRIBUTING.md#cross-browser-compatibility)~~
- [ ] ~~Proper documentation or storybook story was added for features that require explanation or tutorials~~


markov00:
I can see the potential of this, but there are a lot of false negative to deal with:

- everything exported by the library, on our src/index.ts should be removed from the output
- if a function is exported and not used (except the one exported as part of the library from the tsconfig) it means that this function is not tested (jest requires an exported function to work correctly).
- it seems to report multiple time the same export, if it was first exported on it's own file, than from an index.ts

I think we have to wait for that until this https://github.com/nadeesha/ts-prune/issues/14 is solved first

monfera:
No rush on this, we can wait for the fix, though it seems to not cause harm and I like an explicit enlisting of useful if imperfect log commands in `package.json` as I'm bad remembering incantations esp. after a vacation 😃and Ctrl-R isn't ideal either. Such tools are never perfect (incl. tslint whose [lack of concern](https://github.com/pzavolinsky/ts-unused-exports/issues/19) about unused exports was the motivation) - btw. I might check out this library as well

> if a function is exported and not used (except the one exported as part of the library from the tsconfig) it means that this function is not tested (jest requires an exported function to work correctly)

My intent wasn't to automatically cull exports, just to see if there are exports that need _some_ action, such as either removing the exporting, removing the entire `export const/function ...` (if not used) or knowing where to add a test case, at least as a placeholder for real tests. So I saw it as a positive, despite its imperfections like:

> everything exported by the library, on our src/index.ts should be removed from the output

Good issue find in the `ts-prune` repo! I simply skimmed the output as these were typically top-level obvious things on the top. Wondering though, would it be possible to add the shallowest test case to cover the lib exports, even without exercising the API much? Maybe it's advantageous to have tests from the "outside" viewpoint which mimics how our library is used

nickofthyme:
retest

nickofthyme:
jenkins retest this please

nickofthyme:
jenkins retest this please

monfera:
Closing it in favor of Marco's PR [#461](https://github.com/elastic/elastic-charts/pull/461/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R41)

PR #429, opened 2019-10-21
fix: resize observer loop limit exception

markov00:
## Summary

In Kibana, we start getting error messages in TSVB describing that the ResizeObserver has reached the loop limit when used together with react-beautiful-dnd library (used by EUI). This fix the issue in Kibana updating the state in the next frame instead of the current one.

fix https://github.com/elastic/kibana/issues/48589

I don't have found the root cause of that. I've found that the error disappear in Kibana in two way:
- remove all DnD component from TSVB keeping the chart
- remove the ResizeObserver from Chart, keeping EUI DnD

references: 
https://github.com/WICG/ResizeObserver/issues/38

### Checklist

- ~[ ] Unit tests were updated or added to match the most common scenarios`

### DIFF
--- a/src/components/chart_resizer.tsx
+++ b/src/components/chart_resizer.tsx
@@ -11,25 +11,39 @@ class Resizer extends React.Component<ResizerProps> {
   private initialResizeComplete = false;
   private containerRef: RefObject<HTMLDivElement>;
   private ro: ResizeObserver;
+  private animationFrameID: number | null;
   private onResizeDebounced: (entries: ResizeObserverEntry[]) => void = () => {};
 
   constructor(props: ResizerProps) {
     super(props);
     this.containerRef = React.createRef();
     this.ro = new ResizeObserver(this.handleResize);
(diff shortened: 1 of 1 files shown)


src/components/chart_resizer.tsx
```diff
@@ -11,25 +11,39 @@ class Resizer extends React.Component<ResizerProps> {
...
-    this.ro.unobserve(this.containerRef.current as Element);
+    if (this.animationFrameID) {
+      window.cancelAnimationFrame(this.animationFrameID);
+    }
+    this.ro.disconnect();
   }
 
   onResize = (entries: ResizeObserverEntry[]) => {
-    entries.forEach(({ contentRect: { width, height } }) => {
+    if (!Array.isArray(entries)) {
```
nickofthyme: I think once we upgrade to TS 3.7 we should use `Optional Chaining` to do explicit checks for javascript consumers of elastic charts.

PR #430, opened 2019-10-21
fix(tickformatter): add timeZone to tickFormatter

### LINKED ISSUE #427: Not passing the timeZone on tickFormat
**Describe the bug**
When using a different time zone, the tickFormatter needs to use the same timezone to correctly format the ticks.
Right now we are not passing the timeZone to the tickFormat and our exposed `niceTimeFormatter` we are using the default timezone, causing some issues to everyone is using that formatter

**To Reproduce**
1. Create a chart with a different timezone
2. Use the `niceTimeFormatter`
3. You will notice that tick will not be correctly formatted

**Expected behavior**
The `niceTimeFormatter` should use the `timezone` parameter passed to the BarSeries.

**Additional context**
n/a niceTimeFormatter**Errors in browser console**
n/a

**Kibana Cross Issues**
n/a

**Checklist**
- ~[ ] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present~

markov00:
## Summary

to correctly handle a the tick formatting on time based series, we need to share the timeZone
parameter to the tickFormatter. When the tickFormatter is an external function, the user can
manually configure that parameter, but for the niceTimeFormatter we don't have that option already.
Instead of adding a breaking change I've added an optional parameter to the tickFormat props to
share the configured timeZone parameter

fix #427

### Checklist

### DIFF
--- a/src/chart_types/xy_chart/utils/specs.ts
+++ b/src/chart_types/xy_chart/utils/specs.ts
@@ -300,9 +300,9 @@ export interface AxisSpec {
   integersOnly?: boolean;
 }
 
-export type TickFormatterOptions = Partial<{
-  timeZone: string;
-}>;
+export type TickFormatterOptions = {
+  timeZone?: string;
+};
 export type TickFormatter = (value: any, options?: TickFormatterOptions) => string;
 
 export interface AxisStyle {
--- a/src/chart_types/xy_chart/utils/specs.ts
+++ b/src/chart_types/xy_chart/utils/specs.ts
@@ -300,7 +300,10 @@ export interface AxisSpec {
   integersOnly?: boolean;
 }
 
-export type TickFormatter = (value: any) => string;
+export type TickFormatterOptions = Partial<{
+  timeZone: string;
+}>;
+export type TickFormatter = (value: any, options?: TickFormatterOptions) => string;
 
 export interface AxisStyle {
   /** Specifies the amount of padding on the tick label bounding box */
--- a/src/utils/data/formatters.ts
+++ b/src/utils/data/formatters.ts
@@ -2,7 +2,7 @@ import { DateTime, Interval } from 'luxon';
 import { TickFormatter, TickFormatterOptions } from '../../chart_types/xy_chart/utils/specs';
 
 export function timeFormatter(format: string): TickFormatter {
-  return (value: number, options?: Partial<TickFormatterOptions>): string => {
+  return (value: number, options?: TickFormatterOptions): string => {
     const dateTimeOptions = options && options.timeZone ? { zone: options.timeZone } : undefined;
     return DateTime.fromMillis(value, dateTimeOptions).toFormat(format);
   };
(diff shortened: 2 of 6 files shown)


markov00:
I'm expecting some VRT errors because of the use of `local` as timezone

src/utils/data/formatters.ts
```diff
@@ -1,14 +1,14 @@
 import { DateTime, Interval } from 'luxon';
+import { TickFormatter, TickFormatterOptions } from '../../chart_types/xy_chart/utils/specs';
 
-type TimeFormatter = (value: number) => string;
-
-export function timeFormatter(format: string): TimeFormatter {
-  return (value: number): string => {
-    return DateTime.fromMillis(value).toFormat(format);
+export function timeFormatter(format: string): TickFormatter {
+  return (value: number, options?: Partial<TickFormatterOptions>): string => {
```
nickofthyme: This is already a `Partial` type

https://github.com/elastic/elastic-charts/blob/8015f5eef821c9daed4a60aa8212dea5d5130170/src/chart_types/xy_chart/utils/specs.ts#L303-L305

src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -300,7 +300,10 @@ export interface AxisSpec {
   integersOnly?: boolean;
 }
 
-export type TickFormatter = (value: any) => string;
+export type TickFormatterOptions = Partial<{
+  timeZone: string;
+}>;
```
nickofthyme: I think type definitions should set type as is. The advanced types like `Partial`, `Omit`, `Pick` are used with defined types to suit a given need, say I only need one property on the type so I `Pick` that property.

```ts
export interface TickFormatterOptions {
  timeZone?: string;
};
```

nickofthyme:
@rshen91 with the addition of `moment`, it appears to fail gracefully by falling back to the local/browser time zone.

image

markov00:
> @rshen91 with the addition of `moment`, it appears to fail gracefully by falling back to the local/browser time zone.
> 
> image

@nickofthyme @rshen91 do we want to catch that moment exception and throw just a more "nicer" console.warn ?

src/chart_types/xy_chart/utils/specs.ts
markov00: solved in https://github.com/elastic/elastic-charts/pull/430/commits/7cca9e814d3ebbe8362342b51936e177cc45a342

nickofthyme:
> @nickofthyme @rshen91 do we want to catch that moment exception and throw just a more "nicer" console.warn ?

No I don't think that is needed. Wylie pointed out our lack of error catching, which he's right. I think it's a bigger issue to handle all bad inputs to chart "nicely" in both dev and production.

One thing we could add for this is an enum but that would limit the flexibility of kibana to just pass a string. I think `moment.tz` is resilient enough to handle most edge cases 💪 .

src/utils/data/formatters.ts
markov00: solved in https://github.com/elastic/elastic-charts/pull/430/commits/7cca9e814d3ebbe8362342b51936e177cc45a342

PR #431, opened 2019-10-21
ci: update ci to enforce yarn.lock sync

nickofthyme:
## Summary

Enforce `yarn.lock` sync

### DIFF
--- a/.ci/run.sh
+++ b/.ci/run.sh
@@ -95,7 +95,7 @@ yarn install --frozen-lockfile
 ### install codecov dependencies
 ###
 echo " -- installing codecov dependencies"
-yarn add codecov --prefer-offline
+yarn add codecov --prefer-offline --frozen-lockfile
 
 ###
 ### building
--- a/.ci/run.sh
+++ b/.ci/run.sh
@@ -85,19 +85,17 @@ yarn config set yarn-offline-mirror "$cacheDir/yarn-offline-cache"
 yarnGlobalDir="$(yarn global bin)"
 export PATH="$PATH:$yarnGlobalDir"
 
-
+###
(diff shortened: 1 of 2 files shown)


.ci/run.sh
```diff
@@ -85,20 +85,18 @@ yarn config set yarn-offline-mirror "$cacheDir/yarn-offline-cache"
...
+### install dependencies
+###
+echo " -- installing dependencies"
+yarn install --frozen-lockfile
 
 ###
 ### install codecov dependencies
 ###
 echo " -- installing codecov dependencies"
 yarn add codecov --prefer-offline
```
markov00: I think you have to add the `--frozen-lockfile` also to this line, if not you will see the "update lock file" also here

package.json
```diff
@@ -140,7 +140,7 @@
     "mobx-react": "^5.4.3",
     "newtype-ts": "^0.2.4",
     "prop-types": "^15.7.2",
-    "react": "^16.8.3",
+    "react": "^16.10.0",
```
markov00: Why we are upgrading to react 16.10 in this PR?
I think we should check the `konva` implementation first because it usually needs to be aligned with the react version

nickofthyme: To verify it breaks the ci when `package.json` in out of sync with `yarn.lock`

Look like it works now!

```bash
yarn install v1.19.1
02:59:01 [1/5] Validating package.json...
02:59:01 [2/5] Resolving packages...
02:59:02 error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
02:59:02 info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

PR #447, opened 2019-10-31
fix: add clip paths to highlighter component

markov00:
## Summary

The highlighter component doesn't properly handle custom Y domains, displaying the highlighted element outside the chart.

This PR fix it adding a clip area.

**Before**:
Screenshot 2019-10-31 at 17 42 39

**After**:
Screenshot 2019-10-31 at 17 43 04

### Checklist


nickofthyme:
It looks like this causes a small regression with the clipping changes you made a few months ago related to the point on the end.

Looks like it is only clipping the hovered point styles and not the points themselves.

## Current
image


## Your changes

image

markov00:
> It looks like this causes a small regression with the clipping changes you made a few months ago related to the point on the end.
> 
> Looks like it is only clipping the hovered point styles and not the points themselves.
> 
> ## Current
> image
> 
> ## Your changes
> image

@nickofthyme fixed in 27f6ad2

Issue #293, opened 2019-08-06
Provide some color and spectrum creating functions


cchaos:
**Is your feature request related to a problem? Please describe.**
Right now, when you provide an array of colors, the chart simply iterates consecutively through the array, stopping when it's reached the last series. This works great for categorical style charts, but not for quantitative or tend-like charts.

**Describe the solution you'd like**
Provide some color functions that allow consumers to pass 1+ colors, and the number of stops, then returns an array of colors interpolated based on those parameters.

Something like:

```js
const function createSpectrum(colors: [], numSeries: number) {
  if (colors.length === 1) {
    // Create a spectrum using the color[0] as the middle (or near middle)
    'https://d.pr/free/i/o9J63Q'
  } else if (colors.length === 2) {
    // Create a spectrum using the color[0] as the first and color[1] as the last
    'https://d.pr/free/i/4kD8tq'
  } else {
    // Create a spectrum equally spacing out the array of colors
    'https://d.pr/free/i/kQRypX'
  }
}
```

You might even want to consider using [Chroma.js](https://github.com/gka/chroma.js) as is used by the [tool in the screenshots above](https://github.com/gka/palettes).

**Describe alternatives you've considered**
Providing this in EUI instead. But it makes more sense for this library to provide it.

This was my hacky way of creating these functions:

```javascript
import { colorPalette } from '@elastic/eui/src/services';

switch (this.state.colorType) {
  case 'Trend':
    // Three color
    firstColor = 'green';
    lastColor = 'red'
    middleColor = 'gray';

    const half = Math.round(numSeries / 2);

    if (half < 2) {
      vizColors = [firstColor, lastColor];
      break;
    } else {
      let firstHalf = colorPalette(firstColor, middleColor, half);
      let lastHalf = colorPalette(middleColor, lastColor, half);

      if (numSeries % 2) {
        // Number is odd
        const removeFirstColor = lastHalf.shift();
      } else {
        firstHalf = colorPalette(firstColor, middleColor, half + 1);
        lastHalf = colorPalette(middleColor, lastColor, half + 1);
        const removeFirstColor = lastHalf.shift();
        const removeLastColor = firstHalf.pop();
      }
      vizColors = [...firstHalf, ...lastHalf];
      break;
    }

  case 'Quantity':
    // Single color
    firstColor = 'white';
    lastColor = 'green';
    vizColors = colorPalette(
      firstColor,
      lastColor,
      numSeries + 1
    );
[...]

PR #448, opened 2019-11-01
feat: implementing color palette

majagrubic:
## Summary

This is a very WIP implementation of a color palette, but thought it would be easier to start a conversation with a draft PR. Implementation taken from [EUI](https://github.com/elastic/eui/blob/master/src/services/color/color_palette.ts).

When calculated, it looks like this:
Screenshot 2019-11-01 at 19 17 52

Some open questions:
1. What if both colorPalette and custom colors are specified? Does custom color still apply?
Screenshot 2019-11-01 at 19 44 54

2. What if the number of steps is less than number of series? Do we need to safeguard this and always generate at least as many colors as there are series?

3. Are there cases where we don't want to be able to include a color palette? (line chart?)

fix https://github.com/elastic/elastic-charts/issues/293

### Checklist


src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,162 @@
...
+    this.text = createHex(this.collection);
+  }
+}
+
+interface Palette {
+  steps: number;
+  calcPalette(): string[];
+}
+
+export class SequentialPalette implements Palette {
```
markov00: this is my personal preference, so be free to just ignore it:
what if we reimplement this as a simple function instead of a class? This will allow us to:
- remove the Color class, as it's used only internally
- we will have a simple function to call with the colorPalette and steps count instead of creating a new palette object and than call a function on it (I see the calcPalette more as a static function)

I've also tried to clean/optimize it up locally if you want to take a look:
```
function computeSequentialPalette(colors: HexColor[], steps = DEFAULT_NUM_OF_STEPS): HexColor[] {
  if (colors.length !== 2) {
    throw new Error('Valid number of colors for sequential palette is 1 or 2');
  }
  if (steps && steps <= 1) {
    throw new Error('Miminum number of steps is 2');
  }
  const startColor = colors[0];
  const endColor = colors.length === 2 ? colors[1] : startColor;
  if (!isHex(startColor) || !isHex(endColor)) {
    throw new Error('Please provide two valid hex color codes.');
  }
  const count = steps - 1;
  const hexPalette: string[] = [];
  const startHex = hexToRgb(startColor); // get RGB equivalent values as array
  const endHex = hexToRgb(endColor); // get RGB equivalent values as array
  const step = stepCalc(count, startHex, endHex); // create array of step increments
  // build the color palette array
  hexPalette.push(createHex(startHex)); // set the first color in the array
  for (let i = 1; i < count; i++) {
    // set the intermediate colors in the array
    const r = startHex[0] + step[0] * i;
    const g = startHex[1] + step[1] * i;
    const b = startHex[2] + step[2] * i;
    hexPalette[i] = createHex([r, g, b]);
  } // all the colors in between
  hexPalette[count] = createHex(endHex); // set the last color in the array

  return hexPalette;
}
```

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,162 @@
...
+      colorArray[i] = new Color(r, g, b);
+      hexPalette[i] = colorArray[i].text;
+    } // all the colors in between
+    hexPalette[count] = colorArray[count].text; // set the last color in the array
+
+    return hexPalette;
+  }
+}
+
+export class DivergingPalette implements Palette {
```
markov00: because Diverging and Qualitative are not available at the moment I will prefer to remove this mock implementation and directly return something like
```
export function calcColorPalette(colorPalette: ColorPalette) {
  const { colors, steps, type } = colorPalette;
  if (type === 'sequential') {
    return new SequentialPalette(colors, steps).calcPalette();
  }
  return ['red'] // or something that clearly shows that it's something not implemented yet
}
```

src/utils/colors/hex_to_rgb.ts
```diff
@@ -0,0 +1,31 @@
...
+  s: number;
+  v: number;
+}
+export interface RGB {
+  r: number;
+  g: number;
+  b: number;
+}
+
+export type HEX = string;
```
markov00: Would you mind to call this `Hex` or `HexColor` to be a bit more aligned with the current case

stories/bar_chart.tsx
```diff
@@ -635,7 +635,14 @@ storiesOf('Bar Chart', module)
     const formatter = timeFormatter(niceTimeFormatByDay(1));
     return (
       <Chart className={'story-chart'}>
-        <Settings debug={boolean('debug', false)} />
+        <Settings
```
markov00: Would you mind create a new story under `styling` with the `colorPalette`? Maybe adding some knobs to control the colors and/or the number of steps

src/utils/colors/color_palette.ts
majagrubic: Hm, not sure I follow. Are we talking about `Color` class defined here:
https://github.com/elastic/elastic-charts/blob/071f7a00592d22e1064724c8b6233bc0135c0fff/src/utils/colors/color_palette.ts#L8

or the `Palette` interface?
https://github.com/elastic/elastic-charts/blob/071f7a00592d22e1064724c8b6233bc0135c0fff/src/utils/colors/color_palette.ts#L18

I assume you meant Palette?

markov00: We can remove both Palette and Color classes

markov00: @majagrubic I think we can directly remove either Color and Palette/SequentialPalette classes and prefer an utility function over a class.
Having a Class with a single static method doesn't provide us much benefit for us, so I'd rather prefer to go with a single utility function

src/index.ts
```diff
@@ -37,3 +37,4 @@ export {
   RenderChangeListener,
   BasicListener,
 } from './chart_types/xy_chart/store/chart_state';
+export { SequentialPalette } from './utils/colors/color_palette';
```
markov00: I think we should also expose the `HexColor` as it's used as type on the calcPalette function

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,109 @@
...
+    const colorArray: Color[] = [];
+    const hexPalette: string[] = [];
+    const startHex = hexToRgb(startColor); // get RGB equivalent values as array
+    const endHex = hexToRgb(endColor); // get RGB equivalent values as array
+    colorArray[0] = new Color(startHex[0], startHex[1], startHex[2]); // create first color obj
+    colorArray[count] = new Color(endHex[0], endHex[1], endHex[2]); // create last color obj
+    const step = stepCalc(count, colorArray[0], colorArray[count]); // create array of step increments
+    // build the color palette array
+    hexPalette.push(colorArray[0].text); // set the first color in the array
+    for (let i = 1; i < count; i++) {
```
monfera: Hi @majagrubic, Marco asked I also look at this PR. The current loop determines intermediate colors based on a linear interpolation per the separate r, g, b channels between the two color endpoints, which is not quite suitable for data visualization. Linear interpolation will not yield a perceptually uniform distribution of the color intervals (distances), and RGB makes it worse than interpolating in another color space such as CIELab.  

Maybe it's worth a discussion if we want, or need to interpolate between two arbitrary colors. The reason is that such algorithmic interpolation will yield color ramps that'll have weaknesses in any or all of the following:

1. accessibility - if the start/end colors are of a similar luminance, someone with monochromatic vision won't see a difference (extreme case); people with partial color vision will also likely have problems, in part because of the arbitrary start/end points, in part because of the interpolation, esp. in the RGB space
2. perceptual uniformity - it's a desirable property of a color palette (for dataviz) to have even steps between adjacent colors, as perceived by people, even with full color sight, and [CIELab](https://howaboutanorange.com/blog/2011/08/10/color_interpolation/) aspires to be reasonably uniform perceptually, but still won't be as good as a color palette designed for dataviz, with carefully tweaked perceptual uniformity
3. aesthetics: the algorithmic interpolation between two points in the color space will be aesthetically arbitrary, while color palettes designed for data visualization, to the extent other constraints allow, incorporate the design goal of being aesthetic

Having said this, interpolating between two arbitrary colors can be artistic, and I think it'll be eventually useful for very freeform visualizations eg. in Canvas theming (ie. I think `elastic-charts` should, sooner or later, support color interpolation, but in CIELab) and for now, a key goal of Kibana visualizations in general is adherence to data visualization practice.

@markov00 and @VijayDoshi would probably have more info about what we want to do in the color realm for continuous, sequential, divergent and categorical palettes, there are many high quality libraries out there with good dataviz color palettes and scales, for example, [d3-color](https://github.com/d3/d3-color) has CIELab functions, and [d3-scale-chromatic](https://github.com/d3/d3-scale-chromatic) has a ton of palettes

On a more technical level, I think it's good to roll our own functions if there's a benefit such as much more compact and/or more maintainable code, higher performance, or custom requirements, but color calculations can be pretty tricky so I wouldn't be against either depending on pre-made library functions when it comes to color, or at least using conventions, data formats and APIs close to these libraries out there, so if needed, we can switch to, or add color palettes from external sources.

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,109 @@
...
+  collection: RGBArray;
+  text: string;
+
+  constructor(public r: number, public g: number, public b: number) {
+    this.collection = [r, g, b];
+    this.text = createHex(this.collection);
+  }
+}
+
+export class SequentialPalette {
```
monfera: I don't currently see a need for classes for colors or palettes, they could be represented as inert objects

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,109 @@
...
+  static calcPalette(colors: HexColor[], steps = DEFAULT_NUM_OF_STEPS): string[] {
+    if (colors.length !== 2) {
+      throw new Error('Valid number of colors for sequential palette is 1 or 2');
+    }
+    if (steps && steps <= 1) {
+      throw new Error('Miminum number of steps is 2');
+    }
+    const startColor = colors[0];
+    const endColor = colors.length === 2 ? colors[1] : startColor;
```
monfera: `colors.length` is always 2 here, otherwise the function throws in line 21

src/utils/colors/color_palette.ts
majagrubic: As I mentioned earlier in the PR, this code is copy/paste from @cchaos 's implementation in the EUI library: 
https://github.com/elastic/eui/blob/master/src/services/color/color_palette.ts#L27

My understanding was that we want to move this functionality to Elastic Charts, so that EUI can consume it, instead of doing its own thing. I agree with the all points mentioned above, the only question is - do we want to address this now or leave as it is and then figure out a proper algorithm for calculating the appropriate palette?

monfera: I looked at the notes but I'm not clear about why we want to move something that is in EUI, into Elactic Charts, so that EUI can consume it (I suppose, the linear-RGB color palettes rather than the entire `elastic-chart`). As I mentioned, for `elastic-charts` purposes it feels better to rely on pre-made color palettes and maybe CIELab interpolation, via eg. depending on color libraries, so if EUI needs the linear-RGB, I don't currently see a need for it to move into `elastic-charts` as it just adds code and maintenance burden without a ton of benefits. Maybe a call with you and @markov00 ?

monfera: [Linking](https://github.com/elastic/elastic-charts/pull/448) the issue Maya mentioned. The copied RGB interpolation is indeed what it is in EUI now, and Caroline's `chroma.js` based version, using CIELab ❤️ lives only as an issue [note](https://github.com/elastic/elastic-charts/issues/293#issuecomment-518854073).

cchaos: Trying to add some clarification which I may have inadvertently caused some confusion by linking to the wrong function within EUI.

EUI does provide a color palette service for consumers to use. This is the one that lives at 
https://github.com/elastic/eui/blob/master/src/services/color/color_palette.ts#L27

However, the one that was pasted into #293 is what lives in EUI docs only, which means that neither EUI nor consumers of EUI can consume. https://github.com/elastic/eui/blob/389076383d1e4ece1cf8f6329d92fcb0f263f545/src-docs/src/views/elastic_charts/shared.js#L183-L246

This function above is what should be move to Elastic-charts to create the correct color palettes for charts, not the color-palette service or RGB converter.

nickofthyme:
I think this issue will be handled now in kibana via the new palette service in the charts plugin. If not we can revisit these changes in the future, but this PR is too out of date to continue.

cc: @markov00

PR #451, opened 2019-11-14
feat: add PNG export

### LINKED ISSUE #82: Export as PNG function
I want to be able to save the chart as PNG image. This can be used to simplify client side export in Kibana.

Add the possibility to export the chart as a PNG via a function call to the chart.
Konva has the possibility to get the DataURL from the `Stage` object (see https://konvajs.org/docs/data_and_serialization/Stage_Data_URL.html)
```js
 const dataURL = stage.toDataURL();
```
This, btw, will only save the chart into a PNG, not the whole chart with the legend.
A possibility for that is illustrated on https://github.com/tsayen/dom-to-image#how-it-works
but is basically the following:
- clone the DOM element of the legend
- inline all the styles
- create an SVG with a `foreignObject` with the cloned DOM element
- save the SVG as image from js
- copy and paste the image of the legend and the chart into a new image.

The legend can be rendered in canvas, but it's quite difficult and we need to move the whole action logic to a canvas element.
Another option is to render subsequentially a static legend on the chart canvas with only colors and texts of the existing legend.

**Additional context**
n/a

**Kibana Cross Issues**

**Checklist**

rshen91:
## Summary

Thank you Marco for the help on this! This PR will close #82 and allows PNG export of charts:

aintnopartylikeapngparty

A PNG export story was added under the heading `Interactions`. 
Additionally, in the playground.tsx from this PR the following implements the PNG export functionality: 
```js
...
  chartRef: React.RefObject<Chart> = React.createRef();
  onSnapshot = () => {
    if (!this.chartRef.current) {
      return;
    }
    const snapshot = this.chartRef.current.getPNGSnapshot({
// can specify a background color or default is transparent
      backgroundColor: 'white',
      pixelRatio: 2,
    });
    if (!snapshot) {
      return;
    }
    const fileName = 'chart.png';
    switch (snapshot.browser) {
      case 'IE11':
        return navigator.msSaveBlob(snapshot.blobOrDataUrl, fileName);
      default:
        const link = document.createElement('a');
        link.download = fileName;
        link.href = snapshot.blobOrDataUrl;
        document.body.appendChild(link);
        link.click();
        document.body.removeChild(link);
    }
  };
...
```
---
## Next steps:
- add the legend from the chart to the PNG (html2canvas library)

### Checklist


src/components/chart.tsx
```diff
@@ -91,6 +95,56 @@ export class Chart extends React.Component<ChartProps, ChartState> {
       }
     }
   }
+
+  getPNGSnapshot(
+    options = {
+      backgroundColor: 'transparent',
+      pixelRatio: 1,
```
markov00: let's put 2 as default to have it a bit more "retina" friendly

package.json
```diff
@@ -107,6 +107,7 @@
     "eslint-plugin-jest": "^22.6.4",
     "eslint-plugin-prettier": "^3.1.0",
     "eslint-plugin-react": "^7.13.0",
+    "html2canvas": "^1.0.0-rc.5",
```
markov00: Since we are not using this at the moment, I think we can remove it

.playground/playgroud.tsx
```diff
@@ -1,62 +1,67 @@
...
+} from '../src';
+import { KIBANA_METRICS } from '../src/utils/data_samples/test_dataset_kibana';
 export class Playground extends React.Component {
+  chartRef: React.RefObject<Chart> = React.createRef();
+  onSnapshot = () => {
+    if (!this.chartRef.current) {
+      return;
+    }
+    const snapshot = this.chartRef.current.getPNGSnapshot({
+      backgroundColor: 'red',
```
markov00: I know that this is the playground, but maybe change this to `white` so it's a bit more pleasant

.playground/playgroud.tsx
```diff
@@ -1,62 +1,67 @@
...
+    switch (snapshot.browser) {
+      case 'IE11':
+        return navigator.msSaveBlob(snapshot.blobOrDataUrl, fileName);
+      default:
+        const link = document.createElement('a');
+        link.download = fileName;
+        link.href = snapshot.blobOrDataUrl;
+        document.body.appendChild(link);
+        link.click();
+        document.body.removeChild(link);
```
markov00: Can you create a storybook story that allows you to snapshot the chart from a button (a storybook knob button maybe) with some description on how to achieve that?

rshen91: Good idea! I added a story in interactions in 1931d2c

.playground/playgroud.tsx
rshen91: Thanks! Good call - addressed in commit 1931d2c

package.json
rshen91: Thanks! Good point - removed it in commit 1931d2c

src/components/chart.tsx
rshen91: Thanks and makes sense I love my retinas - commit 1931d2c addresses this

stories/interactions.tsx
```diff
@@ -635,4 +635,61 @@ storiesOf('Interactions', module)
     {
       info: 'Sends an event every time the cursor changes. This is provided to sync cursors between multiple charts.',
     },
-  );
+  )
+  .add('PNG export action', () => {
```
markov00: Can you add a description on how to get the snapshot and is a default implementation?
Check storybook API/docs to find a way to add info to the story.
I think we should start add some text on every story we add

rshen91: Good call - I've added an info section to the story in fd354c6. I'll open up an issue to track adding a Docs tab such as in https://www.learnstorybook.com/design-systems-for-developers/react/en/document/ where we can configure and clarify each story 👍

rshen91: Opened issue #457

Issue #369, opened 2019-09-06
X axis should be on 0 as datum


nickofthyme:
**Is your feature request related to a problem? Please describe.**
I would like to make zero the point of reference for any axes if there are positive and negative values.

<img width="796" alt="Image 2019-09-06 at 10 37 34 AM" src="https://user-images.githubusercontent.com/19007109/64440916-806de880-d092-11e9-9042-5445c42ba1e0.png">

see https://elastic.github.io/elastic-charts/?path=/story/bar-chart--scale-to-extent

**Describe the solution you'd like**
The x and or y-axis lines should be on zero as datum.

**Kibana Cross Issues**
Add any [Kibana](https://github.com/elastic/kibana) related issues/feature requests here.

**Checklist**

- [ ] this request is checked against already exist requests
- [ ] every related Kibana issue is listed under `Kibana Cross Issues` list
- [ ] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

PR #456, opened 2019-11-22
feat: add ZeroAxis component 

mbondyra:
## Summary

Closes https://github.com/elastic/elastic-charts/issues/369

This PR adds a new component, ZeroAxis. It is displayed for y=0 through <Settings/> prop `hideZeroAxis`. By default `hideZeroAxis===false`, so the zero axis is shown.

image

I added a story here: https://github.com/elastic/elastic-charts/pull/456/files#diff-053316cd178d68a83ed121a7e1c89e22R557 for easy testing of the component.

Some notes:
1. I only implemented it for y values as I find it weird in vertical (see below, NOT IMPLEMENTED):
image

2. I only implemented it for linear scaleType as the others `y` scales only take values > 0. (Log, Ordinal, Sqrt, Time)

3. I'd appreciate if you let me know what you think about the design for multiple domains, I am not sure if it's understandable enough. See below: 

image

### Checklist

-  ~~Any consumer-facing exports were added to `src/index.ts` (and stories only import from `../src` except for test data & storybook)~~


src/chart_types/xy_chart/utils/zero_axis_utils.ts
```diff
@@ -0,0 +1,33 @@
...
+  const top = chartDimensions.height * zeroPositionRatio;
+  return {
+    left: 0,
+    top,
+    width: chartDimensions.width,
+    height: 0,
+  };
+};
+
+export const computeZeroAxes = (chartDimensions: Dimensions, yDomains: YDomain[]): ZeroAxisSpec[] => {
```
markov00: you can also use the `yScales` instance variable in `chart_state.ts`: each scales already has a function to scale the 0 value to the correct height/width.

stories/axis.tsx
```diff
@@ -553,4 +553,43 @@ storiesOf('Axis', module)
         />
       </Chart>
     );
+  })
+  .add('with zero axes', () => {
```
markov00: Could you add a small description for this story?

mbondyra:
Thank you for your comments @markov00, they are all very valuable! Please recheck once you find a moment 😊

mbondyra:
Currently on hold due to awaiting for some design feedback.

cchaos:
I did a bit a looking around the internet for any guidance and/or examples of how to properly handle these examples, and what it all boils down to is "it depends on the user's data". Just because the `0` line shifts doesn't mean that the `0` is their main baseline.

For example, something [like this image](https://share.getcloudapp.com/QwuGWLDr) shows that their main baseline is actually `100` even though their spread goes from `-20` to `260`.  

Therefore I don't think we need to/should do any automatic styling or adding of this 0 line if it's not the min or max. If the user needs to denote this line they can/should use an annotation which then they can color any way they need.

When it comes to multiple axes, having multiple `0` lines can cause more confusion than just leaving them off. The more difficult part to multiple axes is how to setup the grid. In EUI, we advocate for using the grid to provide helpful markings for scale.

Screen Shot 2019-12-02 at 15 53 43 PM

mbondyra:
Thank you for your feedback @cchaos! @nickofthyme @markov00 I guess we can close this PR then, right?

markov00:
Thanks @cchaos, I also have some doubts about the baseline in some use cases and as always, there are opposite opinions on where to put that baseline (specially for line chart more than on bar charts).
The idea is that bar graphs should always start at 0 because the height of the bar signifies the value of the condition being represented (see https://flowingdata.com/2015/08/31/bar-chart-baselines-start-at-zero/  http://www.storytellingwithdata.com/blog/2012/09/bar-charts-must-have-zero-baseline https://chartio.com/learn/charts/bar-chart-complete-guide/). If you have a different baseline (as in your example) I think we should provide, instead of a fixed 0 baseline, a configurable baseline, that is, by default 0.
Another important thing to consider is that the baseline should be shown only if we are displaying bar charts. For area chart I've mixed feelings, but for lines the baseline can be completely omitted.

Here is an interesting paper about the y-axis range (that doesn't directly mean basline :D) but it's an interesting reading https://open.lnu.se/index.php/metapsychology/article/view/895/2084

mbondyra:
Time to close this one 🔐

Issue #457, opened 2019-11-22
Add Docs to Storybook 


rshen91:
**Is your feature request related to a problem? Please describe.**
This came up a bit in #451 and I'm opening this issue for tracking and  to personally follow-up. Storybook provides a [Docs](https://www.learnstorybook.com/design-systems-for-developers/react/en/document/) add-on which we might want to include to provide a docs section per story. 

It seems like some stories are using the info parameter currently, which is good, but a full blown docs section per story would be clearer for users (hopefully). 

**Describe the solution you'd like**
Mentioned above, there are some resources to get this started: 
https://www.learnstorybook.com/design-systems-for-developers/react/en/document/
https://medium.com/storybookjs/storybook-docspage-e185bc3622bf

**Describe alternatives you've considered**
It looks like there might be a similar issue opened in #379 but with a different approach. I'll keep @nickofthyme in the loop.

**Kibana Cross Issues**
I'm not sure this directly impacts Kibana but would be generally helpful for any consumer. 

**Checklist**

- [x] this request is checked against already exist requests
- [ ] every related Kibana issue is listed under `Kibana Cross Issues` list
- [ ] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

rshen91:
We are on an older version of Storybook 5.1.9 and docs are available on 5.2. Exploring options...

markov00:
@rshen91 we are on a fixed 5.1.9 because there was some small typescript issues switching to 5.2.
Since we are now merged the redux PR I think we can easily switch to 5.2 fixing those issues

nickofthyme:
Some of these may not work at all. Like they used to allow for `object[]` types and now only allow certain types in an array.

Also, the VRT relies on the query params to interpolate the correct value. Like a number will come out as a string not a number.

nickofthyme:
Also #379 is to create a separate docs page apart from storybook.

markov00:
> Some of these may not work at all. Like they used to allow for `object[]` types and now only allow certain types in an array.
> 
> Also, the VRT relies on the query params to interpolate the correct value. Like a number will come out as a string not a number.

I've did a preliminary check few  days ago and seems that there are only few `arrays` to parse back to number or string before using them. There was few differences on adding the info-addon.
All the rest seems to work quite straight forward.

PR #459, opened 2019-11-25
feat(color palette): implementing different color palettes with d3

majagrubic:
## Summary

Implements color palette calcualtions using d3 libraries.
There are 3 different palettes available at the moment:
1. categorical - includes [d3 categorical schemes](https://github.com/d3/d3-scale-chromatic#categorical) and EUI [palettes](https://elastic.github.io/eui/#/utilities/color-palettes)
2. diverging - includes calculating a diverging palette using one of the d3 [interpolators](https://github.com/d3/d3-scale-chromatic#diverging)
3. sequential - includes d3 sequential [single-hue](https://github.com/d3/d3-scale-chromatic#sequential-single-hue) & [multi-hue](https://github.com/d3/d3-scale-chromatic#sequential-multi-hue) palettes. Also, includes a possibility to calculate a custom sequential palette given two colors (starting and ending).
4. cyclical - https://github.com/d3/d3-scale-chromatic#cyclical

This PR does not include an API to be used from the chart itself. This will be added in a subsequent PR.

### Checklist


src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,90 @@
+import * as d3ScaleChromatic from 'd3-scale-chromatic';
+import * as d3Color from 'd3-color';
```
markov00: you can also import only `{rbg, lab}`

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export function getCustomSequentialPalette(colors: string[], steps: number) {
+  // @ts-ignore
+  const startingColor = d3Color.lab(colors[0]).brighter();
+  // @ts-ignore
+  const endingColor = d3Color.lab(colors[1]).darker();
+  const paletteColors = [];
+  // @ts-ignore
+  const scale = scaleLinear()
```
markov00: I think you can remove all of the `@ts-ignore` in this function if you declare add the generics to scale like
`scaleLinear<d3Color.RGBColor, d3Color.RGBColor>()`

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,90 @@
...
+    // @ts-ignore
+    paletteColors.push(d3Color.rgb(scale(i)).hex());
+  }
+  return paletteColors;
+}
+
+export function getCyclicalPalette(steps: number) {
+  const paletteColors = [];
+  for (let i = 0; i < steps; i++) {
+    paletteColors.push(d3Color.rgb(d3ScaleChromatic.interpolateRainbow(i / steps)).hex());
```
markov00: we should safe guard against division by zero and infinite loop with steps < 0

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,90 @@
...
+
+function transformPaletteName(colorPaletteName: ColorPaletteName): string {
+  const colorPaletteNameCapitalized = colorPaletteName.charAt(0).toUpperCase() + colorPaletteName.slice(1);
+  return `interpolate${colorPaletteNameCapitalized}`;
+}
+
+export function getCategoricalPalette(name: CategoricalSchemeName): ReadonlyArray<string> {
+  const schemeName: string = transformSchemeName(name);
+  // @ts-ignore
+  return d3ScaleChromatic[schemeName];
```
markov00: what if we import each scheme separately and we use a switch here to avoid adding `ts-ignore` various places
```
import { schemeCategory10} from  'd3-scale-chromatic';
switch(schemeName){
 case 'schemeCategory10':
   return schemeCategory10;
  default:
  .....
}
```

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,90 @@
+import * as d3ScaleChromatic from 'd3-scale-chromatic';
+import * as d3Color from 'd3-color';
+import { interpolateHcl, quantize, interpolateLab } from 'd3-interpolate';
+import { scaleSequential, scaleLinear } from 'd3-scale';
+
+export type ColorPaletteName = 'blues' | 'greens' | 'greys' | 'oranges' | 'purples' | 'reds';
```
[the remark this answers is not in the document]
majagrubic: Do we really need this or is nice to have?

src/utils/colors/color_palette.ts
majagrubic: That is a lot of hardcoding for something that works fine without it. I am not sure it's worth the benefit.

src/utils/colors/color_palette.ts
majagrubic: I think prefixing the function with `d3-` equivalents improves readability of the code, but if there are performance concerns around importing the whole namespace, I'm happy to do it.

src/utils/colors/color_palette.ts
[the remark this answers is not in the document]
majagrubic: I don't know how are we going to explain `PuBu` and `PuBuGn` color palette names 😂

src/utils/colors/color_palette.ts
markov00: It's for sure hardcoding, but what if scaleChromatic removes one of their scheme names? we will never catch that issue until someone call the function with that removed scale. 
The same happens for the opposite: we add a new Scheme name but it's not exported or available on the scale chromatic library (see for example the typo in the previous comment, one value of the ColorPaletteName was misspelled, with a proper typing we can catch these error much more easily).
Maybe we can find some some way to handle that getting the available keys from d3ScaleChromatic with a type like `keyof typeof d3ScaleChromatic` but I'm not sure how to apply that to the current implementation

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,157 @@
...
+  | 'BuPu'
+  | 'GnBu'
+  | 'OrRd'
+  | 'PuBuGn'
+  | 'PuBu'
+  | 'PuRd'
+  | 'RdPu'
+  | 'YlGnBu'
+  | 'YlOrBr'
+  | 'YlOrRd[';
```
markov00: typo

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,157 @@
+import * as d3ScaleChromatic from 'd3-scale-chromatic';
+import * as d3Color from 'd3-color';
+import { interpolateHcl, quantize, interpolateLab } from 'd3-interpolate';
+import { scaleSequential, scaleLinear } from 'd3-scale';
+
+export type ColorPaletteName =
```
markov00: I'm seeing that some color palette are missing like `sinebow`,`YlGn`,`spectral` and this type mix the sequential with the diverging palettes. I think we should distinguish them in some ways

src/utils/colors/color_palette.ts
markov00: Is not about namespace, but avoid importing all the d3-color library to allow an optimized treeshaking of the library.
You can still use prefix d3 function if you prefer with `import { rgb as d3Rgb } from 'd3-color'`

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+  const paletteColors = [];
+  for (let i = 0; i < steps; i++) {
+    paletteColors.push(d3Color.rgb(d3ScaleChromatic.interpolateRainbow(i / steps)).hex());
+  }
+  return paletteColors;
+}
+
+export function getDivergingPalette(steps: number) {
+  const scale = scaleSequential(d3ScaleChromatic.interpolateRdYlGn).domain([1, steps]);
```
markov00: There are multiple diverging palette that can be used, I think we should apply the same strategy adopted for `getCategoricalPalette`

majagrubic: Not sure what you mean? Categorical uses `interpolateHcl`. Should we use a different interpolator here?

src/utils/colors/color_palette.ts
monfera: Another reason for individual, aliased D3 imports in repos that are not all-in D3 applications - it's a convenient way to find all entry points to D3 use, just by grepping `d3_` such as `d3_rgb`. The underscore might have a slight edge: 

- super precise grepping, as `d3_` is unlikely to occur elsewhere (while `d3` is noisy - it also occurs in colors, UUIDs, tokens, Base64 encoded files eg. encoded SVG, yarn integrity and other hash numbers etc.)
- there's no change to the capitalization of the D3 library functions
- the `_` is slightly weird visually (though easier to spot), but so is the presence of a number inside a camel3Cased identifier

Having said this, either prefix is OK for me (or even their lack thereof), the main point is a slight preference for named, aliased, easy to find imports from D3 in code where it is more of a utility. Sometimes I put all D3 functions into a single file eg. `d3utils.ts` for some given functionality so it's easy to version bump or replace any of them centrally, if needed.

src/utils/colors/color_palette.ts
monfera: It'd be neat to loop in the Tableau palette for the moment for internal testing.

So with Marco's suggestion, both [cyclical palettes](https://github.com/d3/d3-scale-chromatic#cyclical) would be included, right?

src/utils/colors/color_palette.ts
monfera: My 1st impression was `keyof typeof` too, hopefully it can be combined with conditionals - eg. check that the user-specified object key exists on the color object - so [type guarding](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) is possible. In this case, if D3 retires a color, it can fall back onto some default palette. 

Maybe relying on the simplest test cases to detect if a palette was dropped or renamed by D3 is safer than modeling the set of colors as a black box and falling back onto default colors if some go missing

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,157 @@
...
+export function getCategoricalPalette(name: CategoricalSchemeName): ReadonlyArray<string> {
+  if (name.startsWith('eui')) {
+    return euiColorPalette(name);
+  }
+  const schemeName: string = transformSchemeName(name);
+  // @ts-ignore
+  return d3ScaleChromatic[schemeName];
+}
+
+export function getCustomCategoricalPalette(colors: string[], steps: number) {
```
monfera: Not sure if within this PR, but it'd be neat to brainstorm about names, eg. it's a great name overall (no abbreviations to forget etc.) so the only question is, is the `get` adding value. Usually, functions are equivalent (replaceable by) their return value, or to put it differently, I think it should be the norm rather than the exception that a function is referentially transparent, just doing its thing of returning some value determined by its input. Calling it `customCategoricalPalette` would suggest that we're doing value based programming rather than place oriented programming

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,157 @@
...
+  const paletteColors = [];
+  const scale = quantize(interpolateHcl(colors[0], colors[1]), steps);
+  for (let i = 0; i < steps; i++) {
+    paletteColors.push(d3Color.rgb(scale[i]).hex());
+  }
+  return paletteColors;
+}
+
+export function getSequentialPalette(name: ColorPaletteName, steps: number) {
```
monfera: It looks like all current ways for getting the colors require the specification of the `steps`, materializing into a array. We should be able to use continuous color scales too, is it out of scope for this PR? May be useful to add eg. a function that'd just take a D3 continuous color ramp, a `fromValue` and a `toValue` and it'd return a function that can be called with arbitrary numbers between the two values.

src/utils/colors/color_palette.ts
markov00: Here: https://github.com/d3/d3-scale-chromatic#diverging we have various palette to use:
`BrBG`, `PRGn`,`PiYG` etc

markov00: categorical use the scheme name: `d3ScaleChromatic[schemeName];`

src/utils/colors/color_palette.ts
markov00: Maybe we can have that as a next-phase PR

src/utils/colors/color_palette.ts
majagrubic: This doesn't mix the sequential with diverging palettes, but sequential single-hue and multi-hue palettes. Is it important to distinguish the two?

majagrubic:
I've incorporated your comments and added a better check if the color palette exists in d3 namespace. Also, got rid of ts-warnings.
Marking this as ready for review.

package.json
```diff
@@ -139,11 +139,17 @@
     "webpack-dev-server": "^3.3.1"
   },
   "dependencies": {
+    "@types/d3-color": "^1.2.2",
+    "@types/d3-interpolate": "^1.3.1",
+    "@types/d3-scale-chromatic": "^1.3.1",
```
markov00: I don't think we need to have the type deps here since we are not exposing any function, return type or type that comes directly from d3. I think it's safe to move them under `devDependencies`

src/utils/colors/color_palette.ts
markov00: since we have this here and there in the whole library, I think it's ok for now keep the name like this

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,202 @@
...
+
+export function getCustomCategoricalPalette(colors: string[], steps: number) {
+  const paletteColors = [];
+  const scale = quantize(interpolateHcl(colors[0], colors[1]), steps);
+  for (let i = 0; i < steps; i++) {
+    paletteColors.push(d3Rgb(scale[i]).hex());
+  }
+  return paletteColors;
+}
+
```
markov00: A categorical palette is composed by a set of colors where each one differs from the other like the EUI palette or the any other palette in https://github.com/d3/d3-scale-chromatic#categorical.
Interpolating the HCL for the first two colors create a sequential scale rather then a categorical one.
I think we should remove this function and let the consumer just input a custom array of different color if they want a custom categorical palette

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,202 @@
...
+export function getSequentialPalette(name: SequentialColorPaletteName, steps: number) {
```
markov00: can you please add the return type: `ReadonlyArray<string>` to be aligned with the `getCategoricalPalette`

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,202 @@
...
+  const interpolator = getInterpolatorOrThrow(name);
+  const paletteColors = [];
+  const scale = scaleSequential(interpolator).domain([1, steps]);
+  for (let i = 1; i <= steps; i++) {
+    paletteColors.push(d3Rgb(scale(i)).hex());
+  }
+  return paletteColors;
+}
+
+export function getCustomSequentialPalette(colors: string[], steps: number) {
```
markov00: can you please add the return type: `ReadonlyArray<string>` to be aligned with the `getCategoricalPalette`

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,202 @@
...
+    .domain([1, steps])
+    .interpolate(interpolateLab)
+    .range([d3Rgb(startingColor), d3Rgb(endingColor)]);
+  for (let i = 1; i <= steps; i++) {
+    paletteColors.push(d3Rgb(scale(i)).hex());
+  }
+  return paletteColors;
+}
+
+export function getCyclicalPalette(name: CyclicalPaletteName, steps: number) {
```
markov00: return type: `ReadonlyArray<string>`

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+  const interpolator = getInterpolatorOrThrow(name);
+  const paletteColors = [];
+  for (let i = 0; i < steps; i++) {
+    paletteColors.push(d3Rgb(interpolator(i / steps)).hex());
+  }
+  return paletteColors;
+}
+
+export function getDivergingPalette(steps: number, interpolatorName?: DivergingInterpolatorName) {
```
markov00: return type: `ReadonlyArray<string>` and interpolator default to `spectral` directly in the arguments

src/utils/colors/color_palette.ts
majagrubic: but sequential will have very similar colors in the range. the `quantize` in the categorical will make sure that the colors are uniformly distributed. and the first and last color in the categorical palette will be the same as user defined, while in the sequential it will be interpolated to be something similar.
Example, for the input colors: `['#007AFF', '#FFF500']` and 10 steps:
Screenshot 2019-12-04 at 10 37 10

generated categorical palette:
Screenshot 2019-12-04 at 10 36 53

generated sequential palette:
Screenshot 2019-12-04 at 10 37 05

I think it's still useful.

markov00: I see that more as a sequential, multi hue palette than a categorical/qualitative color scale.
It's indeed a nice palette, but doesn't provide the same function as a categorical palette. In the categorical, the color should be easily distinguishable between each other, in your example i can barely distinguish between the 3 yellow, 3 purples, 2 blues, one orange.
I personally don't see that in line with the description of a categorical scale:
http://colorbrewer2.org/learnmore/schemes_full.html#qualitative
https://seaborn.pydata.org/tutorial/color_palettes.html

markov00: cc @cchaos

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,196 @@
...
+  | 'dark2'
+  | 'paired'
+  | 'pastel1'
+  | 'pastel2'
+  | 'set1'
+  | 'set2'
+  | 'set3'
+  | 'euiPaletteColorBlind'
+  | 'euiPaletteForLightBackground'
+  | 'euiPaletteForDarkBackground'
```
cchaos: Can we remove these from being an option? They're specific to Elastic branding and statefulness and shouldn't really be used for charts.

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,196 @@
+import * as d3ScaleChromatic from 'd3-scale-chromatic';
+import { rgb as d3Rgb, lab as d3Lab, RGBColor } from 'd3-color';
+import { interpolateLab } from 'd3-interpolate';
+import { scaleLinear, scaleSequential } from 'd3-scale';
+
+export type SequentialColorPaletteName =
```
cchaos: There are **a lot** of options in this list. While choice is great, it can cripple decision and allow for incorrect choices to be made. Can we reduce this list to just a few most useful ones like "warm", "cool"?

cchaos: Also, are these options going to be documented anywhere with visual examples/representations? I have no idea what `cubehelixDefault` would result as and consumers would need to apply the palette, compile their app, then view the chart (and hope they have a full number of series) in order to see the palette.

src/utils/colors/color_palette.ts
```diff
@@ -0,0 +1,196 @@
...
+export type CategoricalSchemeName =
+  | 'category10'
+  | 'accent'
+  | 'dark2'
+  | 'paired'
+  | 'pastel1'
+  | 'pastel2'
+  | 'set1'
+  | 'set2'
+  | 'set3'
```
cchaos: What are these generically named palettes?

majagrubic: These are the categorical palettes that d3 has to offer: https://github.com/d3/d3-scale-chromatic#schemeSet1
For simplicity, I decided to left out the `scheme` prefix.

src/utils/colors/color_palette.ts
majagrubic: Yes, we'll have a storybook example with palettes being applied to charts and users will get to try it out. This is just the first step of the implementation. 
Caroline, I agree with your suggestion that we keep some of the palettes - but the problem here is that I don't really know what "a few most useful" palettes are, as this is quite far from my domain of expertise. Could you please advise on which palettes should be kept and which removed? You can use d3-scale-chromatic for representation of palettes:
https://github.com/d3/d3-scale-chromatic

cchaos:
> Could you please advise on which palettes should be kept and which removed? 

My advice is to use some sort of object/map that names these color palettes based on their source but also have a few common and shortly named palettes for easy use.

For example:

### Commonly named palettes

Charts would provide a simple list of commonly named palettes like "temperature", "warm", "cool", "grayscale", "category". These would map to hard-coded values provided by Charts (which could also be taken from the D3/EUI options).

The consumer would then write:

```js
const theme = {
  colors: {
    vizColors: 'temperature',
  },
};
```

What these common palettes are still need to be determined. I'll get some thoughts started on this.

### D3 palettes

Providing an easy way for consumers to use these D3 palettes is nice, especially for applications that also allow users to select their own. However, for Elastic engineers we would like to have them stick to the common palettes or EUI palettes so we're consistent throughout.

But for consumers to understand that some palettes are coming from D3, they should have to write it like:

```js
const theme = {
  colors: {
    vizColors: d3Palletes.schemeSet1,
  },
};
```

And the names shouldn't change from the way they're named from D3. This gives consumers the reference point from where these palettes are coming from for them to go view/find them.

### EUI palettes

This will also give us the opportunity to build EUI custom palettes within EUI and provide them to consumers of both EUI and Charts.

```js
const theme = {
  colors: {
    vizColors: euiPaletteForStatus, // Imported from EUI not Charts
  },
};
```

### Custom palettes

And of course still being able to create custom palettes by providing 2+ colors to interpolate.

```js
const theme = {
  colors: {
    vizColors: getCustomSequentialPalette(['#FFFFE0', '#017F75'], 5),
  },
};
```

src/utils/colors/color_palette.ts
markov00: All color palettes, except EUI are taken from d3 scale chromatic -> https://github.com/d3/d3-scale-chromatic
We will document all of them in storybook (we are going to improve the docs soon https://github.com/elastic/elastic-charts/issues/457) 
If you like to select just some of them for us, please feel free

src/utils/colors/color_palette.ts
markov00: the naming comes from https://github.com/d3/d3-scale-chromatic schemas

cchaos:
I have done some looking at palettes and reached out to the other designers ([doc](https://docs.google.com/document/d/1QtdRNJwplyGmWam2eSuYuKXOU4z-WJAk2hnJkH6Ipjs)) to get consensus on a few "common" palettes.

I've then come up with a mapping of these common palettes to the D3 options (or custom) to use as defaults straight from Elastic charts. Links to palettes found in [this section of the doc](https://docs.google.com/document/d/1QtdRNJwplyGmWam2eSuYuKXOU4z-WJAk2hnJkH6Ipjs/edit#heading=h.afvttk9qzt20).

```js
Palettes = {
  'categorical': d3.schemeCategory10 || euiPaletteColorBlind, // The D3 version is not color blind safe
  'grayscale': d3.schemeGreys[k]
  'status': d3.schemeRdYlGn[k]
  'temperature': d3.schemeRdYlBu[k]
  'warm': d3.schemeOrRd[k]
  'cool': d3.schemePuBu[k]
  'complimentary': custom(#a97143, #f1a261, #dddddd, #5795d5, #3d6995)
}
```

In EUI, we will make our own version of the same object using the EUI (Elastic) color palette and ensuring better accessibility.

Consumers of both EUI and Elastic charts can then pass in the EUI version of the palette. Or consumers can grab any of the other D3 palettes or create their own.

majagrubic:
@cchaos one question:
what should `custom (#a97143, #f1a261, #dddddd, #5795d5, #3d6995)` 
return?
should we just return that array of colors? if we're gonna do custom sequential palette, we need a start and end color, not an array of 5 colors

cchaos:
So the custom sequential palette function doesn't allow you to specify more than 2 colors? Hmm, that seems pretty limiting given the fact that you might have at least 3 colors to interpolate from (start, middle, end) so that you can specify the middle to be gray (for example).

src/chart_types/xy_chart/state/selectors/get_color_palette.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (!settingsSpec.colorPalette) {
+      return undefined;
+    }
+    return getColorPalette(settingsSpec.colorPalette);
+  },
+)(getChartIdSelector);
+
+function getColorPalette(colorPalette: ColorPalette): ReadonlyArray<string> {
+  const { name, steps } = colorPalette;
+  switch (name) {
```
majagrubic: Depending on the name, we get the actual color palette.

src/state/selectors/get_chart_theme.ts
```diff
@@ -3,21 +3,33 @@ import { getSettingsSpecSelector } from './get_settings_specs';
...
+  let finalTheme = base;
 
   if (Array.isArray(theme)) {
     const [firstTheme, ...axillaryThemes] = theme;
-    return mergeWithDefaultTheme(firstTheme, base, axillaryThemes);
+    finalTheme = mergeWithDefaultTheme(firstTheme, base, axillaryThemes);
+  } else {
+    finalTheme = theme ? mergeWithDefaultTheme(theme, base) : base;
+  }
+  if (colorPaletteSpec) {
```
majagrubic: If color palette exists, it's applied here.

src/utils/colors/color_palette.ts
```diff
@@ -2,6 +2,22 @@ import * as d3ScaleChromatic from 'd3-scale-chromatic';
...
+  | 'status'
+  | 'temperature'
+  | 'warm'
+  | 'cool'
+  | 'complimentary';
+
+export type ColorPalette = {
+  name: ColorPaletteName;
+  steps?: number;
+};
```
majagrubic: These are the user facing palettes.

nickofthyme:
I think this issue will be handled now in kibana via the new palette service in the charts plugin. If not we can revisit these changes in the future, but this PR is too out of date to continue.

cc: @markov00

PR #491, opened 2019-12-09
refactor: decouple legend and tooltip phase 1 

markov00:
## Summary

This is the first phase to clean and decouple the current code for legend and tooltip from the XYChart dependency.
In this PR I've just moved out 3 types `Rotation`, `Datum` and `Position` from the XYChart to the main `src/utils/commons.ts` and I've also moved 1 level up the `scales` folder

### Checklist


src/utils/commons.ts
```diff
@@ -1,4 +1,19 @@
...
+export type Color = string;
+
+export const Position = Object.freeze({
+  Top: 'top' as 'top',
+  Bottom: 'bottom' as 'bottom',
+  Left: 'left' as 'left',
+  Right: 'right' as 'right',
+});
+
+export type Position = $Values<typeof Position>;
```
monfera: The $Values doesn't seem to add a lot, we could just use 
```
export type Position = keyof typeof Position;
```

src/utils/commons.ts
```diff
@@ -1,4 +1,19 @@
...
+export type Rotation = 0 | 90 | -90 | 180;
+export type Rendering = 'canvas' | 'svg';
+export type Color = string;
+
+export const Position = Object.freeze({
+  Top: 'top' as 'top',
+  Bottom: 'bottom' as 'bottom',
+  Left: 'left' as 'left',
+  Right: 'right' as 'right',
+});
```
monfera: It's good for now but wouldn't there be legit combos eg. `TopRight` etc. and other values like `RightMiddle`? Is it worth eventually separating the alignment to its two independent constituents as in HTML/CSS?

src/utils/commons.ts
```diff
@@ -1,4 +1,19 @@
-import * as uuid from 'uuid';
+import { v1 as uuidV1 } from 'uuid';
+import { $Values } from 'utility-types';
+
+export type Datum = any;
+export type Rotation = 0 | 90 | -90 | 180;
+export type Rendering = 'canvas' | 'svg';
```
monfera: Maybe `Rendering` -> `RenderTarget` or `RenderingSubstrate`?

src/utils/commons.ts
```diff
@@ -1,4 +1,19 @@
...
+export type Color = string;
```
monfera: It may not be for this PR: the type could broaden to include `{r, g, b, a}` etc.

src/utils/commons.ts
```diff
@@ -1,4 +1,19 @@
...
+
+export const Position = Object.freeze({
```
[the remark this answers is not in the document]
markov00: You are right, `Alignment` is much more representative than Position. We can plan this change in a later stage as this will introduce a breaking change for the user facing API

src/utils/commons.ts
markov00: is not the same type  http://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=17&pc=28#code/MYewdgzgLgBACiCBLKTwHkBGArApsWAXhizwIDoAzAJ11wC9cAKAbwCgZOYAlJAcwAWUAFwwA5NX5CxMAIYRxkwVDEAaDlwAyuSiPEAbHSrkKxh3WrYBfAJQBuNmygBPAA64YAEgBqs-QFdcCAAeABUYXAAPKFwwABMFEBx8KAA+GGJQgG0Aa1xnEEoYUIBdBxd3eEQUNDAfP0CFYnqAoOCK3EKq5FQMZIJU8rcPBB7agGl89EpQ4emMmDyCoo6u0Zq+sihHSn8wAlqYGOgARiZXat6wUXWrlsabGHYrNl39q6OgqAAmc8vam7-cCTZzTWbuaaPFgwF6OY5QM63WqkFLkbS6exwr6-JGbVHoqD2IA

we can change that, but it still leads to a breaking change for the API

src/utils/commons.ts
markov00: For sure, we want to limit as much as possible these properties to the one used and implemented inside the code. If we add a `TopRight` for example we should review all the places that use `Position` and limit that type only to the implemented one. The same happens for all the user facing API (Position is used for the Axis position and the Legend)

src/utils/commons.ts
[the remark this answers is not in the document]
monfera: Thanks, sounds fine! If there's a collection of post-PR ideas, can you link it? Eg. like the small post-Redux tweaks such as cache eviction for selectors. At least I forget stuff without lists 😃 Pls. resolve this thread once you read this

src/utils/commons.ts
monfera: Thanks Marco for the super clear example, now I see how the way you're using it (via `$Values`) is more about the _value_ (haha obv. in retrospect) whereas my incorrect assumption was that it'd be about the _key_ so the user can specify charts via serializable things like strings here, something like [this](http://www.typescriptlang.org/play/#code/MYewdgzgLgBACiCBLKTwHkBGArApsWAXhizwIDoAzAJ11wC9cAKAbwCgZOYAlJAcwAWUAFwwA5NX5CxMAIYRxkwVDEAaDlwAyuSiPEAbHSrkKxh3WrYBfAJQBuNmygBPAA64YANVn6ArrggAHgAVGFwADyhcMAATBRAcfCgAPhhiYIBtAGtcZxBKGGCAXQcXd3hEFDQwABJvPwC0rx9-ILLcfIrkVAxEgmTStw8EbuqAaVz0SmChqaacvIL2zpGq3rIoR0pfMAJqmCjoAEYmV0qesFFVi7qWgJsYdi4YUEgQQ3J9ED5T8+r7axbHZ7cAHAJQABMTAWVz+4AmzimM3cUweLBgGk4rwg71wn2+TGu1VISWyuSKAKsjkOUBORPWSXI2l0AKc4KhYmZKnsMAA9LyYBABCBfPoYjBMB50GM2dAOcEQK4xDz+YLhaLxQJZAA3DyyGC0cUQACOvn4fH0zhg+iQYD1CgVrhgSAUYBAsH1CyAA)

It looks fine to have a specification one way or another, so I don't have a change request here, but I'm interested in learning about the tradeoff of expecting a (type checked) string vs. exporting an object with possible values and expecting that the user specifies one of them like in your example.

Btw. my motivation was more about the use of an external library, and to me, unfamiliar dollar sign, in favor of, in this case, just us defining `type Values<T extends object> = T[keyof T];` - should I learn a bit about this library, are we planning to use it more and more?

monfera: ... just looked it again, and if the keys and values are kept same, we can have it [both ways](http://www.typescriptlang.org/play/?ssl=24&ssc=45&pln=24&pc=33#code/MYewdgzgLgBACiCBLKTwHkBGArApsWAXhizwIDoAzAJ11wC9cAKAbwCgZOYAlJAcwAWUAFwwA5L0FQxMAIYRxkoWIA0HLgBlclEYu3S5CsVp2q2AXwCUAbjZsoATwAOuGADVZAGwCuuCAB4AFRhcAA8oXDAAEwUQHHwoAD4YYkCAbQBrXAcQShhAgF1bRxd4RBQ0MAASDx8-FPcvXwCS3Fyy5FQMeIJE4udXBE7KgGls9EpAgYmGrJy81vahiu6yKDtKbzACSpgI6ABGJidyrrBRZbOapr9LGHYuGFBIEE9cck8QPmPTypsLDZbHbgPZ+KAAJiYcwuv3AYwcEymLgmdxYMHUnGeEFe70+30ulVICUy2QK-3Mdn2UCOBNWCXIJig-3sYMhxn0YhsMAA9NyYBABCBvJ4ojBMK50CMWdA2YEQE5OdYeXyBUKRTABLIAG6uWQwWiiiAAR28-D4ngcME8SDAuoUcqcMCQCjAIFgerm0ohTFpYCJFEZd15-MFwtF4pIUqpkN9-qgaXZpjJypDavDEqjrJ9sL9PXjYgdYmTwdVYY12t1+twhpNZotVptdvy8qdLrdchgnqAA) - but maybe there's a deeper reason for varying uppercase/lowercase between key and value. So, with the linked way the user can specify something either via an object reference or a string directly. If you think it's useful, it still needs not be part of this PR. Again, maybe there are tradeoffs I'm overlooking and you've no intention to support strings, glad to learn 🙇

src/utils/commons.ts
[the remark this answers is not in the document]
markov00: I've opened an issue to collect this: https://github.com/elastic/elastic-charts/issues/495

src/utils/commons.ts
markov00: There is no further reasons for that. Just historically we started that way with uppercase key and lowercased value. But, as always, nothing limit us to switch to a different way.
I don't see any benefit having one or the other way + the utility-types lib doesn't add any KB to the current bundle as it's just a typescript thing

PR #506, opened 2020-01-02
fix(specs): shows a chart message without series specified

markov00:
## Summary

This commit fix the use case when no series specs are configured. Instead of throwing an error, it
will just display the "No data to display" message.

Screenshot 2020-01-02 at 17 39 56

### Checklist

### DIFF
--- a/src/components/chart.tsx
+++ b/src/components/chart.tsx
@@ -65,9 +65,10 @@ export class Chart extends React.Component<ChartProps, ChartState> {
 
     const id = uuid.v4();
     const storeReducer = chartStoreReducer(id);
-    const enhancers = (window as any).__REDUX_DEVTOOLS_EXTENSION_COMPOSE__
-      ? (window as any).__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({ trace: true, name: `@elastic/charts (id: ${id})` })()
-      : undefined;
+    const enhancers =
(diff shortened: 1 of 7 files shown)


src/components/chart.tsx
```diff
@@ -65,9 +65,10 @@ export class Chart extends React.Component<ChartProps, ChartState> {
 
     const id = uuid.v4();
     const storeReducer = chartStoreReducer(id);
-    const enhancers = (window as any).__REDUX_DEVTOOLS_EXTENSION_COMPOSE__
-      ? (window as any).__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({ trace: true, name: `@elastic/charts (id: ${id})` })()
-      : undefined;
+    const enhancers =
+      typeof window !== 'undefined' && (window as any).__REDUX_DEVTOOLS_EXTENSION_COMPOSE__
```
nickofthyme: `undefined` as a string?

markov00: jest will complain if I use a simple `window !== undefined && ....` so I'm checking with typeof. the type of an undefined variable is `undefined` as a string
```
typeof anything_that_is_undefined === 'undefined'
```
I can invert that with ` typeof window === 'object'`

nickofthyme: wow that's strange. I guess as long as it works!

Issue #498, opened 2019-12-19
Migrate stories source to Docs


markov00:
Let's remove the deprecated Storybook `addon-info` and let's start moving our story sources to Storybook Docs. 

Let's move our stories to the CSF https://storybook.js.org/docs/formats/component-story-format that is the new format adopted in 5.2 to document stories.

I think we should show a simple static example of each story source in the Docs page instead of the compiled code from the story itself. We should check how we can do that without duplicate too much code between the docs and the story example.

We should check that the props table is correctly rendered with all the props and their jsdocs.

## Tasks
- [ ] POC the CSF format for a small set of stories
- [ ] POC of the Docs (show source + props table) for that story set
- [ ] move all stories to CSF
- [ ] remove `addon-info` and move all sources to Docs

markov00:
done

PR #508, opened 2020-01-03
docs: upgrade storybook to 5.2.8 and refactor stories to CSF 

rshen91:
## Summary
This PR is addressing issue #498. Stories have been refactored to be in CSF format and further documentation improvement has been broken into issue https://github.com/elastic/elastic-charts/issues/523 to address in a subsequent PR. 

### Checklist


rshen91:
Choose any story that uses knobs and it hangs with all knobs available disregarding defaults. After clicking any of the checkboxes, then storybook resolves to the correct knob configuration. Investigating...

rshen91:
Thank you @markov00 for the help with commit a75d477 🙇‍♀  and for reference: https://github.com/smrq/babel-plugin-require-context-hook/blob/master/register.js

rshen91:
@markov00 I have a very sketchy basic table included in bar_chart.tsx with the BarSeries props, and wanted to touch base about other props (in this table or in a new one) that would make sense to be included in the docs section for the basic bar chart. 

In terms of the story source, it looks like there are some good paths to explore with MDX (but seems to be problematic with CSF docs rendering here: https://github.com/storybookjs/storybook/issues/7479#issuecomment-574728632). In commit eac67b2 I'm working on the built in Preview blocks with the docs addon. I'm having issues with the <Story name="basic"></Story> tags with "basic" not being defined. I think this might give us a code sample if I'm understanding correctly (https://github.com/storybookjs/storybook/issues/7607). Let me know if you think this is the right path to go down and if you have any suggestions.

rshen91:
As in earlier commit, a75d477, in recent commit d3596a5 I deleted the visual regression test image snapshots and reran the tests. There were url changes so the old snapshots would be slightly off.

PR #536, opened 2020-02-05
fix: empty domain error for ordinal x scale

nickofthyme:
## Summary

Fixes issue related to empty domain.

If the data is an empty array and the `xScaleType` is `Ordinal` the `scaleBand` from `d3-scale` will run

```ts
new Array(-1)
```

And throw a `RangeError`. This is a quick fix, but ideally, we revisit this to avoid unnecessarily computing the domain if the data is empty for all specs.

### Checklist

### DIFF
--- a/src/chart_types/xy_chart/domains/x_domain.ts
+++ b/src/chart_types/xy_chart/domains/x_domain.ts
@@ -93,11 +93,6 @@ export function mergeXDomain(
     minInterval = customMinInterval || computedMinInterval;
   }
 
-  // TODO: Check for empty data before computing domain
-  if (seriesXComputedDomains.length === 0) {
-    seriesXComputedDomains = [0, 0];
-  }
-
   return {
     type: 'xDomain',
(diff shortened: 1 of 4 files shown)


src/chart_types/xy_chart/domains/x_domain.ts
```diff
@@ -93,6 +93,11 @@ export function mergeXDomain(
     minInterval = customMinInterval || computedMinInterval;
   }
 
+  // TODO: Check for empty data before computing domain
+  if (seriesXComputedDomains.length === 0) {
+    seriesXComputedDomains = [0, 0];
```
markov00: This should be just one single value `[0]` because we are in ordinal domain and we don't want duplicated domain values

markov00: it's also better to move that inside the `computeOrdinalDataDomain` of domains.ts

nickofthyme: c41c0ef

Issue #537, opened 2020-02-05
Add groupIdAccessor


nickofthyme:
**Is your feature request related to a problem? Please describe.**
Allow `groupIdAccessor` props to specify the groupId of a series within a single spec.

**Describe alternatives you've considered**
The alternative is two separate specs that can each have it's own `groupId`. This causes labels to show only split accessor value and not the y accessor labels.

**Additional context**
```tsx
<LineSeries
  id="line"
  xScaleType={ScaleType.Ordinal}
  yScaleType={ScaleType.Linear}
  xAccessor="x"
  yAccessors={['y1', 'y2']}
  splitSeriesAccessors={['z']}
  groupIdAccessor={({ yAccesor }) => yAccesor === 'y1' ? 'group1' : 'group2'}
  data={visData.rows}
/>
```

Then I can define two axes one for `group1` and one for `group2`

PR #539, opened 2020-02-06
feat: add series name settings

nickofthyme:
## Summary

Related to #537 (interim fix)

BREAKING CHANGE:
Remove `customSubSeriesName` prop on series specs in favor of cleaner api using just the `name` prop on `SeriesSpec`. The types `SeriesStringPredicate`, `SubSeriesStringPredicate` have been removed.

The `name` prop is defined below...

```ts
type SeriesNameAccessor = string | SeriesNameFn | SeriesNameConfigOptions;
// where

export interface SeriesSpec extends Spec {
  ...
  /**
   * The name of the spec. Also a mechanism to provide custom series names.
   */
  name?: SeriesNameAccessor;
  ...
}
```

This original function type definition is the same

```ts
type SeriesName = string | number | null;
type SeriesNameFn = (series: SeriesIdentifier, isTooltip: boolean) => SeriesName;
```

However the `SeriesStringPredicate` is now a union type between `SeriesNameFn` and the new `getSeriesNameFromOptions` type defined below.

```ts
export interface SeriesNameConfig {
  /**
   * accessor key (i.e. `yAccessors` and `seriesSplitAccessors`)
   */
  accessor: string | number;
  /**
   * Accessor value (i.e. values from `seriesSplitAccessors`)
   */
  value?: string | number;
  /**
   * New name for Accessor value
   *
   * If not provided, the original value will be used
   */
  name?: string | number;
  /**
   * Sort order of label, overrides order listed in array.
   *
   * lower values - front most
   * higher values - end most
   */
  sortIndex?: number;
}
export interface SeriesNameConfigOptions {
  /**
   * Array of accessor naming configs to replace series names
   *
   * Only provided configs will be included
   * (i.e. if you only provide a single mapping for `yAccessor`, all other series accessor labels will be ignored)
   *
   * The order of configs is the order in which the resulting labels will
   * be joined, if no `sortIndex` is specified.
   *
   * If no values are found for a giving mapping in a series, the mapping will be ignored.
   */
  names?: SeriesNameConfig[];
  /**
   * Delimiter to join values/labels
   *
   * @default ' - '
   */
  delimiter?: string;
}
```

> Note: The order of the `names` array is used to join the names

### Mappings Examples

```ts
const data = [
  {
    x: 0,
    y1: 'speed',
    y2: 'acceleration',
    g: 'ferrari',
  },
  {
    x: 1,
    y1: 'speed',
    y2: 'acceleration',
    g: 'porsche',
  },
]
```

#### Replace yAccessors

```tsx
<LineSeries
  data={data}
  xAccessor="x"
  yAccessors={['speed', 'acceleration']}
  splitSeriesAccessors={['g']}
  customSubSeriesName={{
    names: [
      {
        accessor: 'g',
        value: 'ferrari',
      },
      {
        accessor: 'g',
        value: 'porsche',
      },
      {
        accessor: 'speed',
        name: 'm/s',
      },
      {
        value: 'acceleration',
        name: 'm/s/s',
      },
    ],
  }}
/>
```

| Default names | Custom mapped Names |
| :---: | :---: |
| `ferrari - speed` | `ferrari - m/s` |
| `ferrari - acceleration` | `ferrari - m/s/s` |
| `porsche - speed` | `porsche - m/s` |
| `porsche - acceleration` | `porsche - m/s/s` |

#### Replace splitAccessors and reorder with `delimiter`

```tsx
<LineSeries
  data={data}
  xAccessor="x"
  yAccessors={['speed', 'acceleration']}
  splitSeriesAccessors={['g']}
  customSubSeriesName={{
    names: [
      {
        accessor: 'speed',
      },
      {
        accessor: 'acceleration',
      },
      {
        accessor: 'g',
        value: 'ferrari',
        name: 'Ferrari',
      },
      {
        accessor: 'g',
        value: 'porsche',
        name: 'Porsche',
      },
    ],
    delimiter: ' : ',
  }}
/>
```

| Default names | Custom mapped Names |
| :---: | :---: |
| `ferrari - speed` | `speed : Ferrari` |
| `ferrari - acceleration` | `acceleration : Ferrari` |
| `porsche - speed` | `speed : Porsche` |
| `porsche - acceleration` | `acceleration : Porsche` |

#### Single series

```tsx
<LineSeries
  data={data}
  xAccessor="x"
  yAccessors={['speed']}
  splitSeriesAccessors={['g']}
  customSubSeriesName={{
    names: [
      {
        accessor: 'g',
        value: 'ferrari',
        name: 'Ferrari',
      },
      {
        accessor: 'g',
        value: 'porsche',
        name: 'Porsche',
      },
      // This name be excluded from the name if not in the array
      // {
      //   accessor: 'speed',
      // },
    ],
  }}
/>
```

| Default names | Custom mapped Names |
| :---: | :---: |
| `ferrari` | `Ferrari` |
| `porsche` | `Porsche` |

See [test](https://github.com/elastic/elastic-charts/blob/6e9ee730015843610b102fff2f20128b82f6bf3e/src/chart_types/xy_chart/utils/series.test.ts#L670-L884) cases for more examples

### Checklist


robgil:
# [Codecov](https://codecov.io/gh/elastic/elastic-charts/pull/539?src=pr&el=h1) Report
> Merging [#539](https://codecov.io/gh/elastic/elastic-charts/pull/539?src=pr&el=desc) into [master](https://codecov.io/gh/elastic/elastic-charts/commit/5043725b7581bfea2340ca5323743d9fe03a4e19&el=desc) will **increase** coverage by `0.31%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/elastic/elastic-charts/pull/539?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master     #539      +/-   ##
==========================================
+ Coverage   75.56%   75.88%   +0.31%     
==========================================
  Files         181      194      +13     
  Lines        5689     5879     +190     
  Branches     1103     1136      +33     
==========================================
+ Hits         4299     4461     +162     
- Misses       1374     1400      +26     
- Partials       16       18       +2     
```


| [Impacted Files](https://codecov.io/gh/elastic/elastic-charts/pull/539?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [src/mocks/series/index.ts](https://codecov.io/gh/elastic/elastic-charts/pull/539/diff?src=pr&el=tree#diff-c3JjL21vY2tzL3Nlcmllcy9pbmRleC50cw==) | `100.00% <0.00%> (ø)` | |
| [src/mocks/scale/index.ts](https://codecov.io/gh/elastic/elastic-charts/pull/539/diff?src=pr&el=tree#diff-c3JjL21vY2tzL3NjYWxlL2luZGV4LnRz) | `100.00% <0.00%> (ø)` | |
| [src/mocks/utils.ts](https://codecov.io/gh/elastic/elastic-charts/pull/539/diff?src=pr&el=tree#diff-c3JjL21vY2tzL3V0aWxzLnRz) | `100.00% <0.00%> (ø)` | |
| [src/mocks/series/seriesIdentifiers.ts](https://codecov.io/gh/elastic/elastic-charts/pull/539/diff?src=pr&el=tree#diff-c3JjL21vY2tzL3Nlcmllcy9zZXJpZXNJZGVudGlmaWVycy50cw==) | `93.33% <0.00%> (ø)` | |
| [src/mocks/scale/scale.ts](https://codecov.io/gh/elastic/elastic-charts/pull/539/diff?src=pr&el=tree#diff-c3JjL21vY2tzL3NjYWxlL3NjYWxlLnRz) | `87.50% <0.00%> (ø)` | |
| [src/mocks/series/series.ts](https://codecov.io/gh/elastic/elastic-charts/pull/539/diff?src=pr&el=tree#diff-c3JjL21vY2tzL3Nlcmllcy9zZXJpZXMudHM=) | `81.96% <0.00%> (ø)` | |
| [src/mocks/hierarchical/palettes.ts](https://codecov.io/gh/elastic/elastic-charts/pull/539/diff?src=pr&el=tree#diff-c3JjL21vY2tzL2hpZXJhcmNoaWNhbC9wYWxldHRlcy50cw==) | `100.00% <0.00%> (ø)` | |
| [src/mocks/specs/specs.ts](https://codecov.io/gh/elastic/elastic-charts/pull/539/diff?src=pr&el=tree#diff-c3JjL21vY2tzL3NwZWNzL3NwZWNzLnRz) | `77.27% <0.00%> (ø)` | |
| [src/mocks/series/data.ts](https://codecov.io/gh/elastic/elastic-charts/pull/539/diff?src=pr&el=tree#diff-c3JjL21vY2tzL3Nlcmllcy9kYXRhLnRz) | `100.00% <0.00%> (ø)` | |
| [src/mocks/series/utils.ts](https://codecov.io/gh/elastic/elastic-charts/pull/539/diff?src=pr&el=tree#diff-c3JjL21vY2tzL3Nlcmllcy91dGlscy50cw==) | `88.88% <0.00%> (ø)` | |
| ... and [4 more](https://codecov.io/gh/elastic/elastic-charts/pull/539/diff?src=pr&el=tree-more) | |

------

[Continue to review full report at Codecov](https://codecov.io/gh/elastic/elastic-charts/pull/539?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/elastic/elastic-charts/pull/539?src=pr&el=footer). Last update [cc561a4...a6ead1c](https://codecov.io/gh/elastic/elastic-charts/pull/539?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

markov00:
Hi @nickofthyme I've few doubts on this PR because the same effect can be achieved with the available props like the following:
```ts
customSeriesLabel={(args) => {
    return args.seriesKeys.join(' ~~ ');
}}
```

or 
```ts
customSeriesLabel={(args) => {
    return args.seriesKeys.slice(0, args.seriesKeys.length -1).join(' ~~ ');
}}
```

Users are always a bit reluctant to use Maps object (I remember most of them commenting on a simplified version) that could be in fact achieved with a simple object `{a: 'apple'}`

One other way we could think about this feature: what if we provide a set of utility function for this cases?
like:
```ts
function fullSeries(delimiter: string = ' - ') {
  return (args) => {
      return args.seriesKeys.join(delimiter);
  }
}

function simpleSeries(delimiter: string = ' - ') {
  return (args) => {
      return args.seriesKeys.slice(0, args.seriesKeys.length -1).join(delimiter);
  }
}
```

this will allow much more configurability and doesn't pollute the, already polluted, set of props we have in the chart components.

What do you think?

src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -54,12 +54,60 @@ export type PointStyleAccessor = (datum: RawDataSeriesDatum, seriesIdentifier: S
...
+   *
+   * Only provided mappings will be included
+   * (i.e. if you only provide a single mapping for `yAccessor`, all other series accessor labels will be ignored)
+   *
+   * The order of mappings is the order in which the resulting labels will
+   * be joined, if no `sortIndex` is specified.
+   *
+   * If no values are found for a giving mapping in a series, the mapping will be ignored.
+   */
+  mappings: SeriesLabelMapping[];
```
markov00: what if we specify the mappings as optional, so the user can specify only the delimiter and have the `default` behaviour but with a different delimiter?

src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -54,12 +54,60 @@ export type PointStyleAccessor = (datum: RawDataSeriesDatum, seriesIdentifier: S
...
+/**
+ * Function to create custom series label for a given series
+ */
+export type SeriesLabelFn = (series: SeriesIdentifier, isTooltip: boolean) => SeriesLabel;
+/**
+ * Accessor mapping to replace labels
+ */
+export interface SeriesLabelMapping {
+  /**
+   * accessor key (i.e. `seriesSlittAccessors`)
```
markov00:
```suggestion
   * accessor key (i.e. `seriesSplitAccessors`)
```

src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -54,12 +54,60 @@ export type PointStyleAccessor = (datum: RawDataSeriesDatum, seriesIdentifier: S
...
+   *
+   * ignored for `yAccessor`
+   */
+  accessor?: string;
+  /**
+   * Accessor value/label (i.e. `yAccessors` or values from `seriesSlittAccessors`)
```
markov00:
```suggestion
   * Accessor value/label (i.e. `yAccessors` or values from `seriesSplitAccessors`)
```

src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -239,20 +287,12 @@ export interface SeriesSpec extends Spec {
...
    * @param isTooltip - true if tooltip label, otherwise legend label
    */
-  customSeriesLabel?: SeriesStringPredicate;
-  /**
-   * Custom sub series naming predicate function.
-   *
-   * `customSeriesLabel` takes precedence
-   */
-  customSubSeriesLabel?: SubSeriesStringPredicate;
+  customSeriesLabel?: SeriesLabelAccessor;
```
markov00: The docs is no longer valid as it accept also the `SeriesLabelMappingOptions`

markov00: I've also another idea on this customization. Why don't we move the `SeriesLabelAccessor` to the `name` prop?
Because there we have just a simple string, that is basically used only in a set of small cases (only one dataset)
if we have more, the customSeriesLabel have always precedence.
Small non-trivial proposal: change the `name` prop to accept `name?: string | SeriesNameAccessor` renaming `SeriesLabelAccessor` to a more clear `SeriesNameAccessor`

.playground/playground.tsx
```diff
@@ -1,113 +1,40 @@
...
+    }));
+
     return (
       <>
         <div className="chart">
-          <Chart ref={this.chartRef}>
-            <Settings rotation={90} onBrushEnd={this.onBrushEnd} />
-            <Axis id="y" position={Position.Left} title={'y'} />
+          <Chart>
+            <Settings seriesLabels={{ delimiter: ': ', full: true }} />
```
markov00: `seriesLabels` is no longer valid

src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -54,12 +54,60 @@ export type PointStyleAccessor = (datum: RawDataSeriesDatum, seriesIdentifier: S
...
+ */
+export type SeriesLabelFn = (series: SeriesIdentifier, isTooltip: boolean) => SeriesLabel;
+/**
+ * Accessor mapping to replace labels
+ */
+export interface SeriesLabelMapping {
+  /**
+   * accessor key (i.e. `seriesSlittAccessors`)
+   *
+   * ignored for `yAccessor`
```
markov00: It sounds a bit strange to put the `yAccessor` into the `value` parameter instead into the `accessor` param.
The yAccessor is an accessor, so it should go to the `accessor` property, but it's value is ignored

.playground/playground.tsx
nickofthyme: yeah thats a good point I guess I was thinking of the final label value. I'll change that.

src/chart_types/xy_chart/utils/specs.ts
nickofthyme: this is defined above

```ts
export type SeriesLabelAccessor = SeriesLabelFn | SeriesLabelMappingOptions;
```

src/chart_types/xy_chart/utils/specs.ts
nickofthyme: yeah thats a good point I guess I was thinking of the final label value. I'll change that.

src/chart_types/xy_chart/utils/series.test.ts
```diff
@@ -661,4 +667,219 @@ describe('Series', () => {
...
+                value: 'a',
+              },
+              {
+                accessor: 'y',
+              },
+            ],
+            delimiter: ' ¯\\_(ツ)_/¯ ',
+          },
+        });
+        expect(actual).toBe('a ¯\\_(ツ)_/¯ y');
```
markov00: ❤️

src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -50,12 +50,58 @@ export type PointStyleAccessor = (datum: RawDataSeriesDatum, seriesIdentifier: S
...
+ */
+export type SeriesNameFn = (series: SeriesIdentifier, isTooltip: boolean) => SeriesLabel;
+/**
+ * Accessor mapping to replace labels
+ */
+export interface SeriesNameConfig {
+  /**
+   * accessor key (i.e. `yAccessors` and `seriesSplitAccessors`)
+   */
+  accessor: string;
```
markov00: What if we have our datum as an array of values `[0, 10, 20]` the splitAccessor could be a `number` other then a `string`

src/chart_types/xy_chart/utils/series.ts
```diff
@@ -379,85 +380,79 @@ export function getSplittedSeries(
...
+  if (!options.names) {
+    return null;
   }
 
-  const { seriesKeys } = seriesIdentifier;
+  return (
+    options.names
+      .sort(({ sortIndex: a = Infinity }, { sortIndex: b = Infinity }) => a - b)
+      .map(({ accessor, value, name }) => {
+        const accessorValue = (accessor ? splitAccessors.get(accessor) : null) ?? null;
```
markov00: Isn't the accessor always defined? also please be careful because the accessor should be also a number and this will be null if the accessor is `0`

src/chart_types/xy_chart/utils/series.ts
```diff
@@ -379,85 +380,79 @@ export function getSplittedSeries(
...
+        }
 
-  return isMultipleY ? seriesKeys : seriesKeys.slice(0, -1);
+        if (yAccessor === accessor) {
+          return name ?? accessor;
+        }
+        return null;
+      })
+      .filter((d) => Boolean(d) || d === 0)
+      .slice()
```
markov00: why are you slicing here? maybe it's better to slice just after line 393 to avoid sorting the original options.names object

src/chart_types/xy_chart/utils/series.ts
```diff
@@ -379,85 +380,79 @@ export function getSplittedSeries(
...
-    const { yAccessor, splitAccessors } = seriesIdentifier;
-    const fullKeyPairs: [string | number | null, string | number][] = [...splitAccessors.entries(), [null, yAccessor]];
-    const labelKeys = fullKeyPairs.map(getCustomSubSeriesName(spec.customSubSeriesLabel, isTooltip));
-
-    return isMultipleY ? labelKeys : labelKeys.slice(0, -1);
+const getSeriesNameFromOptions = (
+  options: SeriesNameConfigOptions,
+  { yAccessor, splitAccessors }: SeriesIdentifier,
+  delimiter: string,
+) => {
```
markov00: could you please rewrite this function as a named function instead of an anonymous function expression? We should add a linter rule for that to match the code style already used in the rest of the code

src/chart_types/xy_chart/utils/series.ts
```diff
@@ -379,85 +380,79 @@ export function getSplittedSeries(
...
-export function getSeriesLabel(
+export function getSeriesName(
   seriesIdentifier: SeriesIdentifier,
   hasSingleSeries: boolean,
   isTooltip: boolean,
   spec?: BasicSeriesSpec,
 ): string {
-  if (spec && spec.customSeriesLabel) {
-    const customLabel = spec.customSeriesLabel(seriesIdentifier, isTooltip);
+  let delimiter = ' - ';
```
markov00: Could you please move this as an external constant? In the future we can add it to the `theme` also

src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -50,12 +50,58 @@ export type PointStyleAccessor = (datum: RawDataSeriesDatum, seriesIdentifier: S
 export const DEFAULT_GLOBAL_ID = '__global__';
 
 export type FilterPredicate = (series: SeriesIdentifier) => boolean;
-export type SeriesStringPredicate = (series: SeriesIdentifier, isTooltip: boolean) => string | null;
-export type SubSeriesStringPredicate = (
-  accessorLabel: string | number,
-  accessorKey: string | number | null,
-  isTooltip: boolean,
-) => string | number | null;
+export type SeriesLabel = string | number | null;
```
markov00: should we change that to `SeriesName` ?

src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -50,12 +50,58 @@ export type PointStyleAccessor = (datum: RawDataSeriesDatum, seriesIdentifier: S
...
+   * New name for Accessor value
+   *
+   * If not provided, the original value will be used
+   */
+  name?: string | number;
+  /**
+   * Sort order of label, overrides order listed in array.
+   *
+   * lower values - front most
+   * higher values - end most
```
markov00: We should specify this as `left most`, `right most` because depends on the direction of the language (see RTL languages)

src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -50,12 +50,58 @@ export type PointStyleAccessor = (datum: RawDataSeriesDatum, seriesIdentifier: S
...
 export type FilterPredicate = (series: SeriesIdentifier) => boolean;
-export type SeriesStringPredicate = (series: SeriesIdentifier, isTooltip: boolean) => string | null;
-export type SubSeriesStringPredicate = (
-  accessorLabel: string | number,
-  accessorKey: string | number | null,
-  isTooltip: boolean,
-) => string | number | null;
+export type SeriesLabel = string | number | null;
+/**
+ * Function to create custom series label for a given series
```
markov00: `label` -> `name`

src/chart_types/xy_chart/utils/series.ts
nickofthyme: remnant

src/chart_types/xy_chart/utils/series.ts
nickofthyme: good catch

src/chart_types/xy_chart/utils/specs.ts
nickofthyme: yup

src/chart_types/xy_chart/utils/series.ts
nickofthyme: Sure but idk about adding a linting rule for this. Arrow functions have their purpose and to forbid them entirely is something I would disagree with.

src/chart_types/xy_chart/utils/series.ts
nickofthyme: Yeah I thought about that. I keep questioning what should go in the theme vs be a prop on some component.

For example the `radiusRatio` in the dot accessor code. Should that be on the theme? Even though it is meant to be adjustable?

src/chart_types/xy_chart/utils/specs.ts
nickofthyme: sure that's fine.

src/chart_types/xy_chart/utils/series.ts
```diff
@@ -381,85 +384,78 @@ export function getSplittedSeries(
...
+  { yAccessor, splitAccessors }: XYChartSeriesIdentifier,
+  delimiter: string,
+) {
+  if (!options.names) {
+    return null;
   }
 
-  const { seriesKeys } = seriesIdentifier;
+  return (
+    options.names
```
markov00: I think we should keep the `.slice()` here to avoid mutating the original names array

src/chart_types/xy_chart/utils/series.ts
```diff
@@ -381,85 +384,78 @@ export function getSplittedSeries(
...
-    const { yAccessor, splitAccessors } = seriesIdentifier;
-    const fullKeyPairs: [string | number | null, string | number][] = [...splitAccessors.entries(), [null, yAccessor]];
-    const labelKeys = fullKeyPairs.map(getCustomSubSeriesName(spec.customSubSeriesLabel, isTooltip));
-
-    return isMultipleY ? labelKeys : labelKeys.slice(0, -1);
+export function getSeriesNameFromOptions(
+  options: SeriesNameConfigOptions,
+  { yAccessor, splitAccessors }: XYChartSeriesIdentifier,
+  delimiter: string,
+) {
```
markov00: nit: could you add the return type?

src/chart_types/xy_chart/utils/series.ts
```diff
@@ -381,85 +384,78 @@ export function getSplittedSeries(
...
+    if (typeof spec.name === 'function') {
+      customLabel = spec.name(seriesIdentifier, isTooltip);
+    } else {
+      delimiter = spec.name.delimiter ?? delimiter;
+      customLabel = getSeriesNameFromOptions(spec.name, seriesIdentifier, delimiter);
+    }
 
     if (customLabel !== null) {
-      return customLabel;
+      return customLabel.toString();
```
markov00: nit: this will never be anything other than a `string` because the `.join` always return a string also with an array of numbers

nickofthyme: No `spec.name` could return a number

https://github.com/elastic/elastic-charts/blob/9dcb74010d0a0f49a698120df97baf059ea78925/src/chart_types/xy_chart/utils/specs.ts#L59-L63

markov00: Forget about it, I didn't check the rest of the function, you are right

markov00:
jenkins test this

PR #545, opened 2020-02-10
feat(legend): allow color picker component

### LINKED ISSUE #538: Render prop for color picker component
Allow color picker render prop

Allow `renderColorPicker` prop to pass value. 

```ts
interface Props {
  onChange: (color: string) => void;
}
type renderColorPicker = (props: Props) => React.Component;
```

Build our own color picker. For consistency and desire not to import EUI into elastic charts, the above option is preferable.

**Additional Context**
The logic to store colors in memory is still available in the legend so this would require handling the handshake between these two things.

Related to old issue #23

nickofthyme:
## Summary

Add option to pass color picker component to chart.

Closes #538

> Elastic charts will maintain the color selection in memory beyond chart updates. However, to persist colors beyond browser refresh the consumer would need to manage the color state and use the `color` prop on the `SeriesSpec` to assign a color via the `SeriesColorAccessor`.

with `FunctionComponent<LegendColorPickerProps>`
```tsx
<Chart>
  <Settings
    showLegend
    legendColorPicker={({ anchor, onClose, color, seriesIdentifier }) => (
      <MyCustomColorPicker
        anchor={anchor}
        onClose={onClose}
        color={color}
        seriesIdentifier={seriesIdentifier}
      />
    )}
  />
</Chart>
```

or with `Component<LegendColorPickerProps>`

```tsx
<Chart>
  <Settings
    showLegend
    legendColorPicker={MyCustomColorPicker}
  />
</Chart>
```

where

```ts
export interface LegendColorPickerProps {
  /**
   * Anchor used to position picker
   */
  anchor: HTMLDivElement;
  /**
   * Current color of the given series
   */
  color: string;
  /**
   * Callback to close color picker and set persistent color
   */
  onClose: () => void;
  /**
   * Callback to update temporary color state
   */
  onChange: (color: string) => void;
  /**
   * Series id for the active series
   */
  seriesIdentifier: XYChartSeriesIdentifier;
}
export type LegendColorPicker = ComponentType<LegendColorPickerProps>;
```

The button argument is the legend item that should be used as the `button` prop for `EuiPopover` as the positioning reference.

Remove `seriesColorOverrides` logic which was deprecated, now unused, back when the color picker was removed from elastic charts.

## Demo usage with `EuiColorPicker` and `EuiWrappingPopover`
http://localhost:9001/?path=/story/legend--color-picker

Screen Recording 2020-02-26 at 03 04 PM

### Code

https://github.com/elastic/elastic-charts/blob/400a70be384f3731e59f35149e5422bb02ef9572/stories/legend/9_color_picker.tsx#L13-L58

### Checklist


src/components/legend/legend_item.tsx
```diff
@@ -8,13 +8,21 @@ import { onLegendItemOutAction, onLegendItemOverAction } from '../../state/actio
 import { Position } from '../../chart_types/xy_chart/utils/specs';
 import { SeriesIdentifier } from '../../chart_types/xy_chart/utils/series';
 
+export type RenderColorPicker = (
+  onChange: (color: string) => void,
+  onClose: () => void,
+  isOpen: boolean,
+  button: NonNullable<ReactNode>,
```
markov00: I don't think it's a good idea to give the ColorPicker the right to render or not the LegendItem. I think that we should instead pass a Ref object here to the picker, that can use to determine where the LegendItem is and where to render itself.

src/components/legend/legend_item.tsx
```diff
@@ -106,19 +140,39 @@ export class LegendListItem extends React.Component<LegendItemProps> {
...
     const itemClasses = classNames('echLegendItem', `echLegendItem--${legendPosition}`, {
-      'echLegendItem-isHidden': !isSeriesVisible,
+      'echLegendItem--hidden': !isSeriesVisible,
       'echLegendItem__displayValue--hidden': !isLegendItemVisible,
     });
 
     return (
       <div className={itemClasses} onMouseEnter={this.onLegendItemMouseOver} onMouseLeave={this.onLegendItemMouseOut}>
-        {color && renderColor(color, isSeriesVisible)}
+        {color && this.renderColor(this.state.color ?? color, isSeriesVisible)}
```
markov00: In the way it's built, the color picker is just directly changing the color on the legend item, not on the series itself. Sounds a bit weird that the legend is updated and to update the geometry we need to pass a `customSeriesColors` property.
I think we should align the behaviour in one or the other way:
- the color picker doesn't change colors at all but just allow someone to render their own custom color picker. In this case, the consumer needs always to update/change the colors through the `customSeriesColor` prop
- the color picker also update the internal color state (as it was initially)

src/components/legend/legend_item.tsx
```diff
@@ -106,19 +140,39 @@ export class LegendListItem extends React.Component<LegendItemProps> {
...
+    const { renderColorPicker } = this.props;
+    if (renderColorPicker && this.state.isOpen) {
+      return renderColorPicker(this.onColorChange, this.setToggleIsOpen, this.state.isOpen, this.renderItem());
+    }
+
+    return this.renderItem();
   }
 
+  setToggleIsOpen = () => {
+    this.setState({ isOpen: !this.state.isOpen });
```
markov00: We should not reply on the current state object values (see https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous)

```suggestion
    this.setState((state) => ({ isOpen: !state.isOpen }));
```

src/components/legend/legend_item.tsx
```diff
@@ -106,19 +140,39 @@ export class LegendListItem extends React.Component<LegendItemProps> {
...
+  };
+
+  onColorChange = (color: string) => {
+    this.setState({
+      color,
+      isOpen: !this.state.isOpen,
+    });
```
markov00: Same as previous comment, but I think we should remove the color change here as per few comment before 
```suggestion
    this.setState((state) => ({
      color,
      isOpen: !state.isOpen,
    }));
```

src/components/legend/legend_item.tsx
nickofthyme: Yeah I tried that but the `EuiPopover` requires a `button` prop and can't pass a ref in it's place. Unless you think there is a way to do this that I missed.

https://elastic.github.io/eui/#/layout/popover

src/components/legend/legend.test.tsx
```diff
@@ -131,4 +131,121 @@ describe('Legend', () => {
...
+          <button id="change" onClick={() => onChange(customColor)}>
+            {customColor}
+          </button>
+          <button id="close" onClick={onClose}>
+            close
+          </button>
+          {button}
+        </div>
+      ) : (
+        { button }
```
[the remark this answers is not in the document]
nickofthyme: now passing ref to anchor element with `EuiWrappingPopover`

src/chart_types/xy_chart/state/selectors/get_tooltip_values_highlighted_geoms.ts
```diff
@@ -36,6 +36,10 @@ const EMPTY_VALUES = Object.freeze({
   highlightedGeometries: [],
 });
 
+export interface TooltipData {
+  header: TooltipValue | null;
+  values: TooltipValue[];
+}
```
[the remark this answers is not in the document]
nickofthyme: yup thanks

src/components/legend/legend_item.tsx
```diff
@@ -62,63 +63,104 @@ function renderLabel(
...
+          className={colorClasses}
+          aria-label="series color"
+          title="series color"
+        >
+          <Icon type="dot" color={color} />
+        </div>
+      );
+    }
+
+    // changing the default viewBox for the eyeClosed icon to keep the same dimensions
```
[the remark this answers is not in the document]
nickofthyme: Yup would this effect performance or just cleaner to see?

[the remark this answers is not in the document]
markov00: just easier to read (from the kibana style guide https://github.com/elastic/kibana/blob/master/STYLEGUIDE.md#returnthrow-early-from-functions)

stories/legend/9_color_picker.tsx
```diff
@@ -0,0 +1,53 @@
...
+    return (
+      <EuiWrappingPopover isOpen button={anchor} closePopover={onClose}>
+        <EuiColorPicker color={color} onChange={handleChange}></EuiColorPicker>
+      </EuiWrappingPopover>
+    );
+  };
+
+  return (
+    <Chart className="story-chart">
+      <Settings theme={theme} showLegend legendColorPicker={renderColorPicker} />
```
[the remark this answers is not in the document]
markov00: Another possible way to fix that is to tweak the Legend component and force him to not update its rendering when we are going from a `specsInitialized = true` to `specsInitialized = false` keeping the previous rendering on the DOM. (tweaking the shouldComponentUpdate should to the trick). But I'm not sure if that could cause issues with the rendering when changing specs or values etc

nickofthyme:
jenkins retest this

src/chart_types/xy_chart/utils/series.ts
```diff
@@ -474,17 +478,55 @@ export function getSortedDataSeriesColorsValuesMap(
   return new Map([...seriesColorsArray]);
 }
 
+/**
+ * Helper function to get highest override color.
+ *
+ * from highest to lowest: `temporary`, `customSeriesColors` then `persisted`
```
markov00: `customSeriesColors` => `color` or `customColors`

src/components/legend/legend_item.tsx
```diff
@@ -20,6 +20,9 @@ interface LegendItemProps {
   onLegendItemOverListener?: LegendItemListener;
   legendItemOutAction: typeof onLegendItemOutAction;
   legendItemOverAction: typeof onLegendItemOverAction;
+  clearTemporaryColors: () => void;
+  setTemporaryColor: (key: SeriesKey, color: string) => void;
+  setPersistedColor: (key: SeriesKey, color: string) => void;
```
markov00:
```suggestion
  clearTemporaryColors: typeof clearTemporaryColors;
  setTemporaryColor: typeof setTemporaryColor;
  setPersistedColor: typeof setPersistedColor;
```

src/specs/settings.tsx
```diff
@@ -114,9 +114,25 @@ export interface TooltipProps {
...
 
 export interface LegendColorPickerProps {
+  /**
+   * Anchor used to position picker
+   */
   anchor: HTMLDivElement;
+  /**
+   * Current color of the given series
+   */
   color: string;
```
markov00:
```suggestion
  color: Color;
```

src/specs/settings.tsx
```diff
@@ -114,9 +114,25 @@ export interface TooltipProps {
...
+  /**
+   * Callback to close color picker and set persistent color
+   */
   onClose: () => void;
+  /**
+   * Callback to update temporary color state
+   */
+  onChange: (color: string) => void;
```
markov00:
```suggestion
  onChange: (color: Color) => void;
```

src/state/chart_state.ts
```diff
@@ -116,27 +117,56 @@ export interface ExternalEventsState {
...
   interactions: InteractionsState;
-  // external event state
+  /**
+   * external event state
+   */
   externalEvents: ExternalEventsState;
+  /**
+   * Color map used to persist color picker changes
+   */
+  colors: ColorOverrides;
```
markov00: thanks for this cleanup ❤️

stories/legend/9_color_picker.tsx
```diff
@@ -45,8 +50,8 @@ export const example = () => {
         xAccessor="x"
         yAccessors={['y']}
         splitSeriesAccessors={['g']}
-        data={data}
-        customSeriesColors={({ key }) => colors[key] ?? null}
+        data={BARCHART_1Y1G}
+        color={({ key }) => colors[key] ?? null}
       />
     </Chart>
   );
```
markov00: Please add a description of the mechanism here:
```
example.story = {
  parameters: {
    info: {
      text: 'Elastic charts will maintain the color selection in memory beyond chart updates. However, to persist colors beyond browser refresh the consumer would need to manage the color state and use the color prop on the` SeriesSpec` to assign a color via the `SeriesColorAccessor`.',
    },
  },
};
```

stories/legend/9_color_picker.tsx
```diff
@@ -1,40 +1,45 @@
...
       onClose();
+      onCloseAction();
       setColors({
         ...colors,
         [seriesIdentifier.key]: color,
       });
     };
+    const handleChange = (color: string) => {
+      onChangeAction(color);
+      onChange(color);
```
markov00: nit: just to keep consistency on the code between the handleChange and the handleClose
```suggestion
      onChange(color);
      onChangeAction(color);
```

src/components/legend/legend_item.tsx
nickofthyme: I tried this but for some reason there was a type error last time, idk but it's fixed 👍

Issue #499, opened 2019-12-19
Add eslint snake_case rule for filenames


markov00:
**Is your feature request related to a problem? Please describe.**
Align the repo to the current standard adopted by Kibana on filenaming.

https://github.com/elastic/kibana/blob/master/STYLEGUIDE.md

PR #550, opened 2020-02-13
fix: add snake case check for filenames

rshen91:
## Summary

Closes issue #499 

### Checklist


rshen91:
Ugh, crud, with commit 58d5c57 it looked like using match-exported: 'snake' vs going down the match-regex route was working (vs. the regex issues with *.test.tsx files and single word .ts files (utils.ts)). No errors with adding a testCamel.ts file in src or any errors produced when running yarn lint 😭 . Investigating...

rshen91:
Ok I think I figured out the regex! There are some jest config files in the overrides.

PR #552, opened 2020-02-14
docs: fix filterSeriesInTooltip predicate

markov00:
## Summary

This PR fix a small bug on storybook (story `BarChart - 2y1g`) where an example was using a wrong FilterPredicate on the         `filterSeriesInTooltip` props.

I've also took the opportunity of this PR to remove all the `getSpecId` and `getAxisId` calls as they are no longer needed

### Checklist


markov00:
> Can we also remove the function definitions and make this a breaking change with the tooltip logic? That would finish this once and for all.
> 
> https://github.com/elastic/elastic-charts/blob/2242fca6800995eaa8b7c4402cc8f82cad383c1c/src/utils/ids.ts#L1-L17

@nickofthyme  Sure will do, I will open up a PR for this



> LGTM, finally, has this day actually come where we remove all `getXXXId`??
> 
> Side note: I think we should have a react linting rule in place for string and boolean props, what do you think?
> 
> ```tsx
> // bad
> id={'bottom'}
> // good
> id="bottom"
> 
> // bad
> showLegend={true}
> // good
> showLegend
> ```

@rshen91 do you want to include also this eslint rule on your PR?

PR #553, opened 2020-02-14
refactor: decouple tooltip from XY chart

### LINKED ISSUE #246: Legend should not show values from the chart
**Describe the bug**
The legend shows either the right-most value from the chart or the currently-hovered series. It would make more sense to hide the value, giving more space for the legend or chart.

**To Reproduce**

This behavior is shown in all of the chart samples, including: https://elastic.github.io/elastic-charts/?selectedKind=Bar%20Chart&selectedStory=with%20axis%20and%20legend&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybook%2Fstories%2Fstories-panel

Jun-19-2019 16-33-37

**Expected behavior**
Because there is already the same information available elsewhere, I would expect the numeric value to be hidden on the legend.

**Checklist**
- ~[ ] every related Kibana issue is listed under `Kibana Cross Issues` list~
- ~[ ] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present~

markov00:
## Summary

This PR will decouple the tooltip from the XY chart to allow Partition and other chart type to use it.

**BREAKING CHANGES**
the `SeriesIdentifier` type is generalised into a simple object with two common values: specId and key. A specialised `XYChartSeriesIdentifier` extends now the base `SeriesIdentifier`
The `SettingsSpec` prop `showLegendDisplayValue` is renamed to `showLegendExtra`.
Previously `showLegendDisplayValue` was true by default always showing the legend extra
value. Now it becomes always hidden by default

close #246 this PR will hide the value by default, we will iterate over this to make this parameter customisable https://github.com/elastic/elastic-charts/issues/561

**Note for reviewers**
It's easier to follow the changes one commit a time, as they are self contained.

### Checklist


src/specs/settings.tsx
```diff
@@ -90,7 +90,10 @@ export interface SettingsSpec extends Spec {
   tooltip: TooltipType | TooltipProps;
   debug: boolean;
   legendPosition: Position;
-  showLegendDisplayValue: boolean;
+  /**
+   * Show an extra parameter on each legend item defined by the chart type
+   */
+  showLegendExtra: boolean;
```
nickofthyme: can we rename this to `hideLegendExtra` so the default is `false`? Then it would just be.

```tsx
<Settings hideLegendExtra />
```

nickofthyme: can we rename this to `hideLegendExtra` so the default is `false`? Then it would just be.

```tsx
<Settings hideLegendExtra />
```

src/chart_types/xy_chart/state/chart_state.test.ts
```diff
@@ -1028,21 +1034,27 @@ describe.skip('Chart Store', () => {
     store.annotationDimensions.set(rectAnnotationSpec.id, annotationDimensions);
 
     const highlightedTooltipValue: TooltipValue = {
-      name: 'foo',
+      label: 'foo',
```
nickofthyme: Can we pick `name` or `label` to be consistent? I am having trouble choosing between the two. If we are going to have a name prop where this value is derived from, this should be `name` not `label` right?

src/chart_types/xy_chart/utils/series.ts
```diff
@@ -45,21 +45,23 @@ export interface DataSeriesDatum<T = any> {
...
-export interface SeriesIdentifier {
+export type SeriesIdentifier = {
   specId: SpecId;
+  key: string;
+};
+
+export interface XYChartSeriesIdentifier extends SeriesIdentifier {
   yAccessor: string | number;
   splitAccessors: Map<string | number, string | number>; // does the map have a size vs making it optional
   seriesKeys: (string | number)[];
```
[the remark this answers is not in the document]
markov00: unfortunately not, there is no yAccessor (it's a valueAccessor at the moment and it's a function at the moment, no splitAccessors also)

src/specs/settings.tsx
markov00: I think we should have the extra only visible if the user explicitly enable them (https://github.com/elastic/elastic-charts/issues/246)

src/chart_types/xy_chart/state/chart_state.test.ts
markov00: Do you mean the `name` prop of the Series component?

src/specs/settings.tsx
nickofthyme: Ok in that case shouldn’t `showLegendExtra` default be `false`?

src/chart_types/xy_chart/state/chart_state.test.ts
nickofthyme: Correct, this values is determined from the `customSeriesLabel`, soon to be `name` prop.

https://github.com/elastic/elastic-charts/pull/539#discussion_r378160221

markov00: ok let me switch to `name` then

markov00: from our yesterday call, we will keep the `label` naming to make the tooltip a bit more abstract from the series. The Label should refer to the descriptive part of the value

PR #554, opened 2020-02-18
feat: expose Specs types and clean up annotation types

### LINKED ISSUE #547: Export Typescript definition for what a Spec is
For example, this definition is one that I have written to infer it externally, but it would be more convenient to have an official spec from the chart library itself.

```typescript
type InferPropType<T> = T extends React.FunctionComponent<infer P> ? P : T;
type SeriesSpec = InferPropType<typeof LineSeries> &
  InferPropType<typeof BarSeries> &
  InferPropType<typeof AreaSeries>;
```

markov00:
## Summary

This PR fix some wrongly configured annotation TS types and expose the Specs types as requested by #547
It also expose some more types that can be used by consumers that where buried inside our chart library.
This commit removes also the `getSpecId`, `getAnnotationId`, `getGroupId`, `getAxisId` getters in favour of just using a string

BREAKING CHANGE: The `getSpecId`, `getGroupId`, `getAxisId` and `getAnnotationId` are no more available, you can use directly a `string` for those ids.

close #547

### Checklist


src/index.ts
```diff
@@ -1,41 +1,20 @@
+// polyfill for Path2D canvas
 import 'path2d-polyfill';
-export * from './specs';
+
+export { ChartTypes } from './chart_types';
 export { Chart } from './components/chart';
 export { ChartSize, ChartSizeArray, ChartSizeObject } from './utils/chart_size';
+
+// DEPRECATED: we will remove these types soon
 export { SpecId, GroupId, AxisId, AnnotationId, getAxisId, getGroupId, getSpecId, getAnnotationId } from './utils/ids';
```
nickofthyme: I think you mean remove from export but I saw something in kibana that I started doing in the color picker PR which is for mapped values using `SeriesId` rather than string. It makes no difference to the output or consumer facing types, so we wouldn't need to export them. Something like...

```ts
// Good
type SeriesCollection = Map<SeriesId, `SeriesIdentifier>;

// Bad
type SeriesCollection = Map<string, `SeriesIdentifier>;
```

Any thoughts on this?

src/index.ts
```diff
@@ -1,41 +1,20 @@
...
-export { ScaleType } from './scales';
-export * from './utils/themes/theme';
-export { LIGHT_THEME } from './utils/themes/light_theme';
-export { DARK_THEME } from './utils/themes/dark_theme';
-export * from './utils/themes/theme_commons';
-export { RecursivePartial } from './utils/commons';
+
+// Everything related to the specs types and react-components
+export * from './specs';
```
nickofthyme: 👍

src/index.ts
```diff
@@ -46,3 +25,16 @@ export {
...
+export { ScaleType } from './scales';
+
+// theme
+export * from './utils/themes/theme';
+export * from './utils/themes/theme_commons';
+export { LIGHT_THEME } from './utils/themes/light_theme';
+export { DARK_THEME } from './utils/themes/dark_theme';
+
+// utilities
+export { RecursivePartial } from './utils/commons';
```
nickofthyme: maybe we could have a `utils/index` for all the utils we want to export.

src/index.ts
markov00: You are right, let's keep these

src/index.ts
markov00: I will address this in a different PR. Currently the Utils directory has more than utilities (also some types, the theme....it's a little mess there, I will try to move things around)

Issue #518, opened 2020-01-20
Partitioning visualizations - improvement ideas


monfera:
This is a set of improvements for the partitioning charts introduced with #493

1. consider adding either more structure to the config (like `linkLabel`, which is already nested) - or alternatively, rely on a flatter structure for composable configurability
2. review types for improvement opportunities
3. outside link label improvements (value display; length limitation; avoidance of truncation due to chart port size; consider multirow)
4. feed styling from Theme, for what's possible and suitable

wylieconlon:
I have found some problems/issues as well:

* Missing legend and tooltip https://github.com/elastic/elastic-charts/issues/518
* All of the configuration callback functions need specific type arguments, not `Datum`. For example, `fillColor` is a function called with the argument of type `ShapeTreeNode`, not of type `Datum`. This applies to the formatter functions too, like `valueFormatter`
* Multi-layer treemaps should support labels outside the rectangle, like this: https://www.d3-graph-gallery.com/graph/treemap_custom.html

wylieconlon:
A new piece of feedback from  https://github.com/elastic/kibana/pull/55477:

> Not reporting the percentages and only showing the values is a concern. A pie chart is about percentages. In addition, it is helpful to have the actual value but I'm afraid without percentages it loses meaning. Ideally, we show the percentage on the slices by default and show the actual number on the tooltip.

Converting to percentages is currently done by `visLib`, so this seems like a responsibility of elastic-charts as well.

monfera:
Thanks @wylieconlon, good discussion here and in the referred Lens item.

On some of the items:

- I'll add the percentages around tomorrow (CD today) cc @markov00 
- missing tooltip was what I meant by item 9; it's in progress, Marco took on the more laborious part of uncoupling the tooltip from Cartesians, and likely the legends need to be uncoupled too
- I'll check the type, which changed indeed
- treemap labeling:
  - it's planned that the multilayer treemap gets labels in the grooves; in the meantime, 
  - the non-leaf layer legends should be disabled (returning the empty string), as, especially without formatting like in the storybook items, it doesn't look good
  - as a substitute, a less optimal multirow text can be used for integration testing, ie. each leaf recapitulates the higher group names too, like some Tableau example shows (Jan 31 in the charts channel)
[...]

PR #558, opened 2020-02-19
feat: percentage display in partitioning charts

monfera:
## Summary

Adds the option of using the ratio rather than the value in `data` that's the basis for aggregation.

The ratio is computed as the value of the sector or treemap square divided by the total sum of the value.

Client code may use the ratio to display the ratio, or a percentage, instead of the sector / rectangle value.

image
image

One item in https://github.com/elastic/elastic-charts/issues/518

### Checklist


markov00:
I've scanned the code and LGTM. I just want to make a small comment on the API:
your implementation impose to import a function, the `percentageValueGetter` and add them to the `valueGetter` props of the `<Partition/>` component. I think it's fine, but what if we provide a simpler version also? like allow the user to specify a string/constant or a function like

`type ValueGetter = 'value' | 'percentage' |  (node: ShapeTreeNode) => number;`

This will simplify the API a bit and will make it a bit more readable in few cases

monfera:
image

src/chart_types/partition_chart/state/selectors/tooltip.ts
```diff
@@ -73,7 +76,7 @@ export const getTooltipInfoSelector = createCachedSelector(
           specId: pieSpec.id,
           key: pieSpec.id,
         },
-        value: valueFormatter(valueGetterFun(shape)),
+        value: valueFormatter(primaryValueGetterFun(shape)) + ' (' + percentFormatter(percentValueGetter(shape)) + ')',
```
markov00: nit: can you use a template string here?
```suggestion
        value: `${valueFormatter(primaryValueGetterFun(shape))} (${percentFormatter(percentValueGetter(shape))})`,
```

src/chart_types/partition_chart/layout/config/config.ts
```diff
@@ -28,18 +28,18 @@ export function ratioValueGetter(node: ShapeTreeNode): number {
...
+    ? d.toExponential(Math.min(2, Math.max(0, significantDigitCount(d) - 1)))
+    : d.toLocaleString(void 0, {
+        maximumSignificantDigits: 4,
+        maximumFractionDigits: 3,
+        useGrouping: true,
+      });
+}
+
+export function percentFormatter(d: number): string {
+  return Math.round(d) + '%';
```
markov00: nit: template string?
```suggestion
  return `${Math.round(d)} %`;
```

markov00:
Hey @monfera I've pushed a change to add the `percentFormatter` as part of the spec. This because Kibana has the freedom to change the percentage format depending on the Advanced settings. you can see the changes here: https://github.com/elastic/elastic-charts/pull/558/commits/ae26259862f703f43628c789c39bcd981e5b4eff

I've also removed a small portion of code (the one related to the indices) on the tooltip selector, not sure what was the original requirements for these indices values, see: https://github.com/elastic/elastic-charts/pull/558/commits/1843a16e963ca95ae73ef13681b96abca159d7ac

monfera:
Thanks @marco the new commits look great. We can reconsider the [`datumIndices`](https://github.com/elastic/elastic-charts/commit/6bf9a69b12d3075330a5728b7bdb4443e6244985#diff-ca54637bd0932e9f3ebba47e2b6007e9R87-R95) whenever the need comes back

markov00:
@monfera sure, let me copy the code here also:
```
    const datumIndices = new Set();
    pickedShapes.forEach((shape) => {
      const node = shape.parent;
      const shapeNode = node.children.find(([key]) => key === shape.dataName);
      if (shapeNode) {
        const indices = shapeNode[1][INPUT_KEY] || [];
        indices.forEach((i) => datumIndices.add(i));
      }
    });
```

Issue #502, opened 2019-12-19
Scatter chart with marker shape and size


wylieconlon:
Scatter charts and heatmaps (https://github.com/elastic/elastic-charts/issues/501) are complementary. Scatter charts use continuous scales with discrete data, while heatmaps use discrete scales and data.

Scatter charts should support:

* X and Y positioning
* Filled circle and outline markers at the least
* (optional) Markers that take a dynamic shape for categorical data
* Marker size based on value
* Tooltip labeling

<img width="635" alt="Screenshot 2019-12-19 16 17 19" src="https://user-images.githubusercontent.com/666475/71210521-12cd7d00-227b-11ea-8c20-1160ace5b749.png">
<img width="870" alt="Screenshot 2019-12-19 16 17 24" src="https://user-images.githubusercontent.com/666475/71210523-12cd7d00-227b-11ea-83a2-1d74d499db27.png">
<img width="325" alt="Screenshot 2019-12-19 16 17 30" src="https://user-images.githubusercontent.com/666475/71210524-12cd7d00-227b-11ea-8e4f-a2800ee55478.png">

**Kibana Cross Issues**
https://github.com/elastic/kibana/issues/17076
https://github.com/elastic/kibana/issues/20976
https://github.com/elastic/kibana/issues/36541

**Checklist**

- [x] this request is checked against already exist requests
- [x] every related Kibana issue is listed under `Kibana Cross Issues` list
- [x] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

nickofthyme:
More examples:
- https://vega.github.io/vega-lite/examples/point_2d.html
- https://vega.github.io/vega-lite/examples/tick_dot.html

wylieconlon:
An important element of these example scatter charts is that the legend includes a shape and color indicator without showing every individual value. Supporting this type of legend seems like a nice-to-have.

AlonaNadler:
@markov00 APM team plans to use scatterplot in 7.11 https://github.com/elastic/kibana/issues/81720
Based on their needs, do you think what Elastic charts currently has can help ?

markov00:
@AlonaNadler we can probably take that effort to add a simple scatterplot, but it will just be a throw-away implementation as we can't have a fully-featured implementation of a real scatterplot in few weeks. We can, as we have done for the heatmap, support APM on building that directly in elastic-charts, building together a good starting point

sqren:
[...]

PR #559, opened 2020-02-19
feat(series): markSizeAccessor and BubbleSeries (alpha)

### LINKED ISSUE #533: Dot size based on point data
Allow dot sizing based on data as it is currently done in kibana.

Ideally this is through a `dotAccessor` to pull from data rather than using the `pointStyleAccessor` function prop.

image

Related to https://github.com/elastic/elastic-charts/issues/66

**Checklist**

nickofthyme:
## Summary

Closes #533

Allow for a `markSizeAccessor` to render as radius size for point values. Copied Kibana logic for dotRadio using a step function, but this could be replaced with our own continuous sizing function.

Add `BubbleSeries` spec. This spec only draws points with no lines. The unique aspect of this type is that it uses the [`d3-delaunay`](https://github.com/d3/d3-delaunay) spatial indexing approach. Currently, this does not work well with other chart spec types as the tooltip will show all values in range even on different x values. This is _NOT_ a scatter plot, this type will come in another PR in the future.

Tasks:
- [x] Clip dot radius at boundaries.
- [x] Fix highlighter to select dot in range from left or right (up/down works)
- [x] Sort dots to render the largest radii in the back.
- [x] Fix mysterious x-axis disappearance 👽 (https://github.com/elastic/elastic-charts/pull/563)

### Bubble
Screen Recording 2020-03-31 at 11 17 AM

### Area
Screen Recording 2020-03-31 at 11 15 AM

### Line
Screen Recording 2020-03-31 at 11 16 AM

### Checklist


wylieconlon:
This looks like a first step towards a real scatter chart, right? https://github.com/elastic/elastic-charts/issues/502

wylieconlon:
Instead of calling this `dot`, can you call it something like `markSize`? The mark in this case is a dot, but like I requested in the scatter chart issue it would be nice to support other shapes.

src/chart_types/xy_chart/rendering/rendering.ts
```diff
@@ -42,20 +41,11 @@ import {
...
-) {
-  const existing = mutableGeometriesIndex.get(key);
-  const upsertGeometry: IndexedGeometry[] = Array.isArray(geometry) ? geometry : [geometry];
-  if (existing === undefined) {
-    mutableGeometriesIndex.set(key, upsertGeometry);
-  } else {
-    mutableGeometriesIndex.set(key, [...upsertGeometry, ...existing]);
-  }
-}
+export const DEFAULT_HIGHLIGHT_PADDING = 10;
```
markov00, line 48: Also this can go to the theme

src/chart_types/xy_chart/rendering/rendering.ts
```diff
@@ -104,22 +94,60 @@ export function getBarStyleOverrides(
...
+/**
+ * Get radius function form ratio and min/max dot sixe
+ *
+ * @todo add continuous/non-stepped function
+ *
+ * @param  {Datum[]} radii
+ * @param  {number} lineWidth
+ * @param  {number=50} radiusRatio - 1 to 100
+ */
+function getRadiusFn(data: DataSeriesDatum[], lineWidth: number, radiusRatio: number = 50) {
```
markov00, line 106: can you add a return type please?

src/chart_types/xy_chart/rendering/rendering.ts
```diff
@@ -104,22 +94,60 @@ export function getBarStyleOverrides(
...
+      dot === null
+        ? acc
+        : {
+            min: Math.min(acc.min, dot / 2),
+            max: Math.max(acc.max, dot / 2),
+          },
+    { min: Infinity, max: -Infinity },
+  );
+  const radiusStep = (max - min || max * 100) / Math.pow(radiusRatio, 2);
+  return function getRadius(dot: number | null, defaultRadius = 0): number {
```
markov00, line 121: hey hey, I saw that deep in your heart you prefer named functions instead of anonymous ones ❤️

src/chart_types/xy_chart/rendering/rendering.ts
```diff
@@ -104,22 +94,60 @@ export function getBarStyleOverrides(
...
+  radiusRatio?: number,
 ): {
   pointGeometries: PointGeometry[];
-  indexedGeometries: Map<any, IndexedGeometry[]>;
+  indexedGeometryMap: IndexedGeometryMap;
 } {
-  const indexedGeometries: Map<any, IndexedGeometry[]> = new Map();
+  const indexedGeometryMap = new IndexedGeometryMap();
   const isLogScale = isLogarithmicScale(yScale);
+  const getRadius = getRadiusFn(dataSeries.data, lineWidth, radiusRatio);
```
markov00, line 148: can we avoid running a cycle through the data here if we don't have a markAccessor/dotAccessor and just specify a default radius?

src/chart_types/xy_chart/state/utils.ts
```diff
@@ -710,3 +687,7 @@ export function isAllSeriesDeselected(legendItems: Map<string, LegendItem>): boo
   }
   return true;
 }
+
+export function getDistance(a: Point, b: Point) {
+  return Math.sqrt((b.x - a.x) ** 2 + (b.y - a.y) ** 2);
```
markov00, line 692: the `**` is not supported in IE, could you please check if this get correctly replaced by `Math.pow`?

src/chart_types/xy_chart/utils/indexed_geometry_linear_map.ts
```diff
@@ -0,0 +1,50 @@
...
+
+import { IndexedGeometry } from '../../../utils/geometry';
+
+export class IndexedGeometryLinearMap {
+  private map = new Map<string | number, IndexedGeometry[]>();
+
+  set(geometry: IndexedGeometry) {
+    const { x } = geometry.value;
+    const existing = this.map.get(x);
+    const upsertGeometry: IndexedGeometry[] = Array.isArray(geometry) ? geometry : [geometry];
```
markov00, line 27: If geometry is of type `IndexedGeometry`, I don't see the case where this geometry can be an array.
Also the statement at line 25 `const { x } = geometry.value;` will generate an error if `geometry` is an array

src/chart_types/xy_chart/utils/indexed_geometry_linear_map.ts
```diff
@@ -0,0 +1,50 @@
...
+      this.map.set(x, [...upsertGeometry, ...existing]);
+    }
+  }
+
+  getMergeData() {
+    return [...this.map.values()];
+  }
+
+  keys(): Array<number> {
+    return [...this.map.keys()].filter((key): key is number => typeof key === 'number');
```
markov00, line 40: why this keys returned here is numeric only?

src/chart_types/xy_chart/utils/indexed_geometry_map.ts
```diff
@@ -0,0 +1,93 @@
...
+  spatial: 'spatial' as 'spatial',
+});
+export type GeometryType = $Values<typeof GeometryType>;
+
+export class IndexedGeometryMap {
+  private linearMap = new IndexedGeometryLinearMap();
+  private spatialMap = new IndexedGeometrySpatialMap();
+
+  /**
+   * Renterns voronoi instance to render voronoi grid
```
markov00, line 38: typo

src/chart_types/xy_chart/utils/indexed_geometry_map.ts
```diff
@@ -0,0 +1,93 @@
...
+   *
+   * @param bounds
+   */
+  voronoi(bounds?: Delaunay.Bounds) {
```
markov00, line 42: let's call it triangulation or something similar.
It's a little bit strange that the function is called Voronoi but takes a Delaunay bounds type

src/chart_types/xy_chart/utils/indexed_geometry_map.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  keys(): Array<number> {
+    return [...this.linearMap.keys(), ...this.spatialMap.keys()];
+  }
+
+  set(geometry: IndexedGeometry, type: GeometryType = GeometryType.linear) {
+    if (type === GeometryType.spatial) {
+      if (!isPointGeometry(geometry)) {
+        throw new Error('Spatial geometry must be PointGeometry');
```
markov00, line 53: can we just "forgot" about this, without surfacing an error? or trying to typing the possible throw

src/chart_types/xy_chart/utils/series.ts
```diff
@@ -44,8 +44,10 @@ export interface RawDataSeriesDatum<T = any> {
   x: number | string;
   /** the main y metric */
   y1: number | null;
-  /** the optional y0 metric, used for bars or area with a lower bound */
+  /** the optional y0 metric, used for bars and area with a lower bound */
   y0?: number | null;
+  /** the optional dot metric, used for lines and area series */
+  dot: number | null;
```
markov00, line 50: mark

src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -338,6 +338,8 @@ export interface SeriesAccessors {
   splitSeriesAccessors?: Accessor[];
   /** An array of fields thats indicates the stack membership */
   stackAccessors?: Accessor[];
+  /** Field name of dot size metric on `Datum` */
+  dotAccessor: Accessor | AccessorFn;
```
markov00, line 342: let's callit `markAccessor` or `markSizeAccessor`

src/scales/scale_band.ts
```diff
@@ -68,28 +68,45 @@ export class ScaleBand implements Scale {
...
+    return scaleValue;
   }
-  pureScale(value: any) {
-    return this.d3Scale(value);
+
+  pureScale(value: string | number) {
+    const scaleValue = this.d3Scale(value);
+
+    if (typeof scaleValue !== 'number') {
+      throw new Error(`The value (${value}) was not scalable`);
```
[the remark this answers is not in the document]
nickofthyme: I think these will never fire but you are correct. In the past, this type was any that's why I added this. I'm changing it to return 0 as a fallback. Do you think that would cause any issues?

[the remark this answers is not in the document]
nickofthyme: 46d8f178

src/specs/settings.tsx
```diff
@@ -155,6 +155,11 @@ export interface LegendColorPickerProps {
 }
 export type LegendColorPicker = ComponentType<LegendColorPickerProps>;
 
+/**
+ * Buffer between cursor and point to trigger interaction
+ */
+export type PointBuffer = number | ((radius: number) => number);
```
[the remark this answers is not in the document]
nickofthyme: Do you have any other names for this? I was thinking padding but it's not really padding.

src/chart_types/xy_chart/rendering/rendering.ts
nickofthyme: lol 😝 you caught me. I like it in this case because when using closure.

src/chart_types/xy_chart/utils/series.ts
nickofthyme: 7bbeba86

src/chart_types/xy_chart/utils/specs.ts
nickofthyme: 7bbeba86

src/utils/themes/theme.ts
```diff
@@ -175,6 +175,12 @@ export interface Theme {
   colors: ColorConfig;
   legend: LegendStyle;
   crosshair: CrosshairStyle;
+  /**
+   * Used to scale radius with `dotAccessor`
```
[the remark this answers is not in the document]
nickofthyme: 7bbeba86

src/utils/geometry.ts
```diff
@@ -34,6 +34,7 @@ export type BandedAccessorType = $Values<typeof BandedAccessorType>;
 export interface GeometryValue {
   y: any;
   x: any;
+  dot: number | null;
```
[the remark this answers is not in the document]
nickofthyme: 7bbeba86

src/specs/settings.tsx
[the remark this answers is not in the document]
nickofthyme: 7bbeba86

src/chart_types/xy_chart/rendering/rendering.ts
nickofthyme: I don't know where to put it in the theme. Any ideas? I was thinking this is just related to the highlighter and we don't have a theme for that yet. I could add it to the `pointStyles` but it seems out of place there.

package.json
```diff
@@ -163,11 +163,13 @@
     "webpack-dev-server": "^3.3.1"
   },
   "dependencies": {
+    "@types/d3-delaunay": "^4.1.0",
```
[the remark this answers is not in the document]
nickofthyme: cd88f079

src/chart_types/xy_chart/rendering/rendering.ts
nickofthyme: Normally I would agree but I like the implicit type when returning a function. What do you think?

src/chart_types/xy_chart/rendering/rendering.ts
nickofthyme: d3f1e2bf97cffb0c02892397d4c097755df8f901

src/chart_types/xy_chart/state/utils.ts
nickofthyme: b1d48ffd

src/chart_types/xy_chart/utils/indexed_geometry_linear_map.ts
nickofthyme: 7fc99508

This was a carryover from when the classes were combined.

src/chart_types/xy_chart/utils/indexed_geometry_linear_map.ts
nickofthyme: Because of this...

https://github.com/elastic/elastic-charts/blob/d3a691b2cf5ec7b81dd1ea724f2518d1f08c1fe5/src/chart_types/xy_chart/state/selectors/get_geometries_index_keys.ts#L25-L30

where `compareByValueAsc` is defined as...

https://github.com/elastic/elastic-charts/blob/d3a691b2cf5ec7b81dd1ea724f2518d1f08c1fe5/src/utils/commons.ts#L44-L46

src/scales/scale_band.ts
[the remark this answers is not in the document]
markov00: I think this will fire if the value passed to scale is not part of the range, but I'm not 100% sure.
Instead of return a valid value, what if we return a non valid value, a null for example. This will help us catching this error on the caller

src/chart_types/xy_chart/rendering/rendering.ts
markov00: Ok np, let's leave it here for the moment

src/chart_types/xy_chart/rendering/rendering.ts
markov00: It's just a matter of being always clear of what the function needs to return and it's specially useful when we need to maintain that function: adding a different return type will just update the implicit return types of the function and maybe something bad can slip through. An explicit return type make it difficult to propagate errors when fix/change/maintain  this function

src/chart_types/xy_chart/utils/indexed_geometry_linear_map.ts
markov00: is this a wrong has for the commit? I can't find it

src/chart_types/xy_chart/utils/indexed_geometry_linear_map.ts
markov00: awesome thanks

src/chart_types/xy_chart/rendering/rendering.ts
nickofthyme: Ok I was thinking the same.

src/chart_types/xy_chart/utils/indexed_geometry_linear_map.ts
nickofthyme: Sorry forgot to push this commit.

src/chart_types/xy_chart/utils/indexed_geometry_map.ts
nickofthyme: b32a7e63

src/chart_types/xy_chart/utils/indexed_geometry_map.ts
nickofthyme: b32a7e63

src/chart_types/xy_chart/utils/indexed_geometry_map.ts
nickofthyme: 8ff0c27f

src/chart_types/xy_chart/utils/indexed_geometry_linear_map.ts
nickofthyme: Do you know why this is? seems very strange no?

markov00: Yes, so basically the geometryIndexKeys array is only used by the continuous `invertWithStep` scale function. This scale function takes the value and an array of number[]. With the band scale, this array is basically ignored

src/scales/scale_band.ts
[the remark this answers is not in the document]
nickofthyme: 9f0b7ab4d4478f5f27f467d6c020c551c97df491

src/chart_types/xy_chart/rendering/rendering.ts
nickofthyme: f6a3b9f7

src/chart_types/xy_chart/utils/indexed_geometry_linear_map.ts
nickofthyme: Oh I see thanks for the explanation. I think there are a lot of places like that in our code that are typed differently based on a _shared_ type. Especially with `Scales`. I wonder if typescript has a better way to handle these cases.

src/chart_types/xy_chart/utils/indexed_geometry_spatial_map.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (x === null) {
+      return [];
+    }
+
+    const elements = [];
+    if (this.map !== null) {
+      const index = this.map.find(point.x, point.y, this.searchStartIndex);
+
+      // Set next starting search index for faster lookup
+      this.searchStartIndex = index;
```
markov00, line 89: Why do you need to store locally this index?

src/specs/settings.tsx
```diff
@@ -109,6 +109,10 @@ export interface TooltipValue {
    * The value to display
    */
   value: any;
+  /**
+   * The mark value to display
+   */
+  markValue?: any;
```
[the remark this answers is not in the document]
nickofthyme: Agreed

src/chart_types/xy_chart/utils/indexed_geometry_spatial_map.ts
nickofthyme: I think this will boost the lookup performance. The [`find`](https://github.com/d3/d3-delaunay#delaunay_find) method has an optional third argument as index to start searching for the point. By default it's `0` but I think if we track the cursor as it moves we can pass it the last index to make the lookup faster.

>delaunay.find(x, y[, I]) <>
>
>Returns the index of the input point that is closest to the specified point ⟨x, y⟩. The search is started at the specified point i. If i is not specified, it defaults to zero.

src/chart_types/xy_chart/specs/bubble_series.tsx
```diff
@@ -0,0 +1,58 @@
...
+  hideInLegend: false,
+  yScaleToDataExtent: false,
+};
+type SpecRequiredProps = Pick<BubbleSeriesSpec, 'id' | 'data'>;
+type SpecOptionalProps = Partial<Omit<BubbleSeriesSpec, 'chartType' | 'specType' | 'seriesType' | 'id' | 'data'>>;
+
+/**
+ * @note This series type uses a spatial index that is incompatible with other series types.
+ * Nothing will break, just the tooltip values will be combined across linear and spatial values.
+ * Please avoid mixed series types until this is resolved.
```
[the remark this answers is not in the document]
nickofthyme: 70772d1c211a0a9c04fe86417a98568fad60cfc7

src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -443,6 +446,19 @@ export type LineSeriesSpec = BasicSeriesSpec &
     fit?: Exclude<Fit, 'explicit'> | FitConfig;
   };
 
+/**
+ * This spec describe the dataset configuration used to display a line series.
+ */
```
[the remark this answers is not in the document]
nickofthyme: 70772d1c211a0a9c04fe86417a98568fad60cfc7

src/chart_types/xy_chart/state/utils.ts
```diff
@@ -739,3 +762,51 @@ export function isAllSeriesDeselected(legendItems: Map<string, LegendItem>): boo
...
   return true;
 }
+
+/** @internal */
+export function getDistance(a: Point, b: Point) {
+  return Math.sqrt(Math.pow(b.x - a.x, 2) + Math.pow(b.y - a.y, 2));
+}
+
+/** @internal */
+export function strigifyValue(value?: PrimitiveValue): string | number {
```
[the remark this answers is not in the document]
nickofthyme: 5c45f809e73cd6e2c5fe30a88c10cfa68df869f4

src/chart_types/xy_chart/renderer/canvas/bubbles.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import { getGeometryStateStyle } from '../../rendering/rendering';
+import { BubbleGeometry } from '../../../../utils/geometry';
+import { SharedGeometryStateStyle } from '../../../../utils/themes/theme';
+import { LegendItem } from '../../legend/legend';
+import { withContext, withClip } from '../../../../renderers/canvas';
+import { renderPoints } from './points';
+import { Rect } from '../../../../geoms/types';
+
+interface LineGeometriesDataProps {
```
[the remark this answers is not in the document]
nickofthyme: 5c45f809e73cd6e2c5fe30a88c10cfa68df869f4

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
```diff
@@ -223,6 +227,7 @@ const mapStateToProps = (state: GlobalChartState): ReactiveChartStateProps => {
     isChartEmpty: isChartEmptySelector(state),
     debug: getSettingsSpecSelector(state).debug,
     geometries: computeSeriesGeometriesSelector(state).geometries,
+    geometriesIndex: computeSeriesGeometriesSelector(state).geometriesIndex,
```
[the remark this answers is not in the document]
nickofthyme: 5c45f809e73cd6e2c5fe30a88c10cfa68df869f4

nickofthyme:
I changed the mixed behavior to only use the linear index in that case (70772d1c211a0a9c04fe86417a98568fad60cfc7). Lmk if that elevates some of your concerns.

Also, I checked kibana and it looks like the dots are always behind separate line charts no matter the config stacking order.

Screen Recording 2020-04-13 at 11 36 AM

src/chart_types/xy_chart/utils/series.ts
```diff
@@ -44,8 +44,10 @@ export interface RawDataSeriesDatum<T = any> {
   x: number | string;
   /** the main y metric */
   y1: number | null;
-  /** the optional y0 metric, used for bars or area with a lower bound */
+  /** the optional y0 metric, used for bars and area with a lower bound */
   y0?: number | null;
+  /** the optional mark metric, used for lines and area series */
+  mark: number | null;
```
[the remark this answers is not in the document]
nickofthyme: That is fine here in the `RawDataSeriesDatum` because I see `mark` as similar to the `y0`.

But in the `DataSeriesDatum` I think it's best to keep it as `number | null`. I don't like allowing `undefined` for _processed_ values. 

https://github.com/elastic/elastic-charts/blob/70772d1c211a0a9c04fe86417a98568fad60cfc7/src/chart_types/xy_chart/utils/series.ts#L55-L72

Additionally, the `GeometryValues` is where why most of the tests are failing.

https://github.com/elastic/elastic-charts/blob/70772d1c211a0a9c04fe86417a98568fad60cfc7/src/utils/geometry.ts#L34-L39

Ideally, the tests would have static mock values where this change would only be one line change to fix all the broken tests.

Overall, I agree not cluttering up the tests but I think the type should still be `number | null` if you disagree I can change this account for the test changes in the future.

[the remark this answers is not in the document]
markov00: agree, thanks for the explaination

src/chart_types/xy_chart/utils/indexed_geometry_spatial_map.ts
markov00: thanks

src/chart_types/xy_chart/rendering/rendering.test.ts
```diff
@@ -106,10 +108,10 @@ describe('Rendering utils', () => {
     expect(isPointOnGeometry(10, 10, geometry)).toBe(true);
     expect(isPointOnGeometry(0, 10, geometry)).toBe(true);
     expect(isPointOnGeometry(10, 0, geometry)).toBe(true);
-    expect(isPointOnGeometry(11, 11, geometry)).toBe(false);
+    // expect(isPointOnGeometry(11, 11, geometry)).toBe(false);
     expect(isPointOnGeometry(-10, 0, geometry)).toBe(true);
-    expect(isPointOnGeometry(-11, 0, geometry)).toBe(false);
-    expect(isPointOnGeometry(11, 11, geometry)).toBe(false);
+    // expect(isPointOnGeometry(-11, 0, geometry)).toBe(false);
+    // expect(isPointOnGeometry(11, 11, geometry)).toBe(false);
```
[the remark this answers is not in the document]
nickofthyme: 56649d03

nickofthyme:
@markov00 I agree, we really need to update the tooltip to account for more cases.

PR #560, opened 2020-02-20
chore: tighter partition types

monfera:
## Summary

Tightens up partitioning chart types, as several was, in effect, `any`. See point 3 in https://github.com/elastic/elastic-charts/issues/518#issue-552283560

It's a breaking change as it may need that client libraries update types, even though there's no runtime change.

BREAKING CHANGES:
The `valueFormatter` prop function now receive  only a `number` instead of `any`. 
In the `layers` props:
- the `nodeLabel` is now typed as `(d: PrimitiveValue) => String(d),` where `PrimitiveValue` is `string | number | null`.
- the `fillColor` of a `shape` has a tighter type: `(d: ShapeTreeNode, index: number, array: HierarchyOfArrays) => string;`

### Checklist

- [ ] ~~Any consumer-facing exports were added to `src/index.ts` (and stories only import from `../src` except for test data & storybook)~~


src/chart_types/xy_chart/utils/series.test.ts
```diff
@@ -665,23 +665,28 @@ describe('Series', () => {
...
+    expect(datum).toBeDefined();
+    expect(datum!.y1).toBe(1);
+    expect(datum!.y0).toBe(2);
 
     datum = cleanDatum([0, 'invalid', 'invalid'], 0, 1, 2);
-    expect(datum.y1).toBe(null);
-    expect(datum.y0).toBe(null);
+    expect(datum).toBeDefined();
+    expect(datum!.y1).toBe(null);
+    expect(datum!.y0).toBe(null);
```
nickofthyme: Using a `?` would be the same here. I don't like `!`

```ts
expect(datum?.y1).toBe(1);
```

markov00: Sure, let me change that

markov00: done here e58c787

PR #571, opened 2020-02-24
refactor(SeriesSpec): customSeriesColors to color on SeriesSpec

nickofthyme:
## Summary

Refactor naming of `customSeriesColors` to `color` to be inline with new `name` prop on `SeriesSpec`. The `CustomSeriesColors` type  (which was `CustomSeriesColors`) now accepts a single color as string.

BREAKING CHANGE
`customSeriesColors` prop on `SeriesSpec` is now `color`. The `CustomSeriesColors` type is  replaced with `SeriesColorAccessor`.

```ts
type SeriesColorAccessor = string | SeriesColorsArray | SeriesColorAccessorFn;
```

### Checklist


src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -264,7 +264,7 @@ export interface SeriesSpec extends Spec {
   /** The type of series you are looking to render */
   seriesType: SeriesTypes;
   /** Set colors for specific series */
-  customSeriesColors?: CustomSeriesColors;
+  color?: CustomSeriesColors;
```
markov00: shall we change this type into  `SeriesColors`  or `SeriesColorAccessor` to align it to the SeriesNameAccessor?

nickofthyme: Yeah good idea. I was just trying to avoid type name changes but that's a good idea.

PR #572, opened 2020-02-24
test: update VRT baselines with fixed random generators

markov00:
## Summary

Some stories use the `getRandomNumber()` function to generate data or other variables. Unfortunately if you change the order of execution of the test, this can cause the VRT to fail.
This PR change the instantiation of the generator seed to be instantiated on each VRT story to keep always the same seed status.

### Checklist


markov00:
jenkins test this

PR #573, opened 2020-02-25
fix: missing naming type exports

nickofthyme:
## Summary

Add missing naming types to type export

### Checklist


markov00:
cla/check

PR #574, opened 2020-02-25
feat: allow function x accessors

### LINKED ISSUE #551: xAccessor as a function
Currently, all accessors are a `string` or `number` to be used as lookup keys on an `array` or `object` (aka the `Datum`)

```ts
const accessor = 'x';
const datum = {
  x: 'a',
};
const x = datum[accessor];
```

This proposal would introduce the idea of the `xAccessor` as being a function such as...

```ts
const accessor = (d: Datum) => d.x ?? '_all';
const datum = {
  x: 'a',
};
const x = accessor(datum);
```

nickofthyme:
## Summary

closes #551

Allow functional accessor for `xAccessor` to lookup values from datum.

```tsx
<AreaSeries
  xAccessor={(d) => d.x}
/>
```

### Checklist


src/utils/accessor.ts
```diff
@@ -39,3 +39,17 @@ export function getAccessorFormatLabel(accessor: AccessorFormat, label: string):
...
   return accessor(label);
 }
+
+/**
+ * Helper function to get accessor value from string, number or function
+ *
+ * @param  {Datum} datum
+ * @param  {AccessorString|AccessorFn} accessor
+ */
+export function getAccessorValue(datum: Datum, accessor: AccessorString | AccessorFn) {
```
markov00: since we are already there, what if we rename the `AccessorString` type to`AccessorObjectKey | AccessorArrayIndex` or something similar? Because `AccessorString` is no more a valid as name

PR #578, opened 2020-03-04
feat(partition): add element click, over, out events

markov00:
## Summary

This commit adds the element click event to the Partition chart, returning a groupByRollup value for each configured chart layer and the associated value.

The `ElementClickListener` is now defined with two possible types, one for each chart type:
```ts
type XYChartElementEvent = [GeometryValue, XYChartSeriesIdentifier];
type PartitionElementEvent = [Array<LayerValue>, SeriesIdentifier];
type ElementClickListener = (elements: Array<XYChartElementEvent | PartitionElementEvent>) => void;
```
nothing change for the XYChart, but I've added a new type specific for the Partition chart.
This type is composed by the `SeriesIdentifiers` of the Pie/Sunburst and a set of `LayerValue`s on per each defined layer in the `layers` props of a `Partition` component.

A layer value is an object defined as the following:
```ts
interface LayerValue {
  groupByRollup: PrimitiveValue;
  value: number;
}
```
The `groupByRollup` property returns group by value relative to the slice of the specific layer.
The `value` instead returns the numeric value used to represent percentage as a whole of that specific slice.

Screenshot 2020-03-04 at 10 25 02

In this example, clicking on the first slice on the top/right (`source: CN 33` from the parent layer `dest: US 77`) the object passed to the listener is:

```json5
[
  [
    [
      { groupByRollup: 'US', value: '77'}, // the first layer (the parent layer)
      { groupByRollup: 'CN', value: '33'}, // the outermost layer
    ],
    {
       specId: 'thePieSpecId',
       key:'thePieSpecIdKey'
    }
  ],
// ... there could be multiple slices clicked at the same time, specially on very small slices, so we have left the possibility to return more than one 'PartitionElementEvent'
]
```

### Checklist


src/specs/settings.tsx
```diff
@@ -31,9 +31,18 @@ import { XYChartSeriesIdentifier, SeriesIdentifier } from '../chart_types/xy_cha
...
 
-export type ElementClickListener = (elements: Array<[GeometryValue, XYChartSeriesIdentifier]>) => void;
-export type ElementOverListener = (elements: Array<[GeometryValue, XYChartSeriesIdentifier]>) => void;
+export interface LayerValue {
+  groupByRollup: PrimitiveValue;
+  value: number;
+}
+
+export type XYChartElementEvent = [GeometryValue, XYChartSeriesIdentifier];
+export type PartitionElementEvent = [Array<LayerValue>, SeriesIdentifier];
```
nickofthyme: Is the order of the `LayerValue[]` always hierarchical? If not it should be and would be good to note that.

A similar thought is what if this was recursive instead of an object? Something like...

```ts
export interface LayerValue {
  groupByRollup: PrimitiveValue;
  value: number;
  child?: LayerValue;
}
```

Just a thought, not saying to change it. 🤔

src/chart_types/partition_chart/state/selectors/on_element_over_caller.ts
```diff
@@ -0,0 +1,95 @@
...
+          prevPickedShapes = nextPickedShapes;
+        },
+      )({
+        keySelector: getChartIdSelector,
+      });
+    }
+    if (selector) {
+      selector(state);
+    }
+  };
```
nickofthyme: Looks like there is some common logic between these event listers might be good to provide a factory function at some point. Looks fine for now.

src/chart_types/partition_chart/layout/types/config_types.ts
```diff
@@ -22,8 +22,8 @@ import { $Values as Values } from 'utility-types';
 import { Color, ValueFormatter } from '../../../../utils/commons';
 
 export const PartitionLayout = Object.freeze({
-  sunburst: 'sunburst',
-  treemap: 'treemap',
+  sunburst: 'sunburst' as 'sunburst',
```
nickofthyme: 😃

src/specs/settings.tsx
markov00: It's hierarchical but I'd configured them in the same way you are configuring the hierarchy in the `<Partition />` `layers` prop

src/chart_types/partition_chart/state/selectors/on_element_over_caller.ts
markov00: You are perfectly fine, I will I've already cleaned up a bit this code, I will clean up more in a future PR

Issue #238, opened 2019-06-12
Change value labels colors depending on background


markov00:
**Is your feature request related to a problem? Please describe.**
A value label on a bar chart can be colored with a color that is good for some bars but can be a problem with some other backgrounds.
For example if I'm using a white text color for the value label, but the label fall outside the bar the label will disappear.
The same with a black text color and a dark background.

**To Reproduce**
1. Go to https://elastic.github.io/elastic-charts/?selectedKind=Bar%20Chart&selectedStory=with%20value%20label&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs
2. Click on Knobs, tick show labels, change the color to white
3. The label is there but disappeared beacuse of the white background

<img width="831" alt="Screenshot 2019-06-12 at 18 41 39" src="https://user-images.githubusercontent.com/1421091/59370087-14edd680-8d42-11e9-9d03-6bc541caa6c5.png">
<img width="858" alt="Screenshot 2019-06-12 at 18 41 12" src="https://user-images.githubusercontent.com/1421091/59370088-14edd680-8d42-11e9-80cf-b2e475074c45.png">

(originally from https://github.com/elastic/elastic-charts/issues/454) Labels are always drawn with a single color, default black, so there is a contrast issue. I've taken the storybook example and added offsetY of 10:

<img width="531" alt="Screenshot 2019-11-20 15 35 51" src="https://user-images.githubusercontent.com/666475/69276185-f66eef80-0bab-11ea-8fa8-de099cfbe6f4.png">

As you can see in the example, black text is draw on highly saturated bars which is not as contrasty as it could be.

**Describe the solution you'd like**
Automatically or manually configure colors for labels outside bars

**Describe alternatives you've considered**
few possibilities:
- configure 2 different colors: one for labels over bars, one for labels outside bars
- detect the background color and apply an inverse color or one color from a dark/light palette
- configure 1 main color, and apply a different "standard" color for the labels outside the bars

cc @cchaos I'd like to have also your idea on that

**Additional context**
n/a

**Kibana Cross Issues/PR**
- https://github.com/elastic/kibana/pull/36511 this is a community PR, I've suggested to add two colors a dark and a light one depending on the background color.

**Checklist**

- [x] this request is checked against already exist requests
- [x] every related Kibana issue is listed under `Kibana Cross Issues` list
- [x] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

[...]

PR #608, opened 2020-03-26
feat: add 4.5 contrast for text in partition slices

### LINKED ISSUE #606: Partition: calculate relative luminance for text
Currently in pie charts there's logic in place for white text or black text
```
/** @internal */
export function colorIsDark(color: Color) {
  // fixme this assumes a white or very light background
  const rgba = stringToRGB(color);
  const { r, g, b, opacity } = rgba;
  const a = rgba.hasOwnProperty('opacity') ? opacity : 1;
  return r * 0.299 + g * 0.587 + b * 0.114 < a * 150;
}
``` 
but this needs to account for relative luminance to be in accordance with WCAG standards for text and background contrast.

**Additional context**
This logic may be in place already in this https://github.com/elastic/kibana/pull/60261 and we can potentially just move the logic onto the elastic-charts end. 

**Kibana Cross Issues**

**Checklist**

rshen91:
## Summary

Closes [#606 ](https://github.com/elastic/elastic-charts/issues/606) 
Iterated off some logic from https://github.com/elastic/kibana/pull/60261 into the elastic-charts repo to calculate light or dark text for the partition and tree map charts. 

- Text color changes accounting for default 4.5 contrast within the pie or tree map slices or the link labels outside of the pie charts. If black or white text on whatever slice background can achieve the 4.5 contrast ratio, or user specified contrast ratio, then the text will be either black or white. Otherwise, the text will adjust grayscale to try and achieve the desired contrast ratio. 
- If `textInvertible` is set true, then the contrast of the text on the slices and the container background color is taken into account. 
- This PR also allows the user to provide the background color of the container so that contrast can be calculated effectively for any opacity/transparency of a pie slice. 

New stories added: 
- `Stylings - Partition Background`: which has a pie chart with varying opacity values. Knobs are included to set `textInvertible` and `textContrast` to true or false. Additionally you can specify a number for `textContrast` to try and achieve. 
- `Stylings - Partition Labels`: showing the link labels adjusting in color to different background colors. There is a knob provided for the background to be set to show textColor changing in the labels and the text within the pie slices. 

Main files with changes for reviewers:
- src/chart_types/partition_chart/layout/utils/calcs.ts
- src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
- src/chart_types/partition_chart/layout/viewmodel/link_text_layout.ts
- There is also a [Google sheet](https://docs.google.com/spreadsheets/d/1cLrzc6R6X4lJrsys9KRvjx52YUAjpU3ChTvTtJTikkU/edit?usp=sharing) with color and contrast with black text vs. white text worth checking out to justify why some of the stories changed. 

### Checklist


rshen91:
@cchaos when you have the time, I'd love to hear your thoughts and make sure I'm interpreting this  correctly: 
This PR is meant to address the earlier conversation here about taking into account relative luminosity: https://elastic.slack.com/archives/CFK9BJ248/p1584979319002100

Currently, this PR will take into account contrast (which I believe is closely tied to luminosity) but in the `sunburst heterogenous story` (screenshot below) it looks like the lighter red partitions should be black text instead of white? Or am I misinterpreting and the white text on the `rgba(228, 26, 28, 0.3)` background is actually ok? 
Screen Shot 2020-04-01 at 4 23 14 PM

Is there a different term (like transparency?) that I should be considering in calculating the text color with a certain background color? Thanks for you help in advance!

cchaos:
@rshen91 Yes you are correct. That white text on the transparent sections does not have high contrast which means that your calculations don't take transparency into account. Though the only way it'd be able to do so is if you also knew the background color of the container. 

Opacity is a killer for trying to calculate contrast ratios. In EUI, we don't use transparency but "tint" instead which is when a color is blended with white. Or "shade" for blending with black.

But as user and consumer created object, you'll never truly know the background color unless you **ask** for it. 

Also, it still doesn't look like the calculations are correct as even the inner-most slice, which I'd assume is **not** already transparent, fails at only 3.06

Screen Shot 2020-04-03 at 16 59 49 PM

rshen91:
Issue https://github.com/elastic/elastic-charts/issues/238 is tied to this issue in that if we get the background color in order to determine tint or shade for this PR, that same prop can be used to help solve issue #238 `detect the background color and apply an inverse color or one color from a dark/light palette`

rshen91:
This PR will be relevant for PR https://github.com/elastic/elastic-charts/pull/629 where the background color can be used for the tooltip color icon

src/chart_types/partition_chart/layout/utils/d3_utils.ts
```diff
@@ -106,3 +108,23 @@ export function RGBtoString(rgb: RgbObject): string {
...
+ * @internal */
+export function RGBAToHex(rgba: Color) {
+  return chroma(rgba).hex();
+}
+
+/** convert hex to rgb
+ * @internal */
+export function HexToRGB(hex: string) {
+  return chroma(hex).rgba();
+}
```
markov00, line 130: If we want to abstract away from the chroma library, I think it's better to create a Color class, that handles these transformations.
For now, it's fine having them as separate functions

src/components/chart.tsx
```diff
@@ -100,6 +101,9 @@ export class Chart extends React.Component<ChartProps, ChartState> {
       if (state.internalChartState) {
         state.internalChartState.eventCallbacks(state);
       }
+      this.setState({
+        backgroundStyles: getChartThemeSelector(state).background,
+      });
```
markov00, line 106: we should  possibly call setState only one time in this function, we have another call bit later

markov00: I'm also thinking that from the point of view or re-rendering components, it's probably better to move the background color of the whole chart in a separate component, so we just rerender it, instead of the whole `<Chart` one

markov00: One last thing to consider: in this component there is also the `getPNGSnapshot` that should now consider the effective `background` color configured through the Theme:
I think we should change the signature of the call removing the backgroundColor on the option object and reuse the one coming from the `theme` or 'transparent` if no bg theme color is configured

src/chart_types/partition_chart/layout/utils/calcs.ts
```diff
@@ -49,10 +50,92 @@ export function arrayToLookup(keyFun: Function, array: Array<any>) {
...
+  let rgba2FormattedRGBA = rgba2.startsWith('#') ? RGBATupleToString(HexToRGB(rgba2)) : rgba2;
+
+  rgba1FormattedRGBA = /^[a-z]/.test(rgba1) ? RGBATupleToString(chroma(rgba1).rgba()) : rgba1FormattedRGBA;
+  rgba2FormattedRGBA = /^[a-z]/.test(rgba2) ? RGBATupleToString(chroma(rgba2).rgba()) : rgba2FormattedRGBA;
+  const { red: red1, green: green1, blue: blue1, opacity: alpha1 } = convertRGBAStringToSeparateValues(
+    rgba1FormattedRGBA,
+  );
+  const { red: red2, green: green2, blue: blue2, opacity: alpha2 } = convertRGBAStringToSeparateValues(
+    rgba2FormattedRGBA,
+  );
```
markov00, line 76: we can directly use chroma here, something like this should work:
```suggestion
const [red1, green1, blue1, alpha1] = chroma(rgba1).rgba()
const [red2, green2, blue2, alpha2] = chroma(rgba2).rgba()
```

src/chart_types/partition_chart/layout/utils/calcs.ts
```diff
@@ -49,10 +50,92 @@ export function arrayToLookup(keyFun: Function, array: Array<any>) {
...
+    .replace(/[^\d.,]/g, '')
+    .split(',')
+    .map((x) => parseFloat(x));
+  return { red: red1, green: green1, blue: blue1, opacity: alpha1 };
+}
+
+/** If the user specifies the background of the container in which the chart will be on, we can use that color
+ * and make sure to provide optimal contrast
+/** @internal */
+export function combineColors(rgba1: Color, rgba2: Color) {
```
markov00, line 64: I think it's better to specify which color is the foreground and which is the background because I don't think these two colors are interchangeable.

src/chart_types/partition_chart/layout/utils/calcs.ts
```diff
@@ -49,10 +50,92 @@ export function arrayToLookup(keyFun: Function, array: Array<any>) {
...
+    }
+  }
+  return highContrastTextColor.toString();
+}
+
+/**
+ * show contrast amount
+ * @internal
+ */
+export function showContrastAmount(foregroundColor: string | chroma.Color, backgroundColor: string | chroma.Color) {
```
markov00, line 130: nit: `getContrastAmount` or `getContrast`

package.json
```diff
@@ -111,6 +113,7 @@
     "babel-loader": "^8.0.6",
     "backport": "^4.8.0",
     "canvas": "^2.4.1",
+    "chroma-js": "^2.1.0",
```
markov00, line 116: chroma is a dependency, not a dev dependency, it's used at runtime

monfera:
Great work Rachel and great collab on this PR 🎉 

I'm on the lookout for learning, here, via checking out color changes. As expected the PR swaps over some texts from white to black or the other way around. Well cross this, the examples were never switches from white to black :-) My sense of what's more contrasty is apparently heavily biased toward white text on what I think is fairly dark background :-) 

Using a color picker and a contrast checker, I checked Poland [on this image](https://raw.githubusercontent.com/elastic/elastic-charts/511b5b4ca994181c8b6cf0be32d9a7dfaaa637e4/integration/tests/__image_snapshots__/all-test-ts-baseline-visual-tests-for-all-stories-treemap-two-layers-stress-test-visually-looks-correct-1-snap.png), which is the first country to switch over to black:

image

Indeed, Poland #7875B6 gives an over-4.5 contrast for black (5.02) while white is under-4.5 (4.18) so the switch is warranted 💃 

Testing the one preceding item, Thailand #7773B1 for black (4.87) and white (4.31) would need a similar switch, right? I tested it in a couple of checkers eg. [webaim](https://webaim.org/resources/contrastchecker/) and [contrastchecker](https://contrastchecker.com/) and there was only a single hundredth-digit variation, yet I may still not use a correct picker or evaluator, or I'd need to learn about more nuance.

Also I see a couple of stories had been added, out of interest, what do they do that the preexisting mocks don't do?

One part that was puzzling to me is the switch from black to gray here, on 1st look without a change to the story source:
image
Switching from black to grey reduces contrast from about 11 to about 5.5, still good but most likely I'm overlooking a code change that intends to cause it.

src/chart_types/partition_chart/layout/config/config.ts
```diff
@@ -218,6 +219,7 @@ export const configMetadata = {
       },
       textColor: { dflt: '#000000', type: 'color' },
```
monfera, line 220: If we switch to abbreviated hex code in line 173, should it happen here too? Althogh I'm not sure if it's worth the byte savings; currently it's possible to grep for strings of length 7 starting with `#` to find hex colors :-) (strings starting with `#` can also be URLs etc.)

src/chart_types/partition_chart/layout/config/config.ts
```diff
@@ -238,7 +240,7 @@ export const configMetadata = {
   },
 
   // other
-  backgroundColor: { dflt: '#ffffff', type: 'color' },
+  // backgroundColor: { dflt: '#ffffff', type: 'color' },
```
monfera, line 243: As the `backgroundColor` property is not removed from the `Config` type itself, and many examples use this prop, what's the purpose of not leaving it here? If it interferes with something, it'd be easier to chat. We don't currently make a lot of use of it but this object represents some metadata about the props.

src/chart_types/partition_chart/layout/types/types.ts
```diff
@@ -43,6 +43,8 @@ export interface Font {
   fontVariant: FontVariant;
   fontWeight: FontWeight;
   fontFamily: FontFamily;
+  textColor: string;
+  textOpacity: number;
 }
```
monfera, line 48: It was already possible to color the texts, would be glad to catch up on the usefulness of putting it in this type. I'm not against it and it may be the better model, I'm just curious

src/chart_types/partition_chart/layout/types/viewmodel_types.ts
```diff
@@ -94,19 +93,33 @@ export type ShapeViewModel = {
   config: Config;
   quadViewModel: QuadViewModel[];
   rowSets: RowSet[];
-  linkLabelViewModels: LinkLabelVM[];
+  linkLabelViewModels: LinkLabelsViewModelSpec;
```
monfera, line 96: Good call extracting it out!

src/chart_types/partition_chart/layout/types/viewmodel_types.ts
```diff
@@ -94,19 +93,33 @@ export type ShapeViewModel = {
...
   config: specifiedConfig || config,
   quadViewModel: [],
   rowSets: [],
-  linkLabelViewModels: [],
+  linkLabelViewModels: {
+    links: [],
+    labelFontSpec: defaultFont,
+    valueFontSpec: defaultFont,
+    strokeColor: '',
+  },
```
monfera, line 122: I'm curious what the goal is behind moving font/color specs from the individual `LinkLabelVM`s into a newly created upper layer, let's touch on it in a chat unless it's quicker to type in :-)

src/chart_types/partition_chart/layout/types/viewmodel_types.ts
```diff
@@ -94,19 +93,33 @@ export type ShapeViewModel = {
...
+};
+
 /* @internal */
 export const nullShapeViewModel = (specifiedConfig?: Config, diskCenter?: PointObject): ShapeViewModel => ({
   config: specifiedConfig || config,
   quadViewModel: [],
   rowSets: [],
-  linkLabelViewModels: [],
+  linkLabelViewModels: {
+    links: [],
```
monfera, line 118: I'd call these `linkLabels` or similar, rather than `links`, as the links are the lines that link the labels to the pie

src/chart_types/partition_chart/layout/utils/d3_utils.ts
```diff
@@ -17,6 +17,8 @@
  * under the License. */
 
 import { rgb as d3Rgb, RGBColor as D3RGBColor } from 'd3-color';
+import { Color } from '../../../../utils/commons';
+import chroma from 'chroma-js';
```
monfera, line 21: As this file became more than D3 utils, a file name change is warranted, eg. `color_library_wrappers.ts` or some such; luckily all things in here are already color related, and the new file could express the fact that there's no actual logic here, it's just a central, easy to replace "rail turntable" to not bake in external libraries too much

monfera:
Verified that `makeHighContrastColor` is only ever invoked for cells that'll actually get text, ie. big enough - this is a guarantee that it won't be expensive (cell count is essentially unbounded, but text count is bounded by sensible minimum font size and sufficient room constraint)

src/chart_types/partition_chart/layout/utils/calcs.ts
```diff
@@ -49,10 +50,92 @@ export function arrayToLookup(keyFun: Function, array: Array<any>) {
...
+  const lightness = chroma(background).get('hsl.l');
+  let highContrastTextColor = foreground;
+  const isBackgroundDark = colorIsDark(background);
+  // determine whether white or black text is ideal contrast vs a grey that just passes 4.5 ratio
+  if (isBackgroundDark && chroma.deltaE('black', foreground) === 0) {
+    highContrastTextColor = '#fff';
+  } else if (lightness > 0.5 && chroma.deltaE('white', foreground) === 0) {
+    highContrastTextColor = '#000';
+  }
+  const precision = Math.pow(10, 8);
```
monfera, line 102: shorter and needs no calc: `const precision = 1e8;`

src/chart_types/partition_chart/layout/utils/calcs.ts
```diff
@@ -49,10 +50,92 @@ export function arrayToLookup(keyFun: Function, array: Array<any>) {
...
+  let contrast = showContrastAmount(highContrastTextColor, background);
+  // adjust the highContrastTextColor for shades of grey
+  while (contrast < ratio) {
+    if (isBackgroundDark) {
+      highContrastTextColor = chroma(highContrastTextColor)
+        .brighten()
+        .toString();
```
monfera, line 109: does it need a conversion to string?

src/chart_types/partition_chart/layout/utils/calcs.ts
```diff
@@ -49,10 +50,92 @@ export function arrayToLookup(keyFun: Function, array: Array<any>) {
...
+  // determine whether white or black text is ideal contrast vs a grey that just passes 4.5 ratio
+  if (isBackgroundDark && chroma.deltaE('black', foreground) === 0) {
+    highContrastTextColor = '#fff';
+  } else if (lightness > 0.5 && chroma.deltaE('white', foreground) === 0) {
+    highContrastTextColor = '#000';
+  }
+  const precision = Math.pow(10, 8);
+  let contrast = showContrastAmount(highContrastTextColor, background);
+  // adjust the highContrastTextColor for shades of grey
+  while (contrast < ratio) {
```
monfera, line 105: I'd like to better understand what's happening in the loop, when we hop on a call

src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
```diff
@@ -286,6 +286,40 @@ function getWordSpacing(fontSize: number) {
...
+  // change the contrast for the inverted slices
+  if ((textInvertible && !textContrast) || (textInvertible && typeof textContrast !== 'number')) {
+    const backgroundIsDark = colorIsDark(combineColors(sliceFillColor, containerBackgroundColorFromUser));
+    const specifiedTextColorIsDark = colorIsDark(adjustedTextColor);
+    const inverseForContrast = specifiedTextColorIsDark === backgroundIsDark;
+    const { r: tr, g: tg, b: tb, opacity: to } = stringToRGB(adjustedTextColor);
+    adjustedTextColor = inverseForContrast
+      ? to === undefined
+        ? `rgb(${255 - tr}, ${255 - tg}, ${255 - tb})`
+        : `rgba(${255 - tr}, ${255 - tg}, ${255 - tb}, ${to})`
```
monfera, line 312: This function has these parts:
- supplement of default values (`undefined` conditions, `typeof ... = '...'`  spread around)
- format conversions from and to numbers and strings (`stringToRGB`, string templates)
- actual logic

It'd be great to refactor it such that the logic part is separated from the defaults and format conversions. Also there's some nested ternary that's inside an `if` branch - I like ternaries, but the net effect is that it's hard to follow how we arrive at a value. `let` and subsequent modification of the binding in a larger block of code after the declaration makes it hard to comprehend the value flow. Some variables could have more self explanatory names, and/or more comments could be added, eg. `tr`, `tg` etc -maybe there's room for `textR`, `textOpacity` etc.

For these reasons, would it be possible to decompose it into several functions, each of which addresses a single aspect?

src/chart_types/partition_chart/renderer/canvas/canvas_renderers.ts
```diff
@@ -186,41 +187,36 @@ function renderLinkLabels(
...
-          ctx.scale(1, -1); // flip for text rendering not to be upside down
-          ctx.textAlign = textAlign;
-          ctx.font = `${labelFontSpec.fontStyle} ${labelFontSpec.fontVariant} ${labelFontSpec.fontWeight} ${linkLabelFontSize}px ${labelFontSpec.fontFamily}`;
-          ctx.fillText(text, textAlign === 'right' ? -valueWidth - labelValueGap : 0, 0);
-          ctx.font = `${valueFontSpec.fontStyle} ${valueFontSpec.fontVariant} ${valueFontSpec.fontWeight} ${linkLabelFontSize}px ${valueFontSpec.fontFamily}`;
-          ctx.fillText(valueText, textAlign === 'left' ? width + labelValueGap : 0, 0);
-        });
-      },
-    );
+    links.forEach(({ link, translate, textAlign, text, valueText, width, valueWidth }: LinkLabelVM) => {
```
monfera, line 197: let's please call it `linkLabels` or similar

src/chart_types/partition_chart/layout/config/config.ts
rshen91: ah good catch thanks!

rshen91:
jenkins retest this please

src/chart_types/partition_chart/layout/types/types.ts
rshen91: discussed over zoom

src/chart_types/partition_chart/layout/types/viewmodel_types.ts
rshen91: discussed in zoom

src/chart_types/partition_chart/layout/utils/calcs.ts
```diff
@@ -49,10 +50,92 @@ export function arrayToLookup(keyFun: Function, array: Array<any>) {
 }
 
 /** @internal */
+export function convertRGBAStringToSeparateValues(rgba: Color) {
```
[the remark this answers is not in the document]
rshen91: Commit f8619e2 for changes! Thanks!

src/chart_types/partition_chart/layout/utils/calcs.ts
rshen91: Commit f8619e2 for changes! Thanks!

src/chart_types/partition_chart/layout/utils/calcs.ts
rshen91: Good call commit f8619e2 for changes! Thanks!

src/chart_types/partition_chart/layout/utils/calcs.ts
rshen91: Commit f8619e2 for changes! Thanks!

package.json
rshen91: Nice thanks! Commit f8619e2 for changes

src/chart_types/partition_chart/renderer/canvas/canvas_renderers.ts
rshen91: Changed in b411301

src/chart_types/partition_chart/layout/utils/calcs.ts
rshen91: changed in b411301

src/chart_types/partition_chart/layout/utils/d3_utils.ts
rshen91: changed in b411301 thanks!

src/chart_types/partition_chart/layout/types/viewmodel_types.ts
rshen91: changed in b411301 - thanks!

src/components/chart.tsx
rshen91: Addressed in 545e78e

src/chart_types/partition_chart/layout/utils/__mocks__/fill_text_layout.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License. */
+const module = jest.requireActual('../../viewmodel/fill_text_layout.ts');
+
+export const getTextColor = jest.fn(module.getTextColor);
```
nickofthyme, line 20: 👍

api/charts.api.md
```diff
@@ -1250,6 +1258,8 @@ export const Settings: React.FunctionComponent<SettingsSpecProps>;
 export interface SettingsSpec extends Spec {
     // (undocumented)
     animateData: boolean;
+    // (undocumented)
+    backgroundColor?: string;
```
nickofthyme, line 1262: What does this do vs `theme.background`?

src/utils/themes/theme.ts
```diff
@@ -189,6 +192,7 @@ export interface Theme {
    * value from 1 to 100
    */
   markSizeRatio?: number;
+  background: BackgroundStyles;
```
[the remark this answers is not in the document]
rshen91: Ah cool idea. I added some in 01b7052. Let me know what you think!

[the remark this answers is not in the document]
nickofthyme: Oh yeah docs would be great too.

But I was referring to the JSdocs comment to document the `background` property

```suggestion
  /**
   *  This thing does....
   */
  background: BackgroundStyles;
```

src/chart_types/partition_chart/layout/viewmodel/link_text_layout.ts
```diff
@@ -134,18 +172,19 @@ export function linkTextLayout(
         valueFontSpec,
       };
     })
-    .filter((l: LinkLabelVM) => l.text !== ''); // cull linked labels whose text was truncated to nothing
-}
+    .filter((l: LinkLabelVM) => l.text !== ''); // cull linked labels whose text was truncated to nothing;
```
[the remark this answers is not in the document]
rshen91: Changed in commit 01b7052 thank you!

src/chart_types/partition_chart/layout/viewmodel/link_text_layout.ts
```diff
@@ -82,6 +117,9 @@ export function linkTextLayout(
       const rawText = rawTextGetter(node);
       const labelText = cutToLength(rawText, maxTextLength);
       const valueText = valueFormatter(valueGetter(node));
+
+      // const { width, emHeightAscent, emHeightDescent } = measure(linkLabel.fontSize, [{ ...labelFontSpec, text }])[0];
```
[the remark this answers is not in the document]
rshen91: Changed in commit 01b7052 thank you!

src/chart_types/partition_chart/layout/utils/calcs.test.ts
```diff
@@ -16,7 +16,67 @@
...
 
-import { integerSnap, monotonicHillClimb } from './calcs';
+import { makeHighContrastColor, combineColors, integerSnap, monotonicHillClimb } from './calcs';
+
+describe('calcs', () => {
+  describe('test makeHighContrastColor', () => {
+    it('hex input - should change white text to black when background is white', () => {
+      const expected = '#000';
+      const result = makeHighContrastColor('#fff', '#fff');
+      expect(result).toEqual(expected);
```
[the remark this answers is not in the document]
rshen91: Nice! Changed in commit 01b7052 thank you!

api/charts.api.md
rshen91: I think this is leftover from the config and config_types that were used before the partition charts were merged into Storybook - addressed in commit 36baf2e

src/specs/settings.tsx
```diff
@@ -263,6 +263,7 @@ export interface SettingsSpec extends Spec {
    * @defaultValue 2
    */
   minBrushDelta?: number;
+  backgroundColor?: string;
```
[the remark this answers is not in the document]
rshen91: Yep you're right - removing this in commit 36baf2e

src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
rshen91: Moved the logic more into the calcs file hopefully clarifying some of this. Missed adding the commit earlier, I believe it's mainly in commit 545e78e4332cd9e35dc945919b64b39d5e9ed56f

src/components/_global.scss
```diff
@@ -10,3 +10,11 @@
   height: 0;
   position: absolute;
 }
+
+.echChartBackground {
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  position: absolute;
```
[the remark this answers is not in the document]
nickofthyme: @rshen91 have you tested this ⬆️ ?

src/components/chart_background.tsx
```diff
@@ -36,7 +36,7 @@ export class ChartBackgroundComponent extends React.Component<ChartBackgroundPro
   render() {
     const { backgroundStyles } = this.props;
 
-    return <div className="echChartBackground" style={{ backgroundColor: backgroundStyles!.color }} />;
+    return <div className="echChartBackground" style={{ background: backgroundStyles!.color }} />;
```
nickofthyme, line 39: I would say avoid `!` on objects cuz if `backgroundStyles` is undefined, this will error.

```suggestion
    return <div className="echChartBackground" style={{ background: backgroundStyles?.color }} />;
```

if it still complains, you could pass a default `backgroundStyles?.color ?? 'blue'`

src/utils/themes/dark_theme.ts
```diff
@@ -163,4 +163,7 @@ export const DARK_THEME: Theme = {
       visible: true,
     },
   },
+  background: {
+    color: 'transparent',
```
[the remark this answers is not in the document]
nickofthyme: f14f2fa

src/components/_global.scss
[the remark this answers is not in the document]
nickofthyme: Verified in IE11 👍

src/chart_types/partition_chart/layout/utils/calcs.ts
```diff
@@ -52,8 +52,9 @@ export function arrayToLookup(keyFun: Function, array: Array<any>) {
 
 /** If the user specifies the background of the container in which the chart will be on, we can use that color
  * and make sure to provide optimal contrast
-/** @internal */
```
nickofthyme, line 51: good catch

src/components/chart.tsx
```diff
@@ -156,7 +147,7 @@ export class Chart extends React.Component<ChartProps, ChartState> {
 
   render() {
     const { size, className } = this.props;
-    const containerStyle = getContainerStyle(size);
```
nickofthyme, line 160: I don't really like this cuz even though it is only returning the size, it is getting the style object and not the size alone. Then in the future, people will add some other styles and the method will still be `getChartSize`. That's why I had her change this.

I'm fine with it for now but I think this naming is wrong.

src/components/chart_background.tsx
```diff
@@ -17,13 +17,12 @@
...
 import React from 'react';
 import { connect } from 'react-redux';
-import { BackgroundStyles } from '../utils/themes/theme';
 import { getChartThemeSelector } from '../state/selectors/get_chart_theme';
 import { GlobalChartState } from '../state/chart_state';
 import { getInternalIsInitializedSelector } from '../state/selectors/get_internal_is_intialized';
 
 interface ChartBackgroundProps {
-  backgroundStyles?: BackgroundStyles;
+  background: string;
```
nickofthyme, line 25: I think this limits the future expansion of the background theme.

src/components/chart_background.tsx
```diff
@@ -34,18 +33,19 @@ export class ChartBackgroundComponent extends React.Component<ChartBackgroundPro
   }
 
   render() {
-    return <div className="echChartBackground" style={{ background: this.props?.backgroundStyles?.color }} />;
+    const { background } = this.props;
+    return <div className="echChartBackground" style={{ background }} />;
```
nickofthyme, line 37: This could allow none colors to be passed to the `background` 

```scss
body {
  background: lightblue url("img_tree.gif") no-repeat fixed center;
}
```

src/components/chart_background.tsx
```diff
@@ -34,18 +33,19 @@ export class ChartBackgroundComponent extends React.Component<ChartBackgroundPro
...
   }
 }
 
 const mapStateToProps = (state: GlobalChartState): ChartBackgroundProps => {
   if (!getInternalIsInitializedSelector(state)) {
     return {
-      backgroundStyles: getChartThemeSelector(state).background,
+      background: 'transparent',
```
nickofthyme, line 44: good catch

src/components/chart_background.tsx
nickofthyme: changed here 9c91e74

PR #625, opened 2020-04-10
feat(brush): add multi axis brushing

### LINKED ISSUE #587: Y-axis brushing

When using `onBrushEnd` with a chart, brushing only happens along the x-axis. 

I'd like to allow brushing along the y-axis instead of the x-axis. (Perhaps also both, with a rectangular selection, but this is not needed for my use case.)

**Kibana Cross Issues**

Work using this is happening in elastic/kibana#59496.

**Checklist**

markov00:
## Summary

This commit allows the consumer to configure the direction used by the brush tool from 
`<Settings brushAxis="x / y / both" />` . 
By default, the direction is along the X-axis, but can be configured to go along the Y-axis or along both axes. 
For the Y-axis values, we are returning an array of values, one for each Y-axis defined (this is determined by the groupId associated with different axis/series groups)

Along Y axis:

Apr-10-2020 15-15-33

Along both axis, aka rectangular selection:

Apr-10-2020 15-14-56

**BREAKING CHANGE**: The type used by the `BrushEndListener` is now changed to reflect this change. The  `BrushEndListener` is changed to  
```ts
type BrushEndListener = (brushArea: XYBrushArea) => void;

export interface GroupBrushExtent {
  groupId: GroupId;
  extent: [number, number];
}
export interface XYBrushArea {
  x?: [number, number];
  y?: Array<GroupBrushExtent>;
}
```

fix #587

Summarize your PR.

If it involves visual changes include a screenshot or gif.

### Checklist


integration/page_objects/common.ts
```diff
@@ -132,6 +132,51 @@ class CommonPage {
...
+  async dragMouseRelativeToDOMElement(
+    start: { x: number; y: number },
+    end: { x: number; y: number },
+    selector: string,
+  ) {
+    const element = await this.getBoundingClientRect(selector);
+    await page.mouse.move(element.left + start.x, element.top + start.y);
+    await page.mouse.down();
+    await page.mouse.move(element.left + end.x, element.top + end.y);
+  }
```
nickofthyme, line 150: Do you have any ideas on how to determine positioning based on series? I don't like that we blindly put in x and y coordinates. Especially if we started getting into screen size changes.

For future work...

integration/page_objects/common.ts
```diff
@@ -201,6 +246,27 @@ class CommonPage {
...
 
+  /**
+   * Expect a chart given a url from storybook with mouse move
+   *
+   * @param url Storybook url from knobs section
+   * @param start - the start postion of mouse relative to chart
+   * @param end - the end postion of mouse relative to chart
+   * @param options
+   */
+  async expectChartWithDragAtUrlToMatchScreenshot(
```
nickofthyme, line 257: 👍

src/chart_types/xy_chart/state/selectors/get_brush_area.ts
```diff
@@ -30,31 +35,79 @@ const getCurrentPointerPosition = (state: GlobalChartState) => {
...
-      };
-      return area;
+    switch (brushAxis) {
+      case BrushAxis.Y:
+        return getBrushForYAxis(chartDimensions, chartRotation, start, end);
+      case BrushAxis.Both:
+        return getBrushForBothAxis(chartDimensions, start, end);
+      case BrushAxis.X:
+      default:
+        return getBrushForXAxis(chartDimensions, chartRotation, start, end);
```
nickofthyme, line 60: Nice I always struggled with how to write this and account for the `default` case. This is a good idea.

src/chart_types/xy_chart/state/selectors/on_brush_end_caller.ts
```diff
@@ -124,3 +116,75 @@ export function createOnBrushEndCaller(): (state: GlobalChartState) => void {
...
+  let endPos = getLeftPoint(chartDimensions, lastDrag.end.position);
+  let chartMax = chartDimensions.width;
+
+  if (isVerticalRotation(rotation)) {
+    startPos = getTopPoint(chartDimensions, lastDrag.start.position);
+    endPos = getTopPoint(chartDimensions, lastDrag.end.position);
+    chartMax = chartDimensions.height;
+  }
+
+  let minPos = Math.max(Math.min(startPos, endPos), 0);
```
nickofthyme, line 137: I think this is common enought to put in a helper method like `constrain`

```ts
type Constrain = (min?: number, max?: number): number
```

markov00: done in https://github.com/elastic/elastic-charts/pull/625/commits/289d5290bd82165f43a756c67a7d9b9a8d165ee7

markov00:
@nickofthyme I've added the the `minBrushDelta` prop to `<Settings />` to specify a minimum number of pixel to identify a brush area as valid https://github.com/elastic/elastic-charts/pull/625/commits/7de873988b8d713999f7734fba30db385967d47d
I prefer to keep the brushed values in a sorted array [min, max], then having them as [first, last]
we can then change this behavior later.

nickofthyme:
@markov00 I don't see the min number behavior, It still allows me to make the line invisible. Am I missing something?

Screen Recording 2020-04-20 at 10 58 AM

monfera:
Disappearing area: users are box selecting during a continuous expression of intent, ie. holding down the mouse, dragging and seeing visual feedback, encountering a special case which shows zero area, which even corresponds to the current intent. The user keeps pressing the mouse button and is likely to move further if wanted. Also, click-hold-drag-release is appropriately done on low-risk interaction ie. there isn't much harm if the user releases the mouse and retries (if there are negative consequences, D&D is not a good activating mechanism to begin with). 

Other libraries also have disappearing areas, eg. the lower Highcharts [here](https://www.highcharts.com/demo/dynamic-master-detail), or the Plotly [box select](https://plotly.com/javascript/line-and-scatter/) which is btw. an auto X/Y/box selector in this example.

It's possible to add eg. corner markers so that they fall outside the actual selected area, but then there are some mundane details to consider, eg. it can protrude outside the Cartesian area if a selection border coincides with an extreme (which is implementable but it may be a time cost/benefit tradeoff). 

Besides the effort, code size and complexity, there are _some_ advantages to keeping the visuals simple anyway:
- clean, simple, yet unambiguous feedback for the user
- small corner marks not interfering with the mouse cursor (which is why even the Plotly box markers are made to disappear

So it's a fine detail with some tradeoffs that I think wouldn't block this PR, OK done with the bike shedding for a while 🤣

PR #629, opened 2020-04-11
fix(tooltip): show true opaque colors in tooltips

### LINKED ISSUE #628: True tooltip colors
For the lighter colors, the tooltip shows a darker color. I think this is because they are less opaque against the black tooltip background. We could fix this by adding a white element behind each color to get the true color.

Screen Recording 2020-04-10 at 10 42 AM

nickofthyme:
## Summary
Add `backgroundColor` to tooltip colors to show true color. Uses background color from `theme.background.color` per #608.

> Most noticeable with light colors.

fixes #628 

## Before
Screen Recording 2020-06-08 at 05 13 PM

## After
Screen Recording 2020-06-08 at 05 00 PM


src/components/tooltip/_tooltip.scss
```diff
@@ -24,10 +24,29 @@
...
+      padding: 3px;
+      padding-left: 0;
+      min-width: 1px;
+    }
+
+    &--backgroundColor {
+      position: relative;
+      width: $euiSizeXS;
+      margin-right: 3px;
+      background-color: $euiColorGhost;
```
markov00, line 40: This color will be always `#fff` independently of the background color. I think, for a first fix, we should use a EUI theme color dependant, like the `$euiColorFullShade` that is white with the light theme and near black for the dark theme.

@rshen91 is working on adding a background prop to the chart to handle various background color, this will then needs to be connected with this component to pass the right color if specified

nickofthyme: Ok good idea. I'll wait on #608

nickofthyme:
Good eye @rshen91 I'll test that case when I merge with your changes. 👍

src/components/tooltip/_tooltip.scss
nickofthyme: see 58cf176

nickofthyme:
@monfera thanks, good catch! The first issue I think is expected, what do you mean by `the sunburst tooltip shows the old color version`? The color is the only diff in that png.

The second one I will fix now.

markov00:
> @monfera thanks, good catch! The first issue I think is expected, what do you mean by `the sunburst tooltip shows the old color version`? The color is the only diff in that png.
> 
> The second one I will fix now.

The review was mine :D 
by `the sunburst tooltip shows the old color version` I mean: this error was expected, the screenshot diff shows that we forgot to update that screenshot

nickofthyme:
Sorry Marco, I don't know why I thought that was Robert. Gotcha, thanks I'll update both bullets.

nickofthyme:
@markov00 Fixed both in b0e6eb9. I verified the tooltip style changes in IE11, safari, firefox and chrome, but feel free to check yourself.

PR #630, opened 2020-04-14
feat: add color option to use series color

nickofthyme:
## Summary

Add `ColorVariant` type with `Series` and `None`

```ts
/**
 * Color varients that are unique to `@elastic/charts`. These go beyond the standard
 * static color allocations.
 */
export const ColorVariant = Object.freeze({
  /**
   * Uses series color. Rather than setting a static color, this will use the
   * default series color for a given series.
   */
  Series: '__use__series__color__' as '__use__series__color__',
  /**
   * Uses empty color, similar to transparent.
   */
  None: '__use__empty__color__' as '__use__empty__color__',
});
export type ColorVariant = $Values<typeof ColorVariant>;
```

- Series option can be used for any series-related color in the theme object
- None option can be used to set a default color value to transparent
- Refactor `stringToRGB` method to use opacity and `OpacityFn`
- Add helper method to process `ColorVariant`
- Update theme color string with `Color` type
- Refactor renderer styles to allow for new `ColorVariant` and `OpacityFn`

The following `Theme` properties that used to be `Color` are now `Color | ColorVariant`
- `PointStyle.stroke`
- `PointStyle.fill`
- `LineStyle.stroke`
- `AreaStyle.fill`
- `RectStyle.fill`
- `RectStyle.stroke`

#### Area
```ts
const customTheme: PartialTheme = {
  areaSeriesStyle: {
    point: {
      visible: true,
      radius: 10,
      fill: ColorVariant.Series,
      stroke: ColorVariant.None,
      opacity: 0.5,
    },
    area: {
      opacity: 0.2,
    },
    line: {
      visible: false,
    },
  },
};
```
Image 2020-04-14 at 12 01 59 PM

#### Line
```ts
const customTheme: PartialTheme = {
  lineSeriesStyle: {
    point: {
      radius: 10,
      fill: ColorVariant.Series,
      stroke: ColorVariant.None,
    },
  },
};
```
Image 2020-04-14 at 12 01 44 PM

#### Bar
```ts
const customTheme: PartialTheme = {
  barSeriesStyle: {
    rect: {
      fill: ColorVariant.None,
    },
    rectBorder: {
      visible: true,
      strokeWidth: 10,
      stroke: ColorVariant.Series,
    },
  },
};
```
Image 2020-04-14 at 12 01 33 PM

### Checklist


src/utils/themes/theme.ts
```diff
@@ -58,9 +58,9 @@ export interface SharedGeometryStateStyle {
   unhighlighted: GeometryStateStyle;
 }
 
-export interface StrokeStyle {
+export interface StrokeStyle<C = Color> {
```
markov00, line 61: What is the benefit of having a generic type here? I think we don't want that to have someone using a different generic value other than Color here like `StrokeStyle<number>`

src/utils/commons.test.ts
```diff
@@ -639,4 +641,24 @@ describe('commons utilities', () => {
       });
     });
   });
+
+  describe('#getColorFromVariant', () => {
+    const seriesColor = '#626825';
+    it('should return color if series color is undefined ', () => {
+      expect(getColorFromVariant(seriesColor)).toBe(seriesColor);
+    });
```
markov00, line 649: the text sounds wrong to me, maybe it's more correct rephrasing like:
`should return series color if color is undefined` because is the second parameter(color/colorVariant) to be undefined

stories/stylings/17_bar_series_color_variant.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { Axis, BarSeries, Chart, Position, ScaleType, Settings, PartialTheme } from '../../src';
+import * as TestDatasets from '../../src/utils/data_samples/test_dataset';
+import { SB_SOURCE_PANEL } from '../utils/storybook';
+import { ColorVariant } from '../../src/utils/commons';
+
+export const example = () => {
+  const customTheme: PartialTheme = {
+    barSeriesStyle: {
+      rect: {
+        fill: ColorVariant.None,
```
markov00, line 30: Do you mind adding some knobs to change these colors?

src/chart_types/xy_chart/renderer/canvas/styles/point.test.ts
```diff
@@ -0,0 +1,179 @@
...
+      });
+
+      describe('colors', () => {
+        it('should call stringToRGB with values from getColorFromVariant', () => {
+          expect(stringToRGB).nthCalledWith(1, overrides.fill, expect.any(Function));
+          expect(stringToRGB).nthCalledWith(2, overrides.stroke, expect.any(Function));
+        });
+
+        it('should return fill with color', () => {
+          const { opacity, ...color } = stringToRGB(overrides.fill!);
```
markov00, line 156: nit: what if we mark the `stringToRGB` `cssColorSpecifier` param as optional? it could return our "red flag" if no color is specified.

src/utils/themes/theme.ts
nickofthyme: Yeah I didn't need this but I thought this was shared between more interfaces on the theme.

The intent was to be able to add the `ColorVariant` as an option for color. But in some places in the theme that doesn't apply.

src/utils/commons.test.ts
nickofthyme: sure I can make that clearer, I was going off the function argument names.

src/chart_types/xy_chart/renderer/canvas/styles/point.test.ts
nickofthyme: f3f51a14

src/utils/commons.test.ts
nickofthyme: b126a0aa

nickofthyme:
Thanks for pointing out the linecap issue. I saw that in safari and firefox. Looks to be fixed with this 80da4fc.

stories/stylings/17_bar_series_color_variant.tsx
nickofthyme: 95392ea9

PR #647, opened 2020-04-23
fix: tooltip container scroll issue

### LINKED ISSUE #596: Partition style chart (pie, treemap, etc) tooltips cause scrollbars to appear.
**Describe the bug**
Partition style chart (pie, treemap, etc) tooltips cause scrollbars to appear.

**To Reproduce**
1. Make a treemap not in Storybook
2. Hover a treemap piece

**Expected behavior**
No scrollbars should appear

**Screenshots**
Screen Capture on 2020-03-23 at 15-59-17

nickofthyme:
## Summary

Fixes tooltip render issue
- refactor tooltip logic to account for portal
- separate tooltip and tooltip portal components
- update tests to check for new portal condition
- set max tooltip width from component js

Screen Recording 2020-04-23 at 11 15 AM

fixes #596 

## Temporary fix
For really narrow charts where the tooltip cannot fit on the right nor the left, a temporary fix was added to limit the width of the tooltip portal to a max of 70% of the chart elements width to limit hiding the tooltip on the left. This is temporary until a more robust positioning logic is add in #651.

One issue with this approach is that the tooltip will go out of the element on the right for values in the center of the chart. However, this will not cause the scrollbars to show as described in #596.

Screen Recording 2020-04-27 at 08 58 AM

This issue can also be seen with relatively normal widths where the tooltip portal does not fit in the space but the visible tooltip does, creating an unnecessary flip to the left side, which might cause the tooltip to appear outside of the element.

Screen Recording 2020-04-27 at 09 03 AM

## Root cause
Tooltips rendered towards the right edge of the chart container would compute if the tooltip "fits" based on the `.echTooltip` element and not the `#echTooltipContainerPortal` element. Notice the red bar rendered behind the tooltip as the `#echTooltipContainerPortal` element. Thus, when the `.echTooltip` element width is less than that of the `#echTooltipContainerPortal` the element will still render on the right but the `#echTooltipContainerPortal` created an invisible additional width that pushed the tooltip off the page.

Screen Recording 2020-04-23 at 09 17 AM

The fix is to just account for the `#echTooltipContainerPortal` element width in the "fitting" calculation but still use the `.echTooltip` element width for final positioning.

### Checklist


nickofthyme:
Fixed annotation tooltips with the same fix as the regular tooltip.

Screen Recording 2020-04-27 at 05 30 PM

src/chart_types/xy_chart/annotations/annotation_tooltip.ts
```diff
@@ -27,16 +27,21 @@ export function getFinalAnnotationTooltipPosition(
   tooltip: Dimensions,
   /** the tooltip computed position not adjusted within chart bounds */
   tooltipAnchor: { top: number; left: number },
+  /** the width of the tooltip portal container */
+  portalWidth: number,
   padding = 10,
 ): {
   left: string | null;
   top: string | null;
+  anchor: 'left' | 'right';
```
markov00, line 36: nit: we can use `Position.Left` and `Position.Right` here

src/chart_types/xy_chart/renderer/dom/annotation_tooltips.tsx
```diff
@@ -58,6 +58,12 @@ class AnnotationTooltipComponent extends React.Component<AnnotationTooltipProps>
   static displayName = 'AnnotationTooltip';
   portalNode: HTMLDivElement | null = null;
   tooltipRef: React.RefObject<HTMLDivElement>;
+  /**
+   * Max allowable width for tooltip to grow to. Used to determine container fit.
+   *
+   * @unit px
+   */
+  static maxWidth = 256;
```
markov00, line 66: nit
```suggestion
  static MAX_WIDTH = 256;
```

src/components/tooltip/tooltip_portal.tsx
```diff
@@ -16,38 +16,44 @@
...
 
-class TooltipComponent extends React.Component<TooltipProps> {
+class TooltipPortalComponent extends React.Component<TooltipPortalProps> {
   static displayName = 'Tooltip';
+  /**
+   * Max allowable width for tooltip to grow to. Used to determine container fit.
+   *
+   * @unit px
+   */
+  static maxWidth = 256;
```
markov00, line 52: nit:
```suggestion
  static MAX_WIDTH = 256;
```

src/components/tooltip/utils.ts
```diff
@@ -52,23 +52,28 @@ export function getFinalTooltipPosition(
...
   /** the dimensions of the tooltip container */
   tooltip: Dimensions,
+  /** the width of the tooltip portal container */
+  portalWidth: number,
   /** the tooltip anchor computed position not adjusted within chart bounds */
   anchorPosition: TooltipAnchorPosition,
 ): {
   left: string | null;
   top: string | null;
+  anchor: 'left' | 'right';
```
markov00, line 62: `Position.Left | Position.Right`

src/chart_types/xy_chart/annotations/annotation_tooltip.ts
nickofthyme: 90fc2348

src/chart_types/xy_chart/renderer/dom/annotation_tooltips.tsx
nickofthyme: 90fc2348

src/components/tooltip/tooltip_portal.tsx
nickofthyme: 90fc2348

src/components/tooltip/utils.ts
nickofthyme: 90fc2348

PR #649, opened 2020-04-24
fix: tickFormat called on mark value

nickofthyme:
## Summary

Mark value was being called with the axis `tickFormat` function with null values. I don't think the mark value would ever be called with an axis formatted so I removed that and just pass the mark value.

### Checklist

### DIFF
--- a/src/chart_types/xy_chart/tooltip/tooltip.test.ts
+++ b/src/chart_types/xy_chart/tooltip/tooltip.test.ts
@@ -55,7 +55,7 @@ describe('Tooltip formatting', () => {
     showOverlappingTicks: false,
     tickPadding: 0,
     tickSize: 0,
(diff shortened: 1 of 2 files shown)


src/chart_types/xy_chart/tooltip/tooltip.test.ts
```diff
@@ -125,6 +125,7 @@ describe('Tooltip formatting', () => {
     expect(tooltipValue.isHighlighted).toBe(false);
     expect(tooltipValue.color).toBe('blue');
     expect(tooltipValue.value).toBe('10');
+    expect(YAXIS_SPEC.tickFormat).not.toBeCalledWith(null, undefined);
```
rshen91, line 128: Nice!

PR #651, opened 2020-04-24
feat: improve tooltip positioning with popperjs

nickofthyme:
## Summary

This replaced our current tooltip positioning logic with [popperjs](https://popper.js.org/).

Why bother? Well positioning is tricky and based on the current implementation there are some limitations to positioning relative to the window, this comes into play for really small charts.

View on local storybook: http://localhost:9001/?path=/story/bar-chart--test-tooltip-and-rotation

There is now a `Portal` component that is simply a presentation component that renders it's content to the root of the dom. You are able to pass the portal an `anchor` in which the portal will use to position itself. This anchor can be an `HTMLElement` or a `postion` values relative to a given `ref` `HTMLElement`.

These changes also allow greater flexibility in the size, specifically the width, of the tooltip. Currently the `width` is set manually for the tooltip to grow. However, with these changes, the tooltip can have a `max-width` and grow to that values without collapsing early. This also applies to any `CustomTooltip` component that is used in place of the default.

## New Tooltip options

```ts
export interface TooltipProps {
  /**
   * Preferred placement of tooltip relative to anchor.
   *
   * This may not be the final placement given the positioning fallbacks.
   *
   * @default Placement.Right
   */
  placement?: Placement;
  /**
   * If given tooltip placement is not sutable, these `Placement`s will
   * be used as fallback placements.
   */
  fallbackPlacements?: Placement[];
  /**
   * Render custom tooltip given header and values
   */
  customTooltip?: CustomTooltip;
}
type Placement = "top" | "bottom" | "left" | "right" | "top-start" | "top-end" | "bottom-start" | "bottom-end" | "right-start" | "right-end" | "left-start" | "left-end" | "auto" | "auto-start" | "auto-end"
```

## Changes

This works by creating a pseudo anchor element from the calculated `AnchorPosition` values. 

Screen Recording 2020-04-29 at 02 40 PM

> Notice the blue element behind the chart for illustration purposes.

## Why `Popperjs`?

**Pros:**
- Trusted source used for many applications
- No need to tinker with countless positioning edge cases
- Positions based on window or supplied [boundary element](https://popper.js.org/docs/v2/utils/detect-overflow/#boundary). This would allow for optionally allowing the tooltip to go outside of the chart area (i.e. for smaller charts), or confine the tooltip to the chart element if desired.
- It allows the option to set default placement and multiple fullbacks. Try right, else left, else top, else bottom.
- More control over popover placement without fear of regressions

**Cons:**
- Adds a little bloat to the code. https://bundlephobia.com/result?p=@popperjs/core@2.3.3. But we can use the lite version to reduce the bloat by only using the modifiers we need.

### Checklist


nickofthyme:
@markov00 I fixed the debounce issue. I think we could merge this faster without including the highlight `TooltipType`, and work out issues with that later.

I couldn't figure out how to persist the ref to the `tooltipRef`. Whenever the tooltip renders null instead of the portal, it loses the reference. To get around this, I check if the popper.ref and tooltipRef are the same reference and only then calling `createPopper` again. Do you see a better way to keep the reference?

nickofthyme:
Ready for review but still need to fix tests.

monfera:
Another small thing,`<Portal>` seems to be a specific thing - a wrapper component with `popper.js` coupling -, and it's referenced by only two lines, it'd be useful and simple to rename it so it reflects this coupling, eg. `<TooltipPortal>` or similar.

nickofthyme:
Great points @monfera 

I agree with the VRT changes, I think that's a drawback we will likely run into more and more (i.e. moving around tests) and should be in isolated PRs to ensure baseline is not changed functionally.

Yes I will change the name to `TooltipPortal`, I see your point and in the future might have a use for a `Portal` that is more generic.

src/chart_types/xy_chart/state/selectors/get_tooltip_values_highlighted_geoms.ts
```diff
@@ -56,6 +56,13 @@ const EMPTY_VALUES = Object.freeze({
   highlightedGeometries: [],
 });
 
+/**
+ * Selector path for highlighter overlay
+ *
+ * Used to position tooltip
+ */
+export const HIGHLIGHT_PATH_SELECTOR = 'echHighlighter__selectedElement';
```
markov00, line 64: I think we should remove this, it's no more used

src/components/portal/portal.tsx
```diff
@@ -0,0 +1,216 @@
...
+      setPopper();
+    }
+  }, [visible]);
+
+  const updateAnchorDimensions = useCallback(() => {
+    if (!position || !visible) {
+      return;
+    }
+
+    const { left, top, width, height } = position!;
```
markov00, line 182: The non-null assertion is not required here

src/components/portal/portal.tsx
```diff
@@ -0,0 +1,216 @@
...
+  const portalNode = useRef(getOrCreateNode(`echPortal${scope}`));
+
+  /**
+   * Popper instance used to manage position of tooltip.
+   */
+  const popper = useRef<Instance | null>(null);
+
+  const popperSettings = useMemo(
+    () => mergePartial(DEFAULT_POPPER_SETTINGS, settings, { mergeOptionalPartialValues: true }),
+    [settings, settings],
```
markov00, line 88:
```suggestion
    [settings],
```

src/components/portal/portal.tsx
```diff
@@ -0,0 +1,216 @@
+/*
```
markov00, line 1: I kindly suggest adding the react-hook eslint feedback rules (https://github.com/facebook/react/issues/14920) that provide hints and fixes to the current file

src/components/portal/portal.tsx
```diff
@@ -0,0 +1,216 @@
...
+    anchor,
+    (anchor as PortalAnchorRef)?.position,
+  ]);
+
+  const destroyPopper = useCallback(() => {
+    if (popper.current) {
+      popper.current.destroy();
+      popper.current = null;
+    }
+  }, [popper.current]);
```
markov00, line 101: from eslint feedback:
> React Hook useCallback has an unnecessary dependency: 'popper.current'. Either exclude it or remove the dependency array. Mutable values like 'popper.current' aren't valid dependencies because mutating them doesn't re-render the component.

src/components/tooltip/tooltip.tsx
```diff
@@ -17,59 +17,189 @@
...
+      <div className="echTooltip">
+        {renderHeader(info.header)}
+        {renderValues(info.values)}
+      </div>
+    );
+  };
+
+  const anchorPosition = useMemo((): AnchorPosition | null => {
+    if (!isVisible) return null;
+    const { x0, x1, y0, y1 } = position!;
```
markov00, line 138: I think we should prefer to avoid non-null assertions

integration/page_objects/common.ts
```diff
@@ -28,6 +28,64 @@ const baseUrl = `http://${host}:${port}/iframe.html`;
...
+
+  if (top !== undefined || bottom !== undefined) {
+    if (top !== undefined) {
+      y = element.top + top;
+    } else {
+      y = element.top + element.height - bottom!;
+    }
+  }
+
+  if (left !== undefined || right !== undefined) {
```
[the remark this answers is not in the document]
nickofthyme: I generally agree but in this case I don't think ternaries improve readability since the logic seems to require nested ternaries. Since this is not `src` code I'll keep it as is for now, but feel free to change this if you like.

src/components/tooltip/tooltip.tsx
nickofthyme: What do you mean by this?

src/components/portal/portal.tsx
nickofthyme: e7256b68466237132c6bbc95943cb9ceb01fe6c9

src/components/portal/portal.tsx
nickofthyme: e7256b68466237132c6bbc95943cb9ceb01fe6c9

src/components/portal/portal.tsx
nickofthyme: e7256b68466237132c6bbc95943cb9ceb01fe6c9

src/components/portal/portal.tsx
nickofthyme: ddc44cb3d2de2c179f6ee887c6675e3fe7ead8bd

src/components/tooltip/tooltip.tsx
nickofthyme: ddc44cb3d2de2c179f6ee887c6675e3fe7ead8bd

src/chart_types/xy_chart/renderer/dom/highlighter.tsx
```diff
@@ -56,7 +59,7 @@ class HighlighterComponent extends React.Component<HighlighterProps> {
             <rect x="0" y="0" width={clipWidth} height={clipHeight} />
           </clipPath>
         </defs>
-        <g transform={`translate(${left}, ${top}) rotate(${chartRotation})`}>
+        <g id={HIGHLIGHT_PATH_SELECTOR} transform={`translate(${left}, ${top}) rotate(${chartRotation})`}>
```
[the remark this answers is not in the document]
nickofthyme: 9d43a4bebf69863463ae5e91dd9437980eab4b45

src/chart_types/xy_chart/state/selectors/get_tooltip_values_highlighted_geoms.ts
nickofthyme: 9d43a4bebf69863463ae5e91dd9437980eab4b45

integration/page_objects/common.ts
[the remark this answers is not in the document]
nickofthyme: Will move this code to new branch, per your other comment. But will update this on the new branch.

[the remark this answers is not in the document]
nickofthyme: 26954a8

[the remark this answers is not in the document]
monfera: Nicely done 👍

nickofthyme:
@monfera yeah I added a eslint rule for react hooks that required some file changes, namely the correct component naming of functional components. This required changing all the `example` story variables to `Example`, but that is all contained in a single commit.

The screenshots are off because I was changing the story to allow more uniform padding around the chart. I will add this to the other PR with the other changes to vrt. see 884e6f55

monfera:
Thanks @nickofthyme - makes sense. I suppose it'll cause the tooltip(s) to return to those screengrabs too

monfera:
Eg. here the tooltip seems to have moved away
image
and there's another very similar one.

There's also this where, unlike with most cases, the tooltip positioning is less good than before (top protrudes, is clipped), is it a problem?
image

nickofthyme:
Yup. So the first screenshot is now containing the tooltip beyond the chart container, which is an expected change. This allows placement relative to the window or a defined `boundary` such as the chart container.

The second screenshot is also intended at the cursor is not centered relative to the tooltip. There is an option for `right-start` to position the tooltip how it currently is but I think this placement is better.

@markov00 / @rshen91 any thoughts on this? see demo below. Notice the vertical placement of the tooltip relative to the cursor.

### Current
Screen Recording 2020-05-27 at 12 53 PM


### New
Screen Recording 2020-05-27 at 12 54 PM

monfera:
Thanks @nickofthyme I played with it some more around the edge and it's very nice, there's no cropping of the tooltip so it's all good. That it looks cropped is simply due to where the test mouse position is and what area is screenshot

api/charts.api.md
```diff
@@ -678,9 +678,9 @@ export type HistogramModeAlignment = 'start' | 'center' | 'end';
 //
 // @public (undocumented)
 export const HistogramModeAlignments: Readonly<{
-    Start: HistogramModeAlignment;
-    Center: HistogramModeAlignment;
-    End: HistogramModeAlignment;
+    Start: LineAlignSetting;
+    Center: LineAlignSetting;
+    End: LineAlignSetting;
```
markov00, line 683: weird....I will open a PR to fix this

nickofthyme: Thanks, wasn't sure what this was from.

src/components/portal/types.ts
```diff
@@ -38,6 +39,11 @@ export const Placement = Object.freeze({
   AutoStart: 'auto-start' as 'auto-start',
   AutoEnd: 'auto-end' as 'auto-end',
 });
+
+/**
+ * {@inheritDoc (Placement:variable)}
+ * @public
```
nickofthyme, line 45: @markov00 Is this required? I thought `@public` was the default.

src/specs/settings.tsx
```diff
@@ -183,6 +211,11 @@ export interface TooltipProps {
    * @defaultValue parent scroll container
    */
   boundary?: HTMLElement | 'chart';
+  /**
+   * Unit for event (i.e. `time`, `feet`, `count`, etc.).
+   * Not currently used/implemented
+   * @alpha
+   */
   unit?: string;
```
nickofthyme, line 219: 👍

Issue #611, opened 2020-03-27
Improved category labels for treemap


wylieconlon:
Category labels in treemaps are all drawn in a centered way, which causes them to overlap. An improvement that some treemap implementations use is to shift the labels above the rectangle, such as:

![treemap](https://user-images.githubusercontent.com/666475/77788096-08b7d900-7037-11ea-9ee5-4a0a95b8c80f.png)

Issue #612, opened 2020-03-27
[meta] Lens use of partition charts


wylieconlon:
Trying to organize the feedback that we've given about pie and treemap charts into one place

Minimum expectations:

* [ ] Tooltip scrollbar bug #596 
* [ ] Legend for first layer to show category colors #532
* [ ] Color contrast by luminance #606 

Nice to have:

* [ ] Tooltip to show field + value #532 
* [ ] Nested legend #223 
* [ ] Better label placement #599 
* [ ] Tiny slices should be grouped during hover #381 
* [ ] Labels per group in treemaps #611 

For developer friendliness:

* [ ] Color theming #610

monfera:
@wylieconlon upon today's release of #652, I split one row and checked a couple of boxes in the Description todo list, as the PR didn't address everything (can't be marked Closed)

Issue #599, opened 2020-03-23
Inconsistent placement of `fillLabel`s 


cchaos:
**Describe the bug**
There doesn't seem to be a consistent placement of labels inside of pie, donut, treemap charts.

**Expected behavior**
I'd expect them all to appear to be in a consistent place with the consumer able to specify a region like "center", "top left", etc...

**Screenshots**
![Screen Shot on 2020-03-23 at 16:27:10](https://user-images.githubusercontent.com/549577/77360226-655b8100-6d23-11ea-81e0-1663b68f0f17.png)

monfera:
There are two kinds of sectors: horizontal for better readability in wider sectors, and radial in case of narrow sectors. There's an iteration that tries to maximize readability via largest fitting font size, up to the maximum allowed font size which is relatively small here. 

As your examples show, there's plenty of room for improvement, we should circle back to this. Unfortunately, increasing the consistency of the location, while keeping to the possible font maximization logic is not a quick fix: here the layout is a bit underconstrained as the maximum font size is fairly small, so it's kinda loose in the shape, it'll need a bit of thinking and refinement.

monfera:
Also, maybe it's possible to classify it as an improvement opportunity rather than a bug, especially that fill text for sunburst, and treemaps in general are new features and our text layout is more refined than what's available in other tools, so the above placement inconsistencies don't feel like a blocker.

monfera:
Hi @cchaos https://github.com/elastic/elastic-charts/pull/652 should address the treemap part of this issue. Having said that, there's opportunity for improvement, maye slightly more or configurable text padding is needed, the current one maxes out the rectangular area for text.

The sunburst part is also not that trivial to do - I think many or most of what you suggest there are fairly straightforward, but need time to implement properly, so it should be subject to planning and prioritization so we work through issues in the best sequence.

PR #652, opened 2020-04-27
feat(partition): treemap group text in grooves

monfera:
## Summary

- Puts treemap text for groups into top grooves
- Treemap texts are left-aligned and brought to top (groove texts are on the bottom)
- Font min/max sizes can be configured per Layer

To limit areal distortion, the groove heights are capped by a proportion of the area.

image

Closes https://github.com/elastic/elastic-charts/issues/611
Fixes the treemap part of https://github.com/elastic/elastic-charts/issues/599
Ticks a few boxes in https://github.com/elastic/elastic-charts/issues/612

Further iteration ideas:
- expose constraints via configurability, eg. maximal area distortion
- adherence to text padding - now it exploits maximum area

### Checklist


src/chart_types/partition_chart/layout/viewmodel/viewmodel.ts
```diff
@@ -59,11 +58,29 @@ import {
 } from '../utils/group_by_rollup';
 import { StrokeStyle, ValueFormatter } from '../../../../utils/commons';
 import { percentValueGetter } from '../config/config';
+import { $Values } from 'utility-types';
+
+// todo consider turning it into a config option
+const topGroove = 20;
```
markov00, line 64: I think we should move this into the `config`prop, so at least the consumer can decide to show or hide the groove text changing this value to 0 or to a boolean false.

src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
```diff
@@ -183,32 +174,35 @@ export function getSectorRowGeometry(
 
 /** @internal */
 export function getRectangleRowGeometry(
-  container: any,
+  container: RectangleConstruction,
```
nickofthyme, line 177: 🎉

src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
```diff
@@ -183,32 +174,35 @@ export function getSectorRowGeometry(
...
-    + fontSize / 2
-    - linePitch * rowIndex
-
-  const rowCentroidX = cx;
-  const rowCentroidY = cy - offset;
+  const overhang = 0.05;
+  const padding = fontSize < 6 ? 0 : Math.max(1, Math.min(2, fontSize / 16)); // taper out padding with small fonts
+  if ((container.y1 - container.y0 - 2 * padding) / totalRowCount < linePitch) {
+    return {
+      rowCentroidX: NaN,
```
nickofthyme, line 191: curious why you prefer this over setting this to null or otherwise?

src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
```diff
@@ -183,32 +174,35 @@ export function getSectorRowGeometry(
...
+      rowCentroidY: NaN,
+      maximumRowLength: 0,
+    };
+  }
+  const rowCentroidY =
+    verticalAlignment === VerticalAlignments.top
+      ? -(container.y0 + linePitch * rowIndex + padding + fontSize * overhang)
+      : verticalAlignment === VerticalAlignments.bottom
+      ? -(container.y1 - linePitch * (totalRowCount - 1 - rowIndex) - fontSize * overhang)
+      : -((container.y0 + container.y1) / 2 + (linePitch * (rowIndex - totalRowCount)) / 2);
```
nickofthyme, line 201: This is fine for this PR, but nested ternarys are screaming to be a separate function. Makes it so much cleaner and easier to read. 😱

src/chart_types/partition_chart/layout/viewmodel/viewmodel.ts
```diff
@@ -59,11 +58,29 @@ import {
...
+}
+
+export const VerticalAlignments = Object.freeze({
+  top: 'top' as 'top',
+  middle: 'middle' as 'middle',
+  bottom: 'bottom' as 'bottom',
+});
+
+/** @internal */
+export type VerticalAlignment = CanvasTextBaseline & $Values<typeof VerticalAlignments>;
```
nickofthyme, line 81: This kind of combined type is confusing. I think these should be separate types. This implies that the value `VerticalAlignments` has the same types as the `VerticalAlignments` which is not the case.

Fro example `VerticalAlignments` object does not have `'alphabetic'` as a possible value.

```suggestion
export type VerticalAlignment = $Values<typeof VerticalAlignments>;
export type AllVerticalAlignment = CanvasTextBaseline & VerticalAlignment;
```

src/chart_types/partition_chart/renderer/canvas/canvas_renderers.ts
```diff
@@ -35,20 +35,37 @@ import { clearCanvas, renderLayers, withContext } from '../../../../renderers/ca
...
       ctx.translate(crx, cry);
       ctx.rotate(-rotation);
       ctx.fillStyle = fillTextColor;
+      ctx.textBaseline = verticalAlignment;
       currentRow.rowWords.forEach((box) => {
         ctx.font = cssFontShorthand(box, fontSize);
         ctx.fillText(box.text, box.width / 2 + box.wordBeginning, 0);
       });
     });
+    if (0)
```
nickofthyme, line 58: Is this going to be removed? If not I don't think it's good adding unused code before we need it.

src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
monfera: Nick, fair question, my intent was to stick to the `number` type, ie. I didn't want to loosen the type spec just because of an initial value, and `NaN` is an IEEE as well as JS number. It'd be possible to use `null` and upgrade type once an initial value is promoted to an actual value, but it's a bunch of extra code and tedium.

So it's a bit like using an empty string for a string rather than `null` in some identity or initial object; maybe even better distinction, because an empty string can still be a legit value, eg. no text somewhere.

Also, not for now, and not for this chart type, but if we ever want to speed up `Datum[]` length constrained code sections, one option is to put values in typed JS arrays, eg. Float64Array, in which case `NaN` still works. Even if we end up using a normal, untyped array, the JS engines internally map it to eg. Float64Array if the contents are homogeneous, so it became my habit to not mix numbers with other things when not needed.

src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
monfera: Oops, meant to break it up, thanks for spotting this

src/chart_types/partition_chart/layout/viewmodel/viewmodel.ts
monfera: It can certainly be broken up eg.
```
type PermittedVerticalAlignments = $Values<typeof VerticalAlignments>;
export type VerticalAlignment = CanvasTextBaseline & PermittedVerticalAlignments;
```
which just lifts and names the `$Values` part. Would that be clearer code? Your suggestion is the same extraction but I don't follow the namings as `AllVerticalAlignments` seems to hint that it's a broader / union set, while the purpose of the `&` is to intersect the two sets. Also, both do not need to be exported, but maybe you have some non-local change in mind.

OK why the heck did I do this in the first place, ie. why not just the expected `$Values<typeof VerticalAlignments>`? The `&`-ing will not in fact remove values from `$Values<typeof VerticalAlignments>` (one is a proper subset of the other). The purpose is to show intent and assurance to the compiler and code reader that 
- values are expected to be directly digestible by the Canvas2 API without compile time type juggling or worse, runtime type guards, as it's directly channeled into a `ctx.textBaseline` so it better be of `CanvasTextBaseline`(*)
- yet, as you say, we don't currently expose all the alignments

There is a larger, somewhat theme and it'd be great to talk about it on account of this, we touched on esp. text styling configurability superficially with @markov00 and it'd be opportune to use this as a small example.

src/chart_types/partition_chart/renderer/canvas/canvas_renderers.ts
monfera: Yes, I usually diff the code for items like this before the final push, still a reminder is most welcome

monfera: I ended up commenting out the code:
- exposing it under a `debug` config flag would've been tedious to thread through, and it's not user-exposed (we could turn it into a `debug` config option eventually
- yet if anybody esp. beside me works with text, the snippet is not _that_ trivial, given all the projections going on - and it's very useful because texts don't show the allotted box well, for a number of reasons

monfera:
@markov00 on the horizontal misalignment between group and leaf labels: there's still a padding at the group level, but it's leftward a bit, because the group wraps all the leafs, and to cause space between groups, it's larger (btw. the large groove on the top is just an increase of this grooves, but the left, right and bottom grooves aren't zeros either).

It'd be good to further improve on the alignment but it could be a small, follow-up prioritized item as we can do multiple things here - currently I'm leaning toward changing the groove sizing such that the left group groove is zero and the right group groove is double of what it is now, but it needs a bit of t(h)in(er)king. An alternative is to add an extra constraint on the groove text box but that's some additional code path. Just offsetting the text would be a third option, but it'd feel like a hack (long text might protrude on the right) and the vertical alignment would be accidental.

src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
monfera: I extracted it out into a function with `switch` / `case` 
```
1 file changed, 31 insertions(+), 7 deletions(-)
```
and while there are best practices that morph into rules with broad community approval, eg. don't nest ternaries, I'm not sure if it makes the code much more readable, at least I personally don't mind short, chained ternaries, they're easier for me to read than esp. 
```
let rowCentroidY = null; 
if(....) {
  rowCentroidY = ....
} else if { 
.
.
.
```
because here there's no guarantee that something else further down doesn't modify the let binding.

src/chart_types/partition_chart/layout/viewmodel/viewmodel.ts
monfera: done in https://github.com/elastic/elastic-charts/pull/652/commits/b2da71b07fe37c2c328e701ce5dc674a05b655db

src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
monfera: done in https://github.com/elastic/elastic-charts/pull/652/commits/6d3aceaf5fbd6e6ecb84a59f9639a8d14a4e7980

src/chart_types/partition_chart/layout/viewmodel/viewmodel.ts
monfera: broken up, as mentioned, in a slightly different way in https://github.com/elastic/elastic-charts/pull/652/commits/cb141a6402abf7918c8e54e8e2803d5c872f66dd

src/chart_types/partition_chart/renderer/canvas/canvas_renderers.ts
monfera: done in https://github.com/elastic/elastic-charts/pull/652/commits/91013f955a66ce8f2e6db33246ac54cc3b1d903b

src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
nickofthyme: good analogy to the empty string. Sounds fair, I have normally just assumed number would never be `NaN` but that clearly is no longer valid. Right, that's a good point about homogeneous efficiency!

src/chart_types/partition_chart/layout/viewmodel/viewmodel.ts
nickofthyme: Yeah the `AllVerticalAlignments` was just an arbitrary name I cam up with. 

The change in cb141a6 is close to what I was referring to but the key is that the object with the js values and the type should be called the same, that way ts can import the Type and Values as one import.

```ts
// also using this
type SupportedVerticalAlignments = $Values<typeof VerticalAlignments>;

// makes this look strange
const myThing: SupportedVerticalAlignments = VerticalAlignments.top
```

These values are basically the poor mans `enum` because tsc enums are not ideally transpiled (see https://github.com/elastic/kibana/pull/62957#discussion_r406385788), with this thinking, your enum values in this case would be different than the enum type.

```ts
export const VerticalAlignments = Object.freeze({
  top: 'top' as 'top',
  middle: 'middle' as 'middle',	
  bottom: 'bottom' as 'bottom',
});
type VerticalAlignments = $Values<typeof VerticalAlignments>;

// then it becomes much clearer how the type and enum values relate
type VerticalAlignments = $Values<typeof VerticalAlignments>;
```

Could be something to talk about today at the sync 🤷

src/chart_types/partition_chart/renderer/canvas/canvas_renderers.ts
nickofthyme: I'm ok with this for now but I don't think comments are much better. If this is a debug property and you want this to persist in the code. Then I think we wire it up the the `settings.debug` option and make it a permanent home 🏡 .

src/chart_types/partition_chart/layout/viewmodel/viewmodel.ts
monfera: Yes, the enum would have various weaknesses here, one of which here is that it's not some opaque thing I want here, but a subtype of `CanvasTextBaseline` which is a very real set of strings.

If I understand, your remaining ask is that the object and type be named identically; it's doable, though with a bit less directness here as there's the above mentioned goal of piggybacking on the `CanvasTextBaseline` type, in https://github.com/elastic/elastic-charts/pull/652/commits/9c300904c6cdedf486265ffe042742b4779bafd9

PR #653, opened 2020-04-28
test: improve VRT developer experience

markov00:
## Summary

This PR renames `yarn jest:integration` script to `yarn test:integration` to align the name to the rest of the test scripts we have.

This PR adds one more script to run the VRT directly against a local storybook server:
```
yarn test:integration:local
```
This will reduce the VRTs time removing the build time spent on the storybook build.

I've updated the docs to include the use of `--testNamePattern` to limit the test on a specific regex instead of the previously implemented env variable in this PR.


.ci/run.sh
```diff
@@ -131,4 +131,4 @@ echo " -- upload code coverage"
 ### visual testing
 ###
 echo " -- visual testing"
-yarn jest:integration --ci
+yarn test:integration --ci
```
nickofthyme, line 134: 👏

integration/defaults.js
```diff
@@ -19,4 +19,5 @@
 module.exports = {
   PORT: '9009',
   HOST: 'host.docker.internal',
+  USE_LOCAL_STORYBOOK: false,
 };
```
nickofthyme, line 23: We should also set `ALL_STORIES_FILTER` to prevent this from being defined in the ci somehow where it only runs a subset of the VRT which I don't think would cause an error in the ci.

markov00: done here: https://github.com/elastic/elastic-charts/pull/653/commits/b89d258069a3a8f8561cf8a488e7c690bc20d689

markov00: I've completely removed that env variable in favour of jest arguments: https://github.com/elastic/elastic-charts/pull/653/commits/4f3005b4a298ed90cd8d5b885c01b338dffd57de

integration/helpers.ts
```diff
@@ -72,29 +65,42 @@ function encodeString(string: string) {
...
+export function getStorybookInfo(filters?: string[]): StoryGroupInfo[] {
   configure(requireAllStories(__dirname, '../stories'), module);
-  return getStorybook()
-    .filter(({ kind }) => kind)
+  const regexFilters =
+    filters !== undefined
+      ? filters.map((filter) => {
+          return new RegExp(filter, 'i');
+        })
+      : undefined;
```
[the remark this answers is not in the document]
markov00: :( my bad, I completely forgot about that, I've removed that variable...https://github.com/elastic/elastic-charts/pull/653/commits/4f3005b4a298ed90cd8d5b885c01b338dffd57de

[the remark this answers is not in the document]
nickofthyme: Yeah I wasn't sure if you did this because that flag didn't work for the looped test cases.

wiki/testing.md
```diff
@@ -28,3 +28,52 @@ your_file_name.tz.test.jp.ts
...
+yarn test:integration
+```
+
+This will start a separate storybook server and a docker machine with chromium, controlled by pupetteer and jest, that takes and compare screeshots.
+
+If the screenshot differ from the baseline, a test error is rised and a diff image is stored in `integration/test/__image_snapshots__/__diff_output__`.
+
+If a new test is added, a new screenshot `.png` file is written as part of the baseline.
+
+To update the snapshots to the new version run:
```
nickofthyme, line 50: Not that this really matters but I'd like to be consistent calling these screenshots and referring to snapshots as the actual jest dom snapshots.

If needed we could also change the output directory to `__screenshots__`.

```suggestion
To update all existing screenshot baselines to the new version run:
```

PR #654, opened 2020-04-28
fix(partition): consider legendMaxDepth on legend size

### LINKED ISSUE #639: Partition legend width is not considering legendMaxDepth
I noticed this based on data that I was using in Lens, but this can be reproduced in the storybooks.

When I add a legend and `legendMaxDepth={1}` to to  `8_sunburst_two_layers.tsx`, the legend is given the exact same width as when `legendMaxDepth` is not set, which fits the largest country name in the list:

`<Settings showLegend={true} legendMaxDepth={1} />`

This wastes space, as seen in the screenshot:

Screenshot 2020-04-20 14 54 46

markov00:
## Summary

This PR fix the legend size computation for Partition charts taking in consideration the `legendMaxDepth` value if present.

fix #639

### Checklist


src/chart_types/partition_chart/state/selectors/get_legend_items_labels.ts
```diff
@@ -24,18 +24,37 @@ import { HierarchyOfArrays, CHILDREN_KEY } from '../../layout/utils/group_by_rol
...
+
+    return labels;
   },
 )(getChartIdSelector);
 
+/**
+ * Check if the legendMaxDepth from settings is not a valid number (NaN or <=0)
+ * @param legendMaxDepth - Pick<SettingsSpec, 'legendMaxDepth'>
+ */
+function isInvalidLegendMaxDepth({ legendMaxDepth }: Pick<SettingsSpec, 'legendMaxDepth'>): boolean {
```
nickofthyme, line 54:
```suggestion
function isInvalidLegendMaxDepth(legendMaxDepth: SettingsSpec['legendMaxDepth]): boolean {
```

nit: feel free to ignore...

markov00: done in cee4b63
thanks for the tip ❤️

PR #662, opened 2020-05-04
docs: add api-extractor report

markov00:
## Summary

This PR activates the [api-extractor report](https://api-extractor.com) allowing a semi-automated check about possible breaking changes on the current published API contract. 
The extractor outputs a Markdown report file called `api/charts.api.md`. The following is an excerpt of that file:

## API Report File for "@elastic/charts"

> Do not edit this file. It is a report generated by [API Extractor](https://api-extractor.com/).
```ts
import { $Values } from 'utility-types';
import React from 'react';
// @public
export type AccessorFn = UnaryAccessorFn;
// @public
export type AnnotationDomainType = $Values<typeof AnnotationDomainTypes>;
// @public
export const AnnotationDomainTypes: Readonly<{
    XDomain: "xDomain";
    YDomain: "yDomain";
}>;
// Warning: (ae-missing-release-tag) "AnnotationId" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
//
// @public (undocumented)
export type AnnotationId = string;
// Warning: (ae-missing-release-tag) "AnnotationSpec" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
//
// @public (undocumented)
export type AnnotationSpec = LineAnnotationSpec | RectAnnotationSpec;
// Warning: (ae-missing-release-tag) "AnnotationTooltipFormatter" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
//
// @public (undocumented)
export type AnnotationTooltipFormatter = (details?: string) => JSX.Element | null;
// Warning: (ae-missing-release-tag) "AnnotationType" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
//
// @public (undocumented)
export type AnnotationType = $Values<typeof AnnotationTypes>;

...
...
```

This file shows all the public/beta/alpha type/interface/function exposed as part of our public API (visible when importing elastic/charts into a project).

It includes a set of **warnings** related to missing tags or wrongly used tags, for example:
```ts
// Warning: (ae-missing-release-tag) "AnnotationSpec" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
//
// @public (undocumented)
export type AnnotationSpec = LineAnnotationSpec | RectAnnotationSpec;

// src/chart_types/partition_chart/layout/types/config_types.ts:112:5 - (ae-forgotten-export) The symbol "TimeMs" needs to be exported by the entry point index.d.ts
```
The warnings in this example can be interpreted as:
- `AnnotationSpec` type is undocumented (no TSDoc is provided)
-  The release tag (`@alpha`,`@beta`,`@public`) is missing
- a type, in particular `TimeMs`, used by other exported types, is not exported by the main entry point.

This PR adds also a GH workflow that runs the `yarn api:check` commands as a GH check and warns the PR author about possible changes of the current API document.

Locally, you can run the same check running `yarn api:check --local` adding the `--local` will automatically update the `api/charts.api.md` file.

When this will be merged and during a PR review, it will be always required to take a careful look at the `api/charts.api.md`. If something breaks the existing contracts like a type declaration is changed, renamed, or removed then we should mark the PR as breaking change with a  Screenshot 2020-05-06 at 16 36 52 label.

Future tasks:
- cleanup existing API
- automatically adds the `breaking change` label if the API contract is broken
- generate a small `breaking change` changelog and add that to the PR description


src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -278,7 +280,7 @@ export interface SeriesSpec extends Spec {
    */
   name?: SeriesNameAccessor;
   /** The ID of the spec group
-   * @default __global__
+   * @defaultValue {@link DEFAULT_GLOBAL_ID}
```
nickofthyme, line 283: Nice

src/specs/settings.tsx
```diff
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License. */
 
-import { ComponentType } from 'react';
+import React from 'react';
```
nickofthyme, line 19: Does this not work with selective imports?

tsconfig.lib.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "noUnusedLocals": true,
-    "removeComments": true
+    "removeComments": false
```
nickofthyme, line 4: Does the extractor not work if this is set to true? If so I wonder if we should just have a separate `tsconfig` for the API documenter.

src/specs/settings.tsx
markov00: No, the issue here is different, we should always import `React` on TSX files to avoid issues with the api-extractor that's need that all files using JSX should import React in a consistent way https://github.com/microsoft/rushstack/issues/1765

tsconfig.lib.json
markov00: What I came up with is the following:
we can first generate the declaration files with all the comments, and then we recompile again only generating the js (no declarations) with comments.
This will increase the build type, but also avoid any third party library that can touch the generated code in an unpredicted way

markov00: here the commit: https://github.com/elastic/elastic-charts/pull/662/commits/b4edfa30151570e6c46c2d1072a0d078252a8c8a

PR #670, opened 2020-05-07
feat(partition): linked text overflow avoidance

### LINKED ISSUE #633: Pie chart linked labels are being displayed in cut off way
**Describe the bug**
When the horizontal or vertical space to display the linked labels is too small, the labels are getting cut off. This is visible in the "Most basic" sunburst example, since it has long labels that display as linked labels:

Screenshot 2020-04-16 10 52 57

cc @cchaos 

**Expected behaviour**
I would expect one of two things to happen, depending on which is easier:

1. Labels would always push the main circle smaller to ensure that they are shown (this is what I thought was happening)
2. If the labels are too long, they would be hidden

I think 1. is my preference for Lens, but maybe there are use cases for 2.

**Screenshots**

18.3.0

monfera:
## Summary

Truncates linked label text to fit in the chart box (WIP)

Example with off-center pie chart to avoid centered special case:
image

It culls a link label altogether if
- its text has no horizontal room left
- the text would overflow the rectangular area vertically; eg. no linked label for the green `Food...` slice:
image

Closes #633 

### Checklist


monfera:
Removed the `wip` label as it has the vertical space feature as well.
The image tests capture the desired effect to a greater extent than unit tests could. I also have plans to revisit it for improvements, eg. more accurate maximalization of horizontal space.

The unit testing plan for a subsequent PR is the following:
- extract out and maybe rename the maximizer function which is basically a greedy search, gradient descent function
- switch to this function for the determination of font sizes for fill text, which currently iterates through the font sizes linearly, and not yet modular anyway
- the optimizer thus becomes exported, add unit tests for it

src/chart_types/partition_chart/layout/viewmodel/link_text_layout.ts
```diff
@@ -71,43 +79,117 @@ export function linkTextLayout(
...
+        verticalOffset,
         labelFontSpec,
         valueFontSpec,
       };
-    });
+    })
+    .filter((l: LinkLabelVM) => l.text !== ''); // cull linked labels whose text was truncated to nothing
+}
+
+function monotonicMaximizer(
```
markov00, line 136: nit: Could you please add a jsdoc describing what this function does?
It's a little cryptic and I've difficulties understanding what is going on there

markov00: Please, also add the return type

monfera: Thanks Marco, will do, the upcoming PR deals with and extracts out this function

PR #681, opened 2020-05-25
feat(partition): monotonic font size scaling

### LINKED ISSUE #661: Partition chart text should scale uniformly along sectors/rings
Shortest texts stand out more than longer texts in a treemap or pie chart because their fonts can take up much more space than longer texts. This can distort the represented data when the user gives a quick look without at the chart without giving some time to understand the whole picture.
Check the following examples:
Screenshot 2020-05-04 at 10 55 06

In this case, `Belgium` stands out much more than the Netherlands also if from the data it's the opposite and the difference is very small.

Screenshot 2020-05-04 at 10 51 32

In this other case, Asia and Europe on the second layer look much bigger than the United States one.

I'm imaging a possible configuration option to allow scaling each font uniformly, from the biggest to the smallest. Each sector should compute its max font size depending on the computed font-size of the previous one.

monfera:
## Summary

- Fixes #661 by implementing monotonic font size scaling
- Turns the font size selection from O(N) to O(log(N)), resulting in one order of magnitude speedup for complex cases (many nodes, broad font size range, fine font size increments)
- The hill climber, initially used for just linked label text truncation, is extracted and reused for this
- The hill climber accuracy was improved, and the PR exploits this by truncating less of the linked label texts
- Very large functions in `fill_text_layout.ts` got broken up, and some changes were needed by this (typed arg lists) as well as by switching to the hill climber, but otherwise its logic remained unchanged (fill part of mocks didn't change except the two changed 3-layer ones)
- Improves typing in general, adds some generics and retires a few `any`s

## Details

The feature implements per-layer monotonic font scaling.

image

The 3-layer sunburst and treemap have per layer knobs to see the effect.

image

The tradeoff is reduced scaling in legibility: a single bad case (eg. narrow box, long word) can yield a small, maybe unrenderably small text for all the subsequent boxes too, even if those could otherwise be rendered. For this reason, and also, not to cause a breaking change, the current PR has `layers: [ { fillLabel: { monotonic: false } } ]` as a default; can be changed if it's important enough to cause a breaking change. Also, maybe the current logic could be augmented by still allowing the use of the smallest font size even if there was a larger value with text that didn't fit even with the smallest font size.

The logic currently achieves monotonicity _per layer_. This is uncontroversial with the treemap as groove texts are bound to be small. For sunburst, it looks preferable too, because otherwise a single, small inner slice can lead to small fonts in all the outer rings. However it can still lead to cases where an outer sector has a larger font than an inner sector of lower value, eg. `Mineral fuels` vs `Machinery / North America` (can be the other way around too). Once we introduce global monotonicity, new tradeoffs of diminishing returns arise 😸 

The per layer monotonicity also means that font size constraints do not get reset for groups. In other words, if there are two top level groups A and B in a two-layer chart, then the font size of a node in Group A can't be larger than any of the font sizes of nodes in Group B that have higher values.

There's no constraint for linearity or other proportionality. So it's still possible to have unbalanced texts, eg. if the biggest box is `"China"` and a very close runner-up is `"United States"` then the font size for the latter will be unproportionately small, also depending on box aspect ratios.

The current issue and PR response leaves other aspects of "text ink weight perception" unaddressed: text lengths and even the character shapes in the texts impact visual weight. Further, possible improvements might help, but again, there are diminishing returns, so prioritization would be needed:

- shortening texts with `...`
- line break with hyphenation for long fill text words
- letting the chart choose from an alternative set of text lengths as eg. SAP does, eg. 
  - `"UK"` (or ISO `"GB"` / `"GBR"`)
  - `"United Kingdom"`
  - `"United Kingdom of Great Britain and Northern Ireland"`
- shoot for a horizontal aspect ratio for nodes with long words (general problem might be NP-hard)

These would bring improvements in addition to helping with text ink proportionality

### Checklist


src/chart_types/partition_chart/layout/config/config.ts
```diff
@@ -151,6 +151,10 @@ export const configMetadata = {
     type: 'number',
     reconfigurable: false, // there's no real reason to reconfigure it; finding the largest possible font is good for readability
   },
+  monotonic: {
+    dflt: false,
```
markov00, line 155: I don't have a strong opinion, but to me seems that a `monotonic` font scale avoids misunderstanding when reading the chart. I prefer having it `on` by default. The consumer then can opt to turn it off to maximize the font size depending on space. As you already highlighted in the PR description it's a tradeoff between readability and understandability.
@rshen91 @nickofthyme what do you think?
We should not be afraid of turning it on and mark the PR as breaking change, but we should prefer giving the user a good default.

src/chart_types/partition_chart/layout/types/config_types.ts
```diff
@@ -62,6 +62,7 @@ export interface FillFontSizeRange {
   minFontSize: Pixels;
   maxFontSize: Pixels;
   idealFontSizeJump: Ratio;
+  monotonic: boolean;
```
markov00, line 65: Add a tsdoc to describe the effect of this property `monotonic` 
```suggestion
  /** the fonts are scaled in.... **/
  monotonic: boolean;
```

src/chart_types/partition_chart/layout/utils/calcs.ts
```diff
@@ -56,3 +56,68 @@ export function colorIsDark(color: Color) {
...
+
+/** @internal */
+export function integerSnap(n: number) {
+  return Math.floor(n);
+}
+
+type NumberMap = (n: number) => number;
+
+/** @internal */
+export function monotonicHillClimb(
```
markov00, line 81: even if this method is internal, a quick description of this method in a tsdoc is useful

src/chart_types/partition_chart/layout/viewmodel/viewmodel.ts
```diff
@@ -182,7 +182,7 @@ function rectangleConstruction(treeHeight: number, topGroove: number) {
 }
 
 /** @internal */
-export function shapeViewModel(
+export function shapeViewModel<C>(
```
markov00, line 185: I don't see where we are using this generic, do we really need it?

src/chart_types/partition_chart/layout/viewmodel/link_text_layout.ts
```diff
@@ -98,10 +99,13 @@ export function linkTextLayout(
       };
       const translateX = stemToX + west * (linkLabel.horizontalStemLength + linkLabel.gap);
       const { width: valueWidth } = measure(linkLabel.fontSize, [{ ...valueFontSpec, text: valueText }])[0];
-      const widthAdjustment = valueWidth + 3 * linkLabel.fontSize; // gap between label and value, plus possibly 2em wide ellipsis
-      const allottedLabelWidth = rightSide
-        ? rectWidth - diskCenter.x - translateX - widthAdjustment
-        : diskCenter.x + translateX - widthAdjustment;
+      const widthAdjustment = valueWidth + 2 * linkLabel.fontSize; // gap between label and value, plus possibly 2em wide ellipsis
```
markov00, line 102: qq: the previous value `3` was a typo?

src/chart_types/partition_chart/layout/viewmodel/viewmodel.ts
monfera: Thanks, good catch, it was used at some point but no more, will remove

src/chart_types/partition_chart/layout/viewmodel/link_text_layout.ts
monfera: It was just an extra approx. one character offset on whichever side, as the truncation was less accurate than it is now. With the increase of accuracy there's no need for the extra padding

src/chart_types/partition_chart/layout/config/config.ts
monfera: Glad to switch, I was on the fence about it and the config is there either way. A single long word or narrow box can turn subsequent boxes less legible, if at all (see PR desc.) ie. assumes more about the data, iow more vulnerable input-wise. But, monotonically shrinking text is preattentively preferable 😄 The breaking change tipped the balance for me.

monfera: ...maybe it's not even a breaking change? We don't have an API contract that specifies text maximization. We should retain freedom for changing data and text ink that honors API contracts. Worth a discussion, a strict interpretation would require a breaking change for any pixel level visual change, while a too loose interpretation would throw curveballs to dependency maintainers

nickofthyme: Yeah I agree, should default to `true`

src/chart_types/partition_chart/layout/types/geometry_types.ts
```diff
@@ -31,40 +31,40 @@ export type Radius = Cartesian;
 export type Radian = Cartesian; // we measure angle in radians, and there's unity between radians and cartesian distances which is the whole point of radians; this is also relevant as we use small-angle approximations
 export type Distance = Cartesian;
 
-/* @internal */
+/** @internal */
```
nickofthyme, line 34: 👍

src/chart_types/partition_chart/layout/utils/calcs.test.ts
```diff
@@ -0,0 +1,127 @@
...
+ * under the License. */
+
+import { integerSnap, monotonicHillClimb } from './calcs';
+
+describe('monotonicHillClimb', () => {
+  const arbitraryNumber = 27;
+
+  describe('continuous functions', () => {
+    test('linear case', () => {
+      expect(monotonicHillClimb((n: number) => n, 100, arbitraryNumber)).toBeCloseTo(arbitraryNumber, 6);
```
nickofthyme, line 26: Love it! Someone using `toBeCloseTo` 🎉

src/chart_types/partition_chart/layout/utils/calcs.ts
```diff
@@ -56,3 +56,68 @@ export function colorIsDark(color: Color) {
...
+  const { r: tr, g: tg, b: tb, opacity: to } = stringToRGB(textColor);
+  const backgroundIsDark = colorIsDark(shapeFillColor);
+  const specifiedTextColorIsDark = colorIsDark(textColor);
+  const inverseForContrast = textInvertible && specifiedTextColorIsDark === backgroundIsDark;
+  return inverseForContrast
+    ? to === undefined
+      ? `rgb(${255 - tr}, ${255 - tg}, ${255 - tb})`
+      : `rgba(${255 - tr}, ${255 - tg}, ${255 - tb}, ${to})`
+    : textColor;
+}
```
nickofthyme, line 71: This will conflict with @rshen91 changes, might want to get with her to resolve conflicts.

src/chart_types/partition_chart/layout/config/config.ts
rshen91: I don't have strong feelings, but I think it's great we're making this available. I'm leaning towards the stance stated above, it's probably best to set it on by default and just have the option for consumers to set it off if readability suffers. I don't think it's a breaking change and it's nice that the consumer can always set it to false if they liked the font sizing beforehand.

src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
```diff
@@ -299,189 +307,254 @@ type GetShapeRowGeometry<C> = (
...
+      const {
+        textColor,
+        textInvertible,
+        fontStyle,
+        fontVariant,
+        fontFamily,
+        fontWeight,
+        valueFormatter,
+        padding,
+      } = Object.assign(
```
nickofthyme, line 350: can you add a type here that this merge is going to resolve to?

```suggestion
      }: MyType = Object.assign(
```

src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
```diff
@@ -299,189 +307,254 @@ type GetShapeRowGeometry<C> = (
...
-      const measurements = measure(fontSize, allBoxes);
-      const allMeasuredBoxes: RowBox[] = measurements.map(
-        ({ width, emHeightDescent, emHeightAscent }: TextMetrics, i: number) => ({
-          width,
-          verticalOffset: -(emHeightDescent + emHeightAscent) / 2, // meaning, `middle`,
-          wordBeginning: NaN,
-          ...allBoxes[i],
-          fontSize, // iterated fontSize overrides a possible more global fontSize
-        }),
+      const valueFont = Object.assign(
```
nickofthyme, line 360: can you add a type here that this merge is going to resolve to?

```suggestion
      const valueFont: MyType = Object.assign(
```

src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
```diff
@@ -299,189 +307,254 @@ type GetShapeRowGeometry<C> = (
...
+  leftAlign: boolean,
+  container: C,
+  getShapeRowGeometry: GetShapeRowGeometry<C>,
+  cx: Coordinate,
+  cy: Coordinate,
+  padding: number,
+  node: ShapeTreeNode,
+  boxes: Box[],
+  maxRowCount: number,
+) {
```
nickofthyme, line 412: return type?

src/chart_types/partition_chart/layout/utils/calcs.ts
rshen91: We met via zoom - merge conflicts should be minimal or at least easy to resolve with PR #608

monfera: Good call, we were on a call to discuss this aspect (and other coloring PR questions) today

src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
monfera: There's currently no specific type for this, and making one up for this wouldn't be terribly informative as it's just the reiteration of what's in the `const` bracket. I'm planning further extraction from this function (not in this PR though) as it's still too large, in which case a proper boundary, carrying these resolved/cascaded values, needs to arise. But it needs a bit of design work, glad to chat about it on a call separately or on the team sync, the upshot is that I feel uncomfortable typing this transitional thing without that design. Still open to being convinced if I can learn about the reasons 👍

src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
monfera: See prev. answer

src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
monfera: Thanks, on your feedback, I'm about to add 
```
  return function(initialRowSet: RowSet, fontSize: Pixels): { rowSet: RowSet; completed: boolean } {
```
as it's useful; I feel it wouldn't add more documentational value if I recapitulated the signature of the returned function at the outer function `function tryFontSize` as everything is laid out. Pls. see it the upcoming commit and hit me up if it doesn't suffice

src/chart_types/partition_chart/layout/config/config.ts
monfera: Thanks for the common thinking, done in e166746701376d2810b075dc77d2304584e7ea01

src/chart_types/partition_chart/layout/types/config_types.ts
monfera: 0629ff28970df61b527b674d1d3ea78bf6dded15

monfera: note: it got renamed in a subsequent commit https://github.com/elastic/elastic-charts/commit/e166746701376d2810b075dc77d2304584e7ea01

src/chart_types/partition_chart/layout/utils/calcs.ts
monfera: 2537397333f6bd1b757b505697e0518ba2293c4e

src/chart_types/partition_chart/layout/viewmodel/viewmodel.ts
monfera: 9372a7e4d53afd7e555e2c3e76122635aeffc393

src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
monfera: 853f1fa9c8492d25281bb32350f1547c0eebcda0 adds the type improvement for the inner function; as it's an immediate and unconditional return, typing the outer function feels unnecessary but glad to chat about it

src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
nickofthyme: Sure that's fine, this merging of several values is just hard to decipher. Even if the types were defined inline that would be better, something like...

```suggestion
      }: {
        fontWeight: number,
        valueFormatter: () => string,
        padding: number,
      } = Object.assign(
```

nickofthyme: Or even a combination of types that satisfy the resolved values, surely you have all the types to combine into the correct values.

```suggestion
      }: MyFontType & Pick<MyStyleType, 'style'> = Object.assign(
```

PR #684, opened 2020-05-27
build: remove testing hooks and add random seed 

markov00:
## Summary

This PR removes the testing from the pre-commit hooks: we are already running the same set of tests on CI and we don't need to run them locally on every commit. Everyone has the ability to run tests independently from the commit.

We are still running `eslint` and `prettier` on each commit. `eslint` runs now on a cache to check only changed files (we should carefully check how this behaves when working on different branches) 

I've changed also the order of the tasks on the commit hook, now it executes first the `commitlint` to check the commit conventions and then runs the linters. Unfortunately, there is no other way to execute the `commit-msg` hook before the `pre-commit` hook in Husky (https://github.com/typicode/husky/issues/441)

The PR also adds the `RNG_SEED` env variable when running storybook, so we can run VRTs seamlessly on CI and on the local instance (both local versions: `test:integration` and  `test:integration:local`).

The PR removes the `yarn.lock` file from the shipped release, reducing the npm pack size by half.


.huskyrc.js
```diff
@@ -2,13 +2,6 @@ const tasks = (arr) => arr.join(' && ');
...
   hooks: {
-    'pre-commit': tasks([
-      'pretty-quick',
-      'yarn run typecheck:all',
-      'yarn run lint',
-      'yarn run test',
-      'yarn run test:tz',
-    ]),
-    'commit-msg': 'commitlint -E HUSKY_GIT_PARAMS',
+    'commit-msg': 'commitlint -E HUSKY_GIT_PARAMS && pretty-quick && yarn run lint',
```
nickofthyme, line 5: I think the `typecheck:all` could be included too, pretty quick execution.

package.json
```diff
@@ -26,16 +26,16 @@
...
     "playground:ie": "cd .playground && webpack-dev-server --host=0.0.0.0 --disable-host-check --useLocalIp",
     "pq": "pretty-quick",
     "semantic-release": "semantic-release",
     "start": "yarn storybook",
-    "start-docs": "start-storybook -p 8001 -c .storybook-docs --ci",
-    "storybook": "start-storybook -p 9001 -c .storybook --ci",
-    "storybook:build": "rm -rf .out && build-storybook -c .storybook -o .out",
+    "start-docs": "RNG_SEED='elastic-charts' start-storybook -p 8001 -c .storybook-docs --ci",
+    "storybook": "RNG_SEED='elastic-charts' start-storybook -p 9001 -c .storybook --ci",
+    "storybook:build": "RNG_SEED='elastic-charts' rm -rf .out && build-storybook -c .storybook -o .out",
```
nickofthyme, line 38: I don't think this should be on `build`. I would like to see randomized data on the deployed version of storybook.

package.json
```diff
@@ -26,16 +26,16 @@
...
     "lint:fix": "yarn lint --fix",
     "playground": "cd .playground && webpack-dev-server",
     "playground:ie": "cd .playground && webpack-dev-server --host=0.0.0.0 --disable-host-check --useLocalIp",
     "pq": "pretty-quick",
     "semantic-release": "semantic-release",
     "start": "yarn storybook",
-    "start-docs": "start-storybook -p 8001 -c .storybook-docs --ci",
-    "storybook": "start-storybook -p 9001 -c .storybook --ci",
-    "storybook:build": "rm -rf .out && build-storybook -c .storybook -o .out",
+    "start-docs": "RNG_SEED='elastic-charts' start-storybook -p 8001 -c .storybook-docs --ci",
```
nickofthyme, line 36: What is the purpose of adding this to the docs? Do we take screenshots of this?

markov00: none actually, just to keep that consistent with the other runs

package.json
markov00: ok, I don't have strong opinion on that, I can remove it

.huskyrc.js
markov00: it takes 15-20 seconds, I don't have also a strong opinion on that, I'm adding it

markov00: a5a15ba

package.json
markov00: done https://github.com/elastic/elastic-charts/pull/684/commits/68e429910329ba9cd6cbcd70d7d5118ce24496fd

package.json
markov00: done https://github.com/elastic/elastic-charts/pull/684/commits/68e429910329ba9cd6cbcd70d7d5118ce24496fd

.ci/run.sh
```diff
@@ -109,6 +109,12 @@ yarn build
 echo " -- run linter"
 yarn lint
 
+###
+### run prettier check
+###
+echo " -- run prettier check"
+yarn prettier:check
```
nickofthyme, line 116: 👍

.huskyrc.js
```diff
@@ -2,13 +2,6 @@ const tasks = (arr) => arr.join(' && ');
...
   hooks: {
-    'pre-commit': tasks([
-      'pretty-quick',
-      'yarn run typecheck:all',
-      'yarn run lint',
-      'yarn run test',
-      'yarn run test:tz',
-    ]),
-    'commit-msg': 'commitlint -E HUSKY_GIT_PARAMS',
+    'commit-msg': 'commitlint -E HUSKY_GIT_PARAMS && yarn typecheck:all && yarn pq && yarn lint',
```
nickofthyme, line 5:
```suggestion
    'commit-msg': 'commitlint -E HUSKY_GIT_PARAMS && yarn typecheck:all && yarn pq && yarn lint --cache',
```

package.json
```diff
@@ -26,15 +26,16 @@
     "build:watch": "echo 'Watching build...' && yarn build:clean && yarn build:css && yarn build:compile -w ",
     "concat:sass": "echo 'Concat SASS...' && node scripts/concat_sass.js",
     "cz": "git-cz",
-    "lint": "NODE_ENV=production eslint --ext .tsx,.ts,.js .",
+    "lint": "NODE_ENV=production eslint --cache --ext .tsx,.ts,.js .",
```
nickofthyme, line 29: I think we keep this an just add the cache flag to the hook script.

markov00: done here https://github.com/elastic/elastic-charts/pull/684/commits/d67978a1c025423c441e0fc9297df2051c61bf0c

.huskyrc.js
markov00: https://github.com/elastic/elastic-charts/pull/684/commits/d67978a1c025423c441e0fc9297df2051c61bf0c

src/utils/data_samples/test_dataset_discover_per_30s.ts
```diff
@@ -1,3559 +1,2869 @@
-/* eslint-disable */
+/*
```
nickofthyme, line 1: What is this file??? 😮 

This could be the source of our typescript performance issues.

markov00: an old dataset used to test discover

markov00: removed here https://github.com/elastic/elastic-charts/pull/684/commits/55d53a217eae6fa576caefba691d37aa4a7dc392

PR #686, opened 2020-05-28
test(vrt): improve vrt config and update tooltip screenshots

nickofthyme:
> NOTE: There are no changes to `src/` in this or in order to safely re-group the tooltip vrt

⚠️  These changes are meant to help with writing and debugging the visual regression tests but since tests are run **OUTSIDE** of a docker container, the differences may be inconsistent. So always run `yarn test:integration{:local}` to update screenshots or your updated screenshots will likely fail in ci!

## Improvements to VRT config (851f184)

using [`jest-puppeteer-docker`](https://github.com/gidztech/jest-puppeteer-docker) does not allow for non-headless `launch` config, nor does it require/use any other `launch` config options we were passing it such as `browserUrl`. So these were all deleted.

I like having the ability to run the vrt locally in a non-headless browser to see/inspect the dom before the screenshot. So I created `process.env.DEBUG` to enable this. This flag does all the following...

- uses [`jest-puppeteer`](https://github.com/smooth-code/jest-puppeteer) preset over `jest-puppeteer-docker`
- sets host to `localhost`
- removes the jest timeout, for debugging (see new `debug` option).

> Any of the `launch` settings can be changed in `integration/jest_puppeteer.config.js`

To run the visual regression tests in debug mode, run the new npm script...

```bash
yarn test:integration:debug
```
> Must have a local version of storybook running as if you were to run `yarn test:integration:local`

With this mode enabled you can also call [`jestPuppeteer.debug`](https://github.com/smooth-code/jest-puppeteer/tree/master/packages/jest-environment-puppeteer#globaljestpuppeteerdebug) to pause the puppeteer execution. This is used for the `debug` option.

```ts
it('some visual test', async () => {
  await jestPuppeteer.debug();
  await common.expectChartAtUrlToMatchScreenshot(
    'http://localhost:9001/?path=/story/bar-chart--test-tooltip-and-rotation',
  );
});
```

> Note: you must `await` the call for it to catch

To continue from a debug point you must:
  - resume script execution in dev tools
  - press enter in the terminal running the jest tests

> This is very hacky but at the moment, it seems to be the best option for debugging inside the puppeteer context.

#### `screenshotSelector`
The `screenshotSelector` will be used as an override for the screenshot selector element, _only_ for the final screenshot capture. The regular `selector` value will be used otherwise.

This can be useful when you desire the screenshot to go beyond the bounds of the element, like in the example below, including the full `body` to show the tooltip overflow.

```ts
it('some visual test', async () => {
  await common.expectChartAtUrlToMatchScreenshot(
    'http://localhost:9001/?path=/story/bar-chart--test-tooltip-and-rotation',
    { screenshotSelector: 'body' },
  );
});
```

interactions-test-ts-nick-1-diff

#### `hiddenSelectors`

The `hiddenSelectors` option is used to toggle the visibility of a list of selectors. This would limit future diffs for nonessential parts contained in a screenshot. Toggles visibility of _all_ elements for _each_ selector.

An example of this could be hiding the title of the story as shown below.

```ts
it('some visual test', async () => {
  await common.expectChartAtUrlToMatchScreenshot(
    'http://localhost:9001/?path=/story/bar-chart--test-tooltip-and-rotation',
    { screenshotSelector: 'body', hiddenSelectors: ['#root > div:first-of-type'] },
  );
});
```

interactions-test-ts-nick-1-diff

#### `debug`

Used to pause before a given screenshot, where you can inspect the dom. This only works with `yarn test:integration:debug`.

```ts
it('some visual test', async () => {
  await common.expectChartAtUrlToMatchScreenshot(
    'http://localhost:9001/?path=/story/bar-chart--test-tooltip-and-rotation',
    { debug: true },
  );
});
```

### Improve `MousePosition` options (f225d11)
Before we could only set `x` and `y` values to position the cursor.

Now we can set any of `left`, `right`, `top` or `bottom` to easily manipulate the cursor relative to an element.

## Improve tooltip vrt (4a35434)

- Move around tests to group them in a better way for using `describe.each`.
- Add tests for combinations or `rotation`, `boundary` and `placement`
- Use added options above to cleanup tests
- Add padding around tooltip test story to simplify vrt
- Add margins around chart for screenshots outside of chart element or using `screenshotSelector`
- Remove all chart sibling elements from dom before screenshot
- Tightened threshold vrt config to `0` for both pixel and overall diff

## Update all screenshots (e670ab4)


nickofthyme:
@markov00 good point. Looks much better, I like the white background, looks cleaner

image

I am gonna keep the hidden logic in case we need it down the road, unless you care.

markov00:
> @markov00 good point. Looks much better, I like the white background, looks cleaner
> 
> image
> 
> I am gonna keep the hidden logic in case we need it down the road, unless you care.

 I think we should continue to use a different background color or a border around the chart, so it’s easier to spot the actual dimensions of the chart container, padding and margins are easier to check if we have a different color between the chart background and the page background.
I think we can have a more fancy color than the almond one, a gray or just a shadow should also work

nickofthyme:
@markov00 Adding the background changes you requested changes the dimensions on all the images so I would not suggest going through all 400+. Just FYI

.storybook/style.scss
```diff
@@ -1,34 +1,56 @@
...
 #story-root + div table td {
   border: 1px solid gray;
   padding: 5px;
 }
+
 .Pane.vertical.Pane1 {
   background: red;
 }
+
+// html body.echVisualTesting {
```
markov00, line 44: I think you forgot to uncomment this

integration/jest_env_setup.ts
```diff
@@ -27,5 +27,6 @@ export const toMatchImageSnapshot = configureToMatchImageSnapshot({
 
 expect.extend({ toMatchImageSnapshot });
 
-export const JEST_TIMEOUT = 10000;
-jest.setTimeout(JEST_TIMEOUT);
+if (process.env.DEBUG === 'true') {
+  jest.setTimeout(10 * 60 * 1000); // set timeout to 10 minutes;
```
markov00, line 31: isn't 10 minutes too much? the previous one was 10 seconds and the same appear on the TSDocs

integration/page_objects/common.ts
```diff
@@ -18,26 +18,102 @@
...
+   */
+  left?: number;
+}
+
+interface ElementBBox {
+  left: any;
+  top: any;
+  width: any;
+  height: any;
+}
```
markov00, line 59:
```suggestion
interface ElementBBox {
  left: number;
  top: number;
  width: number;
  height: number;
}
```

integration/page_objects/common.ts
```diff
@@ -69,23 +149,28 @@ class CommonPage {
...
   }
 
+  /**
+   * Toggle element visibility
+   * @param selector
+   */
+  async toggleElementVisibility(selector: string) {
+    await page.$$eval(selector, (elements) => {
+      elements.forEach((element) => {
+        element.classList.toggle('invisible');
```
markov00, line 159: Is this a `storybook` class? If so I think we should change to our own class, to avoid messing up when storybook change this internally

integration/jest_env_setup.ts
nickofthyme: This is for debugging only, this is basically the time we have to pause the debugger before the browser closes because the timeout is reached.

integration/page_objects/common.ts
nickofthyme: This is a global class in our `src/styles`

https://github.com/elastic/elastic-charts/blob/30096361d9b1ae826fe3dc50432ef80fd85a8069/src/components/_global.scss#L1-L3

.storybook/style.scss
nickofthyme: 7945cfd9

integration/page_objects/common.ts
nickofthyme: 7945cfd9

nickofthyme:
@markov00 can you review this again please and thank you. 🙏

PR #692, opened 2020-06-01
fix: missing dash style in line annotation

### LINKED ISSUE #687: Dashed Lines Not Working
When setting the `line.dash` value for annotation styles, it doesn't actually impact the appearance of the line.

**To Reproduce**
1. Go to [the line styling example in the StoryBook](https://elastic.github.io/elastic-charts/?path=/story/annotations-lines--styling)
2. Adjust the dash related values
3. Note there is no change in the appearance of the line.

**Expected behaviour**
I would expect the appearance of the line to change.  Note that I suspected a bug in the StoryBook itself, but I see the same behavior with my own annotations.

 - OS: Whatever is hosting your StoryBook
 - Browser: Chrome
 - Elastic Charts: Version rendering current StoryBook (I couldn't find any indication)

Note - I don't see any examples in your StoryBook (although perhaps I missed them) that demonstrate dashed or dotted lines.

markov00:
## Summary

This fix re-adds the dash style for line annotations. This probably was an existing feature before we removed Konva library, the story `annotations/lines/5_styling.tsx` has references to `dashWidth` and `dashGapWidth` used on the line style.

The VRT was manually updated as with the current change threshold (0,005% of the whole image) the change is not surfaced. I will open a subsequent PR to check if we can safely remove this threshold

fix #687

### Checklist


api/charts.api.md
```diff
@@ -1370,16 +1367,12 @@ export const SpecTypes: Readonly<{
 // @public (undocumented)
 export type SpecTypes = $Values<typeof SpecTypes>;
 
-// Warning: (ae-missing-release-tag) "StrokeDashArray" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
```
nickofthyme, line 1373: It was trying to tell us!

PR #693, opened 2020-06-01
fix: remove double rendering

### LINKED ISSUE #690: Double rendering of charts on initial load
**Describe the bug**
Double rendering of charts, reproducible in Storybook and Playground

**To Reproduce**
1. Put in a `console.log` in a render function, and/or randomize some data in the view model or render function
2. Reload the page
3. See dual console logging for rendering, or points/lines/etc. moving with the randomization approach

**Expected behaviour**
Single rendering

**Additional context**
Looks like the code that gets invoked is 
https://github.com/elastic/elastic-charts/blob/master/src/components/chart_resizer.tsx#L86-L93
```
  private handleResize = (entries: ResizeObserverEntry[]) => {
    // return; // <= this line isn't here
    if (this.initialResizeComplete) {
      this.onResizeDebounced(entries);
    } else {
      this.initialResizeComplete = true;
      this.onResize(entries);
    }
  };
```
If the `return` is uncommented, the double rendering goes away, but then of course, it'll not respond to a resizing

markov00:
## Summary

The `chart_resizer` component sends the first action on mount and then activate the resize observer that subsequentially sends another chart size update.

The PR fixes this behavior letting the ResizeObserver doing his job.

fix #690


src/components/chart_resizer.tsx
```diff
@@ -53,10 +55,8 @@ class Resizer extends React.Component<ResizerProps> {
   componentDidMount() {
     this.onResizeDebounced = debounce(this.onResize, this.props.resizeDebounce);
     if (this.containerRef.current) {
-      const { clientWidth, clientHeight } = this.containerRef.current;
-      this.props.updateParentDimensions({ width: clientWidth, height: clientHeight, top: 0, left: 0 });
```
monfera, line 57: Nice simplification and code removals!

monfera:
Jenkins test this

src/components/chart_resizer.tsx
```diff
@@ -36,6 +36,8 @@ interface ResizerDispatchProps {
 
 type ResizerProps = ResizerStateProps & ResizerDispatchProps;
 
+const DEFAULT_RESIZE_DEBOUNCE = 200;
+
 class Resizer extends React.Component<ResizerProps> {
   private initialResizeComplete = false;
   private containerRef: RefObject<HTMLDivElement>;
```
monfera, line 43: unrelated but could be changed when this file needs an update, `containerRef` could be read only

nickofthyme:
Looks to affect the positioning of annotations

image

src/components/chart_resizer.tsx
```diff
@@ -102,11 +102,9 @@ const mapDispatchToProps = (dispatch: Dispatch): ResizerDispatchProps =>
   );
 
 const mapStateToProps = (state: GlobalChartState): ResizerStateProps => {
-  const settings = getSettingsSpecSelector(state);
-  const resizeDebounce =
-    settings.resizeDebounce === undefined || settings.resizeDebounce === null ? 200 : settings.resizeDebounce;
+  const { resizeDebounce } = getSettingsSpecSelector(state);
   return {
-    resizeDebounce,
+    resizeDebounce: resizeDebounce == null || Number.isNaN(resizeDebounce) ? DEFAULT_RESIZE_DEBOUNCE : resizeDebounce,
```
nickofthyme, line 107: Can we avoid `==`?

markov00: done here: https://github.com/elastic/elastic-charts/pull/693/commits/5c7a317bfdb5039401ffee08c72b02da718cda08

markov00:
> Looks to affect the positioning of annotations
> 
> image

@nickofthyme  fixed 2b0c472

markov00:
> Thanks @markov00 the PR is effective in that there's no double _rendering_, I checked it in storybook mode and playground mode.
> 
> As a side effect, the way eg. the `viewModel` is invoked in partition charts (and maybe in Cartesians) seems to have changed a bit, at least when tested via the playground:
> 
> 1. viewmodel generation (`viewModel.ts` main function) is invoked twice (tbh not sure if `master` does it or not); the first invocation have zeroish width/height so it bails (and the resulting identity viewmodel - emptiness - causes that there's no rendering)
> 2. however, now the first invocation not only doesn't have width/height; it also doesn't seem to have the chart spec data
> 
> While point 2 is not apparent - because it bails anyway due to zero width/height - I'm mentioning this as you might know of some unintended consequences to this. Also, it's unrelated to the PR but ideally there wouldn't be an initial round of rendering with zero dimensions but I'm not sure if it's accidental and we don't care about the dual pass, or there's some hard async problem for which it's the optimal or developer time efficient solution
> 
> Approving the PR as it addresses the issue, and leaving it to your judgment if the apparently different zero-width/height invocation is cause for concern or not; @nickofthyme I see you've made commits in this part of the file, maybe know of something?

@monfera fixed here https://github.com/elastic/elastic-charts/pull/693/commits/72b6c36e4159a6ad4b1eccd6fc635a63c404ef98
I've basically added the check on the size of the parent before doing any computation

src/chart_types/partition_chart/layout/viewmodel/viewmodel.ts
```diff
@@ -69,9 +69,9 @@ function topGrooveAccessor(topGroovePx: Pixels) {
 }
 
 export const VerticalAlignments = Object.freeze({
-  top: 'top' as 'top',
-  middle: 'middle' as 'middle',
-  bottom: 'bottom' as 'bottom',
+  top: 'top' as const,
```
nickofthyme, line 72: 😄

PR #698, opened 2020-06-09
feat: show tooltip for external events

### LINKED ISSUE #695: Add tootip sync mechanism
I am building a page that will have multiples charts, they will all have the same `x` value. I'd like to have a way to sync all tooltips so when a user hovers over any chart the others tooltip would be shown.
Like the APM transaction page.
Screenshot 2020-06-08 at 14 18 51

We could have a `hoverX` property that would be used to pass in the current position the user is hovering on another chart.

**Checklist**

markov00:
## Summary

Adds the tooltip on synced charts.

To do:
- [ ] add custom configuration for external tooltip settings (boundary, placement)
- [ ] add show/hide tooltip from external events

close #695

### Checklist


markov00:
> @markov00 Oh that makes sense because the anchor element is just a strip across the highlighted region.
> 
> Another thing I noticed is the `follow` tooltip type. I think the follow type should not be allowed for external tooltips cuz it is spatially specific. I also think the `ExternalPointerEventsSettings` should include a type which would default to the tooltip settings type but allows them to be different.

Whit [this](https://github.com/elastic/elastic-charts/pull/698/commits/8400be04f22244b3debc01336c33ee3cdda37b0b)  commit I'm fixing the fact that: the TooltipType used for external events is only the VerticalCursor (no follow, no Crosshair). I've also fixed the rendering of the cursor band if the tooltip type of a chart is defined as None, showing always the cursor band if the events comes from an external event.

The tooltip disappearing bug was also fixed https://github.com/elastic/elastic-charts/pull/698/commits/b25e01fc4d3fd1aa43e1f5351bba6fd631400d88

src/specs/constants.ts
```diff
@@ -93,6 +94,14 @@ export const DEFAULT_SETTINGS_SPEC: SettingsSpec = {
     type: DEFAULT_TOOLTIP_TYPE,
     snap: DEFAULT_TOOLTIP_SNAP,
   },
+  externalPointerEvents: {
+    tooltip: {
+      visible: false,
+      boundary: 'chart',
```
nickofthyme, line 100: I don't think this should be `chart` by default. I think the default should be the most flexible positioning outside of the chart.

```suggestion
      boundary: undefined, // undefined uses the closest parent scroll container
```

src/specs/settings.tsx
```diff
@@ -200,25 +200,26 @@ export interface ExternalPointerEventsSettings {
...
      *  {@inheritDoc TooltipProps.fallbackPlacements}
      */
-    fallbackPlacements?: Placement[];
+    fallbackPlacements: Placement[];
     /**
      *  {@inheritDoc TooltipProps.boundary}
+     * @defaultValue `chart`
      */
-    boundary?: HTMLElement | 'chart';
+    boundary: HTMLElement | 'chart';
```
nickofthyme: Do all of these have to be required? Can't we just merge these `Partial` settings with some defaults?

src/specs/settings.tsx
```diff
@@ -456,3 +412,12 @@ export function getTooltipType(settings: SettingsSpec): TooltipType {
...
   return defaultType;
 }
+
+
+/**
+ * Always return a Vertical Cursor for external pointer events or None if hidden
+ * @param settings SettingsSpec
+ */
+export function getExternalTooltipType({ externalPointerEvents: { tooltip: { visible } } }: SettingsSpec): TooltipType {
+  return visible ? TooltipType.VerticalCursor : TooltipType.None;
```
nickofthyme, line 422: Shouldn't this be configurable in the settings?

```ts
export type ExternalTooltipType = Exclude<TooltipType, TooltipType.Follow>;
```

src/specs/settings.tsx
markov00: I've marked them as optional here https://github.com/elastic/elastic-charts/pull/698/commits/40cef6a8e88ef0ce957d96c1b6ff9dafd9619117

src/specs/settings.tsx
markov00: No, as explained on a call, the external tooltip at the moment can only be a VerticalCursor tooltip or a None tooltip

src/components/portal/_portal.scss
```diff
@@ -11,4 +11,6 @@
 .echTooltipPortal__invisible {
   position: fixed;
   visibility: hidden;
+  width: 0;
+  height: 0;
```
nickofthyme, line 15: 👍

src/components/portal/tooltip_portal.tsx
```diff
@@ -185,11 +180,11 @@ const TooltipPortalComponent = ({ anchor, scope, settings, children, visible, ch
   }, [visible, anchorNode, position?.left, position?.top, position?.width, position?.height]);
 
   useEffect(() => {
```
nickofthyme, line 186: I Love love love this, I don't know why I made this so overly complicated. I swear I tried this but couldn’t get it to work right. Maybe I was trying to prevent recreating the portal node every time it's hidden.

I also like that the portal now disappears every time.

PR #704, opened 2020-06-11
fix: graceful scale fallbacks and warnings

### LINKED ISSUE #678: XDomain should fallback to ordinal type
**Describe the bug**
Using a `LineSeries` when defining an `XScaleType` as `linear` or undefined, elastic charts will try to use linear and use `ordinal` as a fallback if needed. This should be the same logic with `AreaSeries`.

 - OS: macOS
 - Browser: chrome
 - Elastic Charts: `latest`

**Additional context**
Related to code below...
https://github.com/elastic/elastic-charts/blob/a79e899a8d194e5559c184ca4f20920baef751bb/src/chart_types/xy_chart/domains/x_domain.ts#L68-L72

**Errors in browser console**
```bash
Uncaught Error: Each X value in a linear x scale needs be be a number. String or objects are not allowed
    at Object.mergeXDomain (x_domain.js:51)
    at Object.computeSeriesDomains (utils.js:136)
    at compute_series_domains.js:14
    at index.js:70
    at index.js:30
    at index.js:84
    at index.js:30
    at selector (index.js:105)
    at index.js:80
```

nickofthyme:
## Summary

fix #678

Adds x scale fallback to `ScaleType.Ordinal` when there is a non-numeric value found in the dataset. This includes a warning for y values that are not numeric values. The current functionality is to error.

```sh
Each X value in a linear x scale needs be be a number. Using ordinal x scale as fallback.
```

Adds warning for y values that are not parable into a numeric value.

```sh
Found non-numeric y values in dataset for spec "areas" ("a", "c")
```

## Example

String value `'a'` among numeric values, converts scale from `linear` to `ordinal`

Image 2020-06-11 at 3 20 14 PM

### Checklist


src/chart_types/xy_chart/domains/x_domain.ts
```diff
@@ -46,7 +47,14 @@ export function mergeXDomain(
   let seriesXComputedDomains;
   let minInterval = 0;
 
-  if (mainXScaleType.scaleType === ScaleType.Ordinal) {
+  if (mainXScaleType.scaleType === ScaleType.Ordinal || fallbackScale === ScaleType.Ordinal) {
+    if (mainXScaleType.scaleType !== ScaleType.Ordinal) {
+      // eslint-disable-next-line no-console
+      console.warn(
```
markov00, line 53: I think we should have a way to enable/disable this warning on development/production, but maybe we can defer this decision later

markov00: We should also prefix our warning/errors with something like `[ech]` or `[@elastic/chart]` so Kibana can also filter them overwriting the console.log function

src/chart_types/xy_chart/domains/x_domain.ts
```diff
@@ -58,7 +66,9 @@ export function mergeXDomain(
   } else {
     seriesXComputedDomains = computeContinuousDataDomain(values, identity, true);
     let customMinInterval: undefined | number;
-    if (!isNumberArray(values)) {
+
+    // The number check is done in mergeXDomain which makes this a dummy type check
+    if (!isNumberArray(values, !fallbackScale)) {
```
markov00, line 71: does exist the situation where we don't have a `fallbackScale` and the array has both number and scales?

src/chart_types/xy_chart/utils/series.ts
```diff
@@ -175,7 +177,14 @@ export function splitSeries({
         }
       });
     } else {
-      const cleanedDatum = cleanDatum(datum, xAccessor, yAccessors[0], y0Accessors && y0Accessors[0], markSizeAccessor);
+      const cleanedDatum = cleanDatum(
+        datum,
+        xAccessor,
+        yAccessors[0],
+        nonNumericValues,
```
markov00, line 184: I'm not a great fan of this mutation, but I know that changing the return type means a lot more changes

src/chart_types/xy_chart/domains/x_domain.ts
nickofthyme: Yeah I agree. I wonder if kibana uses a node env variable to set production or not.

src/chart_types/xy_chart/domains/x_domain.ts
nickofthyme: No but I did this type check over calling `getMinInterval(values as number[])`. What do you think?

src/chart_types/xy_chart/utils/series.ts
nickofthyme: Yeah same. Alternatively I could warn every time a non-numeric value is found. I just thought the combined would be nicer.

src/chart_types/xy_chart/domains/x_domain.ts
nickofthyme: 5346284

src/chart_types/xy_chart/domains/x_domain.ts
nickofthyme: 5346284

src/chart_types/xy_chart/utils/series.ts
markov00: ok np, let's keep like that

nickofthyme:
@markov00 I'm not sure about needed to have them separate. I think eventually the chart should error in dev to alert the developer of bad config then in prod only `console.error` severe issues.

I'll make the change for now and see if my opinion changes too.

PR #705, opened 2020-06-11
ci: add console errors to smoke tests

nickofthyme:
## Summary

Add console errors to smoke tests to help identify the issue. Currently the test just fails and doesn't output the `console.errors`.

### DIFF
--- a/browsers/browsers.test.ts
+++ b/browsers/browsers.test.ts
@@ -19,7 +19,7 @@
 
 import path from 'path';
 
-import webdriver, { By } from 'selenium-webdriver';
+import webdriver, { By, logging } from 'selenium-webdriver';
 
 jest.setTimeout(30000);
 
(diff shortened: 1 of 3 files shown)


browsers/browsers.test.ts
```diff
@@ -60,6 +60,17 @@ describe('smoke tests', () => {
   });
 
   afterAll(async() => {
+    const entries = await driver.manage().logs().get(logging.Type.BROWSER);
+    const entries2 = await driver.manage().logs().get(logging.Type.CLIENT);
+    console.log('ERRRORS BROWSER', JSON.stringify(entries));
+    console.log('ERRRORS CLIENT', JSON.stringify(entries2));
```
markov00, line 66: Aaaaaaaaagaaaaaiiiiinnnn thhhhhiiiiss. ccccccccouldddd beeeeee fffffffiiiiiixxxxxeddd :D

nickofthyme: lol I'm trying to log the console errors from the webdriver on the browser tests. Any ideas? This doesn't log anything.

markov00: I think the logging API has not been implemented in the IE driver

markov00: and since this will be sunsetted very soon I think we should not care that much

PR #708, opened 2020-06-12
ci: lint only staged files

nickofthyme:
## Summary

This PR updates the `commit-msg` hook to only lint staged files on commit.

Before we called `yarn lint` which defines all files to be linted in the project regardless.


nickofthyme:
@markov00 I'm not sure. I only used the `cross-env` because `lint-staged` doesn't support env variables otherwise. See https://github.com/okonet/lint-staged#use-environment-variables-with-linting-commands

PR #712, opened 2020-06-16
fix: popper scroll issue

nickofthyme:
## Summary

Fix scroll bar issue with popper js


nickofthyme:
It looks like this has some very slight changes in the screenshots. Particularly with the box-shadow.

image

nickofthyme:
I've also notice that some screenshots appear to be slightly smaller in width. Notice the sliver of transparency on the left edge of the new screenshot. But yet the dimensions are identical. I saw this on #629 as well. Any ideas? Maybe it's just a github thing.

Image 2020-06-16 at 10 16 27 AM

PR #719, opened 2020-06-24
fix: background color dark mode issue

nickofthyme:
## Summary

Set default background color to transparent.

Currently, most charts in kibana do not handle getting the proper base theme for the chart and use the `transparent` as a default. The color can be set on the corresponding theme once the theme is commonly shared across all charts in kibana.

Changes
- revert existing logic for no background color
- fix divide-by-zero error with `combineColors` function
- fix text text leading line on partition charts
- add warning for invalid or transparent colors
- add warning for alpha values less than 1
- fix tooltip color width and groove colors


nickofthyme:
retest this please

PR #732, opened 2020-06-30
ci: fix git refspec

markov00:
## Summary

fix git refspec


nickofthyme:
@markov00  I'm curious what was the error this caused? Just bad checkouts?

markov00:
> @markov00 I'm curious what was the error this caused? Just bad checkouts?

It wasn't able to correctly checkout the PRs

PR #749, opened 2020-07-13
feat(legend): add legend item actions

### LINKED ISSUE #717: Legend actions
## Legend actions

Adds the option for legend actions to the legend item. Preserves the toggle ability of the series and the color picker option for clicking the color dot itself.

### Hover state
Adds blue highlight and action icon or slot to render react component.
image

### Click/focus state
Adds blue highlight to focused icon

image

> The context menu will be a consumers implementation

### Design for small width legends

image

nickofthyme:
## Summary

closes #717

Adds actions to legend items

New prop `legendAction` that takes a `LegendAction` type defined as...

```ts
export interface LegendActionProps {
  /**
   * Series id for the given series
   */
  series: SeriesIdentifier;
}
export type LegendAction = ComponentType<LegendActionProps>;
```

### Vertical legend
Screen Recording 2020-07-13 at 08 08 AM

### Horizontal legend
Screen Recording 2020-07-13 at 08 08 AM

### Checklist


src/specs/settings.tsx
```diff
@@ -190,6 +190,26 @@ export interface ExternalPointerEventsSettings {
   }
 }
 
+export interface LegendActionProps {
```
markov00, line 193: add `@public`

src/components/legend/_legend_item.scss
```diff
@@ -8,13 +8,44 @@ $legendItemVerticalPadding: $echLegendRowGap / 2;
...
   width: 100%;
+  height: 100%;
 
   &:not(&--hidden) {
     .echLegendItem__color--changable {
       cursor: pointer;
     }
   }
 
+  &__main {
```
markov00, line 19: with this additional div, the height of the legend is less than the height of the echLegendItem.
The problem is that the series are highlighted when moving the mouse over the echLegendItem but the edit button appears only when the mouse is over the __main div
Jul-13-2020 17-42-06

markov00: probably moving the 4px padding of the `echLegendItem` to the `__main` could solve the issue

stories/legend/11_legend_actions.tsx
```diff
@@ -0,0 +1,160 @@
...
+import React, { useState } from 'react';
+
+import { Axis, BarSeries, Chart, Position, ScaleType, Settings, LegendAction, XYChartSeriesIdentifier } from '../../src';
+import * as TestDatasets from '../../src/utils/data_samples/test_dataset';
+import { getPositionKnob } from '../utils/knobs';
+
+export const Example = () => {
+  const hideActions = boolean('Hide actions', false);
+  const [popoverId, setPopoverId] = useState<string | null>(null);
+
```
markov00, line 31: Could you move the Action panel to a separate component on the example? it's easier to read/copy and doesn't cause the chart to re-render on every close/open action (just check the `echChartStatus` you will see that rerenders appear whenever closing/opening the panel)

src/specs/settings.tsx
```diff
@@ -190,6 +190,26 @@ export interface ExternalPointerEventsSettings {
...
+   *
+   * Use when openning a popover where you want to persist the action button
+   */
+  onOpen: () => void;
+  /**
+   * Callback to close action
+   *
+   * Use when openning a popover where you want to persist the action button
+   */
+  onClose: () => void;
```
markov00, line 205: This is semantically correct if we are opening and closing a popup, but since you already implemented the action as an abstract tool to use, it can be to add text and hovering could activate something else. we can call them something like `activateAction` , `deactivateAction`, `isActionActive(boolean)` etc, what do you think?

src/components/legend/_legend_item.scss
nickofthyme: I see what you are saying here but I'm having trouble with solution. You idea of putting the padding on the `__main` would work but the blue background would be extended to the added padding and applying it as a margin will not trigger the hover event.

The solution I came up with was to wrap the `__main` element in another element to apply the padding and use that elements hover state to trigger the action. Not ideal nesting divs but better than other ideas I think.

Screen Recording 2020-07-13 at 05 05 PM

> The series will still be highlighted on the chart when hovering over the `additionalValue` element but I think that look fine.

---

Alternatively, I was playing with moving the mouse event from the whole container to the label but this was not a great solution as it made things very clunky, and still didn't fully solve the issue.

Screen Recording 2020-07-13 at 05 10 PM

src/specs/settings.tsx
nickofthyme: 8b303ea

stories/legend/11_legend_actions.tsx
nickofthyme: 8b303ea

src/components/legend/_legend_item.scss
nickofthyme: 8b303ea

src/specs/settings.tsx
nickofthyme: 8b303ea

src/components/legend/_legend_item.scss
markov00: I've tested that and I'm not convinced, I think we should ask @miukimiu what she thinks about that.
Few doubts I have:
1) I think we should highlight the whole text row, including the value on the left.
2) the visual feedback of the series highlight action (the blue highlight), coming from the mouse hovering an element on the legend, should be rendered visually whenever the mouse enters the active area
2) the highlight should be consistent and should not turn off and on when moving from a label the other (as it's currently doing)
3) the same highlight will be used with when using the keyboard to navigate the legend, probably also @myasonik has some ideas on how it's better to place this focus

In case, at the moment we can add the legend actions without thinking about this visual feedback, and we can reason about it when working on the keyboard navigation, what do you think?

markov00: also, be careful that the current solution leads to this epileptic effect
Jul-14-2020 10-58-49

nickofthyme: Yeah all good points, @miukimiu any thoughts?

miukimiu: Good points @markov00, 

For better a11y we need a different design. The one being implemented I think it would be tricky to make it keyboard navigable. 

So we can separate the actions/highlight. 

- The legend with the value has one action (hide/show)

- The edit icon would open the tooltip

Legends PR suggestion@2x

The edit icon would always be visible. The only concern with this new design is that it will require more space and it will truncate more the legend. But I think for a11y is better. What do you think @myasonik?

myasonik: I think the new design you're proposing much better calls out the ability to interact with it and simplifies the focus management quite a bit. I'm +1 on the new design.

That said, if y'all still _really_ wanted the old design, you _could_ do it with some `:focus-within` and `+` CSS operator trickiness but it would likely become a really brittle area of the code and wouldn't work the same in all browsers...

nickofthyme: Ok I updated to the new design. See screenshot below

Screen Recording 2020-07-15 at 10 37 PM


The final question I have is how to handle the chart hover over the element. Due to the padding between the text, there is a space between the legend items that would not be clickable but we are still highlighting the series in the chart (i.e. graying out all the other series).

Screen Recording 2020-07-15 at 10 41 PM

Possible solutions
- remove highlight until over hoverable area
- keep as is, but when hovering below the action there will be no blue highlight and the series will be still be highlighted in the chart.

nickofthyme: @markov00 updated pr with latest designs per zoom call. See 5b6f1a2

nickofthyme:
IE 11 seems to be somewhat strange. Maybe the EUI components are not fully compatible in playground.

Screen Recording 2020-07-16 at 02 20 PM

src/components/legend/_legend_item.scss
markov00: I think there still gaps between legend actions that create an epileptic effect, I can jump on a call to describe that

markov00: Let's fix this in a subsequent PR

PR #762, opened 2020-07-22
fix(axis): left axis sizing based on title padding

nickofthyme:
## Summary

Fix left axis size calculation. Titles as `''` still included the padding in the axis dimensions but not in the charts dimensions, causing a misalignment.

## Before
Screen Recording 2020-07-22 at 02 55 PM

## After
Screen Recording 2020-07-22 at 02 48 PM


markov00:
jenkins test this please

Issue #770, opened 2020-07-29
Bad tooltip placement for single bars


nickofthyme:
**Describe the bug**
When displaying a single bar, stacked or non-stacked, the tooltip is shown within the highlighted area.

**To Reproduce**
1. Go to https://elastic.github.io/elastic-charts/?path=/story/bar-chart--single-data-stacked-chart
2. Hover over the stacked bars
3. Notice the placement of the tooltip

**Expected behaviour**
The tooltip attempts to place itself outside of the area.

**Screenshots**
![Screen Recording 2020-07-29 at 10 03 AM](https://user-images.githubusercontent.com/19007109/88817537-3e0a3a00-d183-11ea-815e-a4832c1b6ddd.gif)

**Version (please complete the following information):**
 - Elastic Charts: `19.8.2`

**Additional context**

PR #771, opened 2020-07-30
fix(tooltip): placement with left/top legends and single bars

### LINKED ISSUE #769: Bad tooltip placement with left/top legend
**Describe the bug**
When setting `legendPosition` to `'left'` or `'top'` the tooltip placement is not accurate. All other positions are correctly placed.

**To Reproduce**
1. Run master locally
2. Go to `http://localhost:9001/?path=/story/legend--left`
3. Hover over any series

**Expected behaviour**
Correct placement.

**Screenshots**
Screen Recording 2020-07-30 at 11 35 AM

 - Elastic Charts: as of at least `19.9.0`

nickofthyme:
## Summary

Fixes #769 and #770 

Related to #259

### Before
Screen Recording 2020-07-30 at 11 35 AM

### After
Screen Recording 2020-07-30 at 11 25 AM

## Single value bars (#770)
Reverses changes from #259 in favor of new popperjs changes that allow greater tooltip customization.

### Before
Screen Recording 2020-07-30 at 12 03 PM

### After
Screen Recording 2020-07-30 at 12 05 PM

> Note:  The ordinal tooltip uses a `top` placement because there is no room on left or right. This can be fixed by setting the `tooltip.boundary` prop. However, storybook renders the chart in an iframe that prevents us from fixing this.

### Checklist


markov00:
Hey @nickofthyme why the tooltip behavior of a single bar histogram is different than the stacked bar one?
In this gif I've added multiple histograms stacked on each other and seems that this behavior is actually fine and better than pushing the tooltip outside the chart

Jul-31-2020 10-28-55

markov00:
We are going to integrate soon the small multiples, which means that the tooltip has to work over different areas independently from the legend position or its own panel.  Is there a way to fix this without coupling this with the legend size? can we constraint the tooltip anchor to the chart panel only? Looking at the `Before`  screenshots seems that the issue is related on the fact that we are not limiting the anchor inside the chart panel, do you think it's possible to achieve that without knowing the legend position/size?

PR #772, opened 2020-07-30
ci: fix browser-ie failure

nickofthyme:
## Summary

Fix `browser-ie` pr check

fix per https://stackoverflow.com/questions/22575662/filename-too-long-in-git-for-windows


nickofthyme:
@markov00 I merged this in the meantime. If we want to just do away with this check in the future that would be fine with me.

Issue #615, opened 2020-04-01
Tooltip re-design


nickofthyme:
### Is your feature request related to a problem? Please describe.
As the chart functionality has grown over the past several months the current tooltip design has come to show its limitations for more complex charts.

Related to #580 

### Describe the solution you'd like
The tooltip should be redesigned to handle all of the following use cases.

#### Multiple x or other values
Up until now, the tooltip is designed with the x value being constant for all values in the tooltip. This is no longer a valid assumption. Possibilities include a table style tooltip or indented values.

**Table style**
![image](https://user-images.githubusercontent.com/19007109/78155979-c1996180-7403-11ea-8112-95eebf941abc.png)

**Indented style**
![image](https://user-images.githubusercontent.com/19007109/78157711-ea225b00-7405-11ea-868a-eb7fa58f7072.png)

The issue with the table style tooltip is that the field names for x, y, mark, etc. could be very long, creating a very wide tooltip that is less useful.

#### Aggregate values like min, max, etc.
When multiple values are allowed the logical progression would be to allow for aggregate values to be used per tooltip item. The indented files could be a better way to show this alongside the respective value, see mock up below.

![image](https://user-images.githubusercontent.com/19007109/78158135-7c2a6380-7406-11ea-88ff-c305e078dc1c.png)

#### Multiple values from the same series
![image](https://user-images.githubusercontent.com/19007109/78151908-ba238980-73fe-11ea-8f4a-93366345bf4b.png)

Up until now, the tooltip would never have two of the same series in the tooltip with the exception of the banded values.

Having multiple values in the tooltip from series based on an extension of the current tooltip design would repeat the series name for each. This seems redundant and avoidable.

#### Upper and lower (aka banded) values
Currently, we support this style but the items are repeated for each series and the order is not configurable.

![image](https://user-images.githubusercontent.com/19007109/78151560-58631f80-73fe-11ea-944b-fa54ab972698.png)

### Additional context

**Checklist**
- [x] this request is checked against already exist requests
- [x] every related Kibana issue is listed under `Kibana Cross Issues` list
- [x] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

nickofthyme:
Also thought about a collapsed mode where if multiple values are to be displayed, they only appear in detail when hovered over a point/bar. Otherwise, they would be a simplified item, similar to what there is now.

nickofthyme:
[...]

PR #775, opened 2020-07-31
feat(series): add simple mark formatter

nickofthyme:
## Summary

Add simple way to format mark accessor values before tooltip redesign (#615)

Screen Recording 2020-09-10 at 01 11 PM

### Checklist


api/charts.api.md
```diff
@@ -1485,7 +1487,7 @@ export interface Theme {
 }
 
 // @public (undocumented)
-export type TickFormatter = (value: any, options?: TickFormatterOptions) => string;
+export type TickFormatter<V = any> = (value: V, options?: TickFormatterOptions) => string;
```
monfera, line 1490: good improv, I'm curious though, what does it achieve with `any`?

api/charts.api.md
```diff
@@ -1543,11 +1545,12 @@ export type TooltipType = $Values<typeof TooltipType>;
 // @public
 export interface TooltipValue {
     color: Color;
+    formattedMarkValue?: string | null;
     formattedValue: string;
     isHighlighted: boolean;
     isVisible: boolean;
     label: string;
-    markValue?: any;
+    markValue?: number | null;
```
monfera, line 1553: do we sometimes have categorical values in a tooltip, represented as strings?

src/chart_types/xy_chart/tooltip/tooltip.ts
```diff
@@ -93,7 +95,12 @@ export function formatTooltip(
     label,
     value,
     formattedValue: tickFormatter(value, tickFormatOptions),
-    markValue: isHeader || mark === null ? null : mark,
+    markValue,
+    ...(isDefined(markValue) && {
```
monfera, line 99: nice!

stories/mixed/7_marks.tsx
```diff
@@ -90,6 +93,7 @@ export const Example = () => {
         yAccessors={['y']}
         markSizeAccessor="z"
         data={data2.slice(0, size)}
+        markFormat={(d) => `${numeral(d).format(markFormat)}%`}
```
monfera, line 96: would it be worth using a different formatter vs. line 86?

api/charts.api.md
nickofthyme: Nothing really, This just allows the `markFormatter` value to be more specific knowing that all values are numbers.

Ideally the datum type can be inferred from the data and replace the any value but that is more complicated than it sounds.

api/charts.api.md
nickofthyme: In the case of the mark value it will always be a number because it needs to scale based on the values.

Does that answer your question?

stories/mixed/7_marks.tsx
nickofthyme: 54d47d3

api/charts.api.md
monfera: Thanks, it does!

monfera (approved):
Looks good Nick, thanks!

PR #779, opened 2020-08-05
fix: allow graceful error handling

### LINKED ISSUE #776: Gracefully handle bad custom domain
**Describe the bug**
When setting a custom min greater than computed max of the data, the chart will throw an error. Or similarly, when setting a custom max less than computed min of the data.

**Expected behaviour**
Renders empty chart when data is outside of custom bounds. Logs warning to console.

 - Elastic Charts: all

**Errors in browser console**
```
node_modules\@elastic\charts\dist\chart_types\xy_chart\domains\y_domain.js:113 Uncaught Error: custom yDomain for yaxis_2e91f9d0-cd39-11ea-8e53-15bf965b677f_main_group is invalid, custom min is greater than computed max
```

**Kibana Cross Issues**
https://github.com/elastic/kibana/issues/73167

**Checklist**

nickofthyme:
fix #776

Allow gracefully handling/catching of errors at a top level.

Now calling errors as `GracefulError` will still show in the console but return an empty chart, not a blank element.

```ts
throw new GracefulError('Something went wrong');
```

Screen Recording 2020-08-17 at 04 11 PM

### DIFF
--- a/src/components/error_boundary/error_boundary.tsx
+++ b/src/components/error_boundary/error_boundary.tsx
@@ -1,0 +1,63 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import React, { Component, ReactNode } from 'react';
+
+import { isGracefulError } from './errors';
+
+type ErrorBoundaryProps = {
+  children: ReactNode;
+};
+
+interface ErrorBoundaryState {
+  hasError: boolean;
+}
+
+/**
+ * Error Boundary to catch and handle custom errors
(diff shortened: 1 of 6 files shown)


src/components/error_boundary/error_boundary.tsx
```diff
@@ -0,0 +1,63 @@
...
+  render() {
+    if (this.hasError) {
+      return (
+        <div className="echReactiveChart_unavailable">
+          <p>No data to display</p>
+        </div>
+      );
+    }
+
+    return <>{this.props.children}</>;
```
markov00, line 61: can you return directly `return this.props.children` ?

src/components/error_boundary/errors.ts
```diff
@@ -0,0 +1,43 @@
...
+});
+/** @public */
+export type ChartTypes = $Values<typeof ErrorType>;
+
+/**
+ * Error to used to gracefully render empty chart
+ */
+export class GracefulError extends Error {
+  type = ErrorType.Graceful;
+  renderEmptyChart: boolean;
```
markov00, line 33: This flag is not currently used, what was the idea behind that?

nickofthyme: Thanks, I'll remove it for now. The idea was to allow this to be a more generic error. You could throw this error everywhere is charts and customize how the error is handled such as if it should throw, or if the chart should render empty. I think this can be discussed once we improve the error handling.

src/components/error_boundary/error_boundary.tsx
nickofthyme: 446f764

src/components/error_boundary/errors.ts
```diff
@@ -0,0 +1,43 @@
...
+ * under the License.
+ */
+
+import { $Values } from 'utility-types';
+
+export const ErrorType = Object.freeze({
+  Graceful: 'graceful' as const,
+});
+/** @public */
+export type ChartTypes = $Values<typeof ErrorType>;
```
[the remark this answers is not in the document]
nickofthyme: 446f764

PR #782, opened 2020-08-11
feat: support color mappings for series

nickofthyme:
## Summary

To replicate the current color functionality of the `ColorMappings` in kibana, we need to know all the series in the chart to allocate a set of unique colors.

This PR adds an array of all `SeriesIdentifiers` to the `SeriesColorAccessorFn`.

### Checklist

### DIFF
--- a/src/chart_types/xy_chart/state/utils/utils.ts
+++ b/src/chart_types/xy_chart/state/utils/utils.ts
@@ -102,6 +102,8 @@ export function getCustomSeriesColors(
   const updatedCustomSeriesColors = new Map<SeriesKey, Color>();
   const counters = new Map<SpecId, number>();
 
+  let allSeries: XYChartSeriesIdentifier[] | null = null;
+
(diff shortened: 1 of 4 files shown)


src/chart_types/xy_chart/state/utils/utils.ts
```diff
@@ -117,7 +119,16 @@ export function getCustomSeriesColors(
...
       } else {
         const counter = counters.get(seriesIdentifier.specId) || 0;
-        color = Array.isArray(spec.color) ? spec.color[counter % spec.color.length] : spec.color(seriesIdentifier);
+
+        if (Array.isArray(spec.color)) {
+          color = spec.color[counter % spec.color.length];
+        } else {
+          if (!allSeries) {
+            allSeries = [...seriesCollection.values()].map(({ seriesIdentifier: si }) => si);
+          }
```
markov00, line 128: nit: we can move this also outside the forEach loop, the cost of looping through the seriesCollection is very low and we don't mutate data the loop

PR #785, opened 2020-08-19
feat(bar_chart): add shadow prop for value labels

dej611:
## Summary

This is just a small PoC that came up while talking with @markov00 about value labels readability.
This PR introduces a new "stroke" level around the value labels in order to increase the contrast with the bar colour: this is particular efficient when the value text color is white and the stroke is some dark color. Using a darker color for the text fill does not provide the same quality result.

value-stroke

There are still some open questions, like:
* [ ] would it make sense to dynamically assign the stroke color based on the fill color by default?
  * I made some experiments here as well, using brightness to measure the appropriate contrast, and results are not great. But maybe there's a better heuristic.
* [ ] Should `transparent` be the default value? I think so, but maybe there's a better idea.
* [ ] Should the name of the feature be `shadow` or something else? I started with a `box shadow` technique to start with, hence the name `shadow`. But now it is probably something different.

### Checklist

### DIFF
--- a/src/chart_types/xy_chart/renderer/canvas/primitives/text.ts
+++ b/src/chart_types/xy_chart/renderer/canvas/primitives/text.ts
@@ -32,6 +32,7 @@ export function renderText(
     fontSize: number;
     align: TextAlign;
     baseline: TextBaseline;
+    shadow?: string;
   },
   degree: number = 0,
   translation?: Partial<Point>,
@@ -48,6 +49,12 @@ export function renderText(
       ctx.font = cssFontShorthand(font, font.fontSize);
       if (translation?.x || translation?.y) {
         ctx.translate(translation?.x ?? 0, translation?.y ?? 0);
+      }
+      if (font.shadow) {
+        ctx.lineWidth = 1.5;
+        ctx.strokeStyle = font.shadow;
+        ctx.strokeText(text, origin.x, origin.y);
+        ctx.shadowBlur = 0;
       }
       ctx.fillText(text, origin.x, origin.y);
     });
--- a/src/utils/themes/theme.ts
+++ b/src/utils/themes/theme.ts
@@ -293,6 +293,7 @@ export type PartialTheme = RecursivePartial<Theme>;
 export type DisplayValueStyle = TextStyle & {
   offsetX: number;
   offsetY: number;
+  shadowColor?: Color;
 };
 
 export interface PointStyle {
--- a/stories/bar/2_label_value.tsx
+++ b/stories/bar/2_label_value.tsx
@@ -61,6 +61,7 @@ export const Example = () => {
         fontStyle: 'normal',
         padding: 0,
         fill: color('value color', '#000'),
+        shadowColor: color('shadow color', 'transparent'),
         offsetX: number('offsetX', 0),
         offsetY: number('offsetY', 0),
       },
--- a/src/chart_types/xy_chart/renderer/canvas/values/bar.ts
+++ b/src/chart_types/xy_chart/renderer/canvas/values/bar.ts
@@ -38,7 +38,7 @@ interface BarValuesProps {
 /** @internal */
 export function renderBarValues(ctx: CanvasRenderingContext2D, props: BarValuesProps) {
   const { bars, debug, chartRotation, chartDimensions, theme } = props;
-  const { fontFamily, fontStyle, fill, fontSize } = theme.barSeriesStyle.displayValue;
+  const { fontFamily, fontStyle, fill, fontSize, shadowColor } = theme.barSeriesStyle.displayValue;
   const barsLength = bars.length;
   for (let i = 0; i < barsLength; i++) {
     const { displayValue } = bars[i];
@@ -93,6 +93,7 @@ export function renderBarValues(ctx: CanvasRenderingContext2D, props: BarValuesP
           fontSize,
           align,
           baseline,
+          shadow: shadowColor,
         },
         -chartRotation,
       );


wylieconlon:
I would consider increasing the shadow size, or making it configurable.

dej611:
> I would consider increasing the shadow size, or making it configurable.

Technically is possible, but I would leave it as last resort.
I'm curious to check if a "golden" value may exist here - not saying the current is the right one. There's some margin for experimenting here.

stories/bar/2_label_value.tsx
```diff
@@ -61,6 +61,7 @@ export const Example = () => {
         fontStyle: 'normal',
         padding: 0,
         fill: color('value color', '#000'),
```
miukimiu, line 63: I've been testing and I think for the example white fill color looks better with a dark shadow:
 
```suggestion
        fill: color('value color', '#fff'),
```

Screenshot 2020-09-28 at 12 29 58

stories/bar/2_label_value.tsx
```diff
@@ -61,6 +61,7 @@ export const Example = () => {
...
+        shadowColor: color('shadow color', 'transparent'),
```
miukimiu, line 64: I think we should enforce the use of a shadow:

```suggestion
        shadowColor: color('shadow color', 'rgba(0,0,0,0.5)'),
```

miukimiu (changes requested):
Ideally, we should by default set the legend color to white or black based on the contrast of the bar fill. And always ensure a shadow to increase the readability. Like this example from highcharts.

Screenshot 2020-09-28 at 12 40 43

stories/bar/2_label_value.tsx
miukimiu: I also increased the size fo the value font size to 12.

dej611: Do you think it should be a default setting or just a better starting point for this example?

miukimiu: I was seeing the other charts where we have labels/values on top and normally they are black or with `textInvertible: true`

Screenshot 2020-09-28 at 14 42 08


So for consistency, I think we should keep the **value labels black** and just increase the size to **12**. 

How do you feel of then creating a second example to show that the value label can be customized? In this new example, you would use **white value labels** with a **shadow**.

nickofthyme: @miukimiu which would you prefer?

1. Changing color contrast of the shadow
2. Changing color contrast of the text itself
3. Either of the above depending on if a shadow is enabled

stories/bar/2_label_value.tsx
nickofthyme: by default in the chart config or just on this one story?

dej611: Forgotten to report here, just the story.

stories/bar/2_label_value.tsx
dej611: From my understanding when it flips the color on text the shadow should change too.
But you are right, there is also the additional scenario where the shadow is not enabled (`transparent`?) and only the text changes.

nickofthyme: Yeah thats true but I was actually thinking that the shadow be `null` or `undefined`, as it's not needed when the text contrast is dynamic.

dej611:
* the `fill` property for the `DisplayValueStyle` handles 3 different scenarios: font color, font + border color, automatic color assignment (w/o border)
* The second scenario leaves full control to the user:
  Screenshot 2020-10-07 at 15 28 10
* the third scenario lets the library pick the color:
  Screenshot 2020-10-07 at 15 28 24

**Note**: when the automatic flag is tuned on, the border color has a little bit of alpha blending as from some manual testing works better in general. It's one variable so it's easy to tweak

miukimiu (approved):
Tested locally and LGTM! 🎉 

The shadow definitely improves the readability and it makes it possible to read value labels when a bar is too small in height.

api/charts.api.md
```diff
@@ -535,9 +535,17 @@ export interface DisplayValueSpec {
...
     offsetX: number;
     offsetY: number;
+    fill: Color | {
+        color: Color;
+        borderColor?: Color;
+    } | {
+        textInvertible: boolean;
+        textContrast?: number | boolean;
+        textBorder?: boolean;
+    };
```
markov00, line 548: It would be great to add also the border width as a customizable option. This can help create different themes and adjust the theme when consuming the chart

src/chart_types/xy_chart/renderer/canvas/primitives/text.ts
```diff
@@ -49,6 +50,12 @@ export function renderText(
       if (translation?.x || translation?.y) {
         ctx.translate(translation?.x ?? 0, translation?.y ?? 0);
       }
+      if (font.shadow) {
+        ctx.lineWidth = 1.5;
+        ctx.strokeStyle = font.shadow;
+        ctx.strokeText(text, origin.x, origin.y);
+        ctx.shadowBlur = 0;
```
markov00, line 57: shadowBlur is by default 0, probably you want to disable the `shadowBlur` if configured in the context somewhere else and probably before writing the stroke, so moving this before the strokeText call or before the if statement you want to apply that reset also to the fillText

src/utils/themes/theme.ts
```diff
@@ -290,9 +290,17 @@ export interface Theme {
...
   offsetX: number;
   offsetY: number;
+  fill:
+    | Color
+    | { color: Color; borderColor?: Color }
+    | {
+        textInvertible: boolean;
+        textContrast?: number | boolean;
+        textBorder?: boolean;
+      };
```
markov00, line 303: This probably is a breaking change that needs to be notified

stories/bar/bars.stories.tsx
```diff
@@ -35,6 +35,7 @@ export { Example as withLinearXAxisNoLinearInterval } from './6_linear_no_linear
 export { Example as withTimeXAxis } from './7_with_time_xaxis';
 export { Example as withLogYAxis } from './8_with_log_yaxis';
 export { Example as withStackedLogYAxis } from './9_with_stacked_log';
+export { Example as withValueLabelAdvanced } from './50_label_value_advanced';
```
markov00, line 38: can you move this right after the other `withValueLabel` so the user can play directly with the advanced settings?

markov00 (approved):
LGTM, tested locally.
I've just wrote a few comments with small improvements.
I think we can also reduce the scope of the storybook story to a smaller dataset and a smaller number of options or highlighting them on a different tab

stories/bar/50_label_value_advanced.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+  const displayValueSettings = {
+    showValueLabel,
+    isAlternatingValueLabel,
+    isValueContainedInElement,
+    hideClippedValue,
+  };
+
+  const debug = boolean('debug', false);
+  const useInverted = boolean('textInverted', false);
```
wylieconlon, line 55: I was a bit surprised that the inverted colors aren't configurable in any way, was this intentional?

dej611: Yes. There are basically three configurations available for `DisplayValueLabels` text:
```ts
// basic coloring
Color
|
// advanced coloring 
{
     color: Color;
     borderColor?: Color;
     borderWidth?: number;
   } 
|
// let the library compute it 
{
     textInvertible: boolean;
     textContrast?: number | boolean;
     textBorder?: number | boolean;
};
```

The only way to influence the library color pick in case of automatic color assignment is the `textContrast` parameter, which influence the threshold for the automatic switch between black and white.

Issue #788, opened 2020-08-20
Value labels could have a font size as a ratio of the bar width


wylieconlon:
One of the issues we have run into with the [Lens value labeling](https://github.com/elastic/kibana/pull/75426) work is that the font sizes are hard to set correctly for a chart we don't know the width of. I think the chart library should be able to use the geometry of the bars to determine a good font size, instead of the font size being set externally.

Issue #786, opened 2020-08-19
Value labels isValueContainedInElement breaks when value input is sent as number instead of string


dej611:
**Describe the bug**
The `isValueContainedInElement` flag may crash the chart rendering in some scenarios.

**To Reproduce**
1. Go to the `Bar Chart with value label` story
2. Change the `tickFormat` prop in the `LeftAxis` component to return a number rather than a string (just remove the `.toFixed(2)` from the function code)
2. Click on 'contain value label within bar element'
3. 💥 

**Expected behaviour**
The chart is rendered and the value labels are shown on the chart.

**Screenshots**
<img width="1186" alt="Screenshot 2020-08-19 at 12 50 26" src="https://user-images.githubusercontent.com/924948/90625937-97153d00-e21a-11ea-8bec-012d6fb36f90.png">

**Version (please complete the following information):**
 - OS: Mac
 - Browser: Chrome
 - Elastic Charts: 21.0.0

**Additional context**
The API says the return value of the `tickFormat` function should be a `string`, but somehow TS managed to miss that and it happened to have a number output.

*Update*: The same error happens when no yAxis is associated with the current series, therefore no `tickFormat` function is used.

Not super important, but it would be helpful to be resilient to this kind of scenario, when possible.

**Errors in browser console**
```
text.ts:78 Uncaught TypeError: text.split is not a function
    at Object.wrapLines (text.ts:78)
    at Object.renderBarValues (bar.ts:72)
    at renderers.ts:161
    at Object.withContext (index.ts:36)
    at renderers.ts:158
    at index.ts:52
    at Array.forEach (<anonymous>)
    at Object.renderLayers (index.ts:52)
    at renderers.ts:69
    at Object.withContext (index.ts:36)
wrapLines @ text.ts:78
renderBarValues @ bar.ts:72
(anonymous) @ renderers.ts:161
withContext @ index.ts:36
(anonymous) @ renderers.ts:158
(anonymous) @ index.ts:52
renderLayers @ index.ts:52
(anonymous) @ renderers.ts:69
withContext @ index.ts:36
renderXYChartCanvas2d @ renderers.ts:41
push../src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx.XYChartComponent.drawCanvas @ xy_chart.tsx:133
push../src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx.XYChartComponent.componentDidUpdate @ xy_chart.tsx:119
[...]

PR #789, opened 2020-08-20
feat(bar_chart): scaled font size for value labels

wylieconlon:
This is my first attempt at creating scaled font sizes for text in the bar charts, which has similarity to the scaling logic that is implemented in partition charts by @monfera. I've applied the simplest possible logic, which is to take the maximum font size that fits these restrictions:

1. Count the number of characters in the label
2. Divide the available bar width by the number of characters, and take the integer value that's smaller
3. If the font size is less than the minimum in the theme, then use the minimum size instead

It doesn't work correctly for horizontal charts because I don't know how to determine rotation.

Fixes https://github.com/elastic/elastic-charts/issues/788

Screen Shot 2020-08-20 at 4 41 20 PM
Screen Shot 2020-08-20 at 4 41 26 PM
Screen Shot 2020-08-20 at 4 41 36 PM

### Checklist


src/chart_types/xy_chart/rendering/rendering.ts
```diff
@@ -390,8 +390,16 @@ export function renderBars(
           : undefined
         : formattedDisplayValue;
 
-    const computedDisplayValueWidth = bboxCalculator.compute(displayValueText || '', padding, fontSize, fontFamily)
-      .width;
+    const targetFontSize = displayValueText
+      ? Math.max(Math.floor(width / displayValueText.length), fontSize)
+      : fontSize;
```
markov00, line 395: This give a valid estimation only using monospaced font.
Instead of computing the fontSize, you can compute a scale factor to fit starting from the bar width and the computedText width, applying this scale factor when rendering the text:

Here, instead of computing the `targetFontSize` you can compute the scale factor with something like:
`const textScale = width / computedDisplayValueWidth;`
passing this `textScale` to the `renderText` function (used by the `renderBarValues`) in that way:
```
export function renderText(
  ...
  translation?: Partial<Point>,
  scale = 1,
) {
...
// replace the ctx.fillText(text, origin.x, origin.y); with
ctx.scale(textScale, textScale);
ctx.fillText(text, 0, 0);
...
```
with should scale every bar text to the max amount of space available. You can also add a bit of padding around the text reducing the bar width by percentage or pixel and introduce a max/min width threshold to limit the upper scale and lower scale.

markov00 (commented):
Thank you @wylieconlon for the PR.
I've added a suggestion to make this work to compute that target font size in a more precise way than the current proposed solution.
If you need the chart rotation parameter, unfortunately is not available at that level, but you can pass it down from `computeSeriesGeometries` ==> `renderGeometries` ==> `renderBars`

I've also a doubt about the readability of the text using this varying-font-size approach:
longer text (due to more decimal points or higher value) can results in smaller font sizes (see the image as an extreme example). On a data visualization, these font changes can be interpreted in different ways from the users, assuming more importance to bigger texts rather than the smaller ones. Uniform font sizes are preferable if we don't need to communicate any exceptional significance to that value. @monfera @nickofthyme thoughts on that?

wylieconlon:
@markov00 Thanks for the feedback! I agree that uniform font sizes would look much better. To implement your suggestion, I think we would need to calculate the formatted values for all bars, and then scale the font size based on the highest character count of the formatted values. What do you think about that approach?

markov00:
> @markov00 Thanks for the feedback! I agree that uniform font sizes would look much better. To implement your suggestion, I think we would need to calculate the formatted values for all bars, and then scale the font size based on the highest character count of the formatted values. What do you think about that approach?

I think we should compute the bounding box of all the formatted values and then scale the font size based on the maximum width of these boundingboxes.

wylieconlon:
@markov00 Do you think it makes sense to change from `fontSize` on the API level to `minFontSize` and `maxFontSize`?

markov00:
> @markov00 Do you think it makes sense to change from `fontSize` on the API level to `minFontSize` and `maxFontSize`?

@wylieconlon we can have that change without introducing a breaking change if we use a union type that covers two cases: fixed font size or scaled font size. Something like this could work:
```ts
export interface TextStyle {
  fontSize: number | { min: number; max: number };
  ...
}
```
With that, we can handle both use cases: users that want a specific fixed font size + users that look at a scalable font size

dej611:
Reworked the PR to address the various feedback:
* ✨ Add a `min/max` range format for `DisplayValueStyle` type
  * why not `TextStyle`? Because it has a general effect on the whole code base not affected by the change - breaking many tests as well
* ✨ Used the `scale` factor approach to rescale size if a range is passed
  * 🐛 tuned `isValueContainedInElement` to work with the new auto scaling feature
  * ✨ tuned the size scaling based on chart rotation
  * 💄 Labels now use 70% of space width - value hardcoded
* ⚠️ small bugfixes for linting warnings

src/chart_types/xy_chart/rendering/rendering.ts
```diff
@@ -56,6 +57,59 @@ export interface MarkSizeOptions {
...
+  scale: number,
+  fixedFontSize: number,
+  limits: DisplayValueStyle['fontSize'],
+): number {
+  if (typeof limits === 'number') {
+    // it's a fixed size, so it's always ok
+    return 1;
+  }
+  const finalFontSize = scale * fixedFontSize;
+  return finalFontSize <= limits.max && finalFontSize >= limits.min ? scale : 0;
```
wylieconlon, line 74: The fallback scaling factor is 0, that seems like it'll cause no text to render?

src/chart_types/xy_chart/rendering/rendering.ts
```diff
@@ -384,34 +439,37 @@ export function renderBars(
...
-      targetFontSize,
-      fontFamily,
-    ).width;
-    const displayValueWidth =
-      displayValueSettings && displayValueSettings.isValueContainedInElement ? width : computedDisplayValueWidth;
+      { padding, fontSize, fontFamily, bboxCalculator, width },
+      displayValueSettings,
+    );
+
+    const isHorizontalRotation = chartRotation == null || [0, 180].includes(chartRotation);
```
wylieconlon, line 452: Either the logic or naming is wrong, the horizontal charts are 90 and 270: https://elastic.github.io/elastic-charts/?path=/story/rotations--rotations-90-deg-ordinal

dej611: Eheh I had a chat with @markov00 about that the other day. He explained to me that their naming reference where the bars are located (for a vertical bar chart they are located on the x axis, therefore `horizontal`).
I used `isVertical` here before which makes more sense for me as well, then changed after the chat as consitency is best when maintain the codebase.

src/chart_types/xy_chart/rendering/rendering.ts
dej611: It can be changed easily: right now the values disappear when the automatic font size is not within the provided range. While testing it I found it nice behaviour.
It is possible to change in a say that the boundary font size "stick" when the auto font size passes them.

dej611: Updated the feature to have the sticky min/max values. The hiding behaviour can still be simulated tuning the `min`/`max` values.

src/chart_types/xy_chart/rendering/rendering.ts
```diff
@@ -56,6 +57,66 @@ export interface MarkSizeOptions {
...
+  fixedFontSize: number,
+  limits: DisplayValueStyle['fontSize'],
+): number {
+  if (typeof limits === 'number') {
+    // it's a fixed size, so it's always ok
+    return 1;
+  }
+  const finalFontSize = scale * fixedFontSize;
+  if (finalFontSize > limits.max) {
+    return limits.max / fixedFontSize;
```
markov00, line 75: please add a check for the division by zero

src/chart_types/xy_chart/rendering/rendering.ts
```diff
@@ -56,6 +57,66 @@ export interface MarkSizeOptions {
...
+  }
+  if (finalFontSize < limits.min) {
+    // it's technically 1, but keep it generic in case the fixedFontSize changes
+    return limits.min / fixedFontSize;
```
markov00, line 79: same as before

src/chart_types/xy_chart/rendering/rendering.ts
```diff
@@ -56,6 +57,66 @@ export interface MarkSizeOptions {
...
+    fontFamily: string;
+    bboxCalculator: CanvasTextBBoxCalculator;
+    width: number;
+  },
+  displayValueSettings: DisplayValueSpec | undefined,
+): { fixedFontScale: number; displayValueWidth: number } {
+  const fixedFontScale = typeof fontSize === 'number' ? fontSize : fontSize.min;
+
+  const computedDisplayValueWidth = bboxCalculator.compute(text || '', padding, fixedFontScale, fontFamily).width;
+  if (typeof fontSize !== 'number') {
```
markov00, line 108: nit: extract `typeof fontSize === 'number'` into a variable and reuse it

src/chart_types/xy_chart/rendering/rendering.ts
```diff
@@ -384,25 +446,37 @@ export function renderBars(
...
+      displayValueSettings,
+    );
+
+    const isHorizontalRotation = chartRotation == null || [0, 180].includes(chartRotation);
+    // Take 70% of space for the label text
+    const fontSizeFactor = 0.7;
+    // Pick the right side of the label's box to use as factor reference
+    const referenceWidth = isHorizontalRotation ? displayValueWidth : fixedFontScale;
+
+    const tempTextScalingFactor = (width * fontSizeFactor) / referenceWidth;
```
markov00, line 465: check division by zero

src/chart_types/xy_chart/state/utils/utils.ts
```diff
@@ -464,7 +467,8 @@ function renderGeometries(
   axesSpecs: AxisSpec[],
   chartTheme: Theme,
   enableHistogramMode: boolean,
-  stackMode?: StackMode,
+  stackMode: StackMode | undefined,
+  chartRotation: number,
```
markov00, line 471: nit: you can just invert the two and keep the `stackMode` as optional to be consistent with most of the code. Since it's not an exported function the refactoring should be easy

stories/bar/50_simple_value_label.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const maxFontSize = number('Max font size', 25);
+  const minFontSize = number('Min font size', 10);
+
+  const theme = {
+    barSeriesStyle: {
+      displayValue: {
+        fontSize: useFixedFontSize ? fixedFontSize : { max: maxFontSize, min: minFontSize },
+        fontFamily: "'Open Sans', Helvetica, Arial, sans-serif",
+        fontStyle: 'normal',
+        padding: 0,
```
markov00, line 68: I've seen that in the `renderingBars` we are ignoring the theme `padding` value. Do you want to take a look and see if we can enable it back again?

markov00 (commented):
Looks good to me, I've checked the code and tested locally seems to work fine and scales correctly all the values. I've check also the scale with very different value lengths and seems to scale correctly per bar

src/chart_types/xy_chart/rendering/rendering.ts
dej611: I think I did that but TS wasn't good at remembering it as type guard, so I had to make it explicit everytime.

stories/bar/50_simple_value_label.tsx
dej611: From offline discussion, we can delay this as for now.

src/chart_types/xy_chart/rendering/rendering.ts
dej611: Tried again but TS is still not happy about that
Screenshot 2020-10-06 at 14 50 53

nickofthyme (commented):
Looks amazing so far.

I get an error when enabling both the _contain within label_ and  _alternating label_ options.

dej611:
Yes, that's related to this bug: https://github.com/elastic/elastic-charts/issues/786 .

src/chart_types/xy_chart/state/utils/utils.ts
markov00: solved in https://github.com/elastic/elastic-charts/pull/789/commits/b8383bcadc34188f0bbffb7674df170ea1d7bb3d

src/chart_types/xy_chart/rendering/rendering.ts
markov00: solved in https://github.com/elastic/elastic-charts/pull/789/commits/0ad57318e70394906055ca213e27b895c5c5248e

src/chart_types/xy_chart/rendering/rendering.ts
markov00: done in https://github.com/elastic/elastic-charts/pull/789/commits/3165ec4f671b51613ec5e8c9b1d3a04847fe9626

src/chart_types/xy_chart/rendering/rendering.ts
markov00: done in https://github.com/elastic/elastic-charts/pull/789/commits/3165ec4f671b51613ec5e8c9b1d3a04847fe9626

stories/bar/bars.stories.tsx
```diff
@@ -79,3 +79,5 @@ export { Example as testSingleHistogramBarChart } from './44_test_single_histogr
 export { Example as testMinHeightPositiveAndNegativeValues } from './46_test_min_height';
 export { Example as testTooltipAndRotation } from './48_test_tooltip';
 export { Example as testDualYAxis } from './49_test_dual_axis';
+
+export { Example as simpleValuelabel } from './50_simple_value_label';
```
markov00, line 83: typo
```suggestion
export { Example as simpleValueLabel } from './50_simple_value_label';
```

markov00 (approved):
Tested locally and it looks very good!
Just a small typo on the story title

stories/bar/50_simple_value_label.tsx
```diff
@@ -0,0 +1,131 @@
...
+        fill: color('value color', '#000'),
+        offsetX: number('offsetX', 0),
+        offsetY: number('offsetY', 0),
+        alignment: {
```
wylieconlon, line 72: The alignment properties don't appear to do anything yet, maybe remove this?

stories/bar/50_simple_value_label.tsx
```diff
@@ -0,0 +1,131 @@
...
+const frozenDataMediumVolume = generateDataWithAdditional(10);
+const frozenDataHighVolume = generateDataWithAdditional(50);
+
+const frozenData: { [key: string]: any[] } = {
+  s: frozenDataSmallVolume,
+  m: frozenDataMediumVolume,
+  h: frozenDataHighVolume,
+};
+
+export const Example = () => {
```
wylieconlon, line 41: You've created a new storybook, but I noticed that you didn't update the existing storybook for value labels. I found it a little confusing not to have these options on the "value label" storybook. Would it be worth combining into a single story?

wylieconlon (commented):
Looking good @dej611! I can't approve since this is technically my own PR, but left some small comments.

stories/bar/50_simple_value_label.tsx
dej611: I've left this in here as it will probably be merged with other stories as well.

PR #802, opened 2020-09-04
feat(tooltip): series tick formatters

nickofthyme:
## Summary

Fix for https://github.com/elastic/kibana/issues/74951.

Allow `SeriesSpec` to have `tickFormat` for different tooltip formatted values.

### Set different axes formatter
Screen Recording 2020-09-03 at 09 39 PM

### Axis formatter defaults to first series formatter
Screen Recording 2020-09-03 at 09 39 PM

### Axis formatter defaults to default formatter with no series formatters
Screen Recording 2020-09-03 at 09 40 PM

### Checklist
- [x] Any consumer-facing exports were added to `src/index.ts` (and stories only import from `../src` except for test data & storybook)
- [x] This was checked for cross-browser compatibility
- [x] Proper documentation or storybook story was added for features that require explanation or tutorials
- [x] Unit tests were updated or added to match the most common scenarios

PR #810, opened 2020-09-08
fix: build issues and tooltip formatting issues

nickofthyme:
## Summary

Fix x-axis and tooltip header formatting caused by #802, by adding `formattedValue` string to `TooltipValue` in addition to actual `value`.

The issue in #802 changed value from the given value to the formatted string value when using the `tooltip.headerFormatter` prop.

Also fixes...
- add `--no-version-updates` flag to storybook start script
- upgrade `@storybook/` dependencies
- remove `@types/react` causing type error in dist library check
- move redux `optionalDependencies` to `devDependencies`

### Checklist


nickofthyme:
jenkins retest this please

monfera (approved):
LGTM on an less than full length look. Would be good to have a visual regression test that passes with this change but wouldn't have passed before this, but it can be added subsequently to merging this PR

monfera:
jenkins retest this please

nickofthyme:
> Would be good to have a visual regression test that passes with this change but wouldn't have passed before this, but it can be added subsequently to merging this PR

Added in bebb62c

PR #817, opened 2020-09-14
fix: show crosshairs for external pointer events

### LINKED ISSUE #816: Show crosshairs on external pointer events
**Describe the bug**
As of https://github.com/elastic/elastic-charts/pull/698, the `dispatchExternalPointerEvent` call will not render crosshairs for external tooltips. Prior to this change this was the default behavior.

**To Reproduce**
1. Go to http://localhost:9001/?path=/story/interactions--cursor-update-action
2. Change the value of `visible` on the second chart to `false`
3. Notice the absence of an external tooltip on either chart

**Expected behaviour**
If a `dispatchExternalPointerEvent` is used, show the crosshairs for the respective pointer value. Even when `externalPointerEvents.tooltip.visible` is false. When this is false, only the tooltip should be hidden.

**Screenshots**
image

**Kibana Cross Issues**
https://github.com/elastic/kibana/issues/76954

nickofthyme:
## Summary

Fixes #816 
Related to https://github.com/elastic/kibana/pull/77306

If a `dispatchExternalPointerEvent` is used, the crosshairs now show for the respective pointer value.

`externalPointerEvents.tooltip.visible` now only controls the tooltip visibility.

To disable crosshairs, don't use `dispatchExternalPointerEvent` at all.

### Checklist


markov00 (approved):
LGTM

PR #819, opened 2020-09-15
feat: cancel brush event with escape key

nickofthyme:
## Summary

Enable canceling brushing and click of xy charts with <kbd>Escape</kbn>

Screen Recording 2020-09-14 at 07 06 PM

### Checklist


markov00 (approved):
Tested locally and looks good.
nit: if we can avoid that `hasBrushListener` instance variable it will be great, if not I'm fine with the current implementation

nickofthyme:
@markov00 Updated based on our talk.

Issue #580, opened 2020-03-06
Legend re-design


markov00:
## Legend
The current legend design, doesn't scale well. I'm collecting a set of features, constraint and issues that needs to addressed when redesigning the legend

### Features
1. Can be structured in a hierarchical way: used for multilayer piechart or splitted charts or in a flat way (https://github.com/elastic/elastic-charts/issues/295, https://github.com/elastic/elastic-charts/issues/223)
2. Each item can contain:
    1. coloured icons (depending on the element, line, bar, area, annotation, https://github.com/elastic/elastic-charts/issues/555 point https://github.com/elastic/elastic-charts/issues/502 )
    2. Text, arbitrary long, possibly wrapped or truncated (configurable)
    3. An arbitrary number of extra values like te max, current, avg value of the specific series, annotation (https://github.com/elastic/elastic-charts/issues/561)
3. Rendered in two formats: vertically (left/right) or horizontally (top/bottom)
4. We should limit the amount of space used
5. We should avoid when possible scrolling
6. Can be configured to automatically be “collapsed”/”minimized” reducing the extra values, truncating text etc
7. Each item can have multiple actions:
    1. Open a popup to change color (done with an external component)
    2. Hide and show the series/annotations on click (https://github.com/elastic/elastic-charts/issues/299)
    3. Highlight a series on hover
    4. Filter in/out a series
    5. All these actions needs to be accessible from keyboard
8. It can be rendered as a table if we a numerous set of extra values are used (see grafana)
9. It can take up as much as space as needed, or be limited to a specific width in pixel or percentage of the chart.
10. Be able to simplify the legend for printing/png exporting, especially long/scrolling legends (https://github.com/elastic/elastic-charts/issues/458)

### Constraints
1. In case of a dynamic computation of space required for the Legend, it’s necessary to render or compute the width/height of the legend before rendering the chart itself. If possible, computing the required width/height is preferable, to avoid rendering the legend and the chart into two different moment in time
2. The same format, except the aspect, can be applied to a tooltip if necessary, both components brings in exactly the same type of information, so we should be able to reuse that component on the legend

Other related issues:
https://github.com/elastic/elastic-charts/issues/79

markov00:
Closing for now until we don't have a set of design ideas to share

VijayDoshi:
One very useful action that could be taken in a legend is renaming/aliasing the field. We don't support aliases today; however, it would be great to have a field alias or "rename" capability right from the legend eventually.

[...]

PR #820, opened 2020-09-15
fix: legend dark mode hover color

nickofthyme:
## Summary

Fixes https://github.com/elastic/kibana/issues/77283

Fixes bad styles for dark theme after changes to legend item hover styles in https://github.com/elastic/elastic-charts/pull/749

Screen Recording 2020-09-14 at 07 12 PM

### Checklist


markov00 (approved):
LGTM

src/components/legend/_legend_item.scss
```diff
@@ -16,7 +16,7 @@ $legendItemVerticalPadding: $echLegendRowGap / 2;
   }
 
   &:hover {
```
cchaos, line 18: This should actually only get background color on `:focus`. Leaving the underline as the hover state.
```suggestion
  &:focus {
```

cchaos (dismissed):

src/components/legend/_legend_item.scss
nickofthyme: Unfortunately, at the moment we don't have accessible interactions on the legend and thus `focus` would do nothing. This will be revisited in https://github.com/elastic/elastic-charts/issues/580.

Are you ok with it being `hover` until then?

src/components/legend/_legend_item.scss
```diff
@@ -16,7 +16,7 @@ $legendItemVerticalPadding: $echLegendRowGap / 2;
...
-    background-color: #e7f0f7;
+    background-color: $euiFocusBackgroundColor;
```
miukimiu, line 19: I've been testing the colors and we can change to:

```suggestion
    background-color: $euiColorLightestShade;
```

I was first thinking to remove the background but then it seemed confusing because **hovering all the legend highlights the series**. 

I just think we decided on the wrong color on #820. So let's reserve the `$euiFocusBackgroundColor` for the upcoming `focus` states and use the  `$euiFocusBackgroundColor` when hovering the legend.

miukimiu (changes requested):

src/components/legend/_legend_item.scss
nickofthyme: merged suggestion 0d54de7

PR #821, opened 2020-09-15
refactor: decouple the brush component from XY chart

markov00:
## Summary

This PR decouple the brush rendering component from the XY axis chart. This can be then reused by heatmap that requires this feature.

I've added a set of more specific methods on the `InternalChartState` interface to decouple the implementation of each single chart related to chart dimensions and brush area computation.
I've just skipped the implementation of these methods in Partition and Goal charts.

Right now the brush area is just a rect, but we can extend that to be whatever we need.


nickofthyme (approved):
LGTM, tested brush stories locally on chrome and appear unaffected by changes.

Issue #808, opened 2020-09-08
Allow non-primitive data values


flash1293:
**Is your feature request related to a problem? Please describe.**
In Kibana values in a chart are not always primitive values like strings or numbers (e.g. when working with ranges). In these cases, the table looks like this (this would be rendered as a bar chart with an ordinal x axis):

| x | y |
| --- | --- |
| `{ from: 5, to: 20 }` | 4 |
| `{ from: 20, to: 40 }` | 7 |

Of course in this case we have a formatter ready to render a nice string tick label out of the object. One solution would be to map over the data and turn all of the objects into the string representation before passing the data to elastic-charts:

| x | y |
| --- | --- |
| 5 < x < 20 | 4 |
| 20 < x < 40 | 7 |

However the problem with this solution is that it makes it harder to get back to the original object. This is necessary on interactions - e.g. if the user clicks a bar associated with the first row in the table, we need the orginial object to turn it into a filter. The string representation doesn't help in that case.

Currently, elastic-charts seems to filter out rows containing objects, so nothing is rendered:

String values (works)
https://codesandbox.io/s/sweet-sunset-f1s3i?file=/src/App.tsx

Object values (breaks)
https://codesandbox.io/s/festive-jang-z78p6?file=/src/App.tsx

**Describe the solution you'd like**
Ideally elastic-charts would simply be able to work with objects as values in the data and treat them as ordinal values. Equality checks could be done using `JSON.stringify` or a similar way of serializing the object.

**Describe alternatives you've considered**

Another option would be to make it possible to pass "additional data" into the click handler, maybe like this:
```tsx
<Chart size={[500, 200]}>
        <Settings
          onElementClick={([[geometry]]) => {
            console.log(geometry.clickHandlerContext[0]); // will be { name: "trousers " }
          }}
        />
        <Axis id="count" title="count" position={Position.Left} />
        <Axis
          id="x"
          title="goods"
          position={Position.Bottom}
          tickFormat={d => `xXx${d}xXx`}
        />
        <BarSeries
          id="bars"
          name="amount"
          xScaleType={ScaleType.Ordinal}
          xAccessor="x"
         clickHandlerContextAccessors={["additional"]}
          yAccessors={["y"]}
          data={[{ x: "trousers", additional: { name: "trousers " }, y: 390, val: 1222 },
            { x: "bags", additional: { name: "bags" }, y: 750, val: 1222 }]}
        />
      </Chart>
```

[...]

PR #822, opened 2020-09-15
feat: expose datum as part of GeometryValue

markov00:
## Summary

This PR partially address the concerns expressed here: https://github.com/elastic/elastic-charts/issues/808
Is currently not allow using an object as a value for X or Y coordinates, so we kindly suggest to add an additional field into your data set that allows us to uniquely identify that object with a string or a number.
As a workaround for a click handler, a way to get back the original object from the stringified/number version is proposed in this PR, exposing the original `datum` as part of the `GeometryValue` and as a consequence exposing the datum to every interaction handlers that use a `GeometryValue`

### Checklist


nickofthyme (approved):
LGTM, tested locally on chrome.

markov00:
Jenkins test this please

PR #829, opened 2020-09-18
fix(axis): style overrides not applied to axis dimensions

nickofthyme:
## Summary

After improvements from #711, the axis theme was allowed to be overridden per axis. These changes failed to account for these overrides in the `computeAxisTicksDimensionsSelector` and just used the `Theme.axis` styles.

This PR adds the overrides to the axis dimensions calculations.

### Before
Image 2020-09-18 at 11 20 41 AM

### After
Screen Recording 2020-09-18 at 01 53 PM

### Checklist


markov00 (approved):
LGTM

Issue #794, opened 2020-08-26
[meta] Vislib functional test replacement


markov00:
**Is your feature request related to a problem? Please describe.**
The replacement of vislib requires also a replacement of the current functional tests available in Kibana.
These tests are checking also the rendering capabilities inside Kibana picking and analyzing DOM elements.
With the current @elastic/charts library, rendering in Canvas, this is no more possible.

**Describe the solution you'd like**
- [x] find and collect every `PageObject` functions, used by functional tests in Kibana, that are checking values/variables in the DOM rendered by vislib
- [x] analyze these functions and extract a set of possible replacement properties/statistics that can be extracted from the chart state or chart selectors (e.g. number of bars, lines, areas, points, bars values, colors, etc)
- [x] put those statistics into the available `<ChartStatus/>` component as attributes only when the `debug` flag is enabled
- [x] enable the `debug` flag in vislib only when running on the CI (we should be careful here because the debug currently also enable the graphical debug that can break Kibana visual regression tests)
- [x] replace the `PageObject` functions for vislib with similar tests using our exposed statistics/properties
- [ ] where possible, extract from the functional tests the chart state (in particular the specs configurations) and create a set of VRTs in the elastic-charts repo to check the correctness of the rendering.

**Describe alternatives you've considered**
Enable Percy on Kibana, but we should define when we want to run it and how

Related issue in Kibana: https://github.com/elastic/kibana/issues/70176

VladLasitsa:
Collected the following functions that we used in function tests for area, line and bar charts:

- getAreaChartData - Calculate needed data using height of rectangle('rect.background') and data about position of one area chart from 'path' in attribute 'd'. This area chart data which we insert in 'path' starts with a 'M'ove to a x,y location, followed by a bunch of 'L'ines from that point to the next.  Those points are the values we're going to use to calculate the data values we're testing. So git rid of the one 'M' and split the rest on the 'L's.

- getAreaChartPaths - Needed for getting number of segments in chart.

- getLineChartData - Calculate needed data using height of rectangle('clipPath rect') and data about position of one line chart from 'circle' in attribute 'cy'.

- getBarChartData - Calculate needed data using height of rectangle('g > g.series > rect') for one bar in chart.

- getChartYAxisRatio - Returns the range/pixel ratio (max value - min value on y axis / position on max value - poation of min value) - needed for chart data getters. So that we can transform heights in data.

- getYAxisLabels - Get all y - axis labels

[...]

PR #834, opened 2020-09-24
feat: debug state flag added to chart status

nickofthyme:
## Summary

Related to https://github.com/elastic/elastic-charts/issues/794

Adds `debugState` prop to `Settings` component to render textual representation of chart.

see http://localhost:9001/?path=/story/debug-options--debug-state

image

### Checklist


src/state/selectors/get_debug_state.ts
```diff
@@ -0,0 +1,227 @@
...
+  ({ geometries }, legend, axes, axesSpecs): string => {
+    const seriesNameMap = getSeriesNameMap(legend);
+    // console.log(geometries);
+
+    const state: DebugState = {
+      legend: getLegendState(legend),
+      axes: getAxes(axes, axesSpecs),
+      areas: geometries.areas.map(getAreaState(seriesNameMap)),
+      lines: geometries.lines.map(getLineState(seriesNameMap)),
+      bars: getBarState(seriesNameMap, geometries.bars),
```
markov00, line 52: what's the deal of reshaping the generated data and geometry into a different type?
Can't we just pass the data as it is?
I'm wondering what will happen when we have to change a data format of one of these geometry, we should re-adapt this mappers, probably these types and then the functional tests that use them.
Instead if we just throw the data as it is, we just need to change the functional test to adapt to that. What do you think?

src/state/selectors/get_debug_state.ts
```diff
@@ -0,0 +1,227 @@
...
+    ({ color, seriesIdentifier: { key }, seriesStyle: { rect, rectBorder }, value: { x, y, mark } }: BarGeometry) => {
+      if (buckets.has(key)) {
+        buckets.get(key)!.bars.push({ x, y, mark });
+      } else {
+        const name = seriesNameMap.get(key) ?? '';
+        buckets.set(key, {
+          key,
+          name,
+          color,
+          bars: [{ x, y, mark }],
```
markov00, line 104: we should also return the width and height + the original datum as these are the checks done by the functional tests

src/specs/settings.tsx
```diff
@@ -290,7 +290,14 @@ export interface SettingsSpec extends Spec {
...
    */
   externalPointerEvents: ExternalPointerEventsSettings;
+  /**
+   * Show debug shadow elements on chart
+   */
   debug: boolean;
+  /**
+   * Show debug render state on `ChartStatus` component
+   */
+  debugState: boolean;
```
markov00, line 300: Do we want to "advertise" this as an available option?
I think we should keep this as internal or something similar to avoid abuse and to avoid having to deal with breaking changes on that format every time we break something in the debug state or the output format we are using

markov00 (commented):
The code looks good and works perfectly. I've left few comments on the data structure and API that should be discussed

src/specs/settings.tsx
nickofthyme: Added `@beta` flag

src/state/selectors/get_debug_state.ts
nickofthyme: talked about just using value not pixel values

src/state/selectors/get_debug_state.ts
nickofthyme: Discussed usefulness of reshaping geometry to consume easier and have smaller string to add to dom.

src/components/chart_status.tsx
```diff
@@ -22,12 +22,14 @@ import { connect } from 'react-redux';
 
 import { RenderChangeListener } from '../specs';
 import { GlobalChartState } from '../state/chart_state';
+import { getDebugStateSelector } from '../state/selectors/get_debug_state';
```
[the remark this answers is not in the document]
nickofthyme: Changed to internal state selector

markov00 (approved):
LGTM, I've added some minor comments.
Please remember to remove the `elastic-charts-22.0.0.tgz ` file

src/chart_types/xy_chart/state/selectors/get_debug_state.ts
```diff
@@ -0,0 +1,252 @@
...
+          labels: [],
+          visible: hasVisibleStyle(rect) || hasVisibleStyle(rectBorder),
+        });
+      }
+
+      buckets.get(key)!.bars.push({ x, y, mark });
+
+      if (label) {
+        buckets.get(key)!.labels.push(label);
+      }
```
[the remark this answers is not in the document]
nickofthyme: c28e687

api/charts.api.md
```diff
@@ -8,6 +8,12 @@ import { $Values } from 'utility-types';
 import { ComponentType } from 'react';
 import React from 'react';
 
+// Warning: (ae-forgotten-export) The symbol "AccessorObjectKey" needs to be exported by the entry point index.d.ts
+// Warning: (ae-forgotten-export) The symbol "AccessorArrayIndex" needs to be exported by the entry point index.d.ts
```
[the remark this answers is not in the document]
nickofthyme: c28e687

PR #837, opened 2020-09-29
feat: allow serialization of complex values

### LINKED ISSUE #808: Allow non-primitive data values

nickofthyme:
## Summary

closes #808

Allows non-primative values to be passed through a serializer. Allows different serializer function for each series on all non-primative `x`, `y` and `mark` accessor values.

> Ignores all serialized values that are non-primative values or if the serialization throws an error.

```tsx
const data = [
  {
    y: {
      value: 2,
    },
    x: {
      from: 3,
      to: 4,
    },
  },
];

return (
  <BarSeries
    yScaleType={ScaleType.Linear}
    xSerializer={({ from, to }) => `${from} - ${to}`}
    ySerializer={({ value }) => value}
    data={data}
  />
);

const equivalentData = [
  {
    y: 2,
    x: '3 - 4',
  },
];
```

### Checklist


flash1293:
Pretty cool :) cc @dej611 This will simplify the logic in the Lens xy chart renderer because we don't have to map back from the serialized x value.

nickofthyme:
Closing this issue to use `AccessorFn` in place of serializer see #838.

Currently the `xAccessor` allows for `AccessorFn` which is analogous to using a separate serializer.

```tsx
return (
  <BarSeries
    id="bars"
    yScaleType={ScaleType.Linear}
    xAccessor={({ from, to }) => `${from} - ${to}`}
    yAccessors={['y']}
    data={[
      { y: 2, obj: { from: 10, to: 20 } },
      { y: 7, obj: { from: 20, to: 30 } },
      { y: 3, obj: { from: 30, to: 40 } },
      { y: 6, obj: { from: 40, to: 50 } },
    ]}
  />
)
```

Issue #832, opened 2020-09-23
Rect annotations can display beyond the chart area


nickofthyme:
**Describe the bug**
`RectAnnotation` can display beyond the chart area

**To Reproduce**
1. Add `RectAnnotation` with a `x0` value that goes beyond the end of the domain.
2. Notice the annotation rendered beyond the chart area.

**Expected behaviour**

**Screenshots**
<img width="1415" alt="Image 2020-09-23 at 7 57 34 AM" src="https://user-images.githubusercontent.com/19007109/94015617-bdc31800-fd72-11ea-9eda-658accbfb45e.png">

![image](https://user-images.githubusercontent.com/19007109/94024436-0a135580-fd7d-11ea-91fc-caee2b97e4a5.png)

![image](https://user-images.githubusercontent.com/19007109/94027206-367ca100-fd80-11ea-90bb-81b2341814fd.png)

**Version (please complete the following information):**
 - Elastic Charts: `latest`

PR #841, opened 2020-10-01
fix: limit annotation to the current domain extent

markov00:
## Summary

Fixes the domain extent limiting function when rendering annotations avoiding overflows when the user input falls outside the current computed or configured data domain.

## Before:

Screenshot 2020-10-01 at 22 35 33

## After:
Screenshot 2020-10-01 at 22 53 29

fix https://github.com/elastic/elastic-charts/issues/832
### Checklist


nickofthyme (approved):
Tested locally and issue is now fixed.

image

Issue #798, opened 2020-09-02
Rect Annotation should fallback to full height/width


nickofthyme:
**Describe the bug**
When using x or y axes with a `groupId` along with a `RectAnnotation`. If no `groupId` is set on the `RectAnnotation` it will _not_ appear as there is no `__global__` `groupId` fallback to determine the correct fallback height.

**To Reproduce**
1. Go to https://codesandbox.io/s/cool-cookies-s8okq?file=/src/App.tsx
2. Notice how toggling the `groupId` on the `RectAnnotation` hides and shows

**Expected behaviour**
When no x or y axis `groupId` is determinable, it should fallback to the chart dimensions.

nickofthyme:
Closed by https://github.com/elastic/elastic-charts/pull/842

PR #842, opened 2020-10-01
fix: annotation rendering with no yDomain

### LINKED ISSUE #438: Annotations should be able to handle scenarios where the min value matches the max value of the dataset
There is an original explanation of this problem with data examples [here](https://github.com/elastic/kibana/issues/48385#issuecomment-544495079), and a followup comment on the problem [here](https://github.com/elastic/kibana/issues/48385#issuecomment-545504259). 

The overall issue is that annotations should still be rendered if their values match the max value of the dataset, and that max value is `0`.

rshen91:
## Summary

Fixes #438 and #798 where `<Annotations />` were not shown in cases where they should be. 

- The below story shows `<Annotations />` when the y domain is [0, 0] or spans only one value (domain fit is true). In these cases, annotations will now take the chartDimension height: 

PR842_zero

- The below story shows `<Annotations />` rendering on the x axis if there is no provided group Id. If there is no group id provided and two y axes, the annotation will not still not show and this behavior is expected since the annotation will not know what y axis to mount onto: 

PR842_group

Stories (`Zero Domain` and `With Group Id`) are added in Annotations -> Rect section in storybook. 

### Checklist


src/chart_types/xy_chart/annotations/utils.test.ts
```diff
@@ -92,6 +92,7 @@ describe('annotation utils', () => {
     showOverlappingLabels: false,
     position: Position.Left,
     style,
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
     tickFormat: (value: any) => value.toString(),
```
nickofthyme, line 96: I think you can leave this alone, there are likely hundreds of these issues that should be addressed to remove type `any` down the road. That's why I left this rule as a lint warning (not error) to prevent new cases.

rshen91: Ok thanks for pointing this out! I thought it was an error and not a warning 😅 . But then I removed the comments and ran yarn lint and it seems to be passing. 👍

src/chart_types/xy_chart/annotations/rect/dimensions.ts
```diff
@@ -37,50 +39,76 @@ export function isWithinRectBounds({ x, y }: Point, { startX, endX, startY, endY
...
+
+    let scaledY1 = yScale.pureScale(y1);
     const scaledY0 = yScale.pureScale(y0);
     if (scaledY1 == null || scaledY0 == null) {
       return;
     }
-    const height = Math.abs(scaledY0 - scaledY1);
+    height = Math.abs(scaledY0 - scaledY1);
+    // if the annotation height is 0 override it with the height from chart dimension
+    if (height === 0) {
```
nickofthyme, line 107: I think we should also check here that the domain start and end are the same. Because the domain could be `0` to `10` and still have a zero height. In this case we don't want it to show.

rshen91: Applying changes in 9e90993

src/chart_types/xy_chart/annotations/rect/dimensions.ts
```diff
@@ -103,8 +103,8 @@ export function computeRectAnnotationDimensions(
       return;
     }
     height = Math.abs(scaledY0 - scaledY1);
-    // if the annotation height is 0 override it with the height from chart dimension
-    if (height === 0) {
+    // if the annotation height is 0 override it with the height from chart dimension and if the values in the domain are the same
+    if (height === 0 && yScale.domain[0] === yScale.domain[1] && yScale.domain.length === 2) {
```
nickofthyme, line 107: I would flip this around a little
```suggestion
    if (height === 0 && yScale.domain.length === 2 && yScale.domain[0] === yScale.domain[1]) {
```

rshen91: I like that better - addressed in 19bc042

src/chart_types/xy_chart/annotations/rect/dimensions.integration.test.ts
```diff
@@ -256,4 +256,88 @@ describe('Render rect annotation within', () => {
...
+        { x: 4, y: 10 },
+      ],
+    });
+    MockStore.addSpecs([settings, annotation, bar], store);
+    const expected = computeAnnotationDimensionsSelector(store.getState());
+    const [resultAnnotation] = expected.get('rect_annotation_1') ?? [];
+    expect(resultAnnotation).toMatchObject({
+      rect: { height: 200 },
+    });
+  });
```
nickofthyme, line 321: You could make these more explicit that the expected value is the `height` being passed.

```suggestion
  it('annotation with group id should render with x0 and x1 values', () => {
    const height = 200;
    const store = MockStore.default({ top: 0, left: 0, width: 20, height });
    const settings = MockGlobalSpec.settingsNoMargins();
    const annotation = MockAnnotationSpec.rect({
      groupId: 'group1',
      dataValues: [{ coordinates: { x0: 2, x1: 4 } }],
    });
    const bar = MockSeriesSpec.bar({
      data: [
        { x: 1, y: 0 },
        { x: 2, y: 5 },
        { x: 4, y: 10 },
      ],
    });
    MockStore.addSpecs([settings, annotation, bar], store);
    const expected = computeAnnotationDimensionsSelector(store.getState());
    const [resultAnnotation] = expected.get('rect_annotation_1') ?? [];
    expect(resultAnnotation).toMatchObject({
      rect: { height },
    });
  });
```

nickofthyme (approved):
LGTM, I just made a few minor comments to fix before merging.

src/chart_types/xy_chart/annotations/rect/dimensions.integration.test.ts
rshen91: Cool! Makes sense to make this explicit and clear that it's the same as the chartStore height - addressed in 63a2df1

src/chart_types/xy_chart/annotations/utils.test.ts
```diff
@@ -105,6 +106,7 @@ describe('annotation utils', () => {
     showOverlappingLabels: false,
     position: Position.Bottom,
     style,
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
```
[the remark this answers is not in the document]
rshen91: Nope 😅 ! 63a2df1

src/chart_types/xy_chart/annotations/utils.test.ts
```diff
@@ -92,6 +92,7 @@ describe('annotation utils', () => {
     showOverlappingLabels: false,
     position: Position.Left,
     style,
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
```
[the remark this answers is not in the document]
rshen91: Fixed in 63a2df1

PR #843, opened 2020-10-01
fix: legend item label for functional tests

nickofthyme:
## Summary

Add legend item dataset variable to legend items to select legend items in chart.


rshen91 (approved):
LGTM! Thanks!

PR #844, opened 2020-10-02
fix(heatmap): destroy canvas bbox calculator when done

markov00:
## Summary

The heatmap use the CanvasBBoxCalculator to compute the right size for labels. The calculator wasn't destroyed after the completion of the task and caused the used `canvas` element to remain in the dom.
This fix destroy the `canvas` used after the measurement task is completed


darnautov (approved):
Tested and LGTM 👍

nickofthyme (approved):

PR #847, opened 2020-10-05
feat: heatmap tooltip enhancements and fixes

darnautov:
## Summary

- Add configurable names for X and Y axes labels 
- Add extra values to the tooltip info for X and Y axes, fix `formattedValue` for the cell value
- Export `HeatmapConfig` as `RecursivePartial`

### DIFF
--- a/src/chart_types/heatmap/layout/config/config.ts
+++ b/src/chart_types/heatmap/layout/config/config.ts
@@ -48,7 +48,7 @@ export const config: Config = {
   timeZone: 'UTC',
 
   xAxisLabel: {
-    name: '',
+    name: 'X Value',
     visible: true,
     fill: 'black',
     fontSize: 12,
@@ -64,7 +64,7 @@ export const config: Config = {
     formatter: String,
   },
   yAxisLabel: {
-    name: '',
+    name: 'Y Value',
     visible: true,
     width: 'auto',
     fill: 'black',
--- a/src/chart_types/heatmap/layout/config/config.ts
+++ b/src/chart_types/heatmap/layout/config/config.ts
@@ -48,6 +48,7 @@ export const config: Config = {
   timeZone: 'UTC',
 
   xAxisLabel: {
+    name: '',
     visible: true,
     fill: 'black',
     fontSize: 12,
@@ -63,6 +64,7 @@ export const config: Config = {
     formatter: String,
   },
   yAxisLabel: {
+    name: '',
     visible: true,
     width: 'auto',
     fill: 'black',
--- a/src/chart_types/heatmap/state/selectors/tooltip.ts
+++ b/src/chart_types/heatmap/state/selectors/tooltip.ts
@@ -51,7 +51,7 @@ export const getTooltipInfoSelector = createCachedSelector(
           // X-axis value
           tooltipInfo.values.push({
             label: config.xAxisLabel.name,
-            color: RGBtoString(shape.fill.color),
+            color: 'transparent',
             isHighlighted: false,
             isVisible: true,
             seriesIdentifier: {
@@ -65,7 +65,7 @@ export const getTooltipInfoSelector = createCachedSelector(
           // Y-axis value
           tooltipInfo.values.push({
             label: config.yAxisLabel.name,
-            color: RGBtoString(shape.fill.color),
+            color: 'transparent',
             isHighlighted: false,
             isVisible: true,
             seriesIdentifier: {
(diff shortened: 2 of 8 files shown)


src/chart_types/heatmap/layout/config/config.ts
```diff
@@ -48,6 +48,7 @@ export const config: Config = {
   timeZone: 'UTC',
 
   xAxisLabel: {
+    name: '',
```
markov00, line 51: let's use a better and informative default `X Value`

src/chart_types/heatmap/layout/config/config.ts
```diff
@@ -63,6 +64,7 @@ export const config: Config = {
     formatter: String,
   },
   yAxisLabel: {
+    name: '',
```
markov00, line 67: let's use a better and informative default `Y Value`

src/chart_types/heatmap/state/selectors/tooltip.ts
```diff
@@ -48,9 +48,37 @@ export const getTooltipInfoSelector = createCachedSelector(
       pickedShapes
         .filter(({ visible }) => visible)
         .forEach((shape) => {
-          const xValueLabel = config.xAxisLabel.formatter(shape.datum.x);
+          // X-axis value
           tooltipInfo.values.push({
-            label: `${shape.datum.y} - ${xValueLabel}`,
+            label: config.xAxisLabel.name,
+            color: RGBtoString(shape.fill.color),
```
markov00, line 54:
```suggestion
            color: 'transparent',
```
I think X and Y values should not be connected to the cell color

src/chart_types/heatmap/state/selectors/tooltip.ts
```diff
@@ -48,9 +48,37 @@ export const getTooltipInfoSelector = createCachedSelector(
...
+              key: spec.id,
+            },
+            value: `${shape.datum.x}`,
+            formattedValue: config.xAxisLabel.formatter(shape.datum.x),
+          });
+
+          // Y-axis value
+          tooltipInfo.values.push({
+            label: config.yAxisLabel.name,
+            color: RGBtoString(shape.fill.color),
```
markov00, line 68: Same comment as above
```suggestion
            color: 'transparent',
```

src/chart_types/heatmap/state/selectors/tooltip.ts
```diff
@@ -48,9 +48,37 @@ export const getTooltipInfoSelector = createCachedSelector(
...
+              specId: spec.id,
+              key: spec.id,
+            },
+            value: `${shape.datum.y}`,
+            formattedValue: config.yAxisLabel.formatter(shape.datum.y),
+          });
+
+          // Cell value
+          tooltipInfo.values.push({
+            label: '',
```
markov00, line 81: Here, on XYAxis charts we use the `spec.name` or the `spec.id` as fallback
```suggestion
            label: label: spec.name ?? spec.id,
```
You can add an optional `name` parameter on the Heatmap spec to use in this case. an empty `name` can be used if the user don't want the default parameter.

src/chart_types/heatmap/layout/config/config.ts
darnautov: Fixed in https://github.com/elastic/elastic-charts/pull/847/commits/b8081acaaaad1740022c85d4912c52d0ea80085e

src/chart_types/heatmap/layout/config/config.ts
darnautov: Fixed in https://github.com/elastic/elastic-charts/pull/847/commits/b8081acaaaad1740022c85d4912c52d0ea80085e

src/chart_types/heatmap/state/selectors/tooltip.ts
darnautov: Fixed in https://github.com/elastic/elastic-charts/pull/847/commits/b8081acaaaad1740022c85d4912c52d0ea80085e

src/chart_types/heatmap/state/selectors/tooltip.ts
darnautov: Fixed in https://github.com/elastic/elastic-charts/pull/847/commits/b8081acaaaad1740022c85d4912c52d0ea80085e

src/chart_types/heatmap/state/selectors/tooltip.ts
darnautov: Fixed in https://github.com/elastic/elastic-charts/pull/847/commits/b8081acaaaad1740022c85d4912c52d0ea80085e

markov00 (approved):
LGTM

PR #848, opened 2020-10-05
fix(interactions): recognise drag after 100ms and 4px

### LINKED ISSUE #748: Click interactions with moving mouse pointer are not recognised
**Describe the bug**
Looks like `onElementClick` event works only if pointer position during `mousedown` event is the same as during `mouseup` event. If mouse pointer is moved during click, then `onElementClick` interaction isn't firing. This seems to be different from native browser behaviour, where click events are still recognised if mouse pointer is moved. This difference makes clicks in elastic charts flaky when using a mouse, because start-end positions could be different when clicking while the mouse in motion. 

**To Reproduce**
1. Go to https://elastic.github.io/elastic-charts/?path=/story/interactions--sunburst-slice-clicks
2. Mousedown in one of the slices. 
3. Slightly change pointer position, leaving the mouse in the same slice
4. Mouse up
5. See `onElementClick` didn't fire

**Expected behaviour**
`OnElementClick` interactions fires

 - OS: Mac
 - Browser: Chrome 83
 - Elastic Charts: 27.0.0

**Additional context**

Noticed while working in a kibana dashboard.

**Kibana Cross Issues**
Add any [Kibana](https://github.com/elastic/kibana) related issues here.
- [ ] TODO.. didn't create issue, but it is relevant

**Checklist**

markov00:
## Summary

This fix reduces the situation where a drag event is fired instead of a click one.
The fix adds the following thresholds:
- a 100ms threshold before the mouse-down time and the current time to consider the movement for dragging purposes
- a 4-pixel delta between mouse-down and the current position  to consider the movement for dragging purposes

fix #748

### Checklist

### DIFF
--- a/src/state/reducers/interactions.ts
+++ b/src/state/reducers/interactions.ts
@@ -35,6 +35,9 @@ import { InteractionsState } from '../chart_state';
 import { getInitialPointerState } from '../utils';
 
 /** @internal */
+export const DRAG_DETECTION_TIMEOUT = 100;
+
+/** @internal */
 export function interactionsReducer(
   state: InteractionsState,
   action: LegendActions | MouseActions | KeyActions,
@@ -58,7 +61,7 @@ export function interactionsReducer(
         pointer: {
           ...state.pointer,
           // enable the dragging flag only if the time between the down action and the move action is > 100ms
-          dragging: !!(state.pointer.down && action.time - state.pointer.down.time >= 100 && delta),
+          dragging: !!(state.pointer.down && action.time - state.pointer.down.time >= DRAG_DETECTION_TIMEOUT && delta),
           current: {
             position: {
               ...action.position,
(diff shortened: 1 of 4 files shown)


nickofthyme (approved):
LGTM, tested locally.

src/state/reducers/interactions.ts
```diff
@@ -51,11 +55,13 @@ export function interactionsReducer(
       return state;
 
     case ON_POINTER_MOVE:
+      const delta = state.pointer.down && getDelta(state.pointer.down.position, action.position) > 4;
```
darnautov, line 58: I guess it'd be better to create a constant for the delta value to avoid magic numbers

darnautov (approved):
Tested and LGTM! Just one suggestion about the constant. 
It's worth slightly change the PR description by mentioning the delta also taken into account.

src/state/reducers/interactions.ts
markov00: done in https://github.com/elastic/elastic-charts/pull/848/commits/ae8f93510a31dc2f877a66883e195202757f0fa6

PR #849, opened 2020-10-06
fix(heatmap): adjust pageSize based available chart height

markov00:
## Summary

This fix the issue with wrongly computed pageSize: when computing the pageSize parameter we should consider the x axis vertical space as space not available for the rendering rows. 

To test:
- copy the PR playground file into a master branch, reduce the screen width and the top most chart will disappear
- on the PR instead, the chart will renderer correctly


src/chart_types/heatmap/state/selectors/get_grid_full_height.ts
```diff
@@ -0,0 +1,76 @@
...
+    const xAxisHeight = visible ? fontSize : 0;
+    const totalVerticalPadding = padding * 2;
+    let legendHeight = 0;
+    if (showLegend && (legendPosition === Position.Top || legendPosition === Position.Bottom)) {
+      legendHeight = maxLegendHeight ?? legendSize.height;
+    }
+    const verticalRemainingSpace = containerHeight - xAxisHeight - totalVerticalPadding - legendHeight;
+
+    // compute the grid cell height
+    const gridCellHeight = getGridCellHeight(yValues, grid, verticalRemainingSpace);
```
darnautov, line 63: nit: as `getGridCellHeight` is only used here maybe it's worth moving from the `viewmodel`

darnautov (approved):
Tested and LGTM

src/chart_types/heatmap/state/selectors/get_grid_full_height.ts
markov00: done https://github.com/elastic/elastic-charts/pull/849/commits/b1d675015243501b02e1818f6cebf17e79a8fd0f

src/chart_types/heatmap/state/selectors/get_grid_full_height.ts
```diff
@@ -0,0 +1,92 @@
...
+    { xAxisLabel: { padding, visible, fontSize }, grid, maxLegendHeight },
+    { yValues },
+  ): GridHeightParams => {
+    const xAxisHeight = visible ? fontSize : 0;
+    const totalVerticalPadding = padding * 2;
+    let legendHeight = 0;
+    if (showLegend && (legendPosition === Position.Top || legendPosition === Position.Bottom)) {
+      legendHeight = maxLegendHeight ?? legendSize.height;
+    }
+    const verticalRemainingSpace = containerHeight - xAxisHeight - totalVerticalPadding - legendHeight;
```
darnautov, line 60: we should probably check for `xAxisLabel.visible`

PR #850, opened 2020-10-06
fix(heatmap): x-axis labels overlapping for time series data

darnautov:
## Summary

Resolve the max number of ticks based on the formatted label and the chart width. 
A proper fix will require `ScaleContinuous` to take into account this value as a limit during creating nice rounded tick values.


markov00 (approved):
Tested locally, works fine!

PR #854, opened 2020-10-06
ci(release): use gh workflow to manually release and push

markov00:
## Summary

With the idea of dismiss Travis CI for release, I'm adding a new gh workflow that can be manually triggered and do both release and gh pages deploy.

I'm using this https://github.com/marketplace/actions/github-pages-action for deploy.
Few security tokens need to be added for `semantic-release` before merging on master

### DIFF
--- a/.github/workflows/auto_release.yaml
+++ b/.github/workflows/auto_release.yaml
@@ -1,0 +1,38 @@
+name: Publish release
+
+on:
+  workflow_dispatch:
+    branches:
+      - master
+
+jobs:
+  release_package:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: '10.x'
+      - uses: bahmutov/npm-install@HEAD
+      - name: Build check
+        run: yarn build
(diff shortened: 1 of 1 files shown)


.github/workflows/auto_release.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Testing
+        run: yarn test --ci
+      - name: Semantic-Release
+        run: yarn semantic-release
+      - name: Build static storybook
+        run: yarn storybook:build
+      - name: Deploy
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
nickofthyme, line 36: These secrets are built into github?

nickofthyme (approved):
LGTM 🎉

.github/workflows/auto_release.yaml
markov00: it should be https://github.com/marketplace/actions/github-pages-action#%EF%B8%8F-set-runners-access-token-github_token
> GitHub Actions runner automatically creates a GITHUB_TOKEN secret to use in your workflow. You can use the GITHUB_TOKEN to authenticate in a workflow run.

PR #855, opened 2020-10-06
fix: filter highlighted Y values

darnautov:
## Summary

Filter out highlighted data based on presented Y values


markov00 (approved):
LGTM

PR #867, opened 2020-10-13
fix(legend): legend sizes with ordinal data

### LINKED ISSUE #811: Bad legend sizing with long showLegendExtra values - Ordinal
**Describe the bug**
With the removal of default `showLegendExtra ` values on `Ordinal`, the chart now incorrectly sizes legends with long `showLegendExtra` values.

**To Reproduce**
1. Go to http://localhost:9001/?path=/story/axes--different-tooltip-formatter
2. Remove the `spacingBuffer` theme override
3. Notice the screenshot below.

**Additional context**
The `defaultExtra` values are both `null`

https://github.com/elastic/elastic-charts/blob/0e6353f4f49412e399005ba345a3f3dbdbce8114/src/chart_types/xy_chart/legend/legend.ts#L99-L102

When the formatted value is used to determine the labels here

https://github.com/elastic/elastic-charts/blob/0e6353f4f49412e399005ba345a3f3dbdbce8114/src/chart_types/xy_chart/state/selectors/get_legend_items_labels.ts#L28-L37

The `defaultExtra` is thus omitted and never used in the size calculations rendering a narrow legend size.

I suspect a fix would be to calculate the last value even for `Ordinal` scales and use a sample formatted value in the dimensions calculations, but still hide the default values on `Ordinal` scales.

**Expected behaviour**
The legend should find the correct size including the `showLegendExtra` content

**Screenshots**
Screen Recording 2020-09-08 at 04 35 PM

rshen91:
## Summary

Fixes #811 
Before this PR, if the legend has `LegendExtra` and no `spacingBuffer` specified in the `theme`, the labels in the legend will be truncated when hovering over the series. 

Oct-15-2020 07-53-03

After the fix, the legend is sized to correctly show the labels and values for the legend:
Oct-16-2020 15-40-10

### Checklist


src/chart_types/xy_chart/state/utils/utils.ts
```diff
@@ -166,7 +164,9 @@ function getLastValues(
 
       const { y0, y1, initialY0, initialY1 } = last;
       const seriesKey = getSeriesKey(series as XYChartSeriesIdentifier);
-
+      if (xDomain.scaleType === ScaleType.Ordinal) {
+        lastValues.set(seriesKey, { y0: initialY0, y1: initialY1 });
+      }
```
nickofthyme, line 169: I'm not sure this is needed. Wouldn't it just set the correct values without this?

src/chart_types/xy_chart/legend/legend.ts
```diff
@@ -56,16 +65,35 @@ function getBandedLegendItemLabel(name: string, yAccessor: BandedAccessorType, p
     : `${name}${postfixes.y0AccessorFormat}`;
 }
 
+function getDefaultExtra(
+  showLegendExtra: boolean,
+  xScaleType: ScaleType,
+  lastValue: LastValues,
+  formatter: (value: any, options?: TickFormatterOptions | undefined) => string,
+) {
+  return showLegendExtra || xScaleType === 'ordinal'
```
nickofthyme, line 74: You can't check the type here because the `legendSizeSelector` needs to know these values, to fix the issue. You could check the `xScaleType` before rendering in the `LegendItem` component.

src/chart_types/xy_chart/legend/legend.ts
```diff
@@ -110,10 +137,8 @@ export function computeLegend(
         isSeriesHidden,
         isItemHidden: hideInLegend,
         isToggleable: true,
-        defaultExtra: {
-          raw: lastValue && lastValue.y0 !== null ? lastValue.y0 : null,
-          formatted: lastValue && lastValue.y0 !== null ? formatter(lastValue.y0) : null,
-        },
```
nickofthyme, line 116: Careful with extracting this, the value here is `y0` not `y1` like the function returns

src/chart_types/xy_chart/legend/legend.ts
nickofthyme: Another way to fix this would be to add a `sizingLabel` prop to this and use that value in the `legendSizeSelector` instead of `formatted`.

src/chart_types/xy_chart/state/utils/utils.ts
rshen91: Agreed - removed in 24d6e44

src/chart_types/xy_chart/legend/legend.ts
rshen91: Commit 24d6e44 and discussed in zoom 🙇‍♀️

src/chart_types/xy_chart/legend/legend.ts
rshen91: Thanks! Commit 24d6e44 for changes

src/chart_types/xy_chart/legend/legend.test.ts
```diff
@@ -344,4 +353,31 @@ describe('Legends', () => {
...
+  });
+  it('should return legendSizingLabel null with showLegendExtra set to false', () => {
+    const formatter = (d: string | number) => `${Number(d).toFixed(2)} dogs`;
+    const lastValues = { y0: null, y1: 14 };
+    const showLegendExtra = false;
+
+    expect(getLegendExtra(showLegendExtra, ScaleType.Ordinal, formatter, 'y1', lastValues)).toMatchObject({
+      legendSizingLabel: null,
+    });
+  });
```
nickofthyme, line 382: I think you should assert all three parts of the `extraValue`  for all of these

nickofthyme (approved):
Awesome! LGTM, I made one comment on the tests to fix but other than that I approve.

src/chart_types/xy_chart/legend/legend.test.ts
rshen91: Thank you! Addressed in commit 6b57387

PR #877, opened 2020-10-26
feat(legend): allow legend text to be copyable

### LINKED ISSUE #710: Make legend in charts selectable
We currently use `user-select: none` on the legend in elastic charts. I've heard from users, that they would like to select the element items to copy and paste (e.g. there is a legend of ip addresses). I think it's worth discussing if the use-case of copying values from the legend wouldn't be worth keeping the legend selectable.

rshen91:
## Summary

Fixes #710 Text from the legend can be selected and copied.
Oct-26-2020 13-05-32

### DIFF
--- a/src/components/legend/_legend_item.scss
+++ b/src/components/legend/_legend_item.scss
@@ -5,7 +5,7 @@ $legendItemVerticalPadding: $echLegendRowGap / 2;
   display: flex;
   flex-wrap: nowrap;
   justify-content: space-between;
-  user-select: none;
+  user-select: all;
   align-items: center;
   width: 100%;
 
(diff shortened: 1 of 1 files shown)


src/components/legend/_legend_item.scss
```diff
@@ -5,7 +5,7 @@ $legendItemVerticalPadding: $echLegendRowGap / 2;
   display: flex;
   flex-wrap: nowrap;
   justify-content: space-between;
-  user-select: none;
+  user-select: auto;
```
nickofthyme, line 8: I think you could just remove this line entirety

```suggestion
```

nickofthyme (approved):

src/components/legend/_legend_item.scss
rshen91: Thank you - good call. Addressed in 08b5b79

PR #880, opened 2020-10-27
feat(legend): add keyboard navigation

### LINKED ISSUE #613: Keyboard navigation for legend (beta)
A small a11y issue for tracking. Please see #300 for full context on a11y efforts. 

Introduce keyboard navigation for the legend which is within a canvas 

**Additional context**
In speaking with @myasonik, he has a fantastic demo that can be found here: https://codepen.io/myasonik/pen/abOLdoE

rshen91:
## Summary

Closes #613 
This PR adds keyboard navigation to the legend. Screen reader tested using VoiceOver on local mac to read the label and extra if relevant.
Nov-06-2020 12-23-47

### Checklist


rshen91:
There are vrts failing for small pixel amounts is it ok to update the stories @markov00 @nickofthyme ?

nickofthyme:
@rshen91 Sorry, yes. I would say it's always appropriate to update the vrt screenshots, then on the PR we can just review what changed in screenshot. Sometimes minor changes somehow create small pixel differences.

markov00 (commented):
Thanks Rachel, with just few changes to the code it looks great!
I'd like to note a few issues:
1) the focus style doesn't looks like the focus style used in EUI. Here I'd like to ask @miukimiu her opinion about that and if we need background + border in the same style as EUI
2) I've noticed an issue with the keyboard: if you enable the color picker with the keyboard we are not passing the focus to the color picker element. Moreover, as shown on the gif, tabbing to the legend item label with the color picker open and clicking on that label cause a crash on the chart
Nov-11-2020 14-46-13
3) Can you prepare a story that "pass" the focus to the color picker? I'm not sure if this is something that happens automatically or we should implement that through our code
4) if possible instead of updating all these screenshot, it could be better to fix the difference in the padding of the updated DOM structure
5) @myasonik I'd like your opinion on the area-labels and the role applied to the legend.

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
```diff
@@ -166,6 +166,7 @@ class XYChartComponent extends React.Component<XYChartProps> {
           width,
           height,
         }}
+        tabIndex={0}
```
myasonik, line 169: Canvas doesn't seem to be intractable in any way via keyboard. Not sure why we need a tab stop here?

src/components/legend/color.tsx
```diff
@@ -37,10 +37,10 @@ export const Color = memo(
   forwardRef<HTMLDivElement, ColorProps>(({ color, isSeriesHidden = false, hasColorPicker, onClick }, ref) => {
     if (isSeriesHidden) {
       return (
-        <div className="echLegendItem__color" aria-label="series hidden" title="series hidden">
+        <button type="button" className="echLegendItem__color" aria-label="series hidden" title="series hidden">
```
myasonik, line 40: This button seems to have no action. Is that intended? Seems this text and the following button should be a single element so it'll read something like `Series hidden: ${seriesName}`

src/components/legend/color.tsx
```diff
@@ -49,16 +49,17 @@ export const Color = memo(
     });
 
     return (
-      <div
+      <button
+        type="button"
         onClick={hasColorPicker ? onClick : undefined}
```
myasonik, line 54: Similar to above, if there is no action, it shouldn't be a button.

src/components/legend/color.tsx
```diff
@@ -49,16 +49,17 @@ export const Color = memo(
...
         className={colorClasses}
-        aria-label="series color"
+        aria-label={hasColorPicker ? 'change series color' : 'series color'}
```
myasonik, line 56: We should also mention what the current color and buttons are assumed to be interactive so I think we can tweak this to something like:
```suggestion
        aria-label={`series color: ${colorName}`}
```

But putting this together with my comment above, we cannot leave an `aria-label` on a `div`. In that case, we should add the label to the `<svg>` of the icon.

src/components/legend/extra.tsx
```diff
@@ -30,7 +30,7 @@ export function renderExtra(extra: string | number, isSeriesHidden?: boolean) {
     'echLegendItem__extra--hidden': isSeriesHidden,
   });
   return (
-    <div className={extraClassNames} title={`${extra}`}>
+    <div className={extraClassNames} title={`${extra}`} aria-label={`${extra}`}>
```
myasonik, line 33: `aria-label` doesn't work on elements with no semantic value such as `div` or `span` (unless their role is also changed).

Also, this label just restates the inner text so there's really no need to add it. 

```suggestion
    <div className={extraClassNames} title={`${extra}`}>
```

src/components/legend/label.tsx
```diff
@@ -22,19 +22,37 @@ import React, { MouseEventHandler } from 'react';
...
 /**
  * Label component used to display text in legend item
  * @internal
  */
-export function Label({ label, onClick }: LabelProps) {
+export function Label({ label, extra, onClick }: LabelProps) {
   const labelClassNames = classNames('echLegendItem__label', {
     'echLegendItem__label--clickable': Boolean(onClick),
   });
+  const labelWithExtra = getExtra(label, extra);
```
myasonik, line 36: I dunno if we need to add this at all... The `extra` content seems to always be the next element so screen readers will read it next and it seems like it's never visually hidden either.

src/components/legend/legend.tsx
```diff
@@ -107,7 +107,7 @@ function LegendComponent(props: LegendStateProps & LegendDispatchProps) {
   return (
     <div className={legendClasses}>
       <div style={legendContainerStyle} className="echLegendListContainer">
-        <ul style={legendListStyle} className="echLegendList">
+        <ul style={legendListStyle} className="echLegendList" role="listbox">
```
myasonik, line 110: Why are we adding `role=listbox` here? 

`listbox` is the role given to `<select>` elements and has a lot of custom controls required to implement a custom one.

stories/legend/11_legend_actions.tsx
```diff
@@ -113,7 +114,7 @@ const getAction = (hideActions: boolean, anchorPosition: PopoverAnchorPosition):
       onClick={() => setPopoverOpen(!popoverOpen)}
     >
       <EuiIcon size="s" type="pencil" />
-    </div>
+    </button>
   );
 
   return (
```
myasonik, line 120: Can't leave a comment lower than this, but [`ownFocus`](https://elastic.github.io/eui/#/layout/popover#removing-the-focus-trap) needs to be added to this Popover otherwise there is no way for a keyboard user to realistically navigate into it.

myasonik (changes requested):

stories/legend/11_legend_actions.tsx
myasonik: This will fix the first half of 2 and 3 in the issues Marco left above

myasonik: EUI 30.3.0 changes the default of `ownFocus` to `true` so if y'all are good at staying up to date then this change might not be necessary.

src/components/legend/color.tsx
nickofthyme: Yeah this is just an icon here when `isSeriesHidden`, the legend label is the clickable element.

src/components/legend/color.tsx
nickofthyme: @myasonik this is actionable only when `hasColorPicker` is `true`. So you are saying just separate this for action and inactionable elements?

src/components/legend/color.tsx
nickofthyme: Yeah I think we just split this and have the actionable element say ```aria-label={`series color: ${colorName}`}``` and the actionable say ```aria-label={`change series color: ${colorName}`}```

src/components/legend/color.tsx
```diff
@@ -49,16 +49,17 @@ export const Color = memo(
     });
 
     return (
-      <div
+      <button
+        type="button"
```
nickofthyme, line 53: Also this should have a `tabIndex={0}` when `hasColorPicker` is false

```suggestion
        type="button"
        tabIndex={hasColorPicker ? undefined : 0}
        
```

nickofthyme (commented):
Looking good, just a few comments on top of @myasonik's

src/components/legend/color.tsx
myasonik: Yeah, exactly

src/components/legend/color.tsx
myasonik: > Also this should have a `tabIndex={0}` when `hasColorPicker` is false

Why's that? Won't it not have any action when `hasColorPicker == false`? If so, then it shouldn't be a `<button>` and it shouldn't be a tab stop (i.e., `tabIndex={0}`). If it does still have an action, then it should stay a `<button>` which doesn't need a `tabIndex` defined.

src/components/legend/color.tsx
```diff
@@ -34,32 +33,34 @@ interface ColorProps {
...
-  forwardRef<HTMLDivElement, ColorProps>(({ color, isSeriesHidden = false, hasColorPicker, onClick }, ref) => {
+  forwardRef<HTMLButtonElement, ColorProps>(({ color, isSeriesHidden = false, hasColorPicker, onClick }, ref) => {
     if (isSeriesHidden) {
       return (
-        <button type="button" className="echLegendItem__color" aria-label="series hidden" title="series hidden">
+        <div className="echLegendItem__color" title="series hidden">
           {/* changing the default viewBox for the eyeClosed icon to keep the same dimensions */}
-          <Icon type="eyeClosed" viewBox="-3 -3 22 22" />
-        </button>
+          <Icon type="eyeClosed" viewBox="-3 -3 22 22" aria-label="series color is hidden" />
```
myasonik, line 41: 🤔 It's not the "series color" that's hidden, it's the series itself, right?

I think this might be more clear? (But I'm also no wordsmith so would love to hear from others)

```suggestion
          <Icon type="eyeClosed" viewBox="-3 -3 22 22" aria-label="series ${seriesName} is hidden" />
```

Also, how does elastic-charts handle internationalization? We need to make sure these strings are i18n-able for Kibana.

src/components/legend/color.tsx
```diff
@@ -34,32 +33,34 @@ interface ColorProps {
...
+    if (hasColorPicker) {
+      return (
+        <button
+          type="button"
+          onClick={onClick}
+          className="echLegendItem__color--changable"
+          title="change series color"
+          ref={ref}
+        >
+          <Icon type="dot" color={color} aria-label="change series color" />
```
myasonik, line 55: Are names for these colors defined anywhere? Would be good to put that into the label. Worst case, even putting the hex code would be better than nothing... 

```suggestion
          <Icon type="dot" color={color} aria-label="change series color, currently ${colorName}" />
```

src/components/legend/color.tsx
```diff
@@ -34,32 +33,34 @@ interface ColorProps {
...
-        className={colorClasses}
-        aria-label={hasColorPicker ? 'change series color' : 'series color'}
-        title={hasColorPicker ? 'change series color' : 'series color'}
-      >
-        <div ref={ref}>
-          <Icon type="dot" color={color} />
-        </div>
-      </button>
+      <div className="echLegendItem__color" title="series color">
+        <Icon type="dot" color={color} aria-label={hasColorPicker ? 'change series color' : 'series color'} />
```
myasonik, line 62: Can remove the `aria-label` conditional, and would be good to add a color name here

```suggestion
        <Icon type="dot" color={color} aria-label={`series color: ${colorName}`} />
```

stories/legend/11_legend_actions.tsx
```diff
@@ -126,6 +126,7 @@ const getAction = (hideActions: boolean, anchorPosition: PopoverAnchorPosition):
       panelPaddingSize="none"
       withTitle
       anchorPosition={anchorPosition}
+      ownFocus
```
myasonik, line 129: One more popover below to add this too - line 142

rshen91: ah! Thank you

src/components/legend/color.tsx
markov00: Not at the moment, we have just a very limited set of strings at the moment, but one next step will be to add i18n support

src/components/legend/color.tsx
markov00: We don't have a naming mapping for the colors as these are actually passed to the charts by the consumer. A conversion from hex to actual name is out of the scope of the chart lib at the moment. Something like that can be built and shared across all kibana, a approximation mapping like the one provided by https://chir.ag/projects/name-that-color/#6195ED can be an interesting solution and can go into the same direction of the i18n

stories/legend/11_legend_actions.tsx
rshen91: Updated in f7e2f27

src/components/legend/color.tsx
rshen91: I think f7e2f27 is working as expected with your feedback, let me know what you think (and if I'm misunderstanding) when you have a chance. Thank you!

src/components/legend/_legend.scss
```diff
@@ -38,10 +37,17 @@
...
+
   .echLegendListContainer {
     @include euiYScrollWithShadows;
     width: 100%;
     overflow-y: auto;
-    overflow-x: hidden;
+  }
+
+  &:focus {
+    box-shadow: inset 0 0 0 $euiSizeXS / 4 $euiFocusRingColor;
```
miukimiu, line 51: Thanks, @rshen91 for introducing this keyboard navigation. It's a very good improvement! 🎉 

I tested this PR locally and I would suggest use our EUI variables and mixins to create the focus styles:

```suggestion
   @include euiFocusRing;
   background-color: $euiFocusBackgroundColor;
   border-radius: $euiBorderRadius / 2;
```

It will look like this:
Screenshot 2020-11-12 at 16 10 41

Screenshot 2020-11-12 at 16 18 06

I tested these for different examples like the **Legend - Color Picker** and it doesn't look good. It seems that the legends have different styles so the `euiFocusRing` doesn't surround the legend properly. Is it possible to fix this?

Screenshot 2020-11-12 at 16 11 41

miukimiu (changes requested):

stories/legend/11_legend_actions.tsx
nickofthyme, line 129: 👍

src/components/legend/_legend.scss
nickofthyme: @miukimiu Would the hover state change at all?

src/components/legend/color.tsx
myasonik: Then hex code is fine for now

src/components/legend/_legend.scss
rshen91: Thank you so much for the help- please see 5b842fe when you have a chance

src/components/legend/color.tsx
rshen91: Please see bc14508

src/components/legend/label.tsx
rshen91: When I tested with VoiceOver it wasn't saying the extra. Let me know if I should test differently - thank you

myasonik: It won't read it as part of the button when tabbing through the layout, but it's available when reading the list item. CTRL+OPTION+arrow keys lets you navigate forward and backward through content.

myasonik: Running through it with VO, one thing I found confusing is knowing what to expect this button to do. Instead of adding the extra content to the tail end of the label, what do you think about adding some instructions like "Activate to hide series in graph." (And swapping it to "Activate to show series in graph." when it's hidden.)

rshen91: Thank you for the clarification! Please see for edits in commit dc373e8

src/components/legend/label.tsx
```diff
@@ -22,19 +22,26 @@ import React, { MouseEventHandler } from 'react';
...
     'echLegendItem__label--clickable': Boolean(onClick),
   });
   return (
-    <div className={labelClassNames} title={label} onClick={onClick}>
+    <button
+      type="button"
+      className={labelClassNames}
+      title={label}
+      onClick={onClick}
+      aria-label={isSeriesHidden ? `Activate to show series in graph` : `Activate to hide series in graph`}
```
myasonik, line 42: `aria-label` replaces the contents of the button so we need to restate the `label` in here as well.

```suggestion
      aria-label={isSeriesHidden ? `${label}; Activate to show series in graph` : `${label}; Activate to hide series in graph`}
```

myasonik (approved):
Just the two `aria-label` changes left from me and this looks ready to ship on the a11y side! 🎉

src/components/legend/label.tsx
rshen91: Changed in dc373e8! Thanks!

miukimiu (approved):
Tested in Chrome, Edge, and Firefox.  I'm suggesting two small changes but all the rest looks good. 🎉

src/components/legend/_legend.scss
miukimiu: @nickofthyme I don't think we need to change the hover state for now.

src/components/legend/color.tsx
rshen91: Thank you for catching this - edited in 07eadda

src/components/legend/_legend.scss
```diff
@@ -42,6 +42,12 @@
...
     width: 100%;
     overflow-y: auto;
-    overflow-x: hidden;
+
+    :focus {
+      @include euiFocusRing;
+      background-color: $euiFocusBackgroundColor;
+      border-radius: $euiBorderRadius / 2;
+      margin-right: $euiSizeXS;
+    }
```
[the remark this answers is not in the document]
rshen91: 🙇‍♀️ thank you f751a99

src/components/legend/_legend_item.scss
```diff
@@ -31,6 +31,7 @@ $legendItemVerticalPadding: $echLegendRowGap / 2;
 
   &__color {
     margin-right: $euiSizeXS;
+    margin-left: $euiSizeXS;
```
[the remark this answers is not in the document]
rshen91: Thank you for the help! Commit f751a99 for changes

src/components/legend/_legend.scss
```diff
@@ -2,7 +2,7 @@
   &--top,
   &--bottom {
     .echLegendList {
-      display: grid;
+      display: flex;
```
nickofthyme, line 5: Why are we making this change? The display grid is used to spread the columns nicely that is not achievable with `flex`

nickofthyme (changes requested):

src/components/legend/_legend.scss
rshen91: thanks - edited in ef51a44

src/components/legend/color.tsx
markov00: @rshen91 can you fix the wording of this as suggested by Michael?

markov00 (changes requested):
Few points to address:
- the updated screenshot `integration/tests/__image_snapshots__/annotations-stories-test-ts-annotations-stories-rotation-rotation-negative-90-1-snap.png` is actually wrong: the Y axis is reversed, it should appear with the 4 on the top: https://elastic.github.io/elastic-charts/?path=/story/annotations-lines--single-bar-histogram&knob-debug=&knob-chartRotation=-90
- there is one missing thing that we should probably do internally: we should return the focus to the button that opened a color picker/modal on close. @myasonik is there any good practices for that that we can use?

src/components/legend/color.tsx
rshen91: Yes thank you good catch - please see ecc7797

rshen91:
Updated the annotation story in ecc7797 🤞

myasonik:
> there is one missing thing that we should probably do internally: we should return the focus to the button that opened a color picker/modal on close. @myasonik is there any good practices for that that we can use?

EUI usually tries to handle this for you but I think it's breaking because we chnage the button icon depending on what was set in the modal. Anyway, EUI expects that this doesn't work 100% of the time so there is an `closePopover` callback available on `EuiPopover` which you can use to manually call `.focus()` on a DOM element.

src/components/legend/_legend.scss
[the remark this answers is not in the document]
nickofthyme: @miukimiu Is there a way we could add a transparent border to this so there is no jump in height and width when the border is added?

Screen Recording 2020-11-17 at 12 53 PM

integration/page_objects/common.ts
```diff
@@ -347,6 +347,26 @@ class CommonPage {
...
+        animation-delay: -0.0001s !important;
+        animation-duration: 0s !important;
+        animation-play-state: paused !important;
+        caret-color: transparent !important;
+      };`;
+      const storyRoot = document.getElementById('#story-root');
+      if (storyRoot) {
+        storyRoot.appendChild(style);
+      }
+    });
```
nickofthyme, line 369: Rather than adding all the css here you could just add a class to the `#story-root` element. No need to fuss with creating a style tag and appending it to the dom.

```
const storyRoot = document.getElementById('story-root');
if (storyRoot) {
  storyRoot.classList.add('disable-animations')
}
```

> Sidenote: `getElementById` implies the type of selector so using `#` won't work

.storybook/style.scss
```diff
@@ -25,6 +25,18 @@ html {
...
+  // *::after,
+  // *::before,
+  // .echLegendItem {
+  //   transition-delay: 0s !important;
+  //   transition-duration: 0s !important;
+  //   animation-delay: -0.0001s !important;
+  //   animation-duration: 0s !important;
+  //   animation-play-state: paused !important;
+  //   caret-color: transparent !important;
+  // }
```
nickofthyme, line 39: By adding the class you can then do...

```scss
#story-root {
  .disable-animations {
    *,
    *::after,
    *::before{
      transition-delay: 0s !important;
      transition-duration: 0s !important;
      animation-delay: -0.0001s !important;
      animation-duration: 0s !important;
      animation-play-state: paused !important;
      caret-color: transparent !important;
    }
  }
}
```

rshen91:
jenkins test this please

rshen91:
@nickofthyme and @markov00 in 3631426 I added a workaround for the vrts that were failing in legend_stories and area_stories with disableAnimations(). Can you let me know your thoughts?

integration/page_objects/common.ts
```diff
@@ -460,6 +440,13 @@ class CommonPage {
     });
   }
 
+  async disableAnimations(url: string) {
+    await this.loadElementFromURL(url, '#story-root');
```
nickofthyme, line 444: This is a good thought. It must replace the `#story-root` after each navigation.

But as long as you use this in the `action` function you don't need to call `loadElementFromURL`.

```suggestion
```

integration/tests/area_stories.test.ts
```diff
@@ -64,27 +64,37 @@ describe('Area series stories', () => {
...
       );
     });
 
-    it('snows only positive values when hiding negative one', async () => {
-      const action = async () => await page.click('.echLegendItem:nth-child(2) .echLegendItem__label');
+    it('shows only positive values when hiding negative one', async () => {
+      const action = async () => {
+        await common.disableAnimations(
+          'http://localhost:9001/?path=/story/area-chart--with-negative-and-positive&knob-Y scale=log',
+        );
```
nickofthyme, line 82: No need to repeat the URL navigation again. The `disableAnimations` method can just be used as a helper functions in the `action`.

```suggestion
        await common.disableAnimations();
```

integration/tests/area_stories.test.ts
```diff
@@ -64,27 +64,37 @@ describe('Area series stories', () => {
...
       );
     });
 
-    it('snows only negative values when hiding positive one', async () => {
-      const action = async () => await page.click('.echLegendItem:nth-child(1) .echLegendItem__label');
+    it('shows only negative values when hiding positive one', async () => {
+      const action = async () => {
+        await common.disableAnimations(
+          'http://localhost:9001/?path=/story/area-chart--with-negative-and-positive&knob-Y scale=log',
+        );
```
nickofthyme, line 95: Same as above. Lmk if this don't work.

```suggestion
        await common.disableAnimations();
```

integration/tests/legend_stories.test.ts
```diff
@@ -93,6 +93,9 @@ describe('Legend stories', () => {
     });
   });
   describe('keyboard navigation', () => {
+    beforeEach(async () => {
+      await common.disableAnimations('http://localhost:9001/?path=/story/legend--right');
+    });
```
nickofthyme, line 98: Is this needed?

integration/tests/area_stories.test.ts
rshen91: It works! Thank you! Please see 5bdb611

integration/tests/legend_stories.test.ts
rshen91: Good call - totally not needed - removed in 5bdb611

integration/tests/area_stories.test.ts
rshen91: 🎉 Good point - please see 5bdb611

integration/page_objects/common.ts
```diff
@@ -383,6 +440,13 @@ class CommonPage {
     });
   }
 
+  async disableAnimations() {
+    // await this.loadElementFromURL(url, '#story-root');
```
markov00, line 444: remove this line

integration/tests/area_stories.test.ts
```diff
@@ -64,27 +64,35 @@ describe('Area series stories', () => {
...
         'http://localhost:9001/?path=/story/area-chart--with-negative-and-positive&knob-Y scale=log',
       );
     });
 
-    it('snows only positive values when hiding negative one', async () => {
-      const action = async () => await page.click('.echLegendItem:nth-child(2) .echLegendItem__label');
+    it('shows only positive values when hiding negative one', async () => {
+      const action = async () => {
+        await common.disableAnimations();
+        // 'http://localhost:9001/?path=/story/area-chart--with-negative-and-positive&knob-Y scale=log',
```
markov00, line 81: remove these commented URLs

markov00 (approved):
LGTM, tested locally and works perfectly.

integration/tests/area_stories.test.ts
rshen91: removed in 55f1628 thanks!

integration/page_objects/common.ts
rshen91: removed in 55f1628 thanks!

.storybook/style_test.scss
```diff
@@ -0,0 +1,76 @@
+@import '../node_modules/@elastic/eui/src/theme_light.scss';
```
nickofthyme, line 1: Doesn't look like you are using this file

integration/page_objects/common.ts
```diff
@@ -121,7 +131,7 @@ type ScreenshotElementAtUrlOptions = ScreenshotDOMElementOptions & {
   /**
    * any desired action to be performed after loading url, prior to screenshot
    */
-  action?: () => void | Promise<void>;
+  action?: () => void | Promise<void> | Iterable<unknown>;
```
nickofthyme, line 134: What is this for?

integration/page_objects/common.ts
```diff
@@ -17,6 +17,9 @@
  * under the License.
  */
 
+// eslint-disable-next-line eslint-comments/disable-enable-pair
+/* eslint-disable no-await-in-loop */
```
nickofthyme, line 21: Can you just disable the line not the whole file?

integration/page_objects/common.ts
```diff
@@ -261,6 +271,27 @@ class CommonPage {
...
+  /**
+   * Press keyboard keys
+   * @param count
+   * @param actionLabel
+   */
+  async pressKey(actionLabel: string, count: number) {
+    if (actionLabel === 'tab') {
+      let i = 0;
+      while (i < count) {
+        await page.keyboard.press('Tab');
```
nickofthyme, line 283:
```suggestion
        // eslint-disable-next-line no-await-in-loop
        await page.keyboard.press('Tab');
```

integration/page_objects/common.ts
```diff
@@ -261,6 +271,27 @@ class CommonPage {
...
+        i++;
+      }
+    } else if (actionLabel === 'enter') {
+      let i = 0;
+      while (i < count) {
+        await page.keyboard.press('Enter');
```
nickofthyme, line 289:
```suggestion
        // eslint-disable-next-line no-await-in-loop
        await page.keyboard.press('Enter');
```

integration/page_objects/common.ts
```diff
@@ -341,6 +372,32 @@ class CommonPage {
...
+   * @param options
+   */
+  async expectChartWithKeyboardEventsAtUrlToMatchScreenshot(
+    url: string,
+    keyboardEvents: KeyboardKeys,
+    options?: Omit<ScreenshotElementAtUrlOptions, 'action'>,
+  ) {
+    const action = async () => {
+      await this.disableAnimations();
+      await this.clickMouseRelativeToDOMElement({ top: 242, left: 910 }, this.chartSelector);
```
nickofthyme, line 388: This should be a generic method, why are there specific `top` and `left` values? Is the purpose of this to just click the element to the tap?

nickofthyme (changes requested):
The final result is SUPER clean and smooth, way to go!

I have a few nits about the vrt helper methods. Also, I wonder if we could move the mouse before the tab to show the focus state only and not the focused and hover state.

image

integration/page_objects/common.ts
rshen91: Yep! Just to make sure that storybook tabbing doesn't go down all the individual stories in the pane before reaching the iframe

integration/page_objects/common.ts
rshen91: Yep this makes sense - made changes in 02a06fe

nickofthyme (approved):

integration/page_objects/common.ts
rshen91: changed in a5acd93

.storybook/style_test.scss
rshen91: changed in a5acd93

PR #892, opened 2020-11-05
docs: add randomness toggle on fit stacked series story

nickofthyme:
## Summary

Add randomness toggle on fit stacked series story per chat with @wylieconlon. Should be disabled by default to easily see the difference.

Screen Recording 2020-11-05 at 02 01 PM

### Checklist

### DIFF
--- a/stories/mixed/6_fitting_stacked.tsx
+++ b/stories/mixed/6_fitting_stacked.tsx
@@ -26,7 +26,7 @@ import { getRandomNumberGenerator } from '../../src/mocks/utils';
 import { SB_KNOBS_PANEL } from '../utils/storybook';
 
 export const Example = () => {
-  const randomizeBoundingData = boolean('randomize bounding data', false);
+  const randomizeBoundingData = process.env.RNG_SEED !== null ? false : boolean('randomize bounding data', false);
   const dataTypes: Record<string, Array<{ x: number | string; y: number | null }>> = {
     isolated: [
       { x: 0, y: 3 },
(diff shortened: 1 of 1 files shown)


stories/mixed/6_fitting_stacked.tsx
```diff
@@ -158,7 +159,7 @@ export const Example = () => {
   const parsedEndValue: number | 'nearest' = Number.isNaN(Number(endValue)) ? 'nearest' : Number(endValue);
   const value = number('Explicit value (using Fit.Explicit)', 5);
   const xScaleType = dataKey === 'ordinal' ? ScaleType.Ordinal : ScaleType.Linear;
-  const rng = getRandomNumberGenerator();
+  const rng = getRandomNumberGenerator(randomizeBoundingData ? undefined : '__seed__');
```
markov00, line 162: This, running locally doesn't have an effect because, by default, the `getRandomNumberGenerator` picks up the RNG_SEED passed when running the storybook.
It has only an effect on the deployed version because we don't build Storybook with that SEED.

markov00 (commented):
I think we should use the RND_SEED in the Stroybook build process to have the same local and deployed version of storybook

stories/mixed/6_fitting_stacked.tsx
nickofthyme: Good point. I conditionally show the toggle if `RNG_SEED` is defined. I think it would be confusing for that `boolean` to have no effect in such case.

nickofthyme: 9fc5157

markov00 (approved):
LGTM

PR #896, opened 2020-11-06
fix(area_charts): correctly represent baseline with negative data points

### LINKED ISSUE #893: Negative areas doesn't respect the zero baseline
**Describe the bug**
When rendering data with all negative values, the area chart doesn't respect the zero baseline and always render the chart with a baseline as the min domain value.

**To Reproduce**
1. Go to https://codesandbox.io/s/gracious-ride-9f6wn?file=/src/App.tsx
2. the area covers the from the min domain to the max domain, and doesn't consider 0 as baseline

**Expected behaviour**
For area and bar, if not configured otherwise(currently not possible), the baseline should be always zero.

**Screenshots**
Screenshot 2020-11-06 at 10 30 32

 - Elastic Charts: 24

**Additional context**
n/a

**Errors in browser console**
n/a

markov00:
## Summary

This PR fixes the rendering issue when using negative values in area charts. The correct approach
is to draw the area with a zero baseline (or dummy one baseline for log y scale). This correctly
represent the sense of area for a single data point.

Screenshot 2020-11-06 at 19 28 55
Screenshot 2020-11-06 at 19 28 47

fix #893

## Note for reviewers 
It's easier to check the changes if you review them by commit:
- the first commit is an initial code cleanup to consider the correct baseline for negative values
- the second one moves the renderArea, renderPoints, renderLine etc methods into their respective files to reduce the size of the utils.ts file and increase readability
- the third commit fix and refactor the areas of uncertainty for lines/area/points when using log scales

Another thing to notice: the change related to the log scale also changed the way we render areas in the [-1,1] range. I've adopted a different strategy that removes areas/points from showing in that range (possibly adjustable via theme/configuration in a future PR). This adjustment will require the implementation of the #63 request.

### Checklist


src/chart_types/xy_chart/rendering/utils.ts
```diff
@@ -32,19 +35,20 @@ export interface MarkSizeOptions {
...
+  if (value !== null || !returnFilled) {
+    return value;
   }
-
-  if (filled && filled.y1 !== undefined) {
-    return filled.y1;
-  }
-
-  return null;
+  return (datum.filled && datum.filled[valueName]) ?? null;
```
nickofthyme, line 51: nit https://stackoverflow.com/a/58780897/6943587

```ts
return  datum.filled?.[valueName] ?? null;
```

src/mocks/series/utils.ts
```diff
@@ -44,4 +44,7 @@ export const getXValueData = (data: DataSeriesDatum[]): (number | string)[] => d
  * Returns value of `y1` or `filled.y1` or null
  * @internal
  */
-export const getYResolvedData = (data: DataSeriesDatum[]): (number | null)[] => data.map(getYValue);
+export const getYResolvedData = (data: DataSeriesDatum[]): (number | null)[] =>
+  data.map((d) => {
+    return getYDatumValue(d);
+  });
```
nickofthyme, line 50:
```ts
data.map(getYDatumValue);
```

src/scales/scale_continuous.ts
```diff
@@ -62,38 +62,39 @@ const SCALES = {
 export function limitLogScaleDomain(domain: any[]) {
   if (domain[0] === 0) {
     if (domain[1] > 0) {
-      return [1, domain[1]];
+      return [LOG_MIN_ABS_DOMAIN, domain[1]];
```
nickofthyme, line 65: ❤️

nickofthyme (approved):
Code changes LGTM. Tested locally and everything functions as expected.

Why does the band disappear when using log scale? 👇 

Screen Recording 2020-11-10 at 10 18 43 AM

Also, I noticed that for positive and negative values the log scale just shows the positive values. Is there a way to logically/correctly show positive and negative values when using log scale? Like the symmetric log scale, @monfera was talking about?

src/chart_types/xy_chart/rendering/rendering.ts
```diff
@@ -711,7 +710,16 @@ export function renderArea(
       return yScale.isInverted ? yScale.range[1] : yScale.range[0];
     })
     .y0(({ y0 }) => {
-      return y0 === null || (isLogScale && y0 <= 0) ? yScale.range[0] : yScale.scaleOrThrow(y0);
+      if (y0 === null) {
+        if (isLogScale) {
+          return yScale.scaleOrThrow(1);
+        }
+        return yScale.scaleOrThrow(0);
```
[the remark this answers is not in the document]
nickofthyme: Also, the `1` should be `LOG_MIN_ABS_DOMAIN` right?

src/chart_types/xy_chart/rendering/line.ts
```diff
@@ -50,22 +49,14 @@ export function renderLine(
...
-      }
-
-      // this should never happen thanks to the defined function
-      return yScale.isInverted ? yScale.range[1] : yScale.range[0];
-    })
+    .y(y1Fn)
     .defined((datum) => {
-      const yValue = getYValue(datum);
-      return yValue !== null && !(isLogScale && yValue <= 0) && xScale.isValueInDomain(datum.x);
+      return definedFn(datum);
```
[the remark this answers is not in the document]
markov00: mm, typescript is not 100% happy with that, because the defined function accept a function with a single argument where definedFn has multiple arguments

src/chart_types/xy_chart/rendering/utils.ts
markov00: oh thanks I was just looking for that ❤️

src/mocks/series/utils.ts
markov00: same comment as for the `definedFn`, getYDatumValue has a different signature then a map function

src/chart_types/xy_chart/rendering/rendering.ts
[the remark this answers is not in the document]
markov00: already solved on https://github.com/elastic/elastic-charts/pull/896/commits/eca41362f88c9e6a81804e88b962c864311ca3bb

rshen91 (approved):
Sorry about the delay in reviewing! These changes look great to me - I appreciate the additional stories and clarity in reducing file sizes.

markov00:
> Why does the band disappear when using log scale? 👇
> 
> Screen Recording 2020-11-10 at 10 18 43 AM

This happens because the lowest point is below the threshold of 1 (currently 0.9). In this case, the band is not defined and is removed from the path

PR #899, opened 2020-11-10
feat: allow clearing colors from memory

### LINKED ISSUE #898: Allow option to clear color picker colors
When using a custom color picker in the legend. The `onChange` callback does _not_ allow for `null` value or any other way to clear the custom color selection.

The `onChange` callback takes a `null` value to clear the selected color value.

Rerender entire chart to update color selection

nickofthyme:
## Summary

Fix #898

Adds option to pass `null` value to `onChange` callback from custom legend color picker.

Screen Recording 2020-11-10 at 01 52 29 PM

### Checklist


src/state/chart_state.ts
```diff
@@ -399,10 +399,18 @@ export const chartStoreReducer = (chartId: string) => {
...
+              action.color !== null
+                ? {
+                    ...state.colors.persisted,
+                    [action.key]: action.color,
+                  }
+                : (() => {
+                    const colors = { ...state.colors.persisted };
+                    delete colors[action.key];
+
+                    return colors;
```
markov00, line 412: Instead of using delete you can also use the following:
```suggestion
                    const {[action.key] : removed, ...others} = state.colors.persisted;
                    return others;
```

src/components/legend/legend_item.tsx
```diff
@@ -183,17 +184,21 @@ export class LegendListItem extends Component<LegendItemProps, LegendItemState>
     const { seriesIdentifier, color } = item;
 
     const handleClose = () => {
-      setPersistedColorAction(seriesIdentifier.key, color);
+      setPersistedColorAction(seriesIdentifier.key, this.shouldClearPersistedColor ? null : color);
```
markov00, line 187: Just to understand the reason of adding the `shouldClearPersistedColor`: is that because the `handleClose` doesn't receive a color and the color is configured only through `handleChange`?
If the action order is:
- user click on "clear color" in the color picker
- `handleChange(null)` is called
- `setTemporaryColorAction(null)` is called
- the component gets rerendered with the default series color (so `color` in the `item` object is the "default color)
- user close the color picker
- `handleClose()`
- `setPersistedColorAction(color)` if here we pass the that `default` color, are we doing something wrong? can't we rely on the `item.color` instead of the `shouldClearPersistedColor`?

src/specs/settings.tsx
```diff
@@ -248,7 +248,7 @@ export interface LegendColorPickerProps {
   /**
    * Callback to update temporary color state
    */
-  onChange: (color: Color) => void;
+  onChange: (color: Color | null) => void;
```
markov00, line 251: As we are changing the API, this should probably be marked as breaking change and requires an API report update

nickofthyme: Do you think this creates a breaking change? This `onChange` is a callback we provide not that they provide.

```ts
// we control callback function signature
const color = 'red' // as string now
onChange(color); // string is string | null, no type error
```

The issue that we had with the display value type change was that we did not control the function, just the type. So that would be something like....
```ts
// we control callback function type
const onChange = (color) => {
  // color type could be breaking
}
```

src/components/legend/legend_item.tsx
nickofthyme: Yes exactly right. The color gets updated from outside of the component based on the temporary color action. So if I don't add this flag, and the steps above occur, the default color will be set in the persisted colors.

src/state/chart_state.ts
nickofthyme: 8f9a229

src/specs/settings.tsx
markov00: ok fine, no breaking change in this case

markov00 (approved):
LGTM after fixing the VRT where the color picker changed legitimately its height

PR #901, opened 2020-11-12
fix: specs swaps correctly reflected in state

### LINKED ISSUE #868: [xyChart] When switching the seriesProps, the chart doesn't render first series
**Describe the bug**
When I want to programatically reorder `yAccessors` in xy_chart, the first series is not displayed. 
I’ve simplified the example here in codesandbox – toggle the 'click' button - the series should switch and display in order from [1,2,3] to [2,1,3], but only the second and third series is visible. 

https://codesandbox.io/s/charts-interval-bug-l8th9?file=/src/App.tsx

 - current Kibana version (master)

markov00:
## Summary

This PR fixes the wrong state caused by swapping specs props between existing components.

The root cause is the `specComponentFactory`. 
The issue appears in the following situation:
- we have two components with respective ids `A` and `B`
- this got `parsed` and stored in our state
- the consumer switch the ids of the two components: the first component (previously with id `A`) becomes the one with id `B` and the second component (previously with id `B`) becomes `A`.

Because we are processing these specs sequentially following the React life cycle, the first component calls the following part of the `specComponentFactory`
```
if (prevId && prevId !== props.id) {
    removeSpec(prevId);
}
upsertSpec(SpecInstance);
```
this removes the `prevId` spec (the `A` for the first component) and then add the `spec B` to the store (1).
Then it's time for the second component. The id of the component has changed so it follows the same path as the previous one: removes the `spec B` (its saved previous id) and then upsert the spec with id `A`, but the removed spec `B` is not the old one, but a new one added in the step (1), the new one provided by the first component.

The first time we send an`upsertSpec` action to the state, the reducer cleans up the specs array. Because of this, it is not needed to call first the `removeSpec` with the `prevId`, if that is different from the current one. We still clean and re-add all the specs on upserting.

fix #868

### Checklist


src/state/spec_factory.test.tsx
```diff
@@ -0,0 +1,94 @@
+/*
```
nickofthyme, line 1: ❤️

nickofthyme (approved):
Changes LGTM, verified the issue is resolved locally.

Screen Recording 2020-11-16 at 08 48 26 AM

rshen91 (approved):
Looks great! I really like the additional tests - they make it really clear

PR #903, opened 2020-11-12
fix(area_charts): correctly represent baseline with negative data points [23.x]

markov00:
Backports the following commits to 23.x:

fix(area_charts): correctly represent baseline with negative data points (#896)


nickofthyme (approved):

markov00:
jenkins test this please

PR #911, opened 2020-11-20
fix: near and far alignments for orthogonal rotations

### LINKED ISSUE #910: Chart axis label alignment defaults with rotation
**Describe the bug**
The default near/far alignments for axis labels do not account for rotation of the labels and thus create some labels with strange alignment.

image

nickofthyme:
## Summary

fix #910 

Improve axis label alignments for `near` and `far` with `0`, `90`, `-90` and `180` rotations.

Screen Recording 2020-11-20 at 03 18 PM

### Checklist


nickofthyme:
jenkins retest this please

src/chart_types/xy_chart/renderer/canvas/axes/tick_label.ts
```diff
@@ -41,6 +42,7 @@ export function renderTickLabel(ctx: CanvasRenderingContext2D, tick: AxisTick, s
     axisStyle,
     tick.position,
     position,
+    tickLabelRotation as Rotation,
```
markov00, line 45: `tickLabelRotation` is in degree, we don't need to cast its type to `Rotation`.
Just check, whenever need, if the rotation if 90,-90,180 or 0, legit numbers for tickLabelRotation

src/chart_types/xy_chart/utils/axis_utils.ts
```diff
@@ -309,6 +309,7 @@ function getVerticalTextOffset(
 
 function getHorizontalAlign(
   position: Position,
+  rotation: Rotation,
```
markov00, line 312: can be just `number` instead of Rotation

markov00 (approved):
LGTM
I've left a comment to improve a bit the typing, removing the cast that is not necessary

src/chart_types/xy_chart/renderer/canvas/axes/tick_label.ts
nickofthyme: Good call, see aee4096ee9bd5e428b447a912078118053c8da87

src/chart_types/xy_chart/utils/axis_utils.ts
nickofthyme: aee4096ee9bd5e428b447a912078118053c8da87

PR #912, opened 2020-11-23
feat: merge series domain with the domain of another group

markov00:
## Summary

This PR will fix a regression that prevents the correct use of `useDefaultGroupDomain`.
It also includes a small feature that allows the consumer to specify a `groupId` on the same prop to allow merging the series domain with the domain of a different `groupId`.


nickofthyme (approved):
LGTM tested locally, fixes the issue.

PR #913, opened 2020-11-24
feat: add projection click listener

### LINKED ISSUE #846: Enable click on the whole bucket
When the difference between buckets is too big on a Bar chart, clicking on a bucket is very difficult as shown in the gif below.

charts-bar-click

I would like to have the whole bucket area clickable.

Screenshot 2020-10-05 at 17 01 02

markov00:
## Summary

This PR adds a projection click listener called `onProjectionListener` and available in the `<Settings />` component with the following characteristics:
- it will fire every time the user clicks on the projection area (everything within the axes)
- the listener is called with an object that contains the nearest X and Ys values. These values are not screen coordinates but the real data domain values inverted computed inverting the x and y scales.
- ~we don't actually prevent firing the `onProjectionListener` if an `onElementClick` listener is configured and the user clicks directly on a bar/point. Both events could fire one after the other.~ We prevent the `onProjectionListener` from firing an event if a `onElementClick` listener is available and it has fired a click event

The type signature of this new listener is the following:
```ts

/**
 * @public
 * The listener type for click on the projection area.
 */
export type ProjectionClickListener = (values: ProjectedValues) => void;


/**
 * @public
 * An object that contains the scaled mouse position based on
 * the current chart configuration.
 */
export type ProjectedValues = {
  /**
   * The independent variable of the chart
   */
  x: PrimitiveValue;
  /**
   * The set of dependent variable, each one with its own groupId
   */
  y: Array<{ value: PrimitiveValue; groupId: string }>;
  /**
   * The categorical value used for the vertical placement of the chart
   * in a small multiple layout
   */
  smVerticalValue: PrimitiveValue;
  /**
   * The categorical value used for the horizontal placement of the chart
   * in a small multiple layout
   */
  smHorizontalValue: PrimitiveValue;
};

```

You can enable it through the `<Settings  onProjectionListener={yourListenerImplementationHere}/>`

The `Interactions -> Bar clicks` story in storybook is updated to fire with this listener

close #846

### Checklist


src/chart_types/xy_chart/state/selectors/get_projected_scaled_values.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    const xValue = xScale.invertWithStep(x, geometriesIndexKeys);
+    if (!xValue) {
+      return;
+    }
+
+    return {
+      x: xValue.value,
+      y: [...yScales.entries()].map(([groupId, yScale]) => {
+        return { value: yScale.invert(y), groupId };
```
nickofthyme, line 48: Is there a need for the y value in this case?

src/chart_types/xy_chart/state/selectors/on_projection_click_caller.ts
```diff
@@ -0,0 +1,62 @@
...
+  let prevClick: PointerState | null = null;
+  let selector: Selector<GlobalChartState, void> | null = null;
+  return (state: GlobalChartState) => {
+    if (selector === null && state.chartType === ChartTypes.XYAxis) {
+      selector = createCachedSelector(
+        [getLastClickSelector, getSettingsSpecSelector, getProjectedScaledValues],
+        (lastClick, { onProjectionClick }, values): void => {
+          if (!onProjectionClick) {
+            return;
+          }
```
nickofthyme, line 47: is there a way to check if the `onElementClick` was called with the same x value, and if so block this from firing?

I'm just thinking about the use case of having the same action fired for both `onElementClick` and `onProjectionClick` especially with multiple bars. Which would fire the action twice. The user could just debounce the function for 100 ms but I wonder if this not firing a duplicate event would be better.

nickofthyme (approved):
LGTM just two comments.

src/chart_types/xy_chart/state/selectors/get_projected_scaled_values.ts
markov00: This is an extra for our current use case. To avoid naming the listener specifically `onXClickListener` I've thought that probably extending the use case a bit can have some benefits for the future

src/chart_types/xy_chart/state/selectors/on_projection_click_caller.ts
markov00: There is a way, we should merge the element and projection click handlers and use a single method to handle both cases. I can check if we can quickly implement that in this PR

nickofthyme: Ok, if it's more hassle than it's worth, I think this is fine as is.

markov00: done here https://github.com/elastic/elastic-charts/pull/913/commits/5d203f221540c800315e1e875096f67d6a191d5d

nickofthyme (approved):

PR #920, opened 2020-11-25
fix: improve line join rendering

markov00:
## Summary

This PR change the [lineJoin](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineJoin) property used when rendering lines to `round` to avoid spikes with very acute angles

From:
Screenshot 2020-11-25 at 14 22 34

To:
Screenshot 2020-11-25 at 14 22 24

The spike effect mostly visible when using a line with a `strokeWidth` greater than 1 and the artifact created distorts also the actual chart reading:
Screenshot 2020-11-25 at 14 33 06


rshen91 (approved):
LGTM tested storybook locally

nickofthyme (approved):

PR #926, opened 2020-11-30
fix(highlighter): show default highlighted radius with hidden dots

### LINKED ISSUE #679: Show default highlighted radius with hidden dots
Using a line or area series with a `MarkAcessor` and hidden points should use the default radius highlighter.

**Expected behavior**
Highlight geometry with hidden point and `MarkAccessor` should be the default radius size.

**Workarounds**
You can check for hidden points and decide when to add the `MarkAccessor` but this would be nice to avoid.

**Screenshots**
Screen Recording 2020-05-15 at 08 37 AM

 - Elastic Charts: `latest`

markov00:
## Summary

This commit fixed the highlighted radius on a line/area point when the `markAccessor` is configured by
the point visibility on the series is turned off

fix #679

A screenshot test is added to check that

### Checklist


nickofthyme (approved):
LGTM

PR #931, opened 2020-12-02
fix: align tooltip z-index to EUI tooltip z-index

### LINKED ISSUE #922: Expose tooltip z-index
When using the tooltip with a `position: fixed` menu, the z-index context can conflict where the tooltip is above the menu. This should be exposed and set by default in eui theme.

image

Related to https://github.com/elastic/kibana/issues/84178

Tooltip `z-index` should be added to the charts theme object

Override the tooltip css directly

markov00:
## Summary

This PR introduces and uses the `z-index` computed on the chart parent when the stack context split begins.
The z-index is applied to the tooltip portal with a +100 value. It should be enough to correctly display tooltip above the current embeddable panel (above the title/header of a panel in Kibana dashboard) and below the current Navigation bar (Kibana)

The storybook style is adapted to use a few z-index values on the various root objects. This should suffice the tooltip rendering tests.

close #922


nickofthyme:
@markov00 is this still a draft? Or is it ready for review?

markov00:
@nickofthyme it still a draft: we can assign a hardcoded z-index, but we should probably use a different way to compute the right z-index for tooltip.
We probably need the same strategy applied by EUI popover: https://github.com/elastic/eui/blob/43ccf53054cc7ad364778d90c25ad781886d3a09/src/services/popover/popover_positioning.ts#L737 we should compute the z-index of the chart element in the dom and use that index accordingly.
I just wondering what would be the best way to apply that computed z-index: should I compute it during the componentDidMount of the `Chart` component  and update the state with that value?

markov00:
I think everything is ok now for a review @nickofthyme

src/components/portal/utils.ts
```diff
@@ -56,3 +62,67 @@ export function getOrCreateNode(id: string, className?: string, parent: HTMLElem
...
+   *
+   * what z-index does the `div` need to display next to `button`?
+   * the `div` and `section` are where the stacking context splits
+   * so `div` needs to copy `section`'s z-index in order to
+   * appear next to / over `button`
+   *
+   * calculate this by starting at `button` and finding its offsetParents
+   * then walk the parents from top -> down until the stacking context
+   * split is found, or if there is no split then a specific z-index is unimportant
+   */
```
nickofthyme, line 92: 👍

src/components/portal/utils.ts
```diff
@@ -56,3 +62,67 @@ export function getOrCreateNode(id: string, className?: string, parent: HTMLElem
...
+
+  // build the array of the element + its offset parents
+  const nodesToInspect: HTMLElement[] = [];
+  while (true) {
+    nodesToInspect.push(element);
+
+    // AFAICT this is a valid cast - the libdefs appear wrong
+    element = element.offsetParent as HTMLElement;
```
nickofthyme, line 100: 👍

nickofthyme (approved):
LTGM, much nicer with the stacking traversing logic. Tested stories for z-index changes and see no regressions 🎉.

PR #933, opened 2020-12-02
feat: improved domain error handling

### LINKED ISSUE #932: Should use fallback minInterval and not error the chart
Error in custom domain should use fallback and not error the chart

```
Uncaught Error: custom xDomain is invalid, custom minInterval is greater than computed minInterval
    at Object.mergeXDomain (x_domain.js:87)
    at Object.computeSeriesDomains (utils.js:137)
    at compute_series_domains.js:20
    at index.js:70
    at index.js:30
    at index.js:84
    at index.js:30
    at selector (index.js:105)
    at index.js:80
```

nickofthyme:
## Summary

Fixes #932

Improves domain errors to fallback to logical defaults.

Use computed `minInterval` when custom `minInterval` is invalid.

Use consistent rendering for custom domain. Currently, when min and max are provided that show no data, the charts is rendered. However, when setting a custom `min` domain that is greater than the computed max, `no data` was rendered. Now the chart will display with the domain set to `[min, min]`, such as...

image

Similarly, when the custom `max` is less than the computed min, `no data` was being rendered. Now the domain will default to `[max, max]`, such as...

Image 2020-12-02 at 2 37 39 PM

> All previous errors replaced with `Logger.warn` logs.

### Checklist


src/chart_types/xy_chart/domains/x_domain.ts
```diff
@@ -77,44 +79,38 @@ export function mergeXDomain(
...
+            seriesXComputedDomains = [customXDomain.min, computedDomainMax];
+          }
+        } else if (isUpperBound(customXDomain)) {
+          if (computedDomainMin > customXDomain.max) {
+            Logger.warn(
+              'custom xDomain is invalid, computed min is greater than custom max. Custom domain is ignored.',
+            );
+          } else {
+            seriesXComputedDomains = [computedDomainMin, customXDomain.max];
+          }
```
markov00, line 108: Can we move this set if ifs outside this function and clean up the nesting a bit?
I'm concerned that we are doing something wrong probably also here that I've discovered when checking how we can clean this up:
when for example the consumer use an invalid custom domain with a minInterval, the min interval is always used also if the configuration is invalid. Do you think we should ignore the custom min interval if the rest of the configuration is invalid?

src/chart_types/xy_chart/domains/x_domain.ts
```diff
@@ -127,6 +123,24 @@ export function mergeXDomain(
...
+      return computedMinInterval;
+    }
+    if (customMinInterval < 0) {
+      Logger.warn('custom xDomain is invalid, custom minInterval is less than 0. Using computed minInterval.');
+      return computedMinInterval;
+    }
+  }
+
+  return customMinInterval || computedMinInterval;
+}
```
markov00, line 142: this can also be cleaned up a bit
```suggestion
  if (customMinInterval == null) {
    return computedMinInternval;
  }
    // Allow greater custom min if xValues has 1 member.
    if (size > 1 && customMinInterval > computedMinInterval) {
      Logger.warn(
        'custom xDomain is invalid, custom minInterval is greater than computed minInterval. Using computed minInterval.',
      );
      return computedMinInterval;
    }
    if (customMinInterval < 0) {
      Logger.warn('custom xDomain is invalid, custom minInterval is less than 0. Using computed minInterval.');
      return computedMinInterval;
    }
  return customMinInterval || computedMinInterval;
}
```

nickofthyme: See 3c7c2f805b4d9a163b285da552b9d59848e75461

src/chart_types/xy_chart/domains/x_domain.ts
nickofthyme: Per slack convo: Gonna leave as is for now and treat the `domain` and `minInterval` parts independently.

markov00 (approved):
LGTM!

PR #934, opened 2020-12-02
fix: point highlight based on geom position and transform

markov00:
## Summary

This PR fixes a regression on the point highlighter for line and area charts.
Due to recent changes on the data processing pipeline, the geometry canvas coordinates are composed by a chart relative x,y position + a transpose function.
The inverse function that checks if the mouse if over the geometry was not updated to reflect this change.
A functional test is also added and tested against the previous code.

### Checklist


nickofthyme (approved):
LGTM, solves the highlight issue.

Screen Recording 2020-12-03 at 09 52 AM

nickofthyme:
Ugh this is still an issue when using bars and lines/areas. I'll take a look at it.

Screen Recording 2020-12-04 at 01 21 PM

Issue #638, opened 2020-04-20
Partition charts with Legend support doesn't work with `showLegendExtra`


cchaos:
**Is your feature request related to a problem? Please describe.**
I'd expect to still be able to show the value of the pie slice in the legend with `showLegendExtra` on `<Settings>`.

<img width="1300" alt="Screen Shot 2020-04-20 at 12 30 21 PM" src="https://user-images.githubusercontent.com/549577/79775896-e6963b80-8302-11ea-8f2c-01caf782bd16.png">

**Checklist**

- [x] this request is checked against already exist requests

PR #939, opened 2020-12-07
feat(legend):  display pie chart legend extra

shahzad31:
## Summary
Fixes: https://github.com/elastic/elastic-charts/issues/638

Pie chart will display values in legends when legend extra prop is passed

Screen Recording 2021-01-14 at 01 30 PM

### Checklist


nickofthyme (changes requested):
This is awesome, thanks for your help! 🎉 

Everything looks good, I found 2 issues thus far.

First, it looks like the indent is not correctly accounting for the width of the legend item and the value is being pushed to the right. You can fix this by changing `marginLeft` here to `paddingLeft`.

 https://github.com/elastic/elastic-charts/blob/3c823fdbc8437c907c02b58c1aa8e084bc7611d1/src/components/legend/legend_item.tsx#L217-L221

Screen Recording 2020-12-08 at 11 30 AM

Secondly, the values of the nested items appear to be incorrect. Notice below that Asia is `$677 Bn` in the chart but `$174 Bn` in the legend. Not sure where this issue is originating in the code.

Image 2020-12-08 at 11 37 29 AM

markov00:
> Secondly, the values of the nested items appear to be incorrect. Notice below that Asia is `$677 Bn` in the chart but `$174 Bn` in the legend. It looks like using the flattened values will override keys in separate branches, you likely need to do this logic on the tree form because in the flattened version you lose the reference to the parent key.

Hey @monfera can you please check this PR and see if your current work on fixing the hierarchical legend fixes also this value? Thanks

monfera:
@markov00 we touched on this on our call with @nickofthyme and he suggests the values are OK as they are and no action is needed in relation to this item. Nick, please tell if I misinterpreted it

nickofthyme:
@monfera that's correct. @Shahzad I can take a look today if you'd like.

shahzad31:
@nickofthyme sure , please go ahead, i am unable to figure out, how to parse legends info in a tree form.

nickofthyme:
@monfera I lied 🤥. After looking into this, the way we display the legend I don't know the parent "key" so I would need your changes from #947

nickofthyme (dismissed):

src/chart_types/partition_chart/state/selectors/get_legend_items_extra.ts
```diff
@@ -0,0 +1,86 @@
...
+    const legendExtraValues = new Map<SeriesKey, LegendItemExtraValues>();
+
+    if (!pieSpec) {
+      return legendExtraValues;
+    }
+    if (isInvalidLegendMaxDepth(legendMaxDepth)) {
+      return legendExtraValues;
+    }
+
+    return getExtraValueMap(pieSpec, tree);
```
monfera, line 45:
```suggestion
    return pieSpec && !isInvalidLegendMaxDepth(legendMaxDepth)) 
        ? getExtraValueMap(pieSpec, tree)
        : legendExtraValues;
```

monfera: Also, I'd prefer positive predicates in general, eg. `isValidLegendMaxDepth`  over `!isInvalidLegendMaxDepth`

src/chart_types/partition_chart/state/selectors/get_legend_items_extra.ts
```diff
@@ -0,0 +1,86 @@
...
+ */
+function getExtraValueMap(
+  { layers, valueFormatter }: Pick<PartitionSpec, 'layers' | 'valueFormatter'>,
+  tree: HierarchyOfArrays,
+  keys: Map<SeriesKey, LegendItemExtraValues> = new Map(),
+): Map<SeriesKey, LegendItemExtraValues> {
+  if (tree.length === 0) {
+    return keys;
+  }
+
```
monfera, line 69:
```suggestion
```

monfera: ... as the `for` loop will bail anyway

src/commons/legend.ts
```diff
@@ -20,14 +20,19 @@
 import { PrimitiveValue } from '../chart_types/partition_chart/layout/utils/group_by_rollup';
 import { Color } from '../utils/commons';
 import { SeriesIdentifier } from './series_id';
+
 /** @internal */
-export type LegendItemChildId = string;
+export type LegendItemChildId = string | number;
```
monfera, line 25: I'm not sure about it, in the other PR I also redefined it, hoisted the definition to `CategoryKey` but that gets channeled to a lot of places; if you think it really needs to admit a number here too, I'd need to understand more about it before approving this broadening, as it might be important, not sure yet. For this reason maybe it's better for the legendStrategy PR go in first, as it still retains the `string`

src/commons/legend.ts
```diff
@@ -20,14 +20,19 @@
...
 
 /** @internal */
 export type LegendItem = {
   seriesIdentifier: SeriesIdentifier;
   childId?: LegendItemChildId;
   depth?: number;
+  /**
+   * Path to iterm in hierarchical legend
+   */
+  path?: number[];
```
monfera, line 35: The legend strategy PR also adds it, however as a mandatory prop, and also structured, so it'd be useful to rebase this PR on that; in part b/c my day is wrapping up sooner than yours :-D

monfera: Btw. mandatory props have a lot of benefit in the realm of typing and code simplicity; no need for runtime conditions or type guards, so wherever possible, I'd prefer mandatory over optional. Maybe there are counterarguments that'd shift my understanding

monfera (changes requested):
Great PR 🎉  Only requesting some changes because of the attempt at hoisting some definition, namely the concept of `Category`, see `category.ts` in the legend strategy PR. It wasn't a fully arbitrary choice; series keys and legend labels have commonality. Basically, both represent category. So it fell out, besides my slant toward moving to more semantic concepts than `string`. Looking forward to nominal typing, or if it doesn't come soon enough, branded types in `elastic-charts`, because `string` is not very safe, now a lot of things that are semantically related just happen to be all `string`s

src/commons/legend.ts
nickofthyme: I added this because the `childId` can be a `number` or `string` and the map with be type sensitive. Meaning if I convert the `childId` to a string I need to do that everywhere. This just simplifies the type checking.

Do you know a specific case where this would adversely affect the code?

src/commons/legend.ts
nickofthyme: True, I was simply using that to find the right key using a path or key but the path is defined such that it works already (i.e. the `value` is the `seriesIdentifier.key`)

src/chart_types/partition_chart/state/selectors/get_legend_items_extra.ts
nickofthyme: 69c7dc8

src/chart_types/partition_chart/state/selectors/get_legend_items_extra.ts
nickofthyme: 69c7dc8

monfera: Good changes here, though would still prefer a positive predicate over a negative one, though totally not a blocker.

I've a question here: this extracts, and only keys on the concatenated `value` ie. series keys; won't this approach need the structure information downstream of this? The legend strategies PR firmly detached the notion of sufficiency of knowing the label even for just highlighting the value, at least for non-single layer charts such as partition charts. I'll look at it in more detail tomorrow, just wanted to check if/how this info won't be needed, ie. why eg. not key on the concatentated `{index, value}` tuples rather than on just the `value`.

```
keys.set(path.map(({ value: v }) => v).join('__'), values);
```

src/chart_types/partition_chart/state/selectors/get_legend_items_extra.ts
monfera: 👍

src/commons/legend.ts
monfera: I don't immediately understand the need; tried to trace and it appears to be, ultimately, `string`: 

- `values.set(key, formattedValue);`, line 68 in `getExtraValueMap` has `key: string` from line 61, so no number there
- I haven't checked the ton of other uses of though many of them appear to ultimately use `CategoryKey` therefore, now, `string`, eg. `const seriesTooltipValues = new Map<SeriesKey, LegendItemExtraValues>();` in `tooltip.ts`
- `computeLegendSelector` in `src/chart_types/partition_chart/state/selectors/compute_legend.ts` used to not need the number; specifically, the `dataName` inside `ShapeViewModel.quadViewModel`, which extends `ShapeTreeNode`, has `dataName` as `CategoryKey`
- having grepped the code for `childId`, no obvious number assignment turned up; things like `dataName` and `BandedAccessorType` get assigned

It'd be great to avoid the potential leakiness of admitting `number` here, so I'd like to better understand where exactly it's neaded; you said

> the childId can be a number or string

but on cursory look it's not universally true so the list of specific places would be neat; ie. where would a conversion from numbers to strings would need to take place. 

It'd be useful for me to know what the ultimate _meaning_ of the `childId` is and why it's desirable for numbers to be admitted.

I don't immediately see test cases that exercise anything called `childId` with numbers either, but again, it's late hours for me 😄

src/chart_types/partition_chart/state/selectors/get_legend_items_extra.ts
nickofthyme: I just assume not mess with existing naming (i.e. `isInvalidLegendMaxDepth`) because there's likely a lot of this around. I typically use positive or negative such that I don't have to `!` it.

nickofthyme: Yes you are right about the `value` that should be `index` instead. See 13811eb051abb5a2399507d8b10c50b7f24c75bf

src/commons/legend.ts
nickofthyme: Yup you are right. This was a remnant from before I used the `path.join` logic. Thanks for being persistent. See a2972a0

monfera: Haha mostly blind luck, could've been needed, it gets used in numerous places

src/chart_types/partition_chart/state/selectors/get_legend_items_extra.ts
monfera: Thanks! Fair enough not to add possibly sweeping changes to this PR. We could maybe brainstorm about whether we should prefer positive predicates at all, not just the general avoidance of `non` or `un` but also some legit words like `invalid`; there's no specific line we can draw between negative and positive words and it might be something noone else cares about, it's a miniscule stylistic thing

monfera: Re the `value` vs `index` change, the current test suite is insensitive to whether it yields the desired thing, it'd be neat to make it fail if the info is insufficient, I guess it depends on mock structure to trigger a difference

nickofthyme: See dac6d17

monfera (approved):
Looks good and super useful, great job  🎉 and thanks for the changes too! Good to merge on green

PR #940, opened 2020-12-07
fix: empty labels on debug state

nickofthyme:
## Summary

When using the new `debugState`, the axis labels where using the non-filtered ticks when using either `showOverlappingLabels` or `showOverlappingTicks` options. This PR uses only the visible ticks and filters out empty string labels.

### Checklist


rshen91 (approved):
LGTM - tested locally

PR #945, opened 2020-12-09
docs: fix color picker clearing logic

nickofthyme:
## Summary

Fixes the way we use the color picker render function in the [color picker story](https://elastic.github.io/elastic-charts/?path=/story/legend--color-picker).

The issue is when setting the color, the `onChange` is fired to set the color in the `temporary` colors. Then when `onClose` is fired, we save the set color to the `persistent` colors. If you are clearing a color that is in the `temporary` colors, there is no issue. But when clearing a color that has been set in the `persistent` colors, the fallback uses the custom color function from the `color` prop. We needed to update the state of the `colors` object in the story but in doing so the color picker would delete the color dot on the legend. Thus using the color picker with `useMemo` prevents react from rerendering the color picker and fixes the clearing issue.

### Before
Screen Recording 2020-12-09 at 10 12 AM

### After
Screen Recording 2020-12-09 at 10 11 AM

### Checklist


rshen91 (approved):
LGTM. This change makes sense to me 👌

Issue #944, opened 2020-12-09
[minor] larger and uniform legend capture zones


monfera:
**Is your feature request related to a problem? Please describe.**
Currently, the legend items capture the mouse only when hovering over the bounding box of the legend item ink:
![legend capture](https://user-images.githubusercontent.com/1548516/101632350-11660900-3a26-11eb-8c33-b7533c365948.gif)

- this causes flashing when the mouse moves from one item to the next
- the indentation in the nested legend case also leads to flashing

**Describe the solution you'd like**
The entire legend row should be active, such that the
- width use is maximized (100%) for capture, even if a legend item is indented, or one text is shorter than on other lines
- height is also maximized, so there's no gap between rows

**Describe alternatives you've considered**
None

**Additional context**
These help not just with flash avoidance but also, easier and more predictable activation by the user and better Fitts law adherence.

If accepted, this could be added to https://github.com/elastic/elastic-charts/issues/858

**Kibana Cross Issues**
Not aware

**Checklist**

Delete any items that are not applicable to this feature request.

- [X] this request is checked against already exist requests
- [X] every related Kibana issue is listed under `Kibana Cross Issues` list
- [X] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

monfera:
Reopened as auto-closing was due to incorrect issue reference

PR #947, opened 2020-12-11
fix(legend): hierarchical legend order should follow the tree paths

monfera:
## Summary

- Refactoring of only local effect of some label-related lines
- Closes https://github.com/elastic/elastic-charts/issues/944
- Deactivates the anti-iterator rule

Hierarchical legends may currently occur with sunburst and treemap charts, so labeled as partition too.

Actual fix is in the last commit, the preceding ones are just a pass of incremental updates to naturally bump into what needed changing.

The same mock as in the above issue:
image

### Checklist


src/chart_types/partition_chart/state/selectors/compute_legend.ts
```diff
@@ -22,80 +22,80 @@ import createCachedSelector from 're-reselect';
...
-      const key = [dataName, fillColor].join('---');
-      if (!acc[key]) {
-        acc[key] = 0;
-      }
-      acc[key] += 1;
-      return acc;
-    }, {});
-
-    const { flatLegend, legendMaxDepth, legendPosition } = settings;
+    const uniqueNames = new Set(map(({ dataName, fillColor }) => makeKey(dataName, fillColor), quadViewModel));
```
monfera, line 41: Here, it could've been
```
const uniqueNames = new Set(quadViewModel.map(({ dataName, fillColor }) => makeKey(dataName, fillColor)));
```
the only difference is that this version would have materialized an array. Maybe we can start adding a couple of utils for non-materializing filter, reduce etc.

src/chart_types/partition_chart/state/selectors/compute_legend.ts
```diff
@@ -22,80 +22,80 @@ import createCachedSelector from 're-reselect';
...
-    let items = geoms.quadViewModel.filter(({ depth, dataName, fillColor }) => {
+    const items = quadViewModel.filter(({ depth, dataName, fillColor }) => {
       if (legendMaxDepth != null) {
         return depth <= legendMaxDepth;
       }
       if (forceFlatLegend) {
-        const key = [dataName, fillColor].join('---');
-        if (uniqueNames[key] > 1 && excluded.has(key)) {
+        const key = makeKey(dataName, fillColor);
+        if (uniqueNames.has(key) && excluded.has(key)) {
```
monfera, line 51: Eventually we should avoid this filter loop, relying on a new operator that intersects two collections eg. two Sets.
A few ones can be added, eg. intersect, union, subtract

src/specs/settings.tsx
```diff
@@ -481,7 +482,7 @@ export function isTooltipProps(config: TooltipType | TooltipProps): config is To
 
 /** @internal */
 export function isTooltipType(config: TooltipType | TooltipProps): config is TooltipType {
-  return typeof config === 'string';
+  return typeof config !== 'object'; // TooltipType is 'vertical'|'cross'|'follow'|'none' while TooltipProps is object
```
markov00, line 485: 🙏

src/chart_types/partition_chart/state/selectors/get_highlighted_shapes.ts
```diff
@@ -27,6 +27,7 @@ import { partitionGeometries } from './geometries';
 const getHighlightedLegendItemKey = (state: GlobalChartState) => state.interactions.highlightedLegendItemKey;
 
 /** @internal */
+// why is it called highlighted... when it's a legend hover related thing, not a hover over the slices?
```
markov00, line 30: Are you referring to the fact that we don't describe that as a legend related thing?
Something like `getLegendHighlightedSectorsSelector` seems to sound better then the current one

markov00 (approved):
Looks good to me, I've added few comments.
I've noticed that the new introduced visual regression tests baseline are not committed, you should commit them and everything else should be fine

nickofthyme (approved):
LGTM, Tested locally. This looks to be a solution to fix #939.

PR #949, opened 2020-12-14
feat: add alternative text description to xy charts

rshen91:
## Summary

This PR should provide a elastic charts generated text alternative for the canvas xy charts. In a future PR, we will address the `<ScreenReaderDataTable />` as well as providing an option for consumers of elastic charts to provide their own alt text and ultimately make this functionality across all chart types. 

### Checklist


src/components/screen_reader_data_table/screen_reader_data_table.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const dataKeys: JSX.Element[] = [];
+    // go through the number of series in ScreenReaderData
+    for (let seriesIndex = 0; seriesIndex < d.length; seriesIndex++) {
+      if (d.length !== 1) {
+        dataKeys.push(
+          <tr key={`${d[seriesIndex].dataKey}__${seriesIndex}`}>
+            <th key={`${d[seriesIndex].dataValue}__${d[seriesIndex].seriesName}__${seriesIndex}`}>
+              {seriesIndex + 1} series of the total {d.length} series{' '}
+            </th>
+          </tr>,
```
markov00, line 79: This is a bit weird, because it is interpreted as a row with only the first column available.
There is no `scope` here and we should probably span the content to cover all the columns

src/components/screen_reader_data_table/screen_reader_data_table.tsx
```diff
@@ -0,0 +1,139 @@
...
+        );
+      }
+      // get the index of the key in each series and use the index to get the index of it in the values
+      for (let j = 0; j < d[seriesIndex].dataKey.length; j++) {
+        dataKeys.push(
+          <tr
+            key={`${d[seriesIndex].seriesType}__${d[seriesIndex].dataKey[j]}__${seriesIndex}`}
+            aria-labelledby="alt text for chart data"
```
markov00, line 87: please check every `aria` attributes and fix them with their correct usage

markov00 (changes requested):
I don't have a lot of knowledge about how screen readers works but it, from what I've tested I see the potential pitfalls here and we should step back a bit from the problem.
There are multiple cases that we should consider and all of them require an accurate design on the accessible table to represent correctly the data, in particular:
- are we correctly describing multiple series? are we describing correctly their types and values?
- are we representing stacked series? do we need to show summary statistics for each stacked set?

I think we should write down a table format proposal for various use cases, validate it with a11y team and then work on implement it in the correct way

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
```diff
@@ -126,6 +132,61 @@ class XYChartComponent extends React.Component<XYChartProps> {
...
+          role="presentation"
+          {...ariaProps}
+        />
+        {description && (
+          // @ts-ignore
+          <HeadingLevel>
+            {' '}
+            {/* id={altTextLabel}> */}
+            {description}
+          </HeadingLevel>
```
monfera, line 183: Does the empty content (row 180) serve some specific purpose?

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
```diff
@@ -126,6 +132,61 @@ class XYChartComponent extends React.Component<XYChartProps> {
...
+        : // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+          `${generatedDescription}`,
+    };
+
+    if (!initialized || isChartEmpty) {
+      this.ctx = null;
+      return null;
+    }
+    return (
+      <figure>
```
monfera, line 163: Just for my learning for a11y, I'm wondering if `the <figure>` tag should be annotated, eg. with `aria-label`; I see things in the <canvas> element, and right below in `<HeadingLevel>`—is there a reference that tells us to wrap things around with `<figure>`?

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
```diff
@@ -219,6 +253,11 @@ const DEFAULT_PROPS: ReactiveChartStateProps = {
   annotationDimensions: new Map(),
   annotationSpecs: [],
   panelGeoms: [],
+  dataTable: {
+    showDefaultDescription: true,
+    description: undefined,
```
monfera, line 258: Would it help keeping types simple, if `description` got initialized to eg. the empty string like `generatedDescription` right below? Can still conditionalize on empty vs non-empty strings in use

src/chart_types/xy_chart/state/utils/utils.ts
```diff
@@ -585,3 +585,87 @@ export function getBarIndexKey(
 
   return [groupId, specId, ...splitAccessors.values(), yAccessor].join('__-__');
 }
+
+function parseDataForKeysStacked(d: DataSeries) {
+  if (!d) return [];
```
monfera, line 590: If `d` is `DataSeries`, this condition will never be true, so I think there's no need for this bail line

src/chart_types/xy_chart/state/utils/utils.ts
```diff
@@ -585,3 +585,87 @@ export function getBarIndexKey(
...
+  const keysAndValues = Object.entries(d.data[0]);
+  const keys = [];
+  for (let i = 0; i < keysAndValues.length; i++) {
```
monfera, line 593: Minor: `reduce` would probably be a little more streamlined to not need an external accumulator in row 592 (glad to chat/zoom), and `keysAndValues` could be renamed `entries`, and eg. destructuring would be better than [0]/[1] indexing, eg.
```
return entries.reduce((result, [key, value]) => {
     if(key !== 'datum' && value !== null || value !== undefined) {
       result.push(key);
     }
     return result;
  }), []);     
  ```
Also, I'm interested in why there's a need for iterating through `data` as we might be doing that elsewhere too, though maybe I'll run into the reason in the code as I go through

File: src/chart_types/xy_chart/state/utils/utils.ts [hunk 8]
```diff
@@ -585,3 +585,87 @@ export function getBarIndexKey(
...
+    if (keysAndValues[i][0] !== 'datum' && keysAndValues[i][1] !== null && keysAndValues[i][1] !== undefined) {
+      keys.push(keysAndValues[i][0]);
+    }
+  }
+  return keys;
+}
+
+function parseDataForValuesStacked(d: DataSeries) {
+  if (!d) return [];
```
monfera, line 602: Same - may not be needed

src/chart_types/xy_chart/state/utils/utils.ts
```diff
@@ -585,3 +585,87 @@ export function getBarIndexKey(
...
+  const values: any[] = [];
+  // need to go through each data[i] and get the value of the not null undefined
+  d.data.map((value) => {
```
monfera, line 605: Combining functional mapping (to get a result) and side effects (pushing into `values` that's outside the `map` has pros&cons, wondering if it's possible to do separately while still preserving performance

src/chart_types/xy_chart/state/utils/utils.ts
```diff
@@ -585,3 +585,87 @@ export function getBarIndexKey(
...
+    const keysAndValues = Object.entries(value);
+    // eslint-disable-next-line array-callback-return
+    return keysAndValues.map((val) => {
```
monfera, line 608: Same here, a `map` shouldn't be used for just its side effect; we return the `keysAndValues.map` here but the outer `map` isn't using it anywhere, should be a `forEach` for clarity, as the most minimal change.

Although I'd look into some other coding approach; basically the entire 13 row function body is roughly equivalent to the equally not speed optimized 3 row version (I haven't tested)
```
return d.data.flatMap(value => Object.entries(value)
  .filter(([k, v]) => k !== 'datum' && tyeof val === 'number'))
  .map(([k]) => k);
```

src/chart_types/xy_chart/state/utils/utils.ts
```diff
@@ -585,3 +585,87 @@ export function getBarIndexKey(
...
+      if (val[0] !== 'datum' && val[1] !== null && val[1] !== undefined && typeof val[1] === 'number') {
```
monfera, line 609: No need to check if the value is null or undefined, as it won't be a `number` if it's either of them, so it's enough to test if it's a number. Unless we want to retain `NaN` and infinities, a nice predicate is `Number.isFinite()`

src/chart_types/xy_chart/state/utils/utils.ts
```diff
@@ -585,3 +585,87 @@ export function getBarIndexKey(
...
+        return values.push(val[1]);
+      }
+    });
+  });
+  return values;
+}
+
+function parseDataForKeys(d: DataSeries) {
```
monfera, line 617: We have several functions of roughly the same shape, runtime behavior and improvement options; in this case, it's worth looking into bringing them together one common function, may or may not be possible but it's good to try for more DRYness

monfera, line 617: This function looks identical to `parseDataForKeysStacked` two functions above and similar or maybe identical to some others I saw. A need to do copy & paste is warning that we could do something more DRY

src/chart_types/xy_chart/state/utils/utils.ts
rshen91: Thank you so much for this review - I updated based on this feedback mainly in commit cd8c964. 🙇‍♀️

src/chart_types/xy_chart/state/utils/utils.ts
rshen91: Yeah I think your suggestion is more streamlined - incorporated in commit cd8c964

src/chart_types/xy_chart/state/utils/utils.ts
rshen91: 👍 Removed in cd8c964

src/chart_types/xy_chart/state/utils/utils.ts
rshen91: Smart! Incorporated in commit cd8c964

src/chart_types/xy_chart/state/utils/utils.ts
rshen91: Thank you! Commit cd8c964

src/chart_types/xy_chart/state/utils/utils.ts
rshen91: DRY is awesome cd8c964

rshen91: Please see cd8c964 🙇

src/chart_types/xy_chart/state/utils/utils.ts
rshen91: I missed this one sorry! a45b9e2

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
rshen91: Sorry this needed to be cleaned up - 3c3f54e

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
rshen91: This was outdated and I needed to clean it up :yikes: 3c3f54e

src/chart_types/xy_chart/state/utils/utils.ts
```diff
@@ -600,3 +600,38 @@ export function getBarIndexKey(
 
   return [groupId, specId, ...splitAccessors.values(), yAccessor].join('__-__');
 }
+
+/** @internal */
```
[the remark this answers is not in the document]
rshen91: Thank you! TIL - removed these

src/chart_types/xy_chart/state/selectors/is_brush_available.ts
```diff
@@ -32,7 +32,7 @@ import { getComputedScalesSelector } from './get_computed_scales';
 export const isBrushAvailableSelector = createCachedSelector(
   [getSettingsSpecSelector, getComputedScalesSelector],
   (settingsSpec, scales): boolean => {
-    if (!scales.xScale) {
+    if (!scales.xScale || settingsSpec.dataTable?.showDataTable === true) {
```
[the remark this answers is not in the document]
rshen91: Yep this is removed - I was hitting a strange error that I can't reproduce now 😅

src/components/screen_reader_data_table/screen_reader_data_table.tsx
```diff
@@ -0,0 +1,166 @@
...
+        firstXAxisGroupId === undefined
+          ? `Its y-axis does not have a defined domain.`
+          : // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+            `The y-axis domain is ${yScales.get(yAxisGroupId!)?.domain}`
+      }`;
+};
+
+/** @internal */
+export const computeAlternativeChartText = (d: ScreenReaderData[]): string => {
+  const { seriesName, axesTitles, splitAccessor, seriesType, yScales } = d[0];
```
markov00, line 107: This line will fail if the `showDefaultDescription` prop was set to `false` on the `Settings` component

src/components/screen_reader_data_table/screen_reader_data_table.tsx
```diff
@@ -0,0 +1,166 @@
...
+    : ` ${new Date(startDomain).toTimeString()} to ${new Date(endDomain).toTimeString()}`;
+};
+
+/** helper function to read out each title of the axes */
+const axesWithTitles = (titles: (undefined | string)[][], xScale: Scale, yScales: Map<GroupId, Scale>) => {
+  let titleDomain = yScales.size > 1 ? `There are ${yScales.size} y-axes. ` : ' ';
+  const numberOfScales = titles.length > yScales.size ? titles.length : yScales.size;
+  for (let i = 0; i < numberOfScales; i++) {
+    const axisDirection = titles[i][2] === 'top' || titles[i][2] === 'bottom' ? 'x-' : 'y-';
+    const scaleType = axisDirection === 'x-' ? xScale.type : yScales.get(titles[i][1]!)?.type;
```
markov00, line 73: It's better to not infer the scale from the axis direction.
Please just list all the yScales and the xScale, independently if they are on the bottom or top, left or right.

src/components/screen_reader_data_table/screen_reader_data_table.tsx
```diff
@@ -0,0 +1,166 @@
...
+    const axisTitle = titles[i][0] ? titles[i][0] : '__global__';
+    const yDomain =
+      yScales.get(axisTitle!) === undefined ? yScales.get('__global__'!)?.domain : yScales.get(axisTitle!)?.domain;
+    const axisDomain = `${axisDirection === 'y-' ? `${yDomain}` : `${formatForTimeOrOrdinalAxis(xScale)}`}`;
+    titleDomain +=
+      yScales.size === 1
+        ? `The ${axisDirection}axis has the scale type  ${scaleType} and has the title ${axisTitle} with the domain ${axisDomain}. `
+        : `${i}. A ${axisDirection}axis is the scale type ${scaleType} and is titled ${axisTitle} with the domain ${axisDomain}. `;
```
markov00, line 81: if the axis doesn't have a title, we don't need to say that, we don't need to use the `__global__` name for that.
Just say: `The X axis is temporal scale` or with title, `The X axis, named blablabla, is a linear scale`

src/components/screen_reader_data_table/screen_reader_data_table.tsx
```diff
@@ -0,0 +1,166 @@
...
+  let titleDomain = yScales.size > 1 ? `There are ${yScales.size} y-axes. ` : ' ';
+  const numberOfScales = titles.length > yScales.size ? titles.length : yScales.size;
+  for (let i = 0; i < numberOfScales; i++) {
+    const axisDirection = titles[i][2] === 'top' || titles[i][2] === 'bottom' ? 'x-' : 'y-';
+    const scaleType = axisDirection === 'x-' ? xScale.type : yScales.get(titles[i][1]!)?.type;
+    const axisTitle = titles[i][0] ? titles[i][0] : '__global__';
+    const yDomain =
+      yScales.get(axisTitle!) === undefined ? yScales.get('__global__'!)?.domain : yScales.get(axisTitle!)?.domain;
+    const axisDomain = `${axisDirection === 'y-' ? `${yDomain}` : `${formatForTimeOrOrdinalAxis(xScale)}`}`;
+    titleDomain +=
```
markov00, line 78: instead of concatenate the strings like that, collect all of them in an array and then join it with the appropriate separator (a space, a new line or nothing depending on your templates)

src/state/selectors/get_screen_reader_settings.ts
```diff
@@ -0,0 +1,36 @@
...
+export const getScreenReaderDataTableSettingsSelector = createCachedSelector(
+  [getSettingsSpecSelector],
+  ({ dataTable }): DataTableProps => {
+    return {
+      showDefaultDescription: dataTable.showDefaultDescription ?? true,
+      description: dataTable.description ? dataTable.description : undefined,
+      visibleCaption: dataTable.visibleCaption ? dataTable.visibleCaption : undefined,
+      HeadingLevel: dataTable.HeadingLevel ? dataTable.HeadingLevel : 'h2',
+    };
+  },
```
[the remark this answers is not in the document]
rshen91: Thank you! b9aaa1d

src/components/screen_reader_data_table/screen_reader_data_table.tsx
```diff
@@ -0,0 +1,166 @@
...
+};
+
+const formatForTimeOrOrdinalAxis = (xScale: Scale) => {
+  if (xScale.domain === undefined) return 'is undefined';
+  if (xScale.type === 'ordinal') return `is ordinal.`;
+  const [startDomain, endDomain] = xScale.domain;
+  return xScale.type !== ScaleType.Time
+    ? // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+      ` ${startDomain} to ${endDomain}`
+    : ` ${new Date(startDomain).toTimeString()} to ${new Date(endDomain).toTimeString()}`;
```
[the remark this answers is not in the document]
rshen91: Thank you, good catch, please see 9e75754

src/components/screen_reader_data_table/screen_reader_data_table.tsx
```diff
@@ -0,0 +1,166 @@
...
+export const computeAlternativeChartText = (d: ScreenReaderData[]): string => {
+  const { seriesName, axesTitles, splitAccessor, seriesType, yScales } = d[0];
+  const numberOfSeriesByType: { [key: string]: number } = {};
+  d.forEach(() => {
+    if (numberOfSeriesByType[seriesType]) {
+      numberOfSeriesByType[seriesType]++;
+    } else {
+      numberOfSeriesByType[seriesType] = 1;
+    }
+  });
```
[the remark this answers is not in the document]
rshen91: Thank you! cf0cbe6

src/components/screen_reader_data_table/screen_reader_data_table.tsx
rshen91: Good point, fixed in fac7927

src/components/screen_reader_data_table/screen_reader_data_table.tsx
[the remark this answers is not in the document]
rshen91: Ah good catch fixed in cf0cbe6

src/components/screen_reader_data_table/screen_reader_data_table.tsx
rshen91: Updated in b9aaa1d

rshen91:
Closing for new and cleaner issue here https://github.com/elastic/elastic-charts/pull/1084

PR #952, opened 2020-12-16
fix(legend): disable focus and keyboard navigation for legend in partition ch…

rshen91:
In #880 focus was added to the legend component. However this doesn't make sense with partition charts because the functionality to hide slices doesn't yet exist.


nickofthyme (approved):
Nice! Thanks for fixing this so quickly.

PR #960, opened 2021-01-04
refactor: annotation marker size calculation

markov00:
## Summary

A propedeutic refactor of the `LineAnnotation` spec to allow the use of a custom react reconciler.

When using a custom reconciler, every component within the wrapping parent is treated in the same way. Due to the nature of the new reconciler we can't "host" mixed components that are both a Spec component (that store it's props into the redux store) and a React component (as it was the LineAnnotation where the marker was first drawn offscreen, measured and then stored into the redux store).

The `LineAnnotation` was refactored to ignore the marker size, the marker is just rendered when and where needed with a transformation to align itself based on its anchor point.

**Note**
- The Path polyfill is removed from the dependencies, since we dropped the IE11 support there is no need for that anymore.
The Path2D mock in the Enzyme testing was applied after receiving an error message from the testing environment. This should not affect the current unit testing as it's just used on the rendering side not tested through Enzyme.

- The calculation of the marker size was used to compute the right position of the marker and the mouse coordinate hit point within that marker: I've replaced the placement with a simple css transform and I've replaced the computation of the mouse hit within the marker with a mouseenter/mouseleave listener on the marker container, storing the information of the hovered DOM element within the redux store.
- I moved the computation of the formatted details/header within the tooltip since it was a very cleaner to just pass the original datum around and it derivated formatted form.


src/chart_types/xy_chart/annotations/line/dimensions.ts
```diff
@@ -209,19 +213,18 @@ function computeXDomainLineAnnotationDimensions(
           ? {
               icon: marker,
               color: lineColor,
-              dimension: { ...markerDimensions },
+              dimension: markerDimensions,
               position: {
-                top: markerPosition.top,
-                left: markerPosition.left,
+                ...markerPosition,
```
monfera, line 218: Any reason to not do `position: markerPosition`?

src/chart_types/xy_chart/annotations/line/dimensions.ts
```diff
@@ -88,35 +89,35 @@ function computeYDomainLineAnnotationDimensions(
...
         const annotationMarker: AnnotationMarker | undefined = marker
           ? {
               icon: marker,
               color: lineColor,
-              dimension: { ...markerDimensions },
+              dimension: markerDimensions,
               position: {
-                top: markerPosition.top,
-                left: markerPosition.left,
+                ...markerPosition,
```
monfera, line 105: here as well

src/chart_types/xy_chart/annotations/line/dimensions.ts
```diff
@@ -375,13 +372,10 @@ export function getMarkerPositionForXAnnotation(
...
   { width, height }: Size,
   rotation: Rotation,
-  { width: mWidth, height: mHeight }: Size,
   position: Position,
   value: number,
-): {
-  top: number;
-  left: number;
-} {
+  { width: mWidth, height: mHeight }: Size = { width: 0, height: 0 },
```
monfera, line 377: 😻

src/chart_types/xy_chart/annotations/tooltip.ts
```diff
@@ -50,47 +48,24 @@ export function computeAnnotationTooltipState(
...
-
-      if (rectAnnotationTooltipState) {
-        return {
-          ...rectAnnotationTooltipState,
-          tooltipSettings,
-          customTooltip,
-          customTooltipDetails: customTooltipDetails ?? spec.renderTooltip,
-        };
-      }
+    if (rectAnnotationTooltipState) {
```
monfera, line 62: Nice reduction of a bunch of branching

src/chart_types/xy_chart/annotations/tooltip.ts
```diff
@@ -50,47 +48,24 @@ export function computeAnnotationTooltipState(
...
-      }
-      const lineAnnotationTooltipState = computeLineAnnotationTooltipState(
-        cursorPosition,
-        annotationDimension as AnnotationLineProps[],
-        groupId,
-        spec.domainType,
-        axesSpecs,
-        chartDimensions,
-      );
+    const rectAnnotationTooltipState = getRectAnnotationTooltipState(
```
monfera, line 55: Great to see the`if` nesting gone too!

src/chart_types/xy_chart/renderer/dom/annotations/annotations.tsx
```diff
@@ -60,25 +68,56 @@ interface AnnotationsOwnProps {
...
+    case Position.Right:
+      return `translate(-50%, 0%)`;
+    case Position.Left:
+      return `translate(-100%, -50%)`;
+    case Position.Top:
+      return `translate(-50%, -100%)`;
+    case Position.Bottom:
+    default:
+      return `translate(-50%, 0%)`;
+  }
```
monfera, line 85:
```suggestion
  const MARKER_TRANSFORMS = {
    Position.Right: 'translate(-50%, 0%',
    Position.Left: 'translate(-100%, -50%)',
    Position.Top: 'translate(-50%, -100%)',
    Position.Bottom: 'translate(-50%, 0%)',
  };
  return MARKER_TRANSFORMS[alignment];
```

src/chart_types/xy_chart/renderer/dom/annotations/annotations.tsx
```diff
@@ -60,25 +68,56 @@ interface AnnotationsOwnProps {
 
 type AnnotationsProps = AnnotationsDispatchProps & AnnotationsStateProps & AnnotationsOwnProps;
 
+function getMarkerCentredTransform(alignment: Position, hasMarkerDimensions: boolean): string | undefined {
+  if (hasMarkerDimensions) {
+    return undefined;
+  }
+  switch (alignment) {
```
monfera, line 75: reason for changes: compact; O(1); no need for backquotes; no need to play tricks with a TS-enforced `default:` as `Position` can't be any other value

src/chart_types/xy_chart/specs/line_annotation.tsx
```diff
@@ -17,108 +17,25 @@
...
-import { connect } from 'react-redux';
-import { bindActionCreators, Dispatch } from 'redux';
+import React from 'react';
 
 import { ChartTypes } from '../..';
-import { Spec } from '../../../specs';
 import { SpecTypes } from '../../../specs/constants';
-import { upsertSpec as upsertSpecAction, removeSpec as removeSpecAction } from '../../../state/actions/specs';
-import { DEFAULT_ANNOTATION_LINE_STYLE, mergeWithDefaultAnnotationLine } from '../../../utils/themes/theme';
+import { getConnect, specComponentFactory } from '../../../state/spec_factory';
```
monfera, line 24: Big code and complexity removal here 💪

src/chart_types/xy_chart/state/selectors/get_annotation_tooltip_state.ts
```diff
@@ -99,3 +114,59 @@ function getAnnotationTooltipState(
...
+  // and we can safety cast the union types to the respective Line types
+  const spec = annotationSpecs.find(({ id }) => id === hoveredDOMElement.createdBySpecId);
+  if (!spec || spec.hideTooltips) {
+    return null;
+  }
+  const annotations = annotationDimensions.get(hoveredDOMElement.createdBySpecId);
+  if (!annotations) {
+    return null;
+  }
+  const dimension = (annotations as AnnotationLineProps[]).find((d) => {
```
monfera, line 137: As you don't need the `annotations` downstream, it'd be possible to skip the `const annotations / return null` above by something like 
```
cost dimension = (annotationDimensions.get(hoveredDOMElement.createdBySpecId) ?? []).find((d) => ...
```
As the annotations would just be empty; can't find anyting in an empty array. Might work better with shorter variable names, probably it's better as it is

package.json
```diff
@@ -65,7 +65,6 @@
     "d3-scale": "^1.0.7",
     "d3-shape": "^1.3.4",
     "newtype-ts": "^0.2.4",
-    "path2d-polyfill": "^0.4.2",
```
monfera, line 68: Great to see a polyfill gone 🎉

src/chart_types/xy_chart/annotations/line/dimensions.ts
```diff
@@ -88,35 +89,35 @@ function computeYDomainLineAnnotationDimensions(
...
-                top: markerPosition.top,
-                left: markerPosition.left,
+                ...markerPosition,
               },
+              alignment: anchorPosition,
             }
           : undefined;
         const lineProp: AnnotationLineProps = {
+          specId,
+          id: `${lineProps.length}`,
```
nickofthyme, line 112: Is this a unique id?

src/chart_types/xy_chart/renderer/dom/annotations/annotations.tsx
nickofthyme: Agree!

src/chart_types/xy_chart/renderer/dom/annotations/tooltip_content.tsx
```diff
@@ -53,10 +53,14 @@ export const TooltipContent = ({
...
-  }, [details, customTooltipDetails]);
+  }, [datum, customTooltipDetails]);
 
   if (CustomTooltip) {
-    return <CustomTooltip details={details} header={header} />;
+    const { details } = datum;
+    if ('header' in datum) {
+      return <CustomTooltip details={details} header={datum.header} datum={datum} />;
+    }
+    return <CustomTooltip details={details} datum={datum} />;
```
nickofthyme, line 63: Why not just this?

```suggestion
    return <CustomTooltip details={details} header={datum.header} datum={datum} />;
```

nickofthyme (approved):
LGTM, left a few comments. Verified storybook line and rectangular annotation stories a still render and interact the same. 👍🏼

src/chart_types/xy_chart/annotations/types.ts
```diff
@@ -53,20 +54,21 @@ export interface AnnotationMarker {
...
 
 /** @internal */
 export interface AnnotationTooltipState {
   isVisible: true;
   annotationType: AnnotationType;
-  header?: string;
-  details?: string;
+  datum: LineAnnotationDatum | RectAnnotationDatum;
   anchor: {
+    // TODO this seems to be unused as it is never set internally
```
monfera, line 71: Might as well remove it then, while at it?

src/chart_types/xy_chart/renderer/dom/annotations/tooltip_content.tsx
markov00: TS complains about `header` not being an field in the union type `LineAnnotationDatum|RectAnnotationDatum`
A workaround for that is:
```
return <CustomTooltip details={details} header={(datum as LineAnnotationDatum).header} datum={datum} />;
```
but I'm not super happy about that or
```
return <CustomTooltip details={details} header={'header' in datum ? datum.header : undefined} datum={datum} />;
```

src/chart_types/xy_chart/annotations/types.ts
markov00: removed here 5919ef2

src/chart_types/xy_chart/annotations/line/dimensions.ts
markov00: done in https://github.com/elastic/elastic-charts/pull/960/commits/77518370eedf37a2d5e4eacdfb197d34146260d0

src/chart_types/xy_chart/state/selectors/get_annotation_tooltip_state.ts
markov00: fixed here: 7465bb3

src/chart_types/xy_chart/renderer/dom/annotations/annotations.tsx
markov00: fixed here: 7465bb3

src/chart_types/xy_chart/renderer/dom/annotations/annotations.tsx
markov00: fixed here: 7465bb3

src/chart_types/xy_chart/annotations/line/dimensions.ts
markov00: fixed here: 7465bb3

src/chart_types/xy_chart/annotations/line/dimensions.ts
markov00: fixed here: 7465bb3

monfera: 👍

src/chart_types/xy_chart/renderer/dom/annotations/annotations.tsx
monfera: 👍

src/chart_types/xy_chart/state/selectors/get_annotation_tooltip_state.ts
monfera: 👍

src/chart_types/xy_chart/annotations/types.ts
monfera: 👍

src/chart_types/xy_chart/annotations/line/dimensions.ts
monfera: 👍

src/chart_types/xy_chart/annotations/line/dimensions.ts
```diff
@@ -88,35 +89,33 @@ function computeYDomainLineAnnotationDimensions(
...
           anchorPosition,
           annotationValueYPosition,
+          markerDimensions,
         );
+
         const linePathPoints = getYLinePath({ width, height }, annotationValueYPosition);
 
         const annotationMarker: AnnotationMarker | undefined = marker
           ? {
               icon: marker,
```
monfera, line 101: (super minor, just thinking aloud) looks like this object creation is just property renaming; eg. possible to change `const lineColor` to `const color`, it adds to documentation value, though possible with a comment too; not sure which feels better, both are good 👍

src/chart_types/xy_chart/annotations/line/dimensions.ts
```diff
@@ -88,35 +89,33 @@ function computeYDomainLineAnnotationDimensions(
...
           chartRotation,
-          markerDimensions,
           anchorPosition,
           annotationValueYPosition,
+          markerDimensions,
         );
+
         const linePathPoints = getYLinePath({ width, height }, annotationValueYPosition);
 
         const annotationMarker: AnnotationMarker | undefined = marker
```
monfera, line 99: Apologies for the late comment in this regard—would it be possible or even useful to avoid a singleton case with runtype conditioins and type unioning by turning it into an array? Zero element: no annotation marker. One: same as when now one is used. Two: superimpose two, can help make custom or multicolor markers. Not necessarily in this PR, if at all, but maybe the opportunity is more feasible now than if we reify this as a `singleton | undefined`. Just acting on the notions we chatted about the other day 😃

markov00: done here: 10e6ca5

src/chart_types/xy_chart/annotations/line/dimensions.ts
markov00: 10e6ca5

src/chart_types/xy_chart/annotations/line/dimensions.ts
monfera: Thanks, looks great! This way it's also super easy to switch from eg. `icon` to `icons` whenever needed

monfera (approved):
Great and quite compact PR, thanks Marco!

Issue #964, opened 2021-01-06
[minor] Relax the `no-param-reassign` warning


monfera:
Currently, a stringent version is active. Ie. it doesn't just complain on parameter reassignment. It also complains if you want to set some inner property of what gets passed as a parameter.  Which is the intent, and often used, eg. when building tree or generally recursive structures, indices etc. [Example](https://github.com/elastic/elastic-charts/blob/f9218ad842d07a67eef4cbfb1209937db9da6853/src/chart_types/partition_chart/layout/utils/group_by_rollup.ts#L181) 

The `no-param-reassign` is a clear misnomer in this case. The word _parameter_ has precise meaning. In this case, it's the binding of the argument object to the name `mapNode`. The parameter is `mapNode`. That line does not reassign the parameter.

Would be great to deactivate it for the property update case and leave it in place for the reassignment of the binding eg. `const foo = x => { x++ }` though even such use isn't a huge deal in a library (popular libraries sometimes have `if(!Array.isArray(arg2)) arg2 = [arg2]` or some such)

It's shades of gray and subjective: my feel is that leaning on the stricter side with warnings and errors is great practice for web apps but feels less useful for library code. Read support for this kind of lib vs app distinction, maybe I'll run into it again. Eg. is the above referred code line in `group_by_rollup` problematic in data processing code that warrants a noise line? I'm OK if we are explicit about our goals and state that we're inching toward immutability; in this case we could chat about the goal, then put the rule in to help move toward the goal. The linter rule alone won't help much if this is the goal. Also, immutability has a price.

It also feels off to see a linter warning and not do something about it. It's nice to see the gutter with no discolorations, so I'll prefix that line with `// eslint-disable-next-line no-param-reassign` which is better than leaving in a warning (even if I currently disagree with the necessity for the warning for the stricter interpretation).

Some of the most popular libraries that do nontrivial things don't enforce a large fraction of the rules just because they are around. Often for some good reason, and they don't typically excuse them with noisy `// eslint-disable-next-line no-param-reassign` lines. Would be interested in how y'all see the webapp vs library use case, and why there should, or should not be a distinction.

monfera:
While it's definitely possible to work around the sometimes compounding linter warnings, which involve various pros and cons, stuff like this could be looked at together:

![image](https://user-images.githubusercontent.com/1548516/103769880-64cb7700-5025-11eb-9ea7-629b53b5adff.png)

[...]

Issue #957, opened 2020-12-28
Phase out some overly prescriptive code rules: `prefer-destructuring` for array and object


monfera:
I think
```
const [r0] = r;
```
is less preferable than
```
const r0 = r[0];
```
and even if not, it doesn't buy anything. Also, while we should [transition to ES2015](https://github.com/elastic/elastic-charts/issues/956), I don't see a reason for practically enforcing ES2015-only constructs, and currently, the resulting polyfill, where an ES5 construct exists which is often no worse and sometimes better.

I'm eager to learn if there's some justification. The rule description itself lists no benefits but points to clear drawbacks eg. when [indexing arrays](https://eslint.org/docs/rules/prefer-destructuring#when-not-to-use-it). I'm not fond of raising an info or warning for language constructs that seem perfectly legit, so such rules should ideally be activated upon shared conviction.

nickofthyme:
I think overall destructuring (both arrays and objects) is more beneficial from a reading perspective. But admittedly, it does get in the way at times, especially when reassigning values from an array.

```ts
// good
const [first, second] = array;

// bad
const first = array[0];
const second = array[1];
```

monfera:
I agree @nickofthyme and have no plans to discourage the array destructuring, which is often much nicer such as in your example. So my current takeaway is that the linter isn't quite able to judge when one is better than the other, and the developer or code reviewer can. If the linter can't know while we can work out OK usages in what has sometimes grey areas, it ideally isn't enforced one way or the other by the linter

nickofthyme:
Yup agreed

PR #965, opened 2021-01-06
feat(partition): Flame and icicle chart

monfera:
## Summary

Adds icicle chart as one of the `<Partition>` layout options next to treemap and sunburst
Also closes https://github.com/elastic/elastic-charts/issues/964
And closes https://github.com/elastic/elastic-charts/issues/957

image

image

Comparison of icicle/flame chart with treemap:
- icicle chart is suitable for at least a couple dozen layers, while treemap is not well readable past 2-3 layers; ideal is 1-2 layers
- icicle chart layout can be stable with the addition/removal of data, while treemap either gets relayouted (may jump) or sticks to original layout but suboptimal aspect ratio of subdivisions
- icicle chart can represent some dimension horizontally, eg. time (like flame chart) or can be alphabetically ordered aggregates (like flame graph) or even, like existing partitioning charts, in descending order of sizes
- lengths are easier to compare than areas

Comparison of icicle/flame chart with sunburst/pie:
- icicle chart is suitable for at least a couple dozen layers, while sunburst/pie is not well readable past 2-3 layers; ideal is 1 layers (pie) to 3 layers
- icicle chart represents values consistently and in a way that's easy to compare, while sunburst magnifies values (projected as sector arc lengths) as radius gets ever larger for less and less aggregated layers (this can be a pro or cons, depending)
- icicle chart has optimal text boxes (straight, horizontal), which also makes it better for future exploratory action eg. drilldown (eg. click on a bar to show only its descendants) - tweening is possible for text too
- icicle chart makes better use of the space, as the container is rectangular
- ordering of slices is more flexible than with pie/sunburst

In short, icicle, or its upright version, the flame chart, are preferable over pie, sunburst and treemap in many general situations as well as in observability eg. monitoring and APM. Due to its prevalence, user familiarity among would-be users is high.

Icicle vs flame: the choice is mostly down to layout on a dashboard; customs for the task or organization, and aesthetic choices.

Also tightens a few types and implementation bits (see the initial commits)

### Checklist

Todo, in a post `@alpha` PR:
- [ ] Either add flamegraph with this PR or remove references to it
- [ ] Image mocks with special cases
- [ ] Better text positioning
- [ ] (discuss) generalize and expose the `FlatTree` function, currently in the story, that transforms a nested JS object tree into the flat representation used by `<Partition>`


monfera:
retest

monfera:
jenkins retest this please

monfera:
jenkins retest this please

stories/icicle/01_unix_icicle.tsx
```diff
@@ -0,0 +1,116 @@
...
+  nodeLabel: (d: PrimitiveValue) => String(d),
+  showAccessor: (d: PrimitiveValue) => d !== undefined,
+  shape: {
+    fillColor: (d: ShapeTreeNode) => (d.dataName ? categoricalFillColor(color, 0.8)(depth) : 'transparent'),
+  },
+}));
+
+export const Example = () => {
+  return (
+    <Chart className="story-chart" size={{ height: 512 }}>
```
nickofthyme, line 81: Setting the height here makes the story look strange.

image

I created a new issue to deal with this in a separate PR. https://github.com/elastic/elastic-charts/issues/972

stories/icicle/01_unix_icicle.tsx
```diff
@@ -0,0 +1,116 @@
...
+          fillLabel: {
+            valueFormatter: (d: number) => d,
+            // fontStyle: 'italic',
+            textInvertible: true,
+            // fontWeight: 900,
+            valueFont: {
+              // fontFamily: 'Menlo',
+              // fontStyle: 'normal',
+              // fontWeight: 100,
+            },
```
nickofthyme, line 101: Do you need these comments still?

stories/icicle/02_unix_flame.tsx
```diff
@@ -0,0 +1,116 @@
...
+  // looks like our test runner can't run c.flatMap(...)
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  // eslint-disable-next-line prefer-spread
+  const childrenRows: Row[] = [].concat.apply(
+    [],
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    c.map((child) => flatTree(child, depth + 1)),
+  );
```
nickofthyme, line 54: How is this different than this?

```suggestion
  const childrenRows = c.reduce<Row[]>((a, child) => [...a, ...flatTree(child, depth + 1)], []);
```

> Then there's no eslint comments needed.

stories/icicle/02_unix_flame.tsx
```diff
@@ -0,0 +1,116 @@
...
+          fillLabel: {
+            valueFormatter: (d: number) => d,
+            // fontStyle: 'italic',
+            textInvertible: true,
+            // fontWeight: 900,
+            valueFont: {
+              // fontFamily: 'Menlo',
+              // fontStyle: 'normal',
+              // fontWeight: 100,
+            },
```
nickofthyme, line 101: Comments

stories/icicle/icicle.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { SB_SOURCE_PANEL } from '../utils/storybook';
+
+export default {
+  title: 'Flame (@alpha)',
```
nickofthyme, line 23: 🔥

stories/icicle/02_unix_flame.tsx
nickofthyme: We should figure out why `flatMap` is not working too cuz that would be very helpful.

stories/icicle/01_unix_icicle.tsx
monfera: Good idea, it's been a quirk, as a timeboxed attempt back then didn't solve it and we had a chat with @markov00 a long time ago; iirc it happens on a treemap mock too

stories/icicle/01_unix_icicle.tsx
monfera: I've been somewhat relaxed about leaving property uses present but commented out; let's chat if such lines should be cleared up in the numerous mocks where they may be present, it could be a follow-up PR with a bunch of other file changes

stories/icicle/02_unix_flame.tsx
monfera: Your version looks better in all regards; l'll create an issue for no `flatMap` in the runner

monfera:
@nickofthyme thanks for the review. I noticed the comparative slowness. The tree is 18 levels deep and has quite a few nodes (somewhat wide too);  and the legends becomes rather long too (lots of HTML elements). Planning to look into it in the alpha stage

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
```diff
@@ -140,14 +142,13 @@ function getRootArrayNode(): ArrayNode {
     [PATH_KEY]: [] as number[],
   };
   Object.assign(bootstrap, { [PARENT_KEY]: bootstrap });
-  const result: ArrayNode = bootstrap as ArrayNode;
-  return result;
+  return bootstrap as ArrayNode;
```
markov00, line 145: This is just a clean up on the PR, but seriously looking at that `getRootArrayNode` function, it doesn't return strictly an `ArrayNode`, the `STATISTICS_KEY` and `SORT_INDEX_KEY` are not available there. 
It's possible that for the root note these values are ignored in the current code, but that doesn't mean that we can easily remember that fact when refactoring/reworking on that file. It can leads to bug if we believe that also the root node is of type `ArrayNode`
Fixing the type here can go in two direction:
- add missing keys in the bootstrap object with some default values, declare the bootstrap as `Exclude<ArrayNode, parent`
- relax the ArrayNode type for the missing keys
- create a RootArrayNode type without these missing keys

This task doesn't need to be handled in this PR, but it's just an example of why ts ignoring or putting as can deal to issues.

src/chart_types/partition_chart/layout/viewmodel/viewmodel.ts
```diff
@@ -248,31 +303,34 @@ export function shapeViewModel(
...
     const diff = n.x1 - n.x0;
-    return treemapLayout
+    return treemapLayout || icicleLayout || flameLayout
       ? n.x1 - n.x0 > minFontSize && n.y1px - n.y0px > minFontSize
       : (diff < 0 ? TAU + diff : diff) * ringSectorMiddleRadius(n) > Math.max(minFontSize, linkLabel.maximumSection);
   };
 
   const nodesWithRoom = quadViewModel.filter(roomCondition);
-  const outsideFillNodes = fillOutside && !treemapLayout ? nodesWithRoom : [];
+  const outsideFillNodes = fillOutside && !treemapLayout && !icicleLayout && !flameLayout ? nodesWithRoom : [];
```
markov00, line 312: nit: Probably it's better if we change this to
```suggestion
  const outsideFillNodes = fillOutside && sunburstLayout ? nodesWithRoom : [];
```
In the case when a new layout is added, we don't have to think and remember about that in first place

src/chart_types/partition_chart/layout/viewmodel/viewmodel.ts
```diff
@@ -207,25 +262,25 @@ export function shapeViewModel(
...
-        x0: -width / 2,
-        y0: -height / 2,
-        width,
-        height,
-      })
-    : sunburst(tree, sunburstAreaAccessor, { x0: 0, y0: -1 }, clockwiseSectors, specialFirstInnermostSector);
+  const sunburstLayout = partitionLayout === PartitionLayout.sunburst;
+  const icicleLayout = partitionLayout === PartitionLayout.icicle;
+  const flameLayout = partitionLayout === PartitionLayout.flame;
+  const longestPath = ([, { children, path }]: ArrayEntry): number =>
```
markov00, line 269: you can also expose these as separate functions as I see multiple times these used also on other functions
`isTreemap(layout)`

stories/icicle/01_unix_icicle.tsx
markov00: One reason to remove those comments is that those types can change in the future, these commented out text could be outdated without noticing that

stories/icicle/01_unix_icicle.tsx
markov00: in the meantime can we remove the size and decrease the fontsize?

markov00 (approved):
Everything looks good, I suggest to rework the legend items sorting before waiting for the sorting PR I'm working on for two reasons:
- the legend sorting I'm working is implemented only for xy charts
- the legend items structure depends mostly on the data model that varies from chart type to chart type

src/chart_types/partition_chart/layout/viewmodel/viewmodel.ts
monfera: 👍

monfera:
Re https://github.com/elastic/elastic-charts/pull/965#discussion_r556466552: my main concern was the aspect ratio, as the current one is a wide yet very short landscape box, less suitable for icicle/flame. I now made the change as both of you mentioned this, maybe automated testing is cropping, or there's some other reason for avoiding height protrusion even at the expense of less human friendly mocks and losing regular font size, as both the height increase and the width decrease lead to an unusually shaped peach border

stories/icicle/01_unix_icicle.tsx
monfera: My main concern was the aspect ratio, as the current one is a wide yet very short landscape box, less suitable for icicle/flame. I now made the change as both of you mentioned this, maybe automated testing is cropping, or there's some other reason for avoiding height protrusion even at the expense of less human friendly mocks and losing regular font size, as both the height increase and the width decrease lead to an unusually shaped peach border

stories/icicle/01_unix_icicle.tsx
monfera: I made the change; getting out of sync wouldn't bother me, or lib developers, but indeed it'd be confusing for someone who only looks at the source sample in storybook (ie. Kibana developer, by and large)

stories/icicle/02_unix_flame.tsx
monfera: 👍

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
monfera: Thanks, good points!

monfera: Just pondering... the possible issue with points 1 and 3 is that it'll no longer be possible to model the entire tree as having homogeneous type; I'm sure it's solvable with type guards and such, with some more work.

With option 2, it'd relax the type for all uses, and I wouldn't want to water it up.

I think adding the missing properties to this root node would be a compact, local solution, but it'd still need the `as` assertion, because the `[PARENT_KEY]` prop demands an `ArrayNode`. 

So my current plan is to see what I can do locally (eg. just adding the missing props but leaving in place the `as`) and continue exploring and likely, discussing, outside this PR

monfera: OK the current solution is, use `Omit` to exclude `PARENT_KEY` in the assignment:
```
function getRootArrayNode(): ArrayNode {
  const children: HierarchyOfArrays = [];
  const bootstrap: Omit<ArrayNode, typeof PARENT_KEY> = {
    [AGGREGATE_KEY]: NaN,
    [DEPTH_KEY]: NaN,
    [CHILDREN_KEY]: children,
    [INPUT_KEY]: [] as number[],
    [PATH_KEY]: [] as number[],
    [SORT_INDEX_KEY]: 0,
    [STATISTICS_KEY]: { globalAggregate: 0 },
  };
  return { ...bootstrap, [PARENT_KEY]: bootstrap } as ArrayNode; // TS doesn't yet handle bootstrapping but the `Omit` above retains guarantee for all props except `[PARENT_KEY`
}
```
This way, there's still a need for `as`, because TS can't type-bootstrap a self-referential object, but it's way better, because should anything in `ArrayNode` change, the `bootstrap` object must reflect that. So the only part where the type checker is asked to trust the human is the presence of `PARENT_KEY` which is a local, single-line assignment right there

stories/icicle/01_unix_icicle.tsx
monfera: Ended up removing the height, doesn't look bad a57b697b6ae00134a2e9138610d1d0a7687348f7

stories/icicle/01_unix_icicle.tsx
monfera: a57b697b6ae00134a2e9138610d1d0a7687348f7

stories/icicle/02_unix_flame.tsx
monfera: a57b697b6ae00134a2e9138610d1d0a7687348f7

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
monfera: f9e51f9e90ed8cfb8128d5225a79c4606ab4953a

src/chart_types/partition_chart/layout/viewmodel/viewmodel.ts
monfera: 1fa058071467a6ca9aed3cb89454d0a1b54ed51f

stories/icicle/01_unix_icicle.tsx
nickofthyme: Yeah I see your point @monfera but that could be a slippery slope where we just have comments all over the place with some different than others.

I think we could achieve this same effect by creating some mock or story defaults file that came house a specific use case to be used anywhere. Then we can describe the use more explicitly.

nickofthyme (approved):

PR #968, opened 2021-01-12
chore: legend gap removal

monfera:
## Summary

Closes https://github.com/elastic/elastic-charts/issues/944
More important for the upcoming _hierarchical layout hover strategy_ PR and the [flame/icicle chart](https://github.com/elastic/elastic-charts/pull/965) than for existing charts

Before, with gaps, see how the activation is temporarily lost between rows, with all colors flashing back:
Screen Recording 2021-01-14 at 03 06 PM

After, no gaps, no flashing:
no gaps

It shifts images up by 1px or 2px, which seems OK, as it's important that `padding-top` and `padding-bottom` equal (text is in the vertical middle of the mouse capture zone)

### Checklist


nickofthyme:
I think this was working at one point, not 100% sure. This was discussed at length here https://github.com/elastic/elastic-charts/pull/749#discussion_r453747642 when implementing the legend actions.

I don't think this is the right solution for this as it is because it changes the aesthetic of the legend.

monfera:
Thanks @nickofthyme this PR doesn't change aesthetics perceptibly, except a 1-2px shift of the entire legends block upwards, but maybe this is what you referred to (CI detects it) - so it's great to go ahead as we discussed on our call

nickofthyme:
@monfera Checkout this commit d9722b6. I am sorry to force push to your branch but a made a copy just in case.

The trick is to add half of the padding to the `.echLegendItem` element and the other half to a `.background` element that is absolutely positioned behind the legend item. This way the look is the same but the interaction is smooth across all legend items. Now moving the cursor vertically over the legend items will always show an actively highlighted series.

Screen Recording 2021-01-12 at 04 35 PM

monfera:
@nickofthyme looks great, I agree with the `push -f` here too. Could you look into the two CI balks?

nickofthyme:
It looks like the legend overall was slightly shifted down. But other than that is the same ui.

Screen Recording 2021-01-13 at 12 33 PM

nickofthyme:
@miukimiu Could you take a look at these changes? Should be the same UI overall.

monfera:
@nickofthyme how should the still remaining CI issue be solved? Is it a question of regenerating some image, or something deeper? Should we, and/or @miukimiu look into removing the visual difference you show above? My slant is, seeing this merged soon as it's a small code change yet good usability improvement, and I'm not concerned about snapshot changes of a couple of pixels of legend position

nickofthyme:
@monfera I'll look into the small px shift.

nickofthyme (approved):

PR #971, opened 2021-01-12
fix(legend): create unique ids for dot icons

### LINKED ISSUE #970: Chart legend circle elements fail accessibility test as all have identical id
The `circle` elements used in chart legends all use the identical id of `dot-a`, which causes an accessibility issue, as revealed by the [axe](https://chrome.google.com/webstore/detail/axe-web-accessibility-tes/lhdoppojpmngadmnindnejefpokejbdd) web accessibility testing tool.

Can be seen on https://elastic.github.io/elastic-charts/?path=/story/legend--right, or for example:

image
 
This will block accessibility tests being added for https://github.com/elastic/kibana/issues/51456 on any pages which use an Elastic chart where the legend is visible.

rshen91:
## Summary

Fixes #970

I'm using the unique colors of the legend items appended to the string `dot` to make unique ids. Changing to a class vs id prevented the color icons from rendering in storybook. 

### Checklist


rshen91:
jenkins test this please

nickofthyme (approved):
I think this is good. This might come up again if the color is the same bug let's just assume that in unlikely.

PR #973, opened 2021-01-13
fix(legend): remove ids for circles

rshen91:
## Summary

Updating the changes in #971 

### Checklist


src/components/icons/assets/dot.tsx
```diff
@@ -31,13 +31,15 @@ export class DotIcon extends React.Component<IconComponentProps> {
...
-        </defs>
-        <g>
-          <use xlinkHref={`#dot-${color}`} />
-        </g>
+      <svg
+        width={16}
+        height={16}
+        viewBox="0 0 16 16"
+        xmlns="http://www.w3.org/2000/svg"
+        aria-labelledby={color}
```
markov00, line 39: we can remove that `aria-labelledby` because we don't have something to associated this with.
Using `color` there is wrong because labelledby accept a DOM element id

markov00 (changes requested):

src/components/icons/assets/dot.tsx
rshen91: 🙌  thank you - a5096c1

markov00 (approved):
Good to merge as soon as it is green

PR #975, opened 2021-01-13
docs: add histogram small multiples example

nickofthyme:
## Summary

Add example of small multiples with histogram using padding scale options.

Screen Recording 2021-01-13 at 02 14 PM

### Checklist


markov00 (approved):
All good except that I think it is a bit more clear if the `enable histogram mode` is on by default as the title refers to histogram and not bar paddings

Issue #959, opened 2020-12-29
[Debug state] Expose debug state for the swim lane chart 


darnautov:
**Is your feature request related to a problem? Please describe.**
`elsatic-charts` renders swimlane/heatmap with canvas, hence it's not possible to retrieve required data from DOM for functional tests. 

**Describe the solution you'd like**
Implement `getDebugState` for `HeatmapState` exposing the following info:

- X and Y axis data (rendered labels with applied formatting, grid size)
- Rendered cells
- Selected brush area

PR #976, opened 2021-01-14
feat: debug state for the heatmap chart 

darnautov:
## Summary

Resolves https://github.com/elastic/elastic-charts/issues/959 

- Adds debug state support for the heatmap/swimlane chart 
- Extends `DebugState` interface with heatmap related data under the `heatmap` property 
- Adds debug action and cell selection knobs to the storybook


src/chart_types/heatmap/state/selectors/get_debug_state.ts
```diff
@@ -0,0 +1,90 @@
...
+interface DebugStateAxes {
+  x: DebugStateAxis;
+  y: DebugStateAxis;
+}
+
+export type HeatmapDebugState = Pick<DebugState, 'legend'> & {
+  cells: CellDebug[];
+  selectedArea: { x: number; y: number; width: number; height: number } | null;
+  axes: DebugStateAxes;
+};
```
nickofthyme, line 47: We had talked about using a single `DebugState` type for all chart types to simplify the API. In that case, we would add these additional keys on the `DebugState` as optional.

@markov00 any thoughts?

markov00: I'm also for simplifying the API here if possible

darnautov: hey @nickofthyme @markov00 , I've updated the `DebugState` interface, please take a look at the recent changes

src/state/types.ts
```diff
@@ -78,6 +79,13 @@ export type DebugStateBar = DebugStateBase & {
   labels: any[];
 };
 
+type CellDebug = Pick<Cell, 'value' | 'formatted' | 'x' | 'y'> & { fill: string };
+
+type HeatmapDebugState = {
+  cells: CellDebug[];
+  selectedArea: { x: number; y: number; width: number; height: number } | null;
+};
```
nickofthyme, line 87: 👍🏼

nickofthyme (approved):
This is exactly what I was thinking. Thanks for the changes.

markov00 (approved):
Looks good to merge when Jenkins is green

PR #977, opened 2021-01-14
refactor: remove default small multiples value

nickofthyme:
## Summary

Currently, the only way to determine if the value of the small multiple exists is to compare it to the `DEFAULT_SINGLE_PANEL_SM_VALUE ` string which is not exported from the library. This pr removes this default value in place of `null`.

### Checklist


markov00:
I see this as a simplified version of the real solution. 
What if we can define these values as optional? I know that this will require much more work, but it looks much cleaner than checking if the value is equal to a brutal string we are using, what do you think @nickofthyme ?

nickofthyme:
Yeah I think that could be nicer. You are saying like set the value to `undefined` or `null`?

nickofthyme:
@markov00 Let me know what you think about these changes

src/chart_types/xy_chart/state/selectors/compute_small_multiple_scales.ts
```diff
@@ -68,6 +64,6 @@ export const computeSmallMultipleScalesSelector = createCachedSelector(
 
 function getScale(domain: Domain, maxRange: number, padding = DEFAULT_SM_PANEL_PADDING) {
   const singlePanelSmallMultiple = domain.length <= 1;
-  const defaultDomain = domain.length === 0 ? [DEFAULT_SINGLE_PANEL_SM_VALUE] : domain;
+  const defaultDomain = domain.length === 0 ? [null] : domain;
```
nickofthyme, line 67: @markov00 I'm not sure of the implications of doing this. It seems to work fine.

I thought of just returning a `null` but that would require a lot of type guards elsewhere.

markov00: In the series identifier the `smHorizontalAccessorValue` is already an optional param, can we just use undefined instead of null?

nickofthyme: cf3d4cf

src/chart_types/xy_chart/utils/series.ts
```diff
@@ -182,14 +182,12 @@ export function splitSeriesDataByAccessors(
     let sum = xValueSums.get(x) ?? 0;
 
     // extract small multiples aggregation values
-    const smH = smallMultiples?.horizontal?.by
-      ? smallMultiples.horizontal?.by(spec, datum)
-      : DEFAULT_SINGLE_PANEL_SM_VALUE;
+    const smH = smallMultiples?.horizontal?.by?.(spec, datum) ?? null;
```
markov00, line 185: nit: we can keep using undefined also here and 5 lines below
```suggestion
    const smH = smallMultiples?.horizontal?.by?.(spec, datum);
```

markov00 (approved):
All looks good, thanks for taking the time for that change.

src/chart_types/xy_chart/utils/series.ts
nickofthyme: 3c175e1

markov00: awesome, it just probably needs fixing the snapshot tests

markov00: good to merge on green

PR #978, opened 2021-01-14
feat(partition): legend hover options

monfera:
## Summary

Implements 6 strategies in place of the current single strategy:

- `key` (the current and default one): highlight all identically named items, no matter where they are
- `keyInLayer`: highlight all identically named items which are within the same depth (ring) as the hovered legend depth
- `node`: highlight exact match in the path only
- `path`: highlight members of the exact path; ie. exact match in the path, plus all its ancestors
- `nodeWithDescendants`: highlight exact match in the path, and everything that is its descendant in that branch
- `pathWithDescendants` (gif below): highlight exact match in the path, and everything that is its ancestor, or its descendant in that branch

strategy

Example use: 
```
<Settings
        showLegend
        flatLegend
        legendStrategy="pathWithDescendants"
        legendMaxDepth={maxDepth}
        theme={STORYBOOK_LIGHT_THEME}
      />
```

Closes https://github.com/elastic/elastic-charts/issues/982 in addition.

### Checklist

- [x] Improve on the API as needed
- [x] Extract the new API and commit
- [x] Fix the temporary way of getting `legendStrategy` from the global spec
- [x] Add `legendStrategy` to another mock, eg. the three-layered sunburst as on the image above


src/specs/settings.tsx
```diff
@@ -354,6 +355,10 @@ export interface SettingsSpec extends Spec {
    * Display the legend as a flat hierarchy
    */
   flatLegend?: boolean;
+  /**
+   * Display the legend as a flat hierarchy
```
monfera, line 359: text needs an update

src/state/actions/legend.ts
```diff
@@ -44,8 +50,9 @@ export interface ToggleDeselectSeriesAction {
 }
 
 /** @internal */
-export function onLegendItemOverAction(legendItemKey: string | null): LegendItemOverAction {
-  return { type: ON_LEGEND_ITEM_OVER, legendItemKey };
+export function onLegendItemOverAction(legendPath: LegendPath): LegendItemOverAction {
+  // todo remove obsoleted `legendItemKey`
```
monfera, line 54: remove or leave in place: opinion sought

src/commons/category.ts
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * A string key is used to uniquely identify categories
+ *
+ * todo: broaden it; some options:
+ *   - allow other values of `PrimitiveValue` type (now: string | number | null) but should add Symbol
+ *   - allow a descriptor object, eg. `{ key: PrimitiveValue, label: string }`
+ *   - allow an accessor that operates on the key, and maps it to a label
+ */
+export type CategoryKey = string;
+export type CategoryLabel = string;
```
monfera, line 29: The intent with series etc. typing is to start formalizing the types, even if eg. it's `string`, along semantic concepts. There was also a need to tweak some types for compatibility. Needs a discussion about where we're heading toward

src/chart_types/heatmap/state/selectors/get_grid_full_height.ts
```diff
@@ -54,7 +54,7 @@ export const getGridHeightParamsSelector = createCachedSelector(
     const xAxisHeight = visible ? fontSize : 0;
     const totalVerticalPadding = padding * 2;
     let legendHeight = 0;
-    if (showLegend && (legendPosition === Position.Top || legendPosition === Position.Bottom)) {
+    if (showLegend && isHorizontalLegend(legendPosition)) {
```
monfera, line 57: Just extracting out this repeated expression

src/chart_types/partition_chart/layout/types/viewmodel_types.ts
```diff
@@ -148,13 +150,13 @@ interface SectorGeomSpecY {
...
 
-export type DataName = any; // todo consider narrowing it to eg. primitives
+export type DataName = CategoryKey; // todo consider narrowing it to eg. primitives
 
 export interface ShapeTreeNode extends TreeNode, SectorGeomSpecY {
   yMidPx: Distance;
   depth: number;
   sortIndex: number;
-  path: number[];
+  path: LegendPath;
```
monfera, line 159: One key change is that it's not enough to just know the route (index within each layer), so now it's an array of `{index: number; value: CategoryKey}`

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
```diff
@@ -55,11 +57,12 @@ interface MapNode extends NodeDescriptor {
   [PARENT_KEY]?: ArrayNode;
 }
 
-export type PrimitiveValue = string | number | null; // there could be more but sufficient for now
-type Key = PrimitiveValue;
+/** @internal */
+export const hierarchyRootKey: Key = '';
```
monfera, line 61: This replaces multiple uses of `null` that needed a bit of trkbal knowledge, with no central binding

src/chart_types/partition_chart/state/selectors/get_highlighted_shapes.ts
```diff
@@ -18,22 +18,61 @@
...
-    if (!highlightedLegendItemKey) {
+export const legendHoverHighlightNodes = createCachedSelector(
+  [getSpecs, getHighlightedLegendItemPath, partitionGeometries],
+  (specs, highlightedLegendItemPath, geoms): QuadViewModel[] => {
+    if (highlightedLegendItemPath.length === 0) {
       return [];
     }
-    return geoms.quadViewModel.filter(({ dataName }) => dataName === highlightedLegendItemKey);
+    // eslint-disable-next-line no-underscore-dangle
+    const pickedLogic: LegendStrategy = (specs.__global__settings___ as SettingsSpec).legendStrategy ?? defaultStrategy;
```
monfera, line 75: Will need to look up proper use; not just the global settings but also the fact that TS, without the assertion, expects that the `legendStrategy` exists on `Spec` not on `SettingsSpec`, though other legend properties also live where I put them

src/components/legend/legend_item.tsx
```diff
@@ -63,11 +63,7 @@ export interface LegendItemProps {
   toggleDeselectSeriesAction: typeof onToggleDeselectSeriesAction;
 }
 
-/**
- * @internal
- * @param item
- * @param props
- */
+/** @internal */
```
monfera, line 66: I wasn't aware that we're doing JSDoc so deleted some that were out of sync with their definitions

src/state/reducers/interactions.ts
```diff
@@ -138,11 +138,14 @@ export function interactionsReducer(
...
         ...state,
         highlightedLegendItemKey: null,
+        highlightedLegendPath: [],
       };
     case ON_LEGEND_ITEM_OVER:
+      const { legendItemKey: highlightedLegendItemKey, legendPath: highlightedLegendPath } = action;
       return {
         ...state,
-        highlightedLegendItemKey: action.legendItemKey,
+        highlightedLegendItemKey,
```
monfera, line 147: So this can be obsoleted if we want, as the path has upwards compatible info

src/chart_types/partition_chart/state/selectors/get_highlighted_shapes.ts
monfera: Fixed it with guidance from @markov00 and some experimentation in latter day commits

src/chart_types/partition_chart/state/selectors/get_highlighted_shapes.ts
```diff
@@ -19,21 +19,58 @@
...
-import { QuadViewModel } from '../../layout/types/viewmodel_types';
+import { getSettingsSpecSelector } from '../../../../state/selectors/get_settings_specs';
+import { DataName, QuadViewModel } from '../../layout/types/viewmodel_types';
 import { partitionGeometries } from './geometries';
 
-const getHighlightedLegendItemKey = (state: GlobalChartState) => state.interactions.highlightedLegendItemKey;
+const getHighlightedLegendItemPath = (state: GlobalChartState) => state.interactions.highlightedLegendPath;
+
+const legendStrategies = Object.freeze({
+  node: (legendPath: LegendPath) => ({ path }: { path: LegendPath }) =>
```
nickofthyme, line 32: Can we add types here? Also, is `freeze` needed?

```suggestion
type LegendStrategyFn = (legendPath: LegendPath) => (shape: { path: LegendPath; dataName: DataName }) => boolean;
const legendStrategies: Record<LegendStrategy, LegendStrategyFn> = Object.freeze({
  node: (legendPath: LegendPath) => ({ path }: { path: LegendPath }) =>
```

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
nickofthyme: I wonder if this could just be something like `__root_key__` or something. Just a thought.

src/commons/legend.ts
```diff
@@ -18,18 +18,21 @@
...
 /** @internal */
-export type LegendItemChildId = string;
+export type LegendItemChildId = CategoryKey;
 
 /** @internal */
 export type LegendItem = {
   seriesIdentifier: SeriesIdentifier;
   childId?: LegendItemChildId;
   depth?: number;
+  path: LegendPath;
```
nickofthyme, line 33: What is the reason for adding this as required? It seems to create a lot of useless empty values on the other chart types.

src/commons/category.ts
nickofthyme: I think this typing should only be used in places where the value is not obvious such as a map.

```ts
// the key here in not obvious
const seriesMap = new Map<string, number>();

// key is now clear and explicit
type SeriesKey = string;
const seriesMap = new Map<SeriesKey, number>();
```

I'm not sure if this is the best use case here. Doing this consistently would get unwieldy quickly.

src/commons/series_id.ts
```diff
@@ -18,11 +18,12 @@
  */
 
 import { SpecId } from '../utils/ids';
+import { CategoryKey } from './category';
 
 /**
  * A string key used to uniquely identify a series
  */
-export type SeriesKey = string;
+export type SeriesKey = CategoryKey;
```
nickofthyme, line 26: If this is true why not just combine them into one?

src/components/legend/legend_item.tsx
nickofthyme: I don't really think this is used outside of the ide. We could make these auto-update when a better docs system is in place.

src/state/chart_state.ts
```diff
@@ -179,7 +181,8 @@ export interface PointerStates {
 /** @internal */
 export interface InteractionsState {
   pointer: PointerStates;
-  highlightedLegendItemKey: string | null;
+  highlightedLegendItemKey: PrimitiveValue | null;
```
nickofthyme, line 184:
```suggestion
  highlightedLegendItemKey: PrimitiveValue;
```

nickofthyme (changes requested):
Love this so much, it is so pleasant on the eyes to change the strategy and see all the options.

I have a few questions, suggestions and comments.

src/commons/category.ts
monfera: It's desirable to centralize shared declarations via such an alias, because, then changing it, eg. considering adding number or some other key type like symbol, it's trivial to get to the places that hinge on it. In the absence of this, one would change the definition in one place, eg. from `string` to `string | symbol` and the only way to see the impact is to change the declaration in one place, and then resolve type conflicts that likely arise in random other corners.

The `SeriesKey` would be a bit of a misnomer in this case, because
- yes, sometimes the legend labels are one series per color / legend item, but sometimes it's a single series, with multiple categories
- some charts, eg. all the partitioning charts, goal etc. don't have a notion of series, or we can consider them single-series, yet there are a still a bunch of Categories that go into the legend 

So `Category*` attempts to be something that unites things that were each declared locally, like SeriesKey, legend label/key  etc., while they are actually shared. It caused no problem in the past because we didn't build significant content into the legends data, but now we do, and a type mismatch arose.

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
monfera: Yes, it could be that too, Marco or Rachel, any preference? I think our runtime should rule out empty series labels or empty pie slice labels, as it'd lead to weird results for the user (unlabeled things, eg. a solitary unexplained color in the legend)

src/components/legend/legend_item.tsx
monfera: 👍

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
```diff
@@ -46,7 +48,7 @@ export interface ArrayNode extends NodeDescriptor {
   [CHILDREN_KEY]: HierarchyOfArrays;
   [PARENT_KEY]: ArrayNode;
   [SORT_INDEX_KEY]: number;
-  [PATH_KEY]: number[];
+  [PATH_KEY]: LegendPath;
```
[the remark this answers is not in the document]
monfera: We can replace all occurrences of these with the values here, with no change whatsoever to runtime or static type check, and the motive isn't that it's remnant of JavaScript, iirc the code was TS off the bat anyway.
```
export const AGGREGATE_KEY = 'value';
export const STATISTICS_KEY = 'statistics';
export const DEPTH_KEY = 'depth';
export const CHILDREN_KEY = 'children';
export const INPUT_KEY = 'inputIndex';
export const PARENT_KEY = 'parent';
export const SORT_INDEX_KEY = 'sortIndex';
export const PATH_KEY = 'path';
```
The benefit is the following. Certain generic data processing functions, eg. `groupByRollup`, are domain independent, ie. they're not semantically coupled to even data visualization. We might as well import such functions from a 3rd party utility, or we might consider extracting some generic data processing and other utilities into a separate package; even if we never want to do this, it's good practice to not lock in accidental things such as property names.

So, _not_ locking in particular field names is just a way of avoiding tight coupling between the data processing functions and their uses. Now it'd be fairly easy to use whatever field names.

It doesn't currently yield some kind of monetary or other short term benefit, though we could move such very generic functions into the collection of relational operators, so I think it's OK as it's not costly to keep it around this way. Btw. tree processing in partition charts may be the only such area that has this kind of keying.

Still, we can discuss removing them, probably for its own PR if it falls that way.

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
```diff
@@ -180,9 +183,9 @@ export function mapsToArrays(root: HierarchyOfMaps, sorter: NodeSorter | null):
     });
   }; // with the current algo, decreasing order is important
   const tree = groupByMap(root, getRootArrayNode());
-  const buildPaths = ([, mapNode]: ArrayEntry, currentPath: number[]) => {
-    const newPath = [...currentPath, mapNode[SORT_INDEX_KEY]];
-    mapNode[PATH_KEY] = newPath;
+  const buildPaths = ([key, mapNode]: ArrayEntry, currentPath: LegendPath) => {
+    const newPath = [...currentPath, { index: mapNode[SORT_INDEX_KEY], value: key }];
```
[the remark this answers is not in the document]
monfera: It's of structure `LegendPath` which uses `type LegendPathElement = { index: number; value: CategoryKey }` so calling it `key` wouldn't compile. As an alternative, I could call it `key` in `legendPathElement`, though it's not really a key in that context; the `index` is more akin to the key in that structure. Still, doable easily if there's a preference for renaming `value` to `key`

src/chart_types/partition_chart/state/selectors/get_highlighted_shapes.ts
```diff
@@ -19,21 +19,58 @@
...
+
+  pathWithDescendants: (legendPath: LegendPath) => ({ path }: { path: LegendPath }) =>
+    // highlight exact match in the path, and everything that is its ancestor, or its descendant in that branch
+    legendPath
+      .slice(0, path.length)
+      .every(({ index, value }, i) => index === path[i]?.index && value === path[i]?.value),
+});
+
+/** @public */
+export type LegendStrategy = 'node' | 'path' | 'key' | 'keyInLayer' | 'nodeWithDescendants' | 'pathWithDescendants'; // keyof typeof legendStrategies
```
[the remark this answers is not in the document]
monfera: Yes, possible, what are the pros & cons of that? I can start with the cons 😃 
- the single line became 10 lines
- there's much repetition and broilerplate: every prop name occurs twice, and `as const` occurs everywhere
- we make the API more complex by needing to export this object, and the user needs to use it, for no benefit I can currently think of
- it's a nit, but it's not an enum (we're not using `enum` anywhere); it's an enum-inspired object representation

In short, why multiply related code size for both ourselves and userland code, while also reducing the DRYness? I may overlook some technical benefits.

Even as it is, I'm not happy with the repetition of keys of the `legendStrategies` object, as it's not DRY, but one repetition feels better than two 😃 Would be great if some TS code transform could change `keyof typeof legendStrategies` to the corresponding union type of literals for cases when `legendStrategies` is unexported because then we could just say, 
```
export type LegendStrategy = keyof typeof legendStrategies;
```
without having to export `legendStrategies

src/commons/legend.ts
monfera: There's always some pros and cons with increasing genericity. Often, it's not that it's useless on other charts; one of my comments relating to the code is, do we want to remove the parallel, singleton structure. I think it's OK to use the generic, non-singleton version, even if it's always an array of 1 element, as there's no nesting, but can go either way. 

Even if it be separate between many-layered (eg. partition) and single-layered (eg. current xy) charts,  I don't feel that making it optional is necessarily the best, as it should be missing for one, and mandatory for the other.

In any case, currently the legend is generic; it's in the root, not under xy or partitioning, and a variety of charts use it, so there wasn't the option of making it just optional.

src/commons/series_id.ts
monfera: Because they may diverge, and their uses are different. One way of viewing things is from the pov of TS as it is: structural typing, which may be a bit of a transitional duck tape. The alternative pov is, preparation for nominal typing which is way more meaningful. 

For example, currently, we can add an angle in radians with an angle in degrees, b/c both are `number` and most of TS is structural (there's branding, classes etc. though).

src/state/chart_state.ts
monfera: good catch, will need look into it

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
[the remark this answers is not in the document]
monfera: Forgot about something else for the pro side: if we use literal property names eg. `thing.path`, then the key name occurrence is spread over the code base, and it's hard to find all occurrences.
- grepping isn't good, because, if the property name is common, like with these, it'll find all kinds of additional hits; we could offset it with too long and too specific property names, but then it's even more typing, and name obfuscation, and is still error prone
- TS helps, if all the uses are properly typed, and some type alias is used, but it's not trustworthy (due to structural typing; one `number` is just like another `number`) and can be too many indirections even when it works

In contrast, with the square bracket notation, you use your favorite shortcut to jump to the definition of the key eg. `export const PATH_KEY...` and you use another shortcut, and you get a list of *all* places where that property is used in the codebase, and nothing more. It's a bit like with functions: you can always jump to the function definition, and to the places where you use the function, unambiguously.

In theory.

Currently, we don't gain a lot of this benefit in `groupByRollup` where it has been around since inception, and there's occasional direct use eg. `thing.parent` etc. so it's watered up. It could be easily ruled out by using symbols, or symbol.for, but it's not worth it for `groupByRollup`.

In sum, it has some benefits, and minor runtime costs too, and it can go either way for `groupByRollup`.

api/charts.api.md
```diff
@@ -1,1970 +1,1974 @@
...
-// src/chart_types/heatmap/layout/types/config_types.ts:61:5 - (ae-forgotten-export) The symbol "TextBaseline" needs to be exported by the entry point index.d.ts
-// src/chart_types/partition_chart/layout/types/config_types.ts:128:5 - (ae-forgotten-export) The symbol "TimeMs" needs to be exported by the entry point index.d.ts
-// src/chart_types/partition_chart/layout/types/config_types.ts:129:5 - (ae-forgotten-export) The symbol "AnimKeyframe" needs to be exported by the entry point index.d.ts
-// src/chart_types/partition_chart/specs/index.ts:48:13 - (ae-forgotten-export) The symbol "NodeColorAccessor" needs to be exported by the entry point index.d.ts
-// src/commons/series_id.ts:39:3 - (ae-forgotten-export) The symbol "SeriesKey" needs to be exported by the entry point index.d.ts
-
-// (No @packageDocumentation comment for this package)
-
-```
+## API Report File for "@elastic/charts"
```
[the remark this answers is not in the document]
monfera: Indeed, weird, it's 3 very small changes, not this monstrous thing, re-committing as you suggest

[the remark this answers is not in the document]
monfera: Woohoo the re-committing worked, thanks 🎉

src/state/chart_state.ts
monfera: fb25f8fbd3cf37e7caa46e1209f2d5b725f1bf51

api/charts.api.md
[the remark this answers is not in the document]
monfera: 6262d2efefa5a40de45b9d88beebbe7aefceaab7

src/specs/settings.tsx
monfera: f50ffb51b96c1577c2d33725ac66a72c6688c419

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
nickofthyme: CC: @markov00 @rshen91

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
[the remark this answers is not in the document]
nickofthyme: Thanks for the thoughtful explanation.

I see some benefit but imagine doing this for all interfaces across the codebase for better searchability. That would be ridiculous. I think there could be a balance where it's beneficial.

As far as grepping issue, I think if we all be more strict on typing everything, this issue would go away and the code would be better for it.

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
monfera: Followed through on this suggestion, as we don't currently rule out the empty string ebbd65f81c39454f6c07c763889966308a73d40e

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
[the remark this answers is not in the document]
nickofthyme: Yeah just thinking to change `value` with `key` but if you think that is more in line with the context I'm fine with it.

src/chart_types/partition_chart/state/selectors/get_highlighted_shapes.ts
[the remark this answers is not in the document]
nickofthyme: Using "enums" in this case is more explicit and structured than having to set the string values. I think all your cons are really pros 😅 

- _the single line became 10 lines_ -- Yes but this is a non-issue with minification
- _there's much repetition and broilerplate: every prop name occurs twice, and as const occurs everywhere
we make the API more complex by needing to export this object, and the user needs to use it, for no benefit I can currently think of_ -- Yes but this also allows us to communicate the intention of each of the enum properties to the end user with makes the API BETTER not more complex. Otherwise, the end user would have to go into the codebase src to see what the options do beyond the simple string property.
- _it's a nit, but it's not an enum (we're not using enum anywhere); it's an enum-inspired object representation_  -- True but that's just because the `enum` support in babel is not yet sufficient.

What do you mean by repetition? The const enum exports a type and value under the same name making it easy to set it as a type or value.

src/commons/series_id.ts
nickofthyme: If that is the case why not just keep it as os?

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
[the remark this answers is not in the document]
monfera: > I think there could be a balance where it's beneficial.

I agree. In current `master`, the main (or only?) place is the data transforms in `group_by_rollup.ts` and that has been that way it was merged over a year ago, so its use isn't growing afaik.

> As far as grepping issue, I think if we all be more strict on typing everything, this issue would go away and the code would be better for it

The issue with relying on current TypeScript for this is that TS typing is structural (not really, but close enough for this point), and it's always possible for some places to redundantly define something, as long as the prop exists, and is the same name and value type, TS would have no complaint. With the move to nominal typing, we'll be able to have assurance. 

So, it'd look unusual if most of our code were full of square brackets for property names, but the technique may be fine for the few generic data processing utilities we have and that we might add to. We might want to deliberate it together and maybe even remove the current square brackets too.

> imagine doing this for all interfaces across the codebase for better searchability. That would be ridiculous.

Currently, most of our code, even where logic is quite general, is tightly bound to how we use it, in which case it'd be pointless indeed.

However, successful dataviz and data processing libraries such as D3 are doing something similar: you can specify accessor functions, as a way to tell your D3 function, which object property to reach for, as it may be some arbitrary dataset. Bostock wrote on it [here](https://bost.ocks.org/mike/chart/) 

The named object properties a much more compact, lighter weight, less opinionated way of doing the same thing, eg. it doesn't rely on the [module pattern](https://macwright.com/2012/06/04/the-module-pattern.html) ie. closures with getters/setters that underpins D3; it also doesn't rely on heavier weight constructs eg. `class`, or passing property names explicitly for each function.

So, in short, I believe that:
- when writing _generic_ data manipulation functions, ideally it works with arbitrary property names, even if we have one instance of naming scheme in the code
- this might be one of the most lighthanded ways of achieving that
- the cost isn't significant for the low number of such functions we have
- a tangible benefit, ie. not just a notion of generality, is the xref / lookup ability, which isn't assured yet by TS

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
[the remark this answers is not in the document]
monfera: Thanks, yes, indeed the content is often in a `key` role elsewhere, though not within this specific structure

src/chart_types/partition_chart/state/selectors/get_highlighted_shapes.ts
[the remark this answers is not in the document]
monfera: > this is a non-issue with minification

> What do you mean by repetition?

My concern wasn't line count or code size in the bundle, though even there it's bigger; I just don't currently see the value of the "enum" object, compared to the list of strings. Frankly it's tedious to type up the bigger one, given the choice I'd type less, for identical results 😅 

> Yes but this also allows us to communicate the intention of each of the enum properties to the end user with makes the API BETTER

> Otherwise, the end user would have to go into the codebase src to see what the options do beyond the simple string property

How does your alternative snippet with the `{ 'optionName': 'optionName' } ` communicate the intention of each option better than the current commit with `'optionName'`? Looks like the same amount of information; and all of them use strings (as the "enum" object keys are strings).

The string version autocompletes for library users just like the "enum" version because it's a union of literal strings. 

But the "enum" version involves the export of `LegendOptions` as a runtime object, not just as a type. I don't feel that increasing the API surface with an extra object, where a string suffices, is desirable unless it buys something I don't see yet.

Or maybe it's a hypothetical scenario where the "enum" prop keys would change, or become much longer, like `{ 'Highlight Node With All Descendants': 'nodeWithDescendants' } `? Please give me something to support that the "enum" is better for the library user.

> makes the API BETTER not more complex

Under what scenario is `legendOption: 'path'` more complex than `legendOptions: LegendOption.path`? 

> True but that's just because the enum support in babel is not yet sufficient

It may be for the better; I'm not super keen on adding code constructs that don't bring in a lot of value but bifurcate TS from JS. So it's not a critique for not using `enum`—avoidance on transpilation, polyfills etc. is generally a good thing.

It's a small change and I'm glad to switch to an object for library *consistency* (we use the "enum" pattern  a lot), or if it has tangible benefits; I'm trying to learn what the benefits are, so a specific example of the value would be neat.

src/commons/series_id.ts
monfera: Preparation for nominal typing; better documentation of the intent, and meaning of things.

And forced by this PR, because using words like _series_, _seriesKey_ etc. may translate well to legend items in case of most Cartesians, but it would've been misleading (though TS-wise, structurally passing) for partition and maybe other chart types. In case of partitioning charts, each slice/rectangle may get its own legend item, while there's no concept of series (technically, it's a singleton series).

Leaving it named `seriesKey` here and `legendKey` here, or calling both `strings` would of course have worked, because current TypeScript only cares about a few aspects of the structure (hey it's even okay with extra properties except in literals, so it's not even structurally tight). It's great if someone, versed in dataviz and TS, but not privvy of implementation detail, can get a good sense of what the library is about, just by reading its types.
 
It'd also be great to identify core, cross-cutting aspects of the library, to unify things along robust concepts. Our discussions, slides etc. and the domain of dataviz has a bunch of central concepts eg. categories, which are invariant, ie. totally not implementation detail. It's great to lean on solid concepts in the library too, and link it to meaningful terms of the art, like we do with words like axis, scale etc. Meanwhile, the concept of series is related, but not quite identical. 

Would be useful to chat about the topic of elevating such a concept at the type level on one of our regular meetings.

monfera: Oh btw. I might have misunderstood you; it'd be great to put `CategoryKey` or category related (or inherent dataviz related) words elsewhere too, but this PR avoids venturing too much without further discussion. Yet it starts the discussion and common thinking around this, so I didn't shy away from making an initial, localized step that doesn't lock us in and show my slant

.eslintrc.js
```diff
@@ -107,7 +107,7 @@ module.exports = {
     'no-bitwise': 0,
     'no-void': 0,
     yoda: 0,
-    'lines-between-class-members': ['error', 'always', { exceptAfterSingleLine: true }],
+    '@typescript-eslint/lines-between-class-members': ['error', 'always', { exceptAfterSingleLine: true }],
```
monfera, line 110: This isn't prefixed in master with `@typescript-eslint/` so the rule wasn't actually in effect

src/chart_types/partition_chart/state/selectors/get_highlighted_shapes.ts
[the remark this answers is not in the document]
markov00: Few pros about the current "enum" way of exposing constants:
- consistency: we used the real ts enum when we started the library, but then we switched to the "enum" objects pattern later on due to missing transpilation from babel. I rather prefer to do that change all our codebase at the same time, then expose a few things in a way and others in a different one. 
- type safety: we can easily change the value for those enums constants when needed without creating breaking changes for the user. if my enum has `{Path: 'path'}` but I need to change `path` to `path_key` I can do that without breaking changes.
- type safety 2: when used in JS environment, providing an enum object is a bit safer then let the user throw in unwanted strings 
- discoverability: with a union type of strings and in a TS environment you can discover the possible values only within the method that uses that union type. Instead, when using an enum object you can easily declare where ever you need your constant and keep using your IDE to discover what other values are available. 
 - discoverability 2: with an enum object you can enumerate the available options and expose them as they are if needed. This probably is not a very good practice, but for automated testing or within our storybook we can iterate over enum object keys and produce selection knobs easily without the need to refactor them on change
- documentation: you can document each element of the enum object easily, where I have never seen that for a union type.

This is basically very similar to the concept of having generic keys for our objects like you have discussed few comments above with the use of `[SOME_KEY]: value`

And, btw, no one blocks the user from using directly the enum value instead of passing from the object e.g.:
```
export const PartitionLayout = Object.freeze({
  sunburst: 'sunburst' as const,
  treemap: 'treemap' as const,
  icicle: 'icicle' as const,
  flame
})
// I can use the "standard" way
<Partition config={{ partitionLayout: PartitionLayout.sunburst }} />
// or the simplified, no TS error will be thrown
<Partition config={{ partitionLayout: 'sunburst' }} />
```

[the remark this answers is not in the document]
monfera: Thanks folks for the comments so far, I think the consistency argument trumps other aspects–if it's a convention and we don't use strings elsewhere, it's best to either stick to it or revise in bulk, which would be a big change for debatable benefit–, and the _discoverability 2_ item is something I didn't think of. Let's go through the rest at some point.

src/chart_types/partition_chart/state/selectors/get_highlighted_shapes.ts
monfera: Thanks, good catch and recommendation, I also shed the per object value type info with this
a53dd28395a05eafe52ebd09e8b1e2563ec5c6ee 
5ec5535d7a53af28250d716631619b6e4f3e0be1

src/chart_types/partition_chart/state/selectors/get_highlighted_shapes.ts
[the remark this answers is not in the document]
monfera: 8e6bd4590d8d896f61dbcbfddf7b875318b8903e

src/chart_types/partition_chart/state/selectors/get_highlighted_shapes.ts
```diff
@@ -18,22 +18,89 @@
...
+   */
+  PathWithDescendants: 'pathWithDescendants' as const,
+});
+
+/** @public */
+export type LegendStrategy = $Values<typeof LegendStrategy>;
+const defaultStrategy: LegendStrategy = LegendStrategy.Key;
 
 /** @internal */
 // why is it called highlighted... when it's a legend hover related thing, not a hover over the slices?
```
markov00, line 96: you can remove this comment as you actually renamed the function

src/specs/settings.tsx
```diff
@@ -354,6 +355,10 @@ export interface SettingsSpec extends Spec {
    * Display the legend as a flat hierarchy
    */
   flatLegend?: boolean;
+  /**
+   * Choose a partition highlighting strategy for hovering over legend items
```
markov00, line 359: Should we describe that actually is applied only on partition charts?

src/state/actions/legend.ts
markov00: we can keep it like that and remove it in a later stage

markov00 (approved):
Everything is fine for me, I've left just a few minor comments, that can easily be left as they are

src/specs/settings.tsx
monfera: It says, `Choose a partition highlighting strategy for hovering over legend items`, do you mean somewhere else?

src/chart_types/partition_chart/state/selectors/get_highlighted_shapes.ts
monfera: oops, will do 😄

src/state/reducers/interactions.ts
monfera: Based on Marco's feedback: consider it for later

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
```diff
@@ -55,11 +57,12 @@ interface MapNode extends NodeDescriptor {
   [PARENT_KEY]?: ArrayNode;
 }
 
-export type PrimitiveValue = string | number | null; // there could be more but sufficient for now
-type Key = PrimitiveValue;
+/** @internal */
+export const hierarchyRootKey: Key = '__root_key__';
```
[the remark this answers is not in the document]
monfera: fa82675619c006b215ef65697235bfbdb1b67749

src/chart_types/partition_chart/state/selectors/get_highlighted_shapes.ts
monfera: fa82675619c006b215ef65697235bfbdb1b67749

src/chart_types/partition_chart/state/selectors/get_highlighted_shapes.ts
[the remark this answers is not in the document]
nickofthyme: Yeah I can see you points Robert on the added surface of the API but my thought about it being clearer to the user was not `{ 'optionName': 'optionName' }` but rather the const _enum_ allows for each key in the object to be defined clearly, much like the `Fit` enum below. In this case we can add a larger more descriptive explanation of each option. So rather than just the string `carry` we can say `Use the previous non-`null` value `.

https://github.com/elastic/elastic-charts/blob/6c4dafd1cdeed5b61ca4c89790faa707b5c083b5/src/chart_types/xy_chart/utils/specs.ts#L174-L256

Regarding the use of one or the other -- that is the beauty of the const enum, is that it is available when needed and always replaceable by a string value. The const enum can also make the consumer code nicer when looping through all available options for a chart configuration option.

Lastly, these const enums act much like a typescript `enum`, and once support is updated, we can replace all of them to use the _real_ `enum` and have a much better end-user experience.

src/commons/legend.ts
nickofthyme: Ok thanks for the explanation!

src/commons/series_id.ts
nickofthyme: Yeah I was just referring to the coupling of the types here and if that was needed. I think we should discuss how/when we want to get away from the use of series from an API perspective. Fine with me for now 👍🏼

nickofthyme (approved):
LGTM, thanks for the changes.

PR #984, opened 2021-01-19
chore: minor naming unification

monfera:
## Summary

No functional or API change, just unifying some names, see by commit. Was going to slap it to a PR but it'd have made review difficult.

Both `common` and `commons` are legit terms, though the former is more common in current `master` and in most software repos, and these are local utils, so leaned toward the former, as we had both


wiki/overview.md
```diff
@@ -205,9 +205,9 @@ On the current `Visualize Editor`, you can stack or cluster in the following cas
 
 ### Multiple charts/Split charts/Small Multiples (phase 2)
 
-Small multiples are created using the `<SmallMultiples>` component, that takes multiple `<SplittedSeries>` component with the usual element inside. `<SplittedSeries>` can contain only `BarSeries` `AreaSeries` and `LineSeries` Axis and other configuration need to be configured outside the `SplittedSeries` element.
+Small multiples are created using the `<SmallMultiples>` component, that takes multiple `<splitSeries>` component with the usual element inside. `<splitSeries>` can contain only `BarSeries` `AreaSeries` and `LineSeries` Axis and other configuration need to be configured outside the `splitSeries` element.
```
monfera, line 208: I haven't found a `<SplittedSeries>` so this doc line needs an update in a real PR (coming up later) so I avoid carbon waste by not pushing a restore to uppercase

markov00: all good robert, this file should be updated anyway as this is not more the way you can create a small multiple

markov00 (approved):
Thanks for fixing my italianiglish :D
the .mdx files are not used yet so it is fine for me.
The same for the wiki overview file

PR #988, opened 2021-01-20
feat: add point shapes

### LINKED ISSUE #987: Enable different shapes for markers
Timelion replacement requires the ability to render the data points choosing the shape from a predefined set: circle, diamond, cross, square, triangle.

We need to add the ability to render at least those shapes and expose an API to allow change thems per series

Screenshot 2021-01-20 at 18 22 53

markov00:
## Summary

This PR adds the ability to choose the shape of the rendered point as per request for Timelion replacement.
Screenshot 2021-01-20 at 18 28 37
 

fix #987 

### Checklist


rshen91 (approved):
LGTM - tested locally on my mac. I like how this is set up to easily add other point shapes in the future in the theme. Defaulting to circle makes a lot of sense. 👍

nickofthyme (approved):
LGTM. I wonder how we could change the highlighter in the future based on the point shape type so we don't get this...

Screen Recording 2021-01-21 at 10 22 AM

monfera (approved):
LGTM 🎉

markov00:
> LGTM. I wonder how we could change the highlighter in the future based on the point shape type so we don't get this...
> 
> Screen Recording 2021-01-21 at 10 22 AM

@nickofthyme @monfera @rshen91  I've updated the PR after Nick's comment here.
I've matched the highlighter shape with the point shape.
I've also cleaned up the canvas rendering functions and created a set of reusable SVG paths.
We have now the ability to display also bubble charts with different shape sizes like:
Screenshot 2021-01-27 at 10 25 35
 or having a single series with customized shapes per point:
Screenshot 2021-01-27 at 10 31 10

I've also introduced a change on the current PointGeometry, computing the style override at the same time when computing the point position and scales, instead of computing it in the renderer.

monfera:
With the broad generalization introduced with the last commits, we, at some point, might want to move away from singleton legends, because each aesthetic channel may encode different categorical or quantitative data. Quick example:
image

(the size legend is rendered somewhat useless though, due to lack of decent comparability of sizes)

While the marker shape may be redundant encoding for line series, eg. for accessibility and discernibility in print (at the cost of making the lines noisier), scatterplots and connected scatterplots would be more inviting for independent channel assignment. 

So with that, the question is, what are the tradeoffs of introducing broad configurability of markers as an independent categorical-only channel with or without independent legends. Without this, the ability to vary markers may not be super useful, though users can just avoid configuring it.

src/utils/themes/theme.ts
```diff
@@ -306,6 +308,18 @@ export type DisplayValueStyle = Omit<TextStyle, 'fill' | 'fontSize'> & {
   };
 };
 
+export const PointShape = Object.freeze({
+  Circle: 'circle' as const,
+  Square: 'square' as const,
+  Diamond: 'diamond' as const,
+  Cross: 'cross' as const,
```
monfera, line 315: An oft used name is, eg. listed in [ggplot2](http://www.sthda.com/english/wiki/r-plot-pch-symbols-the-different-point-shapes-available-in-r), for `+`: plus; for `x`: cross

monfera (approved):
Good functional addition; asking for renaming "angled cross" to something standard, eg. cross, while renaming current cross to plus. Also, a couple of follow-up issues are warranted: 
- linking back to the independent legend point with examples
- using, or making it configurable, the markers in the legend, if redundant encoding is chosen

nickofthyme (approved):
LGTM. Nice job with the highlighter shapes, that looks much better!

I wonder if in the future we allow the user to pass a path, if we could scale that up and take the outline silhouette of the shape as the highlight geometry rather than just hardcoding each outlined shape. Do you think that would be possible for most shapes?

PR #991, opened 2021-01-25
refactor(partition): improving on chart events and callbacks

monfera:
## Summary

This PR 
- improves a bit on the way we access various properties of hierarchical nodes via callbacks we provide
- replaces a few `any`s and inferred types
- doesn't break the API in this regard, but adds to the API and prepares for some deprecation (glad to assist partition users in the change)
- adds more details for mouse events on the hierarchical data in partitioning, eg. `path`, `depth`
- might push a commit that adds functionality
- shared at this WIP stage to solicitate feedback
- ~~not complete; I've some further simplification plan, subject to discussion~~

### Checklist


src/chart_types/partition_chart/layout/types/types.ts
```diff
@@ -108,3 +108,17 @@ export interface Rectangle extends Origin {
...
+ *   - for recursively linking the parent ArrayNode to an ArrayNode (child) in the tree
+ *
+ * By extracting out the 'MODEL_KEY', we make the distinction clear, while the API, which depends on this, doesn't
+ * change. This makes an eventual API change a single-line change, assuming `[MODEL_KEY]` is used where needed, and just there
+ *
+ *  Todo:
+ *   - replace users' use of `s.parent` with `s[MODEL_KEY]` for the ShapeTreeNode -> ArrayNode access
+ *   - change MODEL_KEY to something other than 'parent' when it's done (might still be breaking change)
+ */
+export const MODEL_KEY = 'whatever';
```
monfera, line 124: This specific string `"whatever"` is of course just for while it's WIP; it'll be `parent` in the final form of this PR for API compat, but then the usage of `.parent` can/should be deprecated, to allow us to eventually rename it to eg. `model`, or even, provide direct access to the `model` props in the callback functions so this internal representation detail doesn't leak out (my preference, for flatness, not sure if in this PR or in the future)

src/chart_types/partition_chart/state/selectors/picked_shapes.test.ts
```diff
@@ -128,8 +148,27 @@ describe('Picked shapes selector', () => {
     expect(onClickListener.mock.calls[0][0]).toEqual([
       [
         [
-          { groupByRollup: 'b', value: 2 },
-          { groupByRollup: 'b', value: 1 },
+          {
+            groupByRollup: 'b',
+            value: 2,
+            depth: 1,
```
monfera, line 154: More props for the event callback!

stories/icicle/02_unix_flame.tsx
```diff
@@ -35,6 +35,10 @@ export const Example = () => {
         legendStrategy={LegendStrategy.PathWithDescendants}
         legendMaxDepth={maxDepth}
         theme={STORYBOOK_LIGHT_THEME}
+        onElementClick={(e) => {
+          // eslint-disable-next-line no-console
+          console.log(e);
```
monfera, line 40: WIP, to see the change in playground

src/chart_types/partition_chart/layout/types/viewmodel_types.ts
```diff
@@ -159,7 +159,7 @@ export interface ShapeTreeNode extends TreeNode, SectorGeomSpecY {
   path: LegendPath;
   dataName: DataName;
   value: number;
-  parent: ArrayNode;
+  [MODEL_KEY]: ArrayNode;
```
nickofthyme, line 162: That seems fine to me. Are you trying to avoid breaking changes? I'd say this would be breaking.

One way to avoid a breaking change would be to keep the `parent` key but add the `@deprecated` flag. Replace all usage in kibana with the `MODEL_KEY` string. Then later we could change the name but it would be a cleaner breaking change when we remove `parent`.

```suggestion
  /** @deprecated use `MODEL_KEY` instead */
  parent: ArrayNode;
  [MODEL_KEY]: ArrayNode;
```

monfera: Thanks Nick! The plan would be:

Full compatibility stage, no breaking API, eg. after merging this PR (`"whatever"` changed back to `"parent"`)
- we still call it `parent`
- everything still works and compliant with our API
- a new value appears in the API, `MODEL_KEY`

Deprecation stage: 
- we and users can start replace `.parent` with `[MODEL_KEY]`
- still compatible with the API as is; something gets a `@deprecated` tag, at least conceptually, as I think we can't have two identically named props at the same time, though haven't tried

Breaking change: 
- When nobody in at least Kibana uses `.parent` for the model, we make the breaking change of `const MODEL_KEY = 'model'`
- though it's a breaking change, it'll no longer impact known users due to our assist during deprecation, while other users simply need to deal with a changing API

I'm not sure if the proposed commit would still work, pls. indicate if that's the case, glad to try. At least from the diff, it's not clear what the deprecation line belongs to

monfera:
Cherrypicked commit https://github.com/elastic/elastic-charts/pull/991/commits/e1bef27c2e131e938d58e8410fa485fc9dc5464d is just tighter types, a rename and solving linter warnings with class methods that don't use `this`. I guess the logic is that it should only be a `function` if it's a class method; otherwise an arrow function is clearer as arrow functions have no access to `this`. Probably these are kept as methods rather than static methods or just plain functions for future extensibility.

src/chart_types/partition_chart/state/chart_state.tsx
```diff
@@ -54,66 +54,50 @@ export class PartitionState implements InternalChartState {
...
-  isInitialized(globalState: GlobalChartState) {
-    return getPieSpec(globalState) !== null ? InitStatus.Initialized : InitStatus.SpecNotInitialized;
-  }
+  isInitialized = (globalState: GlobalChartState) =>
+    getPartitionSpec(globalState) !== null ? InitStatus.Initialized : InitStatus.SpecNotInitialized;
 
-  isBrushAvailable() {
-    return false;
-  }
+  isBrushAvailable = () => false;
```
nickofthyme, line 60: I love fat arrow functions as much as the next guy/girl, but I tend to avoid this in classes unless the binding is absolutely necessary. The reasoning is that the methods are treated as function properties that are hoisted into the constructor upon transpilation.

```ts
// something like this...
class Test {
    m1 () {
        return 'test'
    }

    m2 = () => 'test';
}

// becomes this...
"use strict";
class Test {
    constructor() {
        this.m2 = () => 'test';
    }
    m1() {
        return 'test';
    }
}
```

see [playground](https://www.typescriptlang.org/play?#code/MYGwhgzhAEAqCmEAu0DeAoaXoFsCM0AFAJRqbYUBO8SArpQHbQDkSiSz5WAvul7gCZoAXiKlhAPhZtkzANzpuQA)

This is a good [medium article](https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1) on the matter. In brief...

> - The initialization of arrow functions in class properties are transpiled into the constructor.
> - Arrow functions in class properties won’t be in the prototype and we can’t call them with super.
> - Arrow functions in class properties are much slower than bound functions, and both are much slower than the usual functions.
> - You should only bind with .bind() or arrow function a method if you’re going to pass it around.

monfera: Thanks Nick, you made me compare JS vs TS classes, useful thing to learn about. 

I will revert this change, also for consistency's sake, as there's no team consensus about fat arrows being preferable. And mostly, to cut down on code transformations. It had to be merely my IDE that complained, not our linter. 

The easy choice of reversal is not really because of the above concerns, as, in order,: 
- non-function valued properties (strings, numbers etc.) I assume, are also transpiled into the constructor, as JS classes don't have any property "members", only TS classes (afaik); I don't currently see why it's a problem if they're in the constructor (maybe merely the next point)
- we don't have access to non-function valued properties in `super` either; and we don't do much with inheritance anyway
- speed isn't a measurable issue for us as we don't have hundreds of millions of method calls per second with these classes where the article shows a difference, and it may well have been optimized since 2017 (Chrome 62)
- their last prescription (we should...) seems not to be a supported advice, and I'm not aware of support for it either; only alludes at the speed arg

As my reason to switch back anyway is consistency, team preference and to minimize transpilation (as the article's reasons don't seem to apply to us) I wonder if we should phase out using non-function valued instance properties too, and if not, why.

Also, I'm not sure why these methods aren't declared static, do you know? I might turn them static and see what happens, unless you know of a reason not to do so. Would be good to discuss and lock in class conventions, maybe we can touch on it on tomorrow's call.

monfera: ... yup, non-function valued props are also hoisted: [Playground Link](https://www.typescriptlang.org/play?#code/MYGwhgzhAEAqCmEAu0DeAoaXoFsCM0AFAJRqbYUBO8SArpQHbQDkSiSz5WAvul7gCZoAXiKlhAPhZtkzANx8KOAMwjoAIgAS8ECAD26hdyA)

nickofthyme: Just to be clear, I **DO** prefer fat arrows functions with implicit returns but @markov00 prefers traditional `function` declarations. Just not as class methods _unless_ the binding is necessary, in which case maybe `@autobind` decorator is more performant.

Your reasoning seems fine to me.

I think you make a good point, most of the methods could be static but I think making them a non-static method provides greater flexibility across all the different chart types. That said I don't know what would be accessed on the chart state instance that isn't in the `globalState` object.

monfera: Thanks Nick for the added detail. FWIW I also prefer fat arrows, due to less clutter, and the modern/FP aspects it brings in, eg. no `this`, no `arguments` are possible.  Yet, indeed the function declarations feel more natural for class methods, will revert in an upcoming commit

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
```diff
@@ -60,6 +60,7 @@ interface MapNode extends NodeDescriptor {
 /** @internal */
 export const HIERARCHY_ROOT_KEY: Key = '__root_key__';
 
+/** @public */
 export type PrimitiveValue = string | number | null; // there could be more but sufficient for now
 type Key = CategoryKey;
```
nickofthyme, line 65: Not a huge fan of redefining types like this. But can be changed later.

nickofthyme (approved):
LGTM, provided the class methods are changed to non-fat arrow functions. 👍🏼

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
monfera: yes, let's chat

PR #995, opened 2021-01-28
feat(partition): drilldown

monfera:
## Summary

Adds optional drilldown inside flame and icicle charts, an almost universal interaction for data exploration in observability and monitoring including APM, where the numerous layers and often, granular breakdown, may otherwise hide interesting details.

It not only magnifies the ever current interesting part during exploration, but also provides the space for in-rectangle labels. Otherwise, such charts are largely unreadable, due to the large number of items, and the legend's inherent difficulty in coping with it.

It's also an example for direct manipulation, where some change is effected by directly interacting with the components or shapes in the chart, rather than via external configuration. So it fits into the family of tooltip, shape and legend hover highlight. 

drilldown

### Checklist

It can be enhanced with 
- animation (transitioning)
- alternative strategies similar to, or piggybacking on the legend options, ie. not just expand a node, but expand all identically named nodes (usually, the same monitored software function) in the same layer or in all layers
- elimination of specific names, or values (or value ranges) eg. via context menu, as with "Exclude..." in common EDA tools
- options for eg. hiding the 100% bottom layers, and if so, maybe initially showing a determinate number of layers eg. 10
- better text formatting
- reset options, eg. a transient reset button for when the user eliminated some branches
- an indicator that tells the user that it's not currently showing all the data  (maybe combined with the reset button)
- maybe showing the currently selected path eg. above the legend


src/state/reducers/interactions.ts
```diff
@@ -79,8 +82,10 @@ export function interactionsReducer(
         },
       };
     case ON_MOUSE_DOWN:
+      const layerValues: LayerValue[] = getPickedShapesLayerValues(globalState)[0];
```
monfera, line 85: I'm not 100% sure this is the best approach, though it's certainly compact, and fully reuses the existing data flow subgraph

src/state/chart_state.ts
```diff
@@ -186,6 +187,7 @@ export interface InteractionsState {
   highlightedLegendPath: LegendPath;
   deselectedDataSeries: SeriesIdentifier[];
   hoveredDOMElement: DOMElement | null;
+  drilldown: CategoryKey[];
```
monfera, line 190: Maybe it should be named `drilldownSelection` here too, and below, so it's not confused with the boolean `drilldown` that's present in the config in the API

src/state/reducers/interactions.ts
```diff
@@ -30,8 +32,8 @@ import {
   ON_TOGGLE_DESELECT_SERIES,
   ToggleDeselectSeriesAction,
 } from '../actions/legend';
-import { MouseActions, ON_MOUSE_DOWN, ON_MOUSE_UP, ON_POINTER_MOVE } from '../actions/mouse';
-import { InteractionsState } from '../chart_state';
+import { ON_MOUSE_DOWN, ON_MOUSE_UP, ON_POINTER_MOVE, MouseActions } from '../actions/mouse';
```
monfera, line 35: No PR is complete without mysteriously reshuffling imports. Glad to undo!

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
```diff
@@ -108,33 +108,44 @@ export function groupByRollup(
...
-        [DEPTH_KEY]: i,
-        ...(!last && { [CHILDREN_KEY]: childrenMap }),
+    keyAccessors
+      .filter(
+        () =>
+          !drilldown ||
+          keyAccessors
+            .slice(0, drilldownSelection.length)
+            .map((ka) => ka(n, index))
+            .join(' | ') === drilldownSelection.slice(0, drilldownSelection.length).join(' | '),
```
monfera, line 127: Planning to replace it with a semishallow array element comparator, so the string based path check is temporary

src/chart_types/partition_chart/state/selectors/tree.ts
```diff
@@ -20,33 +20,43 @@
...
 const getSpecs = (state: GlobalChartState) => state.specs;
 
+const getDrilldownSelection = (state: GlobalChartState) => state.interactions.drilldown || [];
+
 /** @internal */
 export const getTree = createCachedSelector(
-  [getSpecs],
-  (specs): HierarchyOfArrays => {
+  [getSpecs, getDrilldownSelection],
+  (specs: SpecList, drilldownSelection: CategoryKey[]): HierarchyOfArrays => {
```
nickofthyme, line 40: This typing is inferred from the selectors.
```suggestion
  (specs, drilldownSelection): HierarchyOfArrays => {
```

src/chart_types/partition_chart/layout/utils/group_by_rollup.ts
```diff
@@ -108,33 +108,44 @@ export function groupByRollup(
...
-        [INPUT_KEY]: [...inputIndices, index],
-        [DEPTH_KEY]: i,
-        ...(!last && { [CHILDREN_KEY]: childrenMap }),
+    keyAccessors
+      .filter(
+        () =>
+          !drilldown ||
+          keyAccessors
+            .slice(0, drilldownSelection.length)
+            .map((ka) => ka(n, index))
```
nickofthyme, line 126: nit for consistency 
```suggestion
            .map((keyAccessor) => keyAccessor(n, index))
```

nickofthyme (commented):
Code changes LGTM, still need to play with it some later.

monfera:
jenkins retest this please

monfera:
@markov00 https://github.com/elastic/elastic-charts/pull/995/commits/7e32a70e56b715cdc013195d34f070d9f58485d2 addresses your legit concern that the interaction reducer shouldn't directly depend on chart type dependent functions or selectors, so I extracted it out, making distant future extensibility easier

src/chart_types/partition_chart/layout/types/config_types.ts
```diff
@@ -116,14 +118,19 @@ export interface StaticConfig extends FillFontSizeRange {
 export type EasingFunction = (x: Ratio) => Ratio;
 
 export interface AnimKeyframe {
+  /** @alpha */
```
markov00, line 121: You can export the whole interface as public/alpha

src/state/chart_state.ts
markov00: it's fine to callit `drilldown`

src/state/reducers/interactions.ts
```diff
@@ -199,3 +202,11 @@ function toggleDeselectedDataSeries(
...
   return [...deselectedDataSeries, id];
 }
+
+function getDrilldownData(globalState: GlobalChartState) {
+  if (globalState.chartType !== ChartTypes.Partition) {
+    return [];
+  }
+  const layerValues: LayerValue[] = getPickedShapesLayerValues(globalState)[0];
+  return layerValues ? layerValues[layerValues.length - 1].path.map((n) => n.value) : [];
+}
```
markov00, line 212: IMHO I think we can reduce the unnecessary call to `getPickedShapesLayerValues` if we follow the same principle applied to clicks and other events (check the `createOnElementClickCaller`), building a specific "state" Selector that tracks the click and call the various functions only if the configuration and/or callback listeners are configured

monfera: Thanks @markov00; what is going on in the `createOnElementClickCaller` and the few similar things was on our list of topics today 👍 though for another reason: we could eventually move the `captured variables `prev` and `selector` into the `InternalChartState` objects, removing one type of special state management place. Let's chat, I could use an explanation for a clearer idea behind those generated selectors

src/chart_types/partition_chart/layout/types/config_types.ts
monfera: done, in the merge commit, here's the link to this file within that commit: [link](https://github.com/elastic/elastic-charts/pull/995/commits/e02960ac6b04caf7d27094c0b377cac41113f3df#diff-ab853368fd9bc73166d0839e54efb4f483ce3293c663d91e477ae8e10c214842)

src/state/reducers/interactions.ts
markov00: I've checked and we can't apply the caller pattern here. Ok to leave this as it is for now until we don't find a good global strategy for handling events

src/chart_types/partition_chart/layout/types/config_types.ts
```diff
@@ -115,13 +117,15 @@ export interface StaticConfig extends FillFontSizeRange {
...
 
+/** @alpha */
 export interface AnimKeyframe {
   time: number;
   easingFunction: EasingFunction;
   keyframeConfig: Partial<StaticConfig>;
 }
 
 export interface Config extends StaticConfig {
+  /** @alpha */
```
markov00, line 128: Moving things back to an alpha version after exporting this as public is not a good practice (we should mark this as breaking change) but I don't think anyone is using that config at the moment so we leave with it now

markov00 (approved):
Good to merge for me!

PR #996, opened 2021-01-29
fix: add theme min radius to point shape

markov00:
## Summary

I've fixed now the minimum radius size for point is computed: only in bubble chart  I'd like to keep the theme point size as the minimum radius value to be used in bubble chart, to avoid very small points rendered on chart.

### Checklist


nickofthyme (approved):
LGTM

PR #997, opened 2021-01-29
build: upgade node to 14

markov00:
## Summary

Upgrading node to latest LTS one (also used by Kibana)

I've removed the `engines` requirement from the `package.json` because that engine is only required for the developer environment and not when using the library as a dependency on another app.

Please update your local version of `node` with
```
nvm install 14.15.4
// to make it the default one
nvm alias default 14

```


rshen91 (approved):
LGTM thanks!

nickofthyme (approved):

nickofthyme:
Looks like you need to update the `.ci/global_setup.sh` file

markov00:
jenkins test this

PR #993, opened 2021-01-26
fix: external tooltip legend extra value sync

nickofthyme:
## Summary
fix #992

### Before
Screen Recording 2021-01-26 at 01 04 PM

### After
Screen Recording 2021-01-26 at 12 31 PM

### Checklist

- [x] Any consumer-facing exports were added to `src/index.ts` (and stories only import from `../src` except for test data & storybook)
- [x] Proper documentation or storybook story was added for features that require explanation or tutorials
- [x] Unit tests were updated or added to match the most common scenarios

PR #1003, opened 2021-02-01
fix: external tooltip legend extra value sync (#993) [24.4.x]

nickofthyme:
Backports the following commits to 24.4.x:
 - fix: external tooltip legend extra value sync (#993)


markov00 (approved):
All good to me

Issue #891, opened 2020-11-05
[small multiples] Render axis on every panel


markov00:
**Describe the solution you'd like**
We should let the user configure the ability to show axes on every panel.
This should be in some degree configurable:
- when showing only the main set of axes at the edges of the small multiples grid we should also display by default just the vertical and horizontal axis line
- showing the axis labels on every panel as an option (reconsidering the current padding percentage option to accommodate longer text in the inner panels of the grid)

PR #1004, opened 2021-02-02
feat(axis): small multiples axis improvements

### LINKED ISSUE #985: Allow axis label accessor for small multiples
Using small multiples, the only way to change the axis labels is to change the split accessor. It would be great to have a way to set the value for the axis label across all panels.

A prop on the `Axis` component to set the label for each panel.

Setting split chart accessor to the desired label.

image

### LINKED ISSUE #986: Allow secondary axis label
Allow secondary axis label for small multiples

A prop on `Axis` to set secondary axis label.

image

nickofthyme:
## Summary

fixes #985, fixes #986

Related to #891

This PR adds functionality around the axis for small multiples including the following...
- Allow a single main axis to spread the length of all panels defined by `Axis.title`
- Enable axis title for each panel  with own style `AxisPanelTitle`
- Show axis lines on all non-primary (aka `secondary`) axes
- Add title prop to GroupBy for formatting panel titles by value
- Simplify title dimensioning logic
- Fix check for small multiples with single bucket issue related to [this](https://github.com/nickofthyme/elastic-charts/blob/d28820858d013326b3c660381e70696e9382166d/src/chart_types/xy_chart/state/selectors/compute_per_panel_axes_geoms.ts#L56-L58) code.
- Fix bug with sm value of zero, where a falsy `smV` or `smH` value would not render the series correctly.

## Demo

Screen Recording 2021-02-01 at 08 27 PM

Screen Recording 2021-02-01 at 08 29 PM

### Checklist


monfera:
Good additions and improvements here!

> Allow a single main axis to spread the length of all panels defined by Axis.title

It refers to a single text label only, ie. not a full axis, so no line, ticks and tick values, correct? Though I guess the per row category labels can be seen as tick values of the outside axis

stories/small_multiples/2_vertical_areas.tsx
```diff
@@ -54,14 +54,14 @@ export const Example = () => {
...
-        title="timestamp"
+        title="Timestamp"
         position={Position.Bottom}
         gridLine={{ visible: false }}
         tickFormat={timeFormatter(niceTimeFormatByDay(numOfDays))}
       />
       <Axis
         id="y"
-        title="metric"
+        title="Metric"
```
monfera, line 64: The empty string is handled as if there were no title at all; I think it's nice, whether it was the intent or it's just due to the empty string being falsey. Probably there's no situation when the space needs to be reserved with no text, and even there, a space would work

api/charts.api.md
```diff
@@ -936,12 +938,16 @@ export type GroupBySort = Predicate;
...
-    // (undocumented)
     by: GroupByAccessor;
-    // (undocumented)
     sort: GroupBySort;
+    title?: GroupByTitleFormatter;
 }
 
+// Warning: (ae-incompatible-release-tags) The symbol "GroupByTitleFormatter" is marked as @public, but its signature references "GroupByAccessor" which is marked as @alpha
+//
+// @public (undocumented)
```
monfera, line 948: Nice to see more things documented, maybe this one could be covered too

src/chart_types/xy_chart/renderer/canvas/axes/global_title.ts
```diff
@@ -41,93 +46,89 @@ export function renderTitle(ctx: CanvasRenderingContext2D, props: AxisProps) {
...
     size: { height },
-    axisSpec: { position, hide: hideAxis },
+    axisSpec: { position, hide: hideAxis, title },
     dimension: { maxLabelBboxWidth },
-    axisStyle: { axisTitle, tickLine, tickLabel },
+    axisStyle: { axisTitle, axisPanelTitle, tickLine, tickLabel },
+    anchorPoint,
     debug,
-    title,
+    panelTitle,
```
monfera, line 57: I ❤️ such informative names

src/chart_types/xy_chart/renderer/canvas/axes/panel_title.ts
```diff
@@ -0,0 +1,149 @@
...
+export function renderPanelTitle(ctx: CanvasRenderingContext2D, props: PanelTitleProps) {
+  const {
+    axisSpec: { position },
+    axisStyle: { axisPanelTitle },
+    panelTitle,
+  } = props;
+  if (!panelTitle || !axisPanelTitle.visible) {
+    return null;
+  }
+  if (isHorizontalAxis(position)) {
```
monfera, line 42: might be 
```
return isHorizontalAxis(position) ? renderHorizontalTitle(ctx, props) : renderVerticalTitle(ctx, props)
```
in part b/c our linter insists (I think not ideally) on no `else` branch if the `then` branch returns, which makes it asymmetric.

or
```
const renderTitle = isHorizontalAxis(position) ? renderHorizontalTitle : renderVerticalTitle;
return renderTitle(ctx, props);
```

src/chart_types/xy_chart/utils/axis_type_utils.ts
```diff
@@ -40,12 +40,12 @@ export function isBounded(domain: Partial<CompleteBoundedDomain>): domain is Dom
 }
 
 /** @internal */
-export function isVerticalAxis(axisPosition: Position) {
+export function isVerticalAxis(axisPosition: Position): axisPosition is Extract<Position, 'left' | 'right'> {
```
monfera, line 43: 👍

src/specs/group_by.ts
```diff
@@ -29,10 +29,26 @@ export type GroupByAccessor = (spec: Spec, datum: any) => string | number;
 /** @alpha */
 export type GroupBySort = Predicate;
 
+/** @public */
+export type GroupByTitleFormatter = (value: ReturnType<GroupByAccessor>) => string;
+
 /** @alpha */
 export interface GroupBySpec extends Spec {
+  /**
```
monfera, line 37: ❤️

src/utils/common.ts
```diff
@@ -545,3 +545,18 @@ export const getPercentageValue = <T>(ratio: string | number, relativeValue: num
  * @example [1, 2, 4, 2, 4, 0, 3, 2].filter(keepDistinct) ==> [1, 2, 4, 0, 3]
  */
 export const keepDistinct = <T>(d: T, i: number, a: T[]): boolean => a.indexOf(d) === i;
+
+/**
+ * Safely format values with error handling
+ */
+export const safeFormat = <V = any>(value: V, formatter?: (value: V) => string): string => {
+  if (formatter) {
+    try {
```
monfera, line 554: `try`/`catch` causes a deoptimization of the function, at least, it was the case last time I checked it in Chrome. It may not have measurable effect if the `formatter` function does _not_ get inlined; I'm unsure about whether it'd happen. Of course, if the circumstances are such that an error can arise, and we can't check it beforehand outside the hot inner loop, there's no other option

monfera (approved):
Looks great! A couple of smaller optional changes linked to rows

stories/small_multiples/2_vertical_areas.tsx
nickofthyme: Yes this was intentional, it used to be `title !== undefined` but I agree with you. Also if people want to just the spacing that can be handled via the padding on a visible element in the axis.

src/chart_types/xy_chart/renderer/canvas/axes/global_title.ts
nickofthyme: Yeah thanks @markov00 for the `panel` naming! ❤️

src/utils/common.ts
nickofthyme: Yeah good point. I think we could try this on the sm formatted as it is fewer interactions than saw the label tick formatted. Before adding this to other formatters we gound check the performance.

api/charts.api.md
nickofthyme: Ya, I wish you could document the type, and then the property would also pickup that definition but not the case. Added a definition here 7155cf6.

src/chart_types/xy_chart/renderer/canvas/axes/panel_title.ts
```diff
@@ -0,0 +1,149 @@
...
+    return renderHorizontalTitle(ctx, props);
+  }
+  return renderVerticalTitle(ctx, props);
```
nickofthyme, line 45: Good eye 👁️ 

```suggestion
  return isHorizontalAxis(position) ? renderHorizontalTitle(ctx, props) : renderVerticalTitle(ctx, props);
```

src/chart_types/xy_chart/renderer/canvas/axes/panel_title.ts
nickofthyme: See 67db527

markov00 (commented):
Left few comments, in particular I'd like to discuss the `title` prop

stories/small_multiples/2_vertical_areas.tsx
monfera: I don't think trim is needed

stories/small_multiples/4_horizontal_bars.tsx
```diff
@@ -75,8 +75,8 @@ export const Example = () => {
   return (
     <Chart className="story-chart">
       <Settings onElementClick={onElementClick} rotation={90} showLegend={showLegend} />
-      <Axis id="time" title="metric" position={Position.Bottom} gridLine={{ visible: false }} />
-      <Axis id="y" title="Day of week" position={Position.Left} gridLine={{ visible: false }} />
+      <Axis id="time" position={Position.Bottom} gridLine={{ visible: false }} />
+      <Axis id="y" title="Day of week" position={Position.Right} gridLine={{ visible: false }} />
```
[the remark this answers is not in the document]
nickofthyme: 60ab456

src/specs/group_by.ts
```diff
@@ -29,10 +29,29 @@ export type GroupByAccessor = (spec: Spec, datum: any) => string | number;
...
+   * Sort predicate used to sort grouped data
+   */
   sort: GroupBySort;
+  /**
+   * Title for each panel. Use to display unique title for panel.
+   *
+   * If both axis title and this value are set, the axis title will be
+   * treated as the primary title and this as the secondary.
+   */
+  title?: GroupByTitleFormatter;
```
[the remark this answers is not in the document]
nickofthyme: 60ab456

PR #1005, opened 2021-02-02
fix: missing exported types

markov00:
## Summary

Export some missing types used in Kibana but not exported from the main module

### DIFF
--- a/src/chart_types/heatmap/specs/heatmap.ts
+++ b/src/chart_types/heatmap/specs/heatmap.ts
@@ -72,11 +72,10 @@ export interface HeatmapSpec extends Spec {
   name?: string;
 }
 
-type SpecRequiredProps = Pick<HeatmapSpec, 'id' | 'data'>;
-type SpecOptionalProps = Partial<Omit<HeatmapSpec, 'chartType' | 'specType' | 'id' | 'data'>>;
-
 /** @alpha */
-export const Heatmap: React.FunctionComponent<SpecRequiredProps & SpecOptionalProps> = getConnect()(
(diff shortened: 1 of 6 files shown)


src/chart_types/heatmap/specs/heatmap.ts
```diff
@@ -72,11 +72,10 @@ export interface HeatmapSpec extends Spec {
...
 }
 
-type SpecRequiredProps = Pick<HeatmapSpec, 'id' | 'data'>;
-type SpecOptionalProps = Partial<Omit<HeatmapSpec, 'chartType' | 'specType' | 'id' | 'data'>>;
-
 /** @alpha */
-export const Heatmap: React.FunctionComponent<SpecRequiredProps & SpecOptionalProps> = getConnect()(
+export const Heatmap: React.FunctionComponent<
+  Pick<HeatmapSpec, 'id' | 'data'> & Partial<Omit<HeatmapSpec, 'chartType' | 'specType' | 'id' | 'data'>>
+> = getConnect()(
```
markov00, line 78: Here I've tried to remove the type alias to avoid exporting an alias type.
We should do the same for all the other Components

src/chart_types/heatmap/specs/index.ts
```diff
@@ -17,4 +17,4 @@
  * under the License.
  */
 
-export { Heatmap, HeatmapSpec } from './heatmap';
+export * from './heatmap';
```
nickofthyme, line 20: 😃

nickofthyme (approved):
LGTM!

PR #1015, opened 2021-02-05
chore: add favicon

flash1293:
## Summary

This PR adds a favicon to the storybook as well as to the readme.
Screenshot 2021-02-05 at 09 45 20

This is how the readme looks (it doesn't work in the PR preview because the URL doesn't exist yet, but automatic rewrite doesn't work in Github markdown within HTML - it will work fine once merged)
Screenshot 2021-02-05 at 09 53 43

To test, I had to switch to an incognito tab because Chrome aggressively caches favicons.


miukimiu (approved):
Wow @flash1293, this looks amazing! Thanks for adding this. 🎉

markov00 (approved):
Thanks for that! Looks great!

PR #1016, opened 2021-02-05
refactor: code improvements to non-cartesian charts

monfera:
## Summary

No functional or API change.

Implementational changes:

- some `any`s and `as`es are [gone](https://github.com/elastic/elastic-charts/pull/1016/commits/bf41b2375488cd16cd592a5720a67520eb57d804), in particular, relating to the partition/goal config objects
- some class changes, eg. moving private methods to the end
- `goal_chart` used to import many things from `partition_chart`, these are extracted to a higher level, ie. `src/common`
- this was accompanied by some divvying up or merger of files, where eg. an existing `common` util file existed
- as `partition_chart` lost files, its dir structure is modestly simplified
- extracted out eg. `DEFAULT_CURSOR`
- added comment `/** potential internal` to some exports that do not, or do not yet feel necessary to export; otherwise, this PR doesn't attempt to remove imports

Though a bunch of shareable things are now in `src/common`, this specific PR doesn't aim to do a unification of said types with analogous common or `xy_chart` types, where they exist. I intentionally keep it superficial and incremental, to keep to a PR focus.

### Checklist

- [X] Unit tests were ~~updated or added to match the most common scenarios~~ relocated and jest-wise relinked


src/common/optimize.ts
```diff
@@ -0,0 +1,80 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/** @internal */
+export function integerSnap(n: number) {
+  return Math.floor(n);
+}
```
markov00, line 23: Ok for now, not sure if such a function is really required as importing and calling `integerSnap(n)` over `Math.floor(n)` doesn't bring any particular benefit

src/common/constants.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/** @internal */
+export const DEFAULT_CURSOR = 'default';
```
markov00, line 21: can we call it `DEFAULT_CSS_CURSOR` as this is a specific CSS value https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

src/chart_types/partition_chart/specs/index.ts
```diff
@@ -20,8 +20,9 @@
 import React from 'react';
 
 import { ChartTypes } from '../..';
+import { Pixels } from '../../../common/geometry';
 import { Spec } from '../../../specs';
-import { SpecTypes } from '../../../specs/constants';
+import { SpecTypes } from '../../../specs/constants'; // kept as unshortened import on separate line otherwise import circularity emerges
```
markov00, line 25: import circularity can be fixed here, but let's work on that on a different PR

src/chart_types/partition_chart/state/chart_state.tsx
```diff
@@ -38,6 +39,17 @@ import { createOnElementOverCaller } from './selectors/on_element_over_caller';
...
+  return (
+    <>
+      <Tooltip getChartContainerRef={containerRef} />
+      <Partition forwardStageRef={forwardStageRef} />
+      <HighlighterFromHover />
+      <HighlighterFromLegend />
+    </>
+  );
+}
+
```
markov00, line 52: What's the benefit of moving that outside?

src/chart_types/xy_chart/state/selectors/get_cursor_pointer.ts
```diff
@@ -53,18 +54,15 @@ export const getPointerCursorSelector = createCachedSelector(
...
+      return DEFAULT_CURSOR;
     }
     if (highlightedGeometries.length > 0 && (settingsSpec.onElementClick || settingsSpec.onElementOver)) {
       return 'pointer';
     }
     if (projectedValues !== null && settingsSpec.onProjectionClick) {
       return 'pointer';
     }
-    return isBrushAvailable ? 'crosshair' : 'default';
+    return isBrushAvailable ? 'crosshair' : DEFAULT_CURSOR;
```
markov00, line 66: since we moved out DEFAULT_CURSOR what if we move out also crosshair into `CROSSHAIR_CSS_CURSOR`

src/components/chart_container.tsx
```diff
@@ -44,7 +45,7 @@ interface ChartContainerComponentStateProps {
   isChartEmpty?: boolean;
   pointerCursor: string;
   isBrushing: boolean;
-  initalized?: boolean;
+  initialized?: boolean;
```
markov00, line 48: 🇮🇹 initaliaized :D

stories/sunburst/30_largest_circle.tsx
```diff
@@ -20,7 +20,7 @@
 import React from 'react';
 
 import { Chart, Datum, Partition, PartitionLayout } from '../../src';
-import { config } from '../../src/chart_types/partition_chart/layout/config/config';
+import { config } from '../../src/chart_types/partition_chart/layout/config';
```
markov00, line 23: There a lot of these on all the partition examples. 
If that `valueFormatter` is so useful probably we should export it as part of API and/or have it in the storybook utils folder

src/components/chart_container.tsx
```diff
@@ -234,11 +235,12 @@ const mapStateToProps = (state: GlobalChartState): ChartContainerComponentStateP
 
   return {
     status,
-    initalized,
+    initialized,
     isChartEmpty: isInternalChartEmptySelector(state),
     pointerCursor: getInternalPointerCursor(state),
     isBrushingAvailable: getInternalIsBrushingAvailableSelector(state),
     isBrushing: getInternalIsBrushingSelector(state),
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
```
markov00, line 243: I'm testing locally and doesn't seem to be required

src/chart_types/heatmap/state/selectors/get_cursor_pointer.ts
```diff
@@ -26,6 +27,6 @@ import { getPickedShapes } from './picked_shapes';
 export const getPointerCursorSelector = createCachedSelector(
   [getPickedShapes, isBrushingSelector],
   (pickedShapes, isBrushing) => {
-    return isBrushing || (Array.isArray(pickedShapes) && pickedShapes.length > 0) ? 'pointer' : 'default';
+    return isBrushing || (Array.isArray(pickedShapes) && pickedShapes.length > 0) ? 'pointer' : DEFAULT_CURSOR;
```
markov00, line 30: `pointer` could go in the same direction of the `default`
We can also build a object enum with these CSS properties

markov00 (approved):
All good to me, I've left few comments that can be addressed here or on a future PR

src/common/config_objects.ts
```diff
@@ -0,0 +1,97 @@
...
+
+interface PlainConfigItem {
+  type: 'group' | 'color' | 'string' | 'boolean' | 'number';
+  dflt?: any;
+  min?: number;
+  max?: number;
+  reconfigurable?: boolean | string;
+  values?: unknown;
+  documentation?: string;
+}
```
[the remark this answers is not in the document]
monfera: Yes, my intent was to give it some type now, and if we proceed with the config object concept, or esp. if we want to expose it, it should be more intricately tailored, by type as you say, with `is` predicate functions. Plan is to squeeze in stepwise improvement into various PRs as code is touched

src/common/optimize.ts
monfera: Indeed, no runtime difference, it just makes it explicit that the optimizer is using this, as there'd be other options eg. round or ceil, so it's a bit like extracting out, and making parametric, a magic constant. I'll think about removing it though, maybe there wouldn't be much docu loss

src/chart_types/partition_chart/state/chart_state.tsx
monfera: The motive was to not mix state (`chart_state.tsx`) with rendering, so, planning to move this function from under `state/` to under `render/` - would probably be in a subsequent PR. 

Besides just sharper state/render separation, there's value in not tightly coupling, or associating, _management_ code - eg. a class, which is mostly scaffolding in this case - with _doer_ code (calculations, imperative drawers etc.)

src/chart_types/xy_chart/state/selectors/get_cursor_pointer.ts
monfera: Good idea, maybe it could be `CssCursor.Default` and `CssCursor.Crosshair`, unless this CSS prop is covered by a TS library type definition; will look into it for a follow-up PR

src/components/chart_container.tsx
monfera: in italianized 🇮🇹

stories/sunburst/30_largest_circle.tsx
monfera: Sounds like a good idea, I'll probably put it on my list

src/components/chart_container.tsx
monfera: You don't get any warnings?
...hmm not needed anymore, wondering if some interim code changes temporary arose a warning. Removing it now.

src/chart_types/heatmap/state/selectors/get_cursor_pointer.ts
monfera: Haha great minds! Will put it in the next PR

src/components/chart_container.tsx
monfera: 859d13ee829eb351ebdb2fd2100b37c526585501

src/chart_types/partition_chart/specs/index.ts
monfera: I'll need help, didn't work for me on a couple of quick attempts, seems like you have a plan

src/common/constants.ts
monfera: 859d13ee829eb351ebdb2fd2100b37c526585501

src/common/config_objects.ts
```diff
@@ -0,0 +1,97 @@
...
+  values: Record<string, PlainConfigItem>;
+}
+
+// switching to `io-ts` style, generic way of combining static and runtime type info - 1st step
+class Type<A> {
+  dflt: A;
+
+  reconfigurable: boolean | string;
+
+  documentation = 'string';
```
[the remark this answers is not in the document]
monfera: Thanks, good catch! 859d13ee829eb351ebdb2fd2100b37c526585501

src/chart_types/partition_chart/specs/index.ts
markov00: I have it, let's talk on our monday weekly call

src/chart_types/partition_chart/state/chart_state.tsx
markov00: Great, thanks for the explaination

PR #1020, opened 2021-02-06
refactor(partition): code compliance and improvements

monfera:
## Summary

No external changes. Preparation for small multiples via general code improvements.

Implementation changes, easiest to see by commit:
- stronger separation of API from logic: partition selector files contain no logic; previous code contained in selector bodies got split to 
  - an API related part (remained in the selector file) and 
  - a calculation part (got moved out of selector files)  
- extracted out common partition and goal event handler selector makers https://github.com/elastic/elastic-charts/pull/1020/commits/2503531f4f3952d3b9ad23f93017d9419a84a134
- linting implementation files to new coding rules
- more precise and explicit types
- breaking up some overly large functions
- formulating, extracting out text and color utils, and moving them out
- more documentation and restructuring for better readability
- follow-up to #1016: moving out a renderer

### DIFF
--- a/src/chart_types/partition_chart/renderer/canvas/partition.tsx
+++ b/src/chart_types/partition_chart/renderer/canvas/partition.tsx
@@ -29,12 +29,13 @@ import { Dimensions } from '../../../../utils/dimensions';
 import { MODEL_KEY } from '../../layout/config';
 import { nullShapeViewModel, QuadViewModel, ShapeViewModel } from '../../layout/types/viewmodel_types';
 import { INPUT_KEY } from '../../layout/utils/group_by_rollup';
-import { partitionGeometries } from '../../state/selectors/geometries';
+import { partitionMultiGeometries } from '../../state/selectors/geometries';
 import { renderPartitionCanvas2d } from './canvas_renderers';
 
 interface ReactiveChartStateProps {
   initialized: boolean;
   geometries: ShapeViewModel;
+  multiGeometries: ShapeViewModel[];
   chartContainerDimensions: Dimensions;
 }
 
@@ -46,10 +47,11 @@ interface ReactiveChartOwnProps {
 }
 
 type PartitionProps = ReactiveChartStateProps & ReactiveChartDispatchProps & ReactiveChartOwnProps;
+
 class PartitionComponent extends React.Component<PartitionProps> {
   static displayName = 'Partition';
 
-  // firstRender = true; // this'll be useful for stable resizing of treemaps
+  // firstRender = true; // this will be useful for stable resizing of treemaps
   private ctx: CanvasRenderingContext2D | null;
 
   // see example https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio#Example
@@ -137,13 +139,16 @@ class PartitionComponent extends React.Component<PartitionProps> {
   }
 
   private drawCanvas() {
-    if (this.ctx) {
-      const { width, height }: Dimensions = this.props.chartContainerDimensions;
-      renderPartitionCanvas2d(this.ctx, this.devicePixelRatio, {
-        ...this.props.geometries,
-        config: { ...this.props.geometries.config, width, height },
-      });
-    }
+    const { width, height }: Dimensions = this.props.chartContainerDimensions;
+    this.props.multiGeometries.forEach((geometries) => {
+      if (this.ctx) {
+        // TS weirdness for the ctx check to need to be inside the forEach
+        renderPartitionCanvas2d(this.ctx, this.devicePixelRatio, {
+          ...geometries,
+          config: { ...geometries.config, width, height },
+        });
+      }
+    });
   }
 
   private tryCanvasContext() {
@@ -163,6 +168,7 @@ const mapDispatchToProps = (dispatch: Dispatch): ReactiveChartDispatchProps =>
 const DEFAULT_PROPS: ReactiveChartStateProps = {
   initialized: false,
   geometries: nullShapeViewModel(),
+  multiGeometries: [],
   chartContainerDimensions: {
     width: 0,
     height: 0,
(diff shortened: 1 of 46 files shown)


src/common/text_utils.ts
```diff
@@ -128,3 +129,33 @@ export const VerticalAlignments = Object.freeze({
 });
 
 export type VerticalAlignments = Values<typeof VerticalAlignments>;
+
+/** internal */
```
markov00, line 133: why not `@internal`? there are a few other ones with the same comment

src/state/chart_state.ts
```diff
@@ -410,17 +410,14 @@ function chartTypeFromSpecs(specs: SpecList): ChartTypes | null {
   return nonGlobalTypes[0];
 }
 
+const constructors: Record<ChartTypes, () => InternalChartState | null> = {
+  [ChartTypes.Goal]: () => new GoalState(),
+  [ChartTypes.Partition]: () => new PartitionState(),
+  [ChartTypes.XYAxis]: () => new XYAxisChartState(),
+  [ChartTypes.Heatmap]: () => new HeatmapState(),
+  [ChartTypes.Global]: () => null,
+}; // with no default, TS signals if a new chart type isn't added here too
```
markov00, line 419: 👍

src/chart_types/partition_chart/renderer/canvas/partition.tsx
```diff
@@ -139,16 +140,16 @@ class PartitionComponent extends React.Component<PartitionProps> {
   }
 
   private drawCanvas() {
-    const { width, height }: Dimensions = this.props.chartContainerDimensions;
-    this.props.multiGeometries.forEach((geometries) => {
-      if (this.ctx) {
-        // TS weirdness for the ctx check to need to be inside the forEach
+    if (this.ctx) {
```
markov00, line 143: what if we reduce nesting and return earlier:
```
if(!this.ctx){
  return
}
```

markov00 (approved):
Everything looks in better shape than before, thanks for this huge cleanup!
I've also seen the preparation part for the multi-geometries and looks a good start

src/common/text_utils.ts
monfera: Thanks for the catch! My best guess is that, whenever I put the `/** @internal */` before using the refactoring tool, it removed the `@`. Not 100% sure, will keep an eye on this

src/chart_types/partition_chart/renderer/canvas/partition.tsx
monfera: Oddly (TS seems to be full of holes that we need to work around, so maybe not that odd) it signals with that, or even with
```typescript
  private drawCanvas() {
    if (this.ctx === null) {
      return;
    }
    const { width, height }: Dimensions = this.props.chartContainerDimensions;
    clearCanvas(this.ctx, width * this.devicePixelRatio, height * this.devicePixelRatio);
    this.props.multiGeometries.forEach((geometries) => {
      renderPartitionCanvas2d(this.ctx, this.devicePixelRatio, {
        ...geometries,
        config: { ...geometries.config, width, height },
      });
    });
  }
```
despite the fact that it's clearly inferrable.

I ended up solving it with the TS-wise advantageous iterator based `for` loop, which enjoys more love by TS maintainers:
```typescript
  private drawCanvas() {
    if (this.ctx) {
      const { width, height }: Dimensions = this.props.chartContainerDimensions;
      clearCanvas(this.ctx, width * this.devicePixelRatio, height * this.devicePixelRatio);
      for (const geometries of this.props.multiGeometries) {
        renderPartitionCanvas2d(this.ctx, this.devicePixelRatio, {
          ...geometries,
          config: { ...geometries.config, width, height },
        });
      }
    }
  }
```

Once I chased down the reason for this counterintuitive difference, and there's the argument that "`.forEach()` uses a callback function, we've no idea what that callback function does, so can't be certain" which is superficially true, but there's no reason for TS to handle the callback function as a black box.

To summarize,
- it's improved in the follow-up PR to a single check
- if you want, I can replace the `if(this.ctx} draw()` with `if(!this.ctx) { return; } draw();`

src/common/text_utils.ts
monfera: https://github.com/elastic/elastic-charts/pull/1020/commits/f65ce9fcbe34640cabb195ff61138deada3e2066

monfera: ...more likely, it was a typo I managed to copy/paste within one file

src/state/chart_state.ts
monfera: Btw. the `switch/default/never` pattern would've worked too if the only purpose was to avoid missing a new chart type: https://github.com/elastic/elastic-charts/blob/89393aeb448fcc9aab6b1a95dc918f69284477fa/src/chart_types/partition_chart/layout/viewmodel/viewmodel.ts#L190-L231

src/chart_types/partition_chart/renderer/canvas/partition.tsx
markov00: All good to me, If you want to replace it in a different PR ok, if not is not a priority atm

src/chart_types/partition_chart/state/selectors/get_partition_specs.ts
```diff
@@ -0,0 +1,33 @@
...
+import { GlobalChartState } from '../../../../state/chart_state';
+import { getSpecsFromStore } from '../../../../state/utils';
+import { PartitionSpec } from '../../specs';
+
+const getSpecs = (state: GlobalChartState) => state.specs;
+
+/** @internal */
+export const getPartitionSpecs = createCachedSelector([getSpecs], (specs) => {
+  return getSpecsFromStore<PartitionSpec>(specs, ChartTypes.Partition, SpecTypes.Series);
+})((state) => state.chartId);
```
nickofthyme, line 33: Use `getChartIdSelector` for consistency 

```suggestion
})(getChartIdSelector);
```

nickofthyme (approved):
Code changes LGTM, just one comment. Tested locally and all partition stories seem to work as expected. 🎉 🚢

src/chart_types/partition_chart/state/selectors/get_partition_specs.ts
monfera: Thanks Nick, I'll put it in the next PR to avoid a CI run

monfera: This and 10 more like this are in a commit in the upcoming PR

PR #1023, opened 2021-02-10
feat: sort values in actions by closest to cursor

### LINKED ISSUE #835: Order TooltipValues in CustomTooltip by distance to pointer
Tooltips with multiple values on hover. It would be nice to have the closest value to the pointer first to display that one first if there is only one to be displayed.

Order `TooltipValue[]` array elements by distance to geometry element, closest first.

Render multiple elements for all highlighted elements.

nickofthyme:
## Summary

closes #835

Sorts values when finding indexed geometries by closest to cursor `Point`. This is determined by the spatial distance between the index points and the cursor point. Will only compare across different x values in spatially indexed charts (i.e. bubble).

### Example
Image 2021-02-10 at 2 09 50 PM

Clicking near the yellow arrow will now return the values in the array as `7.0` then `6.8` because the cursor is closest to `7.0`

Clicking near the pink arrow will now return the values in the array as `6.8` then `7.0` because the cursor is closest to `6.8`

This will also be performed on spatial maps with bubble charts and will use the closest distance, including across different x values.

### Checklist


markov00 (approved):
Tested locally and works fine when pushing out events on click/over.
Can you describe what is the current use case for that sorting?

nickofthyme:
For now, the only use case is in `vislib` replacement for area and line charts where points are close together and clicked to filter. This will now filter the nearest point.

PR #1031, opened 2021-02-17
feat(xy_charts): render legend inside the chart

### LINKED ISSUE #861: Legend positioned inside the charts
To replace timelion's flot implementation in Kibana by elastic-chart we'd need to be able to place the legend inside the chart. After a quick chat with @markov00 we decided this might be good to be implemented inside elastic charts.

screenshot-20201008-100941

You currently have 4 positions for the legend inside charts in Timelion: northeast, northwest, southeast, southwest.

**Kibana cross issue:** https://github.com/elastic/kibana/issues/79984

markov00:
## Summary

This PR adds the ability to display the legend inside the XY cartesian charts.
The `<Settings>` property `legendPosition` will take now an extended version of the position that allow an increased customizability:
```ts
... // on SettingsSpec
legendPosition: Position | LegendPositionConfig;

...

type LegendPositionConfig = {
  /**
   * The vertical alignment of the legend
   */
  vAlign: typeof VerticalAlignment.Top | typeof VerticalAlignment.Bottom; // TODO typeof VerticalAlignment.Middle
  /**
   * The horizontal alignment of the legend
   */
  hAlign: typeof HorizontalAlignment.Left | typeof HorizontalAlignment.Right; // TODO typeof HorizontalAlignment.Center
  /**
   * The direction of the legend items.
   * `horizontal` shows all the items listed one a side the other horizontally, wrapping to new lines.
   * `vertical` shows the items in a vertical list
   */
  direction: LayoutDirection;
  /**
   * Remove the legend from the outside chart area, making it floating above the chart.
   * @default false
   */
  floating: boolean;

};
```

The legend outside the chart only works with the following cases:
```
direction: vertical, vAlign: top, hAlign: left/right
direction: horizontal, hAlign:left, vAlign: top/bottom
```

This in the future can be expanded to consider:
- centred alignment positions like `VerticalAlignment.Middle` `HorizontalAlignment.Center`
- add more customization to the legend, like the number of columns
- add a grow factor like the flex box layout: shrink, grow, fill can be used to specify how to display the legend items

Screenshot 2021-02-17 at 16 55 39

fix #861


markov00:
@miukimiu can you please double check from the UX prospective if we need to add a background layer below the legend and/or change the current highlight background on each legend item on hover, thanks

markov00:
@monfera @rshen91 @nickofthyme  I'd like your opinion on the API as written on the PR description

miukimiu (approved):
I've been experimenting on https://github.com/miukimiu/elastic-charts/tree/2021_01_21-legend_inside_chart how the legend would look like without the light gray background.

On hover, we could make the `font-weight` bolder and the `color` darker. But with a bolder `font-weight`, the text would sometimes shift a little bit. To avoid that we could use a `text-shadow`. But after experimenting with this solution, I came to the conclusion that:

- **It's hard to notice the hover state in dark themes**

So I think the light gray background works better for most scenarios. 

legend-hover@2x

Conclusion, LGTM! 🎉

rshen91:
> @monfera @rshen91 @nickofthyme I'd like your opinion on the API as written on the PR description

I think it might make sense to have an internalLegendPosition property in addition to the legendPosition property to avoid confusion.

nickofthyme (commented):
Sorry this took so long to get to.

I think most of this api is fine but I really don't like the array approach. I listed the negatives below:

- IMHO I think it is more complicated for the chart consumer to interoperate how to position the legend
- None of the positions mention the legend being internal but `['top', 'left']`
- It creates more checking internally to determine if the position is internal and handling the `legendPosition` as a tuple or a `Position` type
- `top`/`bottom` must come first then `left`/`right`

I think the best approach is to add these to the `Position` enum or likely another `LegendPosition` to avoid collisions.

```ts
LegendPosition = {
  ...Position,
  InsideTopLeft: 'ITL',
  InsideTopRight: 'ITR',
  InsideBottomLeft: 'IBL',
  InsideBottomRight: 'IBR',
}
```

simple API, explicit internal designation, cleaner internal values checks.

monfera:
Is it a full understanding of the distinct envisioned use cases?

| Name | Dimension | Exists? | Shape | Chart landmark | Legend landmark 
| :--      |:--:                |:--:          |:--        |:--                           |:--                       
| Top | Y | yes | horizontally spans<br />vertically minimizes | X:left<br />Y: top | X:left<br />Y: bottom
| Bottom | Y | yes | horizontally spans<br />vertically minimizes | X:left<br />Y: bottom | X:left<br />Y: top
| Left | X | yes | vertically spans<br />horizontally minimizes | X:left<br />Y: top | X:right<br />Y: top
| Right | X | yes | horizontally spans<br />vertically minimizes | X:right<br />Y: top | X:left<br />Y: top
| InsideTopLeft | XY | no | vertically spans<br />horizontally minimizes | X:left<br />Y: top | X:left<br />Y: top
| InsideTopRight | XY | no | vertically spans<br />horizontally minimizes | X:right<br />Y: top | X:right<br />Y: top
| InsideBottomLeft | XY | no | vertically spans<br />horizontally minimizes | X:left<br />Y: bottom | X:left<br />Y: bottom
| InsideBottomRight | XY | no | vertically spans<br />horizontally minimizes | X:right<br />Y: bottom | X:right<br />Y: bottom

A single enum has the benefit of one, primitive type for the user (no union type), at the slight cost of bundling independent properties (into not too many, easy to understand, most frequently used options).

Where it runs into limit is if we want to add pretty much anything; eg. 
- bring in [middle/center](https://user-images.githubusercontent.com/1548516/110001373-9e161f80-7d14-11eb-99cb-9ca4c404eda9.png) for every option as an alternative to left/right/top/bottom - leads to much more than the advised 5-9 options
- add the possibility of [relative](https://plotly.com/javascript/legend/) or absolute coordinates
- [multiple legends](https://www.datanovia.com/en/wp-content/uploads/dn-tutorials/ggplot2/figures/026-ggplot-legend-guides-change-the-order-of-legends-1.png)
- separating the legends from the charts (because then we can have multiple charts on a dashboard that share common legends; also, non-`elastic-charts` visuals eg. cell-colored tables or maps can use `elastic-charts` legends) 

Addressing the last two are larger exercises that needs more planning anyway.

So either way looks good if the table is a correct summary of the ~1year outlook, maybe enums are simpler (more than 7-9 enums and they become unmanageable) and involve a smaller API change; we could rethink for all these other use cases in a next round. But if we need relative coordinates now, maybe it's a good time to go for the more ambitious undertaking.

api/charts.api.md
```diff
@@ -1210,6 +1210,17 @@ export interface LayerValue {
     value: number;
 }
 
+// Warning: (ae-missing-release-tag) "LayoutDirection" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
```
monfera, line 1213: I also kept getting this message, which seems incorrect. I think it can be made go away by documenting the individual properties

markov00: yep

api/charts.api.md
```diff
@@ -1250,6 +1261,37 @@ export type LegendPathElement = {
...
+    hAlign: typeof HorizontalAlignment.Left | typeof HorizontalAlignment.Right;
+    direction: LayoutDirection;
+    floating: boolean;
+};
+
+// @public
+export interface LegendSpec {
+    flatLegend?: boolean;
+    legendAction?: LegendAction;
+    // (undocumented)
```
monfera, line 1276: Minor, not needed for this PR, just a thought. As we might be contemplating the separation of the API into an outer layer (lots of optional or defaulting props) maybe we could start defining "hard" specs, ie. nothing is optional, and then layer some soft spec over it, by eg. combining `<Partial<HardLegendSpec>>` with the list of mandatory ones (eg. apparently `legendMaxDepth`, `legendPosition` etc. which are mandatory already).

Then the optionals can get default values, eg. `onLegendItemClick` right away (eg. `()=>{}`) and thus restore well-defined, reliable typing. The result is that in the code, there won't be a need for the ever present question of "is it mandatory? should we do optional chaining? null coalescing?"

The biggest win, I think, is that there's one defined place for the defaults to go in, and remains safely DRY. I find myself in temptation when resolving optionality deeper in the code logic. "Have I not coded in some logic for this?"

So I believe the dual layering is a big win in collecting the people-oriented, somewhat mushy defaulting logic into one place, while letting the core scintillate, unburdened with branching and defaulting. I should practice it more haha

api/charts.api.md
```diff
@@ -1918,7 +1960,7 @@ export type SeriesTypes = $Values<typeof SeriesTypes>;
 export const Settings: React_2.FunctionComponent<SettingsSpecProps>;
 
 // @public
-export interface SettingsSpec extends Spec {
+export interface SettingsSpec extends Spec, LegendSpec {
```
monfera, line 1963: Nit: I might have misunderstood; I thought the `LegendSpec` would be its own object, not folded into `SettingsSpec`. Eg. `SettingsSpec.legend = { show: true, maxDepth: 2, strategy: ...}`

It works as it is, though `SettingsSpec` is getting a bit long and overwhelmed with legend options and many names start with, or maybe less ideally, have in other places in their name (eg. `showLegend`); what do we do when we make the next thing sophisticated with 10+ props, would those also be extended into `SettingsSpec`? Admittedly the flat representation has some pros too, though for some reason I'd lean toward structuring

markov00: This was just a first pass toward exposing the legend configuration as its own component.
Just didn't want to create a breaking change right now for that

api/charts.api.md
markov00: > Minor, not needed for this PR, just a thought. As we might be contemplating the separation of the API into an outer layer (lots of optional or defaulting props) maybe we could start defining "hard" specs, ie. nothing is optional, and then layer some soft spec over it, by eg. combining <Partial<HardLegendSpec>> with the list of mandatory ones (eg. apparently legendMaxDepth, legendPosition etc. which are mandatory already).

Yes I think we should apply the same approach to every other specs in charts

monfera:
Some specifications don't combine, eg. 

- outside chart + bottom
- outside chart + left

Maybe it's expected though, left for a future PR

image

src/chart_types/heatmap/state/selectors/get_grid_full_height.ts
```diff
@@ -54,7 +54,7 @@ export const getGridHeightParamsSelector = createCachedSelector(
     const xAxisHeight = visible ? fontSize : 0;
     const totalVerticalPadding = padding * 2;
     let legendHeight = 0;
-    if (showLegend && isHorizontalLegend(legendPosition)) {
+    if (showLegend && isHorizontalLegend(Array.isArray(legendPosition) ? legendPosition[1] : legendPosition)) {
```
monfera, line 57: If `legendPosition` can be an array, where is it specified? On a cursory search I only see
```
legendPosition: Position | LegendPositionConfig
```

markov00: Oh thanks, good catch, it was the previous implementation, Let me fix this

src/components/legend/legend_item.tsx
```diff
@@ -47,7 +53,7 @@ export const LEGEND_HIERARCHY_MARGIN = 10;
 export interface LegendItemProps {
   item: LegendItem;
   totalItems: number;
-  position: Position;
+  legendPositionConfig: LegendPositionConfig;
```
monfera, line 56: As we are already inside `LegendItemProps`, and the type is prefixed with `Legend`, maybe the name could shorten to `PositionConfig`; the other props aren't prefixed with `legend` either

src/components/legend/position_style.ts
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * @internal
+ */
+export function getLegendPositionConfig(position: LegendSpec['legendPosition']): LegendPositionConfig {
+  if (typeof position === 'object') {
+    return position;
+  }
+  return LEGEND_TO_FULL_CONFIG[position];
+}
```
monfera, line 101: Or just 
```
return typeof position === 'object' ? position : LEGEND_TO_FULL_CONFIG[position];
```

src/components/legend/style_utils.ts
```diff
@@ -74,8 +73,8 @@ export function getLegendListStyle(
  * Get the legend global style
  * @internal
  */
-export function getLegendStyle(position: Position, size: BBox, margin: number): LegendStyle {
-  if (position === Position.Left || position === Position.Right) {
+export function getLegendStyle(direction: LegendPositionConfig['direction'], size: BBox, margin: number): LegendStyle {
+  if (direction === 'vertical') {
```
monfera, line 77: Minor: here and elsewhere, using the "enum" feels a bit more reified:
```
if (direction === LayoutDirection.Vertical)
```

markov00: thanks I'm gonna change every missing enums

src/state/selectors/get_legend_config_selector.ts
```diff
@@ -0,0 +1,59 @@
...
+    onLegendItemMinusClick,
+    onLegendItemOut,
+    onLegendItemOver,
+    onLegendItemPlusClick,
+    showLegendExtra,
+  }) => {
+    return {
+      flatLegend,
+      legendAction,
+      legendColorPicker,
```
monfera, line 46: This extraction of the listed legend props here wouldn't be needed if the legends spec props belonged in its cohesive object, via a single composite prop in `SettingsSpec`. As is, every future prop change/addition needs maintenance here too, so a bit less DRY

src/utils/legend.ts
```diff
@@ -17,10 +17,11 @@
...
 
+import { SettingsSpec } from '../specs/settings';
 import { Position } from './common';
 
-export const isHorizontalLegend = (legendPosition: Position) =>
+export const isHorizontalLegend = (legendPosition: SettingsSpec['legendPosition']) =>
   legendPosition === Position.Bottom || legendPosition === Position.Top;
 
-export const isHierarchicalLegend = (flatLegend: boolean | undefined, legendPosition: Position) =>
+export const isHierarchicalLegend = (flatLegend: boolean | undefined, legendPosition: SettingsSpec['legendPosition']) =>
```
monfera, line 26: 👍 the referencing of the `legendPosition` is great, helps keep types DRY

src/chart_types/partition_chart/layout/utils/legend.ts
```diff
@@ -45,7 +46,7 @@ export function getLegendItems(
   layers: Layer[],
   flatLegend: boolean | undefined,
   legendMaxDepth: number,
-  legendPosition: Position,
+  legendPosition: SettingsSpec['legendPosition'],
```
monfera, line 49: 👍

src/state/selectors/get_legend_config_selector.ts
markov00: I know, hope to get to this soon

src/chart_types/xy_chart/utils/axis_utils.ts
```diff
@@ -66,8 +66,14 @@ export interface TickLabelProps {
   offsetY: number;
   textOffsetX: number;
   textOffsetY: number;
-  align: Extract<HorizontalAlignment, 'left' | 'center' | 'right'>;
-  verticalAlign: Extract<VerticalAlignment, 'top' | 'middle' | 'bottom'>;
+  align: Extract<
```
monfera, line 69: Is it intentionally just `align` rather than `horizontalAlign`?

markov00: I've changed only the type declaration here, I can change the propname to reflect it better

api/charts.api.md
monfera: Makes sense to do it in a dedicated PR, it represents a very breaking change

src/chart_types/heatmap/state/selectors/get_grid_full_height.ts
monfera: Weird that TS doesn't complain here that it narrowed to `never` and the test will be always falsey; structural typing without structural support 😄

src/chart_types/xy_chart/utils/axis_utils.ts
markov00: fixed 95d59c9

src/components/legend/style_utils.ts
markov00: fixed 95d59c9

monfera (approved):
Works well for me for the inside placed Cartesians. Thanks for assessing the individual comments. Unless there's concern about some of the options not yet functional (see comment with screenshot) it's good to go, it adds useful functionality and doesn't remove existing capabilities

src/components/legend/position_style.ts
markov00: fixed in 95d59c9 (not sure why this doesn't shows up here as outdated)

src/components/legend/legend_item.tsx
markov00: done in 95d59c9

src/chart_types/heatmap/state/selectors/get_grid_full_height.ts
markov00: done in 95d59c9

nickofthyme (commented):
lol was just about to approve! 🟢

Verified with rotations on chrome. One thing we might want to add in the future is a theme item to set the box-shadow of the text and color to avoid hidden dots.

Image 2021-03-24 at 11 29 33 AM

PR #1033, opened 2021-02-18
feat(partition): clip text in partition chart fill label

dgieselaar:
## Summary

Optionally clip text for fill label in partition chart, and fix an alignment issue.

Before:

image

After:
image

### Checklist


src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
```diff
@@ -111,7 +111,7 @@ function getVerticalAlignment(
     case VerticalAlignments.bottom:
       return -(container.y1 - linePitch * (totalRowCount - 1 - rowIndex) - paddingBottom - fontSize * overhang);
     default:
-      return -((container.y0 + container.y1) / 2 + (linePitch * (rowIndex - totalRowCount)) / 2);
+      return -((container.y0 + container.y1) / 2 + (linePitch * (rowIndex + 1 - totalRowCount)) / 2);
```
markov00, line 114: :D

src/chart_types/partition_chart/renderer/canvas/canvas_renderers.ts
```diff
@@ -58,6 +66,7 @@ function renderTextRow(
         ctx.font = cssFontShorthand(box, fontSize);
         ctx.fillText(box.text, box.width / 2 + box.wordBeginning, 0);
       });
+      ctx.closePath();
```
markov00, line 69: I don't think we need to call `closePath` here, and probably we don't even need the `ctx.beginPath();` at line 60 as `fillText` doesn't  need to instantiate a new path

src/chart_types/partition_chart/layout/config.ts
```diff
@@ -212,6 +212,48 @@ export const configMetadata: Record<string, ConfigItem> = {
...
+            dflt: 2,
+            min: 0,
+            max: 20,
+            type: 'number',
+            reconfigurable: true,
+            documentation: 'Right padding for fill text',
+          },
+        },
+      },
+      clip: {
```
markov00, line 252: we rename it `clipText` within the code as it clips the text, not the geometry. I know that is part of the `fillLabel` config, but we also have other attributes as `textColor` `textContrast` etc

markov00 (approved):
LGTM, I've left a few minor comments

src/chart_types/partition_chart/renderer/canvas/canvas_renderers.ts
monfera: Indeed! The follow-up PR significantly restructures the Canvas render code, will explain reasons (motive: reduce context save/restore count)

src/chart_types/partition_chart/layout/config.ts
monfera: https://github.com/elastic/elastic-charts/pull/1033/commits/8853edff2457f22dad6dbe7b86ce13c4366ba46b

PR #1035, opened 2021-02-18
chore: add lint rules for release tags

nickofthyme:
## Summary

Adds three new lint rules to help control exports and api documentation. Pertains only to `.ts` and `.tsx` files. Adds line break after header to be consistent across files and help discern comments just below the header.

### `require-tsdocs` (fixable)
Requires [tsdocs](https://tsdoc.org/) for all `export` types and values.

Converts single `Line` comment into single tsdoc comment
```diff
- // single comment
+ /**
+  * single comment
+  */
export const SomeThing = 'test';
```

Ignores all but the closest comments
```diff
// comment 1
- // comment 2
+ /**
+  * comment 2
+  */
export const SomeThing = 'test';
```

Fixes  `Block` comment formatting to tsdoc format (aka starts with `/**`)
```diff
- /*
+ /**
 * comment
 */
export const SomeThing = 'test';
```

### `require-release-tag` (fixable)
Requires a [release tag](https://api-extractor.com/pages/tsdoc/doc_comment_syntax/#release-tags) on all `export` tsdoc comments.

Adds `@internal` tag to bottom of tsdoc comment if not tag provided
```diff
/**
 * Some thing
+ * @internal
 */
export const SomeThing = 'test';
```

> Default fix is to add `@internal` tag.

### `require-documentation`
Requires a non-empty tsdoc [comment](https://api-extractor.com/pages/tsdoc/doc_comment_syntax/#comment-structure) on all `@public` `export` type and values.

image

> Right now this is just set as a warning and is _not_ fixable. We can add documentation to these as we go.

### `no-different-release-tag` (fixable)
Requires the same release tag for types and variables exported from the same file per [`ae-different-release-tags`](https://api-extractor.com/pages/messages/ae-different-release-tags/) rule.

If release tags differ, the rule will replace the tag with the lowest significance. Tags from highest to lowest significance are as follows: `@alpha`, `@beta`, `@public`, `@internal`.

```diff
- /** @internal */
+ /** @public */
export const BandedAccessorType = Object.freeze({
  Y0: 'y0' as const,
  Y1: 'y1' as const,
});

/**
 * @public
 */
export type BandedAccessorType = $Values<typeof BandedAccessorType>;
```


.eslintrc.js
```diff
@@ -29,34 +29,22 @@ module.exports = {
     'react-hooks',
     'jsx-a11y',
     'prettier',
+    'elastic-charts',
   ],
   rules: {
-    /**
+    /*
```
markov00, line 35: IF the rule is `Fixes Block comment formatting to tsdoc format (aka starts with /**)` why this is not applied here? and the opposite is applied?

markov00: I'm reading it correctly that the double start comment is only applied to exported members?

package.json
```diff
@@ -52,7 +52,11 @@
     "typecheck:src": "yarn build:ts --noEmit",
     "typecheck:all": "tsc -p ./tsconfig.json --noEmit",
     "ts:prune": "ts-prune",
-    "watch": "yarn test --watch"
+    "watch": "yarn test --watch",
+    "index:check": "ts-node scripts/index_check.ts",
+    "tsdocs:check": "ts-node scripts/tsdoc_check.ts",
+    "lint:test": "node --inspect ./node_modules/eslint/bin/eslint src/utils/geometry.ts",
+    "lint:check": "eslint src/utils/geometry.ts"
```
markov00, line 59: This was probably just a for testing right? can we remove this?

package.json
```diff
@@ -151,6 +155,7 @@
     "eslint": "^7.17.0",
     "eslint-config-airbnb-typescript": "^12.0.0",
     "eslint-config-prettier": "^7.1.0",
+    "eslint-plugin-elastic-charts": "file:./eslint-plugin-elastic-charts",
```
markov00, line 158: I rather prefer to put this into a different folder. If in the near future we are going to use yarn workspaces what about moving the folder into a `packages/eslint-plugin-elastic-charts`?

src/chart_types/goal_chart/layout/types/config_types.ts
```diff
@@ -21,8 +21,10 @@ import { Pixels, SizeRatio } from '../../../../common/geometry';
 import { FontFamily } from '../../../../common/text_utils';
 import { Color } from '../../../../utils/common';
 
-// todo switch to `io-ts` style, generic way of combining static and runtime type info
-/** potential internal */
+/**
+ * todo switch to `io-ts` style, generic way of combining static and runtime type info
+ * @public
+ */
```
markov00, line 27: looks like in this case the rule doesn't apply correctly: it should have put `/** potential internal */` as the comment for the Config interface, leaving the todo as it is

src/chart_types/index.ts
```diff
@@ -30,4 +30,5 @@ export const ChartTypes = Object.freeze({
   XYAxis: 'xy_axis' as const,
   Heatmap: 'heatmap' as const,
 });
+/** @public */
```
markov00, line 33: Can we also assure to leave 1 blank line between the comment and the previous code?

src/chart_types/partition_chart/layout/types/config_types.ts
```diff
@@ -77,7 +82,10 @@ export interface FillFontSizeRange {
   maximizeFontSize: boolean;
 }
 
-// todo switch to `io-ts` style, generic way of combining static and runtime type info
+/**
+ * todo switch to `io-ts` style, generic way of combining static and runtime type info
+ * @public
+ */
```
markov00, line 88: In this case, let's keep the todo as a separate comment

src/common/__mocks__/calcs.ts
```diff
@@ -16,7 +16,10 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+
 const module = jest.requireActual('../color_calcs.ts');
 
+/** @internal */
 export const getBackgroundWithContainerColorFromUser = jest.fn(module.getBackgroundWithContainerColorFromUser);
+/** @internal */
```
markov00, line 24: Can we make a special rule to avoid TSDocs on this particular folder, or in every folder/file not part of the final bundle?
Like everything related to testing and for the `data` folder for example

src/scales/scale_continuous.ts
```diff
@@ -229,7 +229,7 @@ const defaultScaleOptions: ScaleOptions = {
 
 /**
  * Continuous scale
- * @internal
+ * @public
```
markov00, line 232: Do we need to expose it? I don't think so because this brings in a lot of internal details that we don't want other use it

api/charts.api.md
```diff
@@ -764,23 +674,20 @@ export type DisplayValueStyle = Omit<TextStyle, 'fill' | 'fontSize'> & {
 // @public (undocumented)
 export type DomainRange = LowerBoundedDomain | UpperBoundedDomain | CompleteBoundedDomain | UnboundedDomainWithInterval;
 
-// Warning: (ae-missing-release-tag) "ElementClickListener" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
+// Warning: (ae-forgotten-export) The symbol "Ratio" needs to be exported by the entry point index.d.ts
 //
+// @public (undocumented)
+export type EasingFunction = (x: Ratio) => Ratio;
```
markov00, line 680: Do we need to export the `EasingFunction` at all? it's actually an alpha implementation for flame graphs

api/charts.api.md
```diff
@@ -2102,16 +1854,77 @@ export const StackMode: Readonly<{
 // @public
 export type StackMode = $Values<typeof StackMode>;
 
-// Warning: (ae-missing-release-tag) "Statistics" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
+// Warning: (ae-forgotten-export) The symbol "FillFontSizeRange" needs to be exported by the entry point index.d.ts
 //
+// @public
+export interface StaticConfig extends FillFontSizeRange {
```
markov00, line 1860: Also this one I don't think we should export this as it is, also because the name doesn't reflect the chart type

api/charts.api.md
```diff
@@ -2126,6 +1939,11 @@ export interface StrokeStyle<C = Color> {
     strokeWidth: number;
 }
 
+// Warning: (ae-forgotten-export) The symbol "TEXT_ALIGNS" needs to be exported by the entry point index.d.ts
+//
+// @public (undocumented)
+export type TextAlign = typeof TEXT_ALIGNS[number];
```
markov00, line 1945: I think we should not expose this

api/charts.api.md
```diff
@@ -2134,15 +1952,18 @@ export interface TextAlignment {
     vertical: VerticalAlignment;
 }
 
+// Warning: (ae-forgotten-export) The symbol "TEXT_BASELINE" needs to be exported by the entry point index.d.ts
+//
+// @public (undocumented)
+export type TextBaseline = typeof TEXT_BASELINE[number];
```
markov00, line 1958: same here, not expose

.eslintrc.js
nickofthyme: This was a manual change as these comments are not useful to be a JSDocs comment.

package.json
nickofthyme: Yup great idea

api/charts.api.md
nickofthyme: No this was an attempt to fix some of the https://github.com/elastic/elastic-charts/pull/1035/commits/353e7bb43bd780e0075ad09ae99df2e812ff4ece so called `forgotten exports` but I can revert them if you think they should not be exported.

src/chart_types/index.ts
nickofthyme: No this would be a separate rule which is not yet converted to typescript. https://github.com/typescript-eslint/typescript-eslint/issues/1150

Even so, I think this would also be hard to constrain to only the right case and @monfera would not like the limited freedom it would impose 😄

package.json
nickofthyme: f04ffc6

package.json
nickofthyme: d31d140

src/chart_types/goal_chart/layout/types/config_types.ts
nickofthyme: Works now...


https://user-images.githubusercontent.com/19007109/113173818-644f1080-920f-11eb-8d6e-dbd15cead5fc.mp4

src/chart_types/partition_chart/layout/types/config_types.ts
nickofthyme: d42ea3b

src/common/__mocks__/calcs.ts
nickofthyme: 7308fd1

src/scales/scale_continuous.ts
nickofthyme: c356066 & bd33ede

api/charts.api.md
nickofthyme: c356066

api/charts.api.md
nickofthyme: c356066

api/charts.api.md
nickofthyme: c356066

api/charts.api.md
nickofthyme: c356066

nickofthyme:
If you have trouble running this pr locally, try running `yarn install --force`. If this does not work, try running `rm -fr node_modules/eslint-plugin-elastic-charts/` and then run `yarn add -D file:./packages/eslint-plugin-elastic-charts`. Whenever updating static files related to eslint config you may need to quit and reopen vscode or whatever ide you use.

src/chart_types/partition_chart/layout/types/viewmodel_types.ts
```diff
@@ -202,7 +204,9 @@ export interface LayerFromTo {
   y1: TreeLevel;
 }
 
-/** potential internal */
+/**
+ * @public
```
rshen91, line 208: I'm noticing because the comment before was formatted `/** ` these seem to be multiline/doc comments with release tags?

rshen91 (approved):
Thanks for setting this up! I left one comment but I don't think it should be blocking or necessarily addressed before this is merged (more just a clarification). Thanks!

src/chart_types/partition_chart/layout/types/viewmodel_types.ts
nickofthyme: Yeah it will use single when it can. In this case it `fixed` it like this...

```diff
- /** potential internal */
+ /**
+  * potential internal
+  * @public
+  */
```

And then I deleted the `potential internal` line manually

src/chart_types/index.ts
markov00: :D

markov00 (approved):
Changes LGTM
Thanks Nick for that, it will improve a lot the DX

PR #1037, opened 2021-02-28
fix: clippedRanges when complete dataset is null

anishagg17:
## Summary

Fixes: #1047  

Before 

Screenshot 2021-03-03 at 11 59 05 PM

After 

Screenshot 2021-03-03 at 11 58 05 PM

### Checklist


stories/legend/6_hide_legend.tsx
```diff
@@ -60,6 +60,18 @@ export const Example = () => {
...
+        id="bars"
+        fit="zero"
+        name="amount"
+        xScaleType={ScaleType.Ordinal}
+        xAccessor="x"
+        yAccessors={['y']}
+        data={[
+          { x: 'trousers', y: null },
+          { x: 'pants', y: null },
+        ]}
```
anishagg17, line 74: I have added this part just for review purposes, will remove it later on.

anishagg17:
I have partially resolved this issue, looking for maintainers' insights on how they would like to proceed with it.

markov00:
Hi @anishagg17, thank you for your contribution, we really appreciated it.
We usually discuss the issue to better define the constraints and use cases before start working on it. 
Some issues/requests look fine at a first glance but they probably need a better and thoughtful discussion before being processed and implemented, some others may be outdated or invalid.
In this case, the issue was a bit vague: it actually adds an additional behavior to a specific edge case, increasing the number of internally managed edge cases that are not immediately understandable and known from the outside.
I think we should not proceed further on this issue and close it as it is. @nickofthyme what do you think?

nickofthyme:
Hey @anishagg17 Thanks for opening this PR. I can't recall what the impetus was for opening this issue.

I looked through your solution and after thinking more about this issue I am not sure this is something we need to change. The issue is suggesting we show the [empty state UI](https://elastic.github.io/elastic-charts/?path=/story/test-cases--no-series) when all values are null. After discussing this with the team, we think this is not the best approach as `null` values could be useful to visualize the data.

That being said. I do see an issue with how the line is being rendered see #1047. Would you like to take this issue? I can guide you where to look to fix this.

anishagg17:
> Would you like to take this issue?

Will really like to work on it, as it could help me gain more knowledge regarding the codeBase.

I think that I should make changes for #1047 within this pr only. But could you please let me know where to make changes at, I observed several functions, within which the fix could be implemented 

- https://github.com/elastic/elastic-charts/blob/85b6ed3cf54ccd9a21219ba16d5d9ac3b4b065bc/src/chart_types/xy_chart/state/utils/utils.ts#L308
- https://github.com/elastic/elastic-charts/blob/85b6ed3cf54ccd9a21219ba16d5d9ac3b4b065bc/src/chart_types/xy_chart/rendering/line.ts#L40

nickofthyme:
Sweet! So the idea behind the fitting functions is that there is data that may be `null` or `undefined` and we provide a few options on how these values can be _filled_. Filled values appear with subdued areas and dashed lines. You can see this better in this [story](https://elastic.github.io/elastic-charts/?path=/story/mixed-charts--fitting-functions-non-stacked-series).


https://user-images.githubusercontent.com/19007109/109844477-c9fcac80-7c11-11eb-85f6-feb3e1c368a7.mp4

Since there are possibly curved lines we must render the full line/area `path` for both the "real" data and the fitted data. Then with these paths, we determine the `clippedRanges` or the segments along the x-axis that represent fitted data, then this can be used to clip the paths and merge them into a single chart. So an illustration of that would be something like...

image

Where the red represents the `clippedRanges` as a list of intervals (i.e. `[[0,1], [3,4]]`) which is used to clip the "real" path and the blue represents the inverse of the `clippedRanges` to clip the "fitted" path.

These `clippedRanges` are computed here for area series...

https://github.com/elastic/elastic-charts/blob/85b6ed3cf54ccd9a21219ba16d5d9ac3b4b065bc/src/chart_types/xy_chart/rendering/area.ts#L74

And here for line series...

https://github.com/elastic/elastic-charts/blob/85b6ed3cf54ccd9a21219ba16d5d9ac3b4b065bc/src/chart_types/xy_chart/rendering/line.ts#L83

So with the example in the issue, the clipped ranges return `[]` which clips nothing and hence the issue. So I would look into the `getClippedRanges` function here...

https://github.com/elastic/elastic-charts/blob/85b6ed3cf54ccd9a21219ba16d5d9ac3b4b065bc/src/chart_types/xy_chart/rendering/utils.ts#L67-L98

Maybe we could check if all the values are `null` then return the full domain within the `getClippedRanges` function or it might need to be done outside of that function (i.e. where that function is called) to have access to the necessary values. I;m not sure.

Let me know if you have any questions about that and if you run into any issues.

nickofthyme:
Sorry forgot to mention, it's totally fine to make the changes in this pr, just remove the filtering logic on the data.

nickofthyme:
jenkins test this please

src/chart_types/xy_chart/rendering/utils.ts
```diff
@@ -67,6 +67,25 @@ export function isDatumFilled({ filled, initialY1 }: DataSeriesDatum) {
 export function getClippedRanges(dataset: DataSeriesDatum[], xScale: Scale, xScaleOffset: number): ClippedRanges {
   let firstNonNullX: number | null = null;
   let hasNull = false;
+
+  const completeDatasetIsNull = dataset.filter((data) => !isDatumFilled(data)).length === 0;
```
nickofthyme, line 71: This could be simplified a little

```suggestion
  const completeDatasetIsNull = dataset.every((datum) => isDatumFilled(datum));
```

src/chart_types/xy_chart/rendering/utils.ts
```diff
@@ -67,6 +67,25 @@ export function isDatumFilled({ filled, initialY1 }: DataSeriesDatum) {
...
+      }
+
+      const xValue = xScaled - xScaleOffset + xScale.bandwidth / 2;
+      const endXValue = xScale.range[1] - xScale.bandwidth * (2 / 3);
+      if (xValue === endXValue) {
+        acc.push([0, xValue]);
+      }
+
+      return acc;
+    }, []);
```
nickofthyme, line 87: This would work just fine, however since we know we need to clip the whole `range` we can just pull that off the `xScale`.

```suggestion
  if (completeDatasetIsNull) {
    const [min, max] = xScale.range;
    return [[min, max - xScale.step / 2]];
  }
```

I don't think we _really_ need to account for the `step` at the end since this would only apply to this series, but rather be safe than sorry.

nickofthyme (commented):
Nicely done! Just a few simplifications.

src/chart_types/xy_chart/rendering/utils.ts
```diff
@@ -67,6 +67,14 @@ export function isDatumFilled({ filled, initialY1 }: DataSeriesDatum) {
 export function getClippedRanges(dataset: DataSeriesDatum[], xScale: Scale, xScaleOffset: number): ClippedRanges {
   let firstNonNullX: number | null = null;
   let hasNull = false;
+
+  const completeDatasetIsNull = dataset.every((datum) => isDatumFilled(datum));
+
+  if (completeDatasetIsNull) {
+    const [min, max] = xScale.range;
+    return [[min, max - xScale.step / 2]];
+  }
```
nickofthyme, line 76: Sorry @anishagg17 I edited this to exclude the step. You are too quick 💨 

```suggestion
  if (completeDatasetIsNull) return [xScale.range];
```

nickofthyme:
jenkins test this please

nickofthyme:
jenkins test this please

nickofthyme (approved):
LGTM

anishagg17:
Screenshot 2021-03-05 at 12 12 32 AM

Due to this error, I made a small change.

nickofthyme:
Sorry about all the updating I had to make some updates to `master`. I'm done now.

nickofthyme:
jenkins test this please

nickofthyme:
@anishagg17 all good to merge this whenever you want 👍🏼

markov00 (approved):
LGTM, thanks @anishagg17 for your work on this, we really appreciate it

anishagg17:
> @anishagg17 all good to merge this whenever you want 👍🏼

I can't merge this, Since this is a community submitted pull request.

nickofthyme:
Oh yeah true. Done! Thanks again 🎉

PR #1038, opened 2021-02-28
feat(wordcloud): wordcloud

Kati-dev-hu:
## Summary

Initial wordcloud implementation

image
image
image

### Checklist


markov00:
Hi @Kati-dev-hu thanks so much for that, we really appreciate it!
I've played a bit with the examples and I've to say that it looks amazing!
I've compared it to the Kibana implementation and looks like we have nearly all required features. What is missing is the following:
- the user can select the type of scale applied to each text weight: linear, log and square root are the options here. 
- There are some specific cases where, due to the applied constraints like the number of tags, font size, etc, some words are filtered out (you can try that just by reducing the width of the chart to a very small size). Kibana notify the user about that with a small icon over imposed on the chart with a tooltip
Screenshot 2021-03-05 at 11 39 15
There can be several ways to achieve this, for now, it also fine living without it, but it will be great at least to have an internal logic that checks if all the words will be rendered or not. We can then use that somehow to expose that information to Kibana/other consumers

Anyway I think this already works very well and looks very clean, thanks for this big help!

Kati-dev-hu:
Hi @markov00, thanks for your comments. Text weight is resolved, and the logic for missing words is coming up too, pushing commits soon

image

Kati-dev-hu:
Commit pushed:
1/ warning in case not all words can be placed, 
2/ text weight amongst knobs (weightFun):

image

monfera:
jenkins test this please

monfera:
jenkins test this please

src/chart_types/wordcloud/layout/types/viewmodel_types.ts
```diff
@@ -0,0 +1,162 @@
...
+  fontStyle: 'italic',
+  minFontSize: 10,
+  maxFontSize: 50,
+  spiral: 'archimedean',
+  exponent: 3,
+  data: [],
+  weightFun: 'exponential',
+  outOfRoomCallback: (wordCount, renderedWordCount) => {
+    Logger.warn(`Not all words have been placed: ${renderedWordCount} words rendered out of ${wordCount}`);
+  },
```
markov00, line 142: I'm not 100% sure about having that console log by default. We can probably have a no-op here

src/chart_types/wordcloud/renderer/svg/connected_component.tsx
```diff
@@ -0,0 +1,329 @@
...
+  private drawCanvas() {
+    if (this.ctx) {
+      /*      const { width, height }: Dimensions = this.props.chartContainerDimensions;
+                renderCanvas2d(this.ctx, this.devicePixelRatio, {
+             ...this.props.geometries,
+             config: { ...this.props.geometries.config, width, height },
+           });
+      */
+    }
+  }
```
markov00, line 295: cam we completely remove the canvas element here and on the rendering? If we are not using it right now it's better to clean this up

src/chart_types/wordcloud/state/chart_state.tsx
```diff
@@ -0,0 +1,142 @@
...
+  getPointerCursor() {
+    return DEFAULT_CSS_CURSOR;
+  }
+
+  isTooltipVisible(globalState: GlobalChartState) {
+    return { visible: isTooltipVisibleSelector(globalState), isExternal: false };
+  }
+
+  getTooltipInfo(globalState: GlobalChartState) {
+    return getTooltipInfoSelector(globalState);
```
markov00, line 105: The tooltip doesn't bring any new info as it is proposed today. We can easily disable it

markov00 (approved):
Everything looks great, I've added few comments that we can easily fix on a subsequent PR.
Thank you very much for this great contribution!

Kati-dev-hu:
Hi @markov00 Thanks for the review, I created the follow-up [issue](#1083) with these steps

PR #1039, opened 2021-03-01
fix: rounding values on stacked w percentage charts

markov00:
## Summary

With stacked charts, in percentage mode, we use a default data domain of `[0,1]` instead of computing it from the stacked data.
On the rendering side, we remove geometries in which the data point falls outside the computed domain.
There are cases where the stack computes function, returns floating-point errors, and in particular, with such a small interval `[0,1]` this can cause a geometry not to be rendered because the value falls, for example, right above the `1` by a floating-point error. If the floating-point error falls internally on the domain, is fine, the rendering will take place and the error is barely noticeable if the error falls externally then the geometry is not rendered at all as in the example below.
This PR fixes this issue, clamping the stacked as percentage values in the exact interval between 0 and 1.

The floating-point error on the computed value in the example is `1.0000000000000002`

Before:
Screenshot 2021-03-01 at 14 01 08

After:
Screenshot 2021-03-01 at 14 00 44

### Checklist


nickofthyme (approved):
LGTM

PR #1057, opened 2021-03-05
fix: filter out zero values on fitted log domains

### LINKED ISSUE #1043: Using logMinLimit should filter out zero values from domain
When using `logMinLimit` with `domain.fit` set to `true`, we are computing the fitted domain to apply the `logMinLimit`. The problem is that if there are any `0` values in the data set the domain become `[0, y]` and the `logMinLimit` now acts like setting the `domain.min`.

I propose that if `domain.fit` is `true`, a `logMinLimit` value is set and `Scale.type === 'log'` then we ignore zero values when computing the domain.

The reasoning is that the domain is set to **INCLUDE** certain values inside of the domain. However with log scales, any value of `0` is never **INCLUDABLE** thus removing them is the more accurate functionality.

image

> Notice the screenshot above with the codesandbox example. With the `logMinLimit` set to `0.0001`, the min domain value is also set to `0.0001` because the domain extents is `[0, 100]`. But if the zeros are excluded from the domain extents, the domain becomes `[0.01, 100]` and with a with a `logMinLimit` set to `0.0001` the min value would be `0.01` because that is the smallest non-zero value in the dataset greater than `0.0001`.

See [codesandbox](https://codesandbox.io/s/stupefied-swirles-rkq98?file=/src/App.tsx) example

**Additional context**
Line `104` should account for the scale type and `fit` for computing the extent.

https://github.com/elastic/elastic-charts/blob/85b6ed3cf54ccd9a21219ba16d5d9ac3b4b065bc/src/utils/domain.ts#L99-L111

**Checklist**

nickofthyme:
## Summary

fix #1043 

- fit log data to only non-zero values
- fix `logFormatter` utils method sign (+/-)

Fitting the domain to the data was a feature added some time ago but the log data fit does not abide by the same rules as the others. Thus if we have a dataset of `[1, 0, 10, 0, 100]` to domain is computed as `[0,100]` which on a log scale is never possible and hence _breaks_ the ability to _fit_ the domain.

This PR filters out zero values before computing the domain to more accurately fit the data.

https://user-images.githubusercontent.com/19007109/110048453-7ffcef00-7d15-11eb-8e71-6a20e2de6873.mp4

### Checklist


stories/test_cases/2_filter_zero_values_log.tsx
```diff
@@ -0,0 +1,60 @@
...
+        id="count"
+        position={Position.Left}
+        domain={{ fit, logMinLimit }}
+        tickFormat={logFormatter(LogBase.Common)}
+      />
+      <Axis id="x" position={Position.Bottom} />
+      <AreaSeries
+        id="bars"
+        name="amount"
+        xScaleType={ScaleType.Ordinal}
```
markov00, line 44: We should not use an area or line series with an ordinal X scale, as area/line describe continuity along the X, that is not exactly what a categorical scale is intended, there can be exceptions but I rather prefer having cleaner examples than edge cases

markov00 (commented):
I think this is fine, but I've doubts about applying filters when computing the data domain.
I think that the domain of my data set should be computed independently from the applied scale. The domain should be always the same, then you can apply transformations to limit what is rendered or to adjust specific calculations when a scale is applied.
A log scale could easily also compute the `log(0)` is then the renderer that needs to take an active role on rendering something nicely for the user

stories/utils/formatters.ts
```diff
@@ -47,8 +47,9 @@ export const logBaseMap = {
   [LogBase.Natural]: Math.E,
 };
 
-export const logFormatter = (base: LogBase = LogBase.Common) => (n: number) => {
-  const sign = n < 1 ? '-' : '';
+export const logFormatter = (base: LogBase = LogBase.Common) => (n: number): string => {
+  if (n === 0) return '0';
+  const sign = n < 0 ? '-' : '';
```
monfera, line 52: OK so right below is the place where the absolute value is taken. Not sure if we should take the absolute value and log that one:

- if there's a mix of negative and positive x values to `log(abs(x))` then both 7 and -7 will map to 1.94, so a chart will be misleading
- if we reapply the sign, ie. in effect show `sign * log(abs(x))` and/or format strings by potentially attaching a `"-"` sign, then it's misleading too, because negatives are already allowed to come from `log`. Eg. `log(0.143)` is -1.94
- also, in this case, assuming a reattachment of the sign, and a _shown_ value ends up being `-1.94` then did it come from logging a positive 0.143 or a negative 7? How can the user tell?
- If we allow logging positives and negatives (via absolute values), and they all go "above" the x axis, then they'll be unduly interspersed; if the negatives go "below" the horizontal axis, then there's essentially an infinite distance between Y=1 and Y=-1

So maybe I'm just not fully grasping what we're trying to solve by admitting negatives, and apparently, a mixture of negative and positive numbers. 

There is a somewhat standard scale type for negative and positive numbers that try to logarithmic-ishly compress both sides of 0 examples [here](https://codepen.io/stopyransky/pen/OdREzG) but that's not a log scale, it's a [bisymmetric log transformation](https://observablehq.com/@d3/continuous-scales#scale_symlog) which handles the issue that numbers in [0, 1] would be negative with a regular, simple log function. This has configurability for the aforementioned problem: the distances around zero are arbitrary. Illustrated in the above links

monfera (commented):
Asked for clarification as the preexisting code seems to aim at handling negative values too; I suppose there is a spot which rules out mixing positives and negatives, even then I'm not clear on how the double negation would work with e.g. -1, ie. `log(0.1)` which is negative, then we'd apply another negative sign due to `const sign = n < 0 ? '-' : '';` but maybe we prevent this from happening

src/utils/domain.ts
```diff
@@ -99,9 +100,11 @@ export function computeDomainExtent(
 export function computeContinuousDataDomain(
   data: any[],
   accessor: (n: any) => number,
+  scaleType: ScaleType,
   domainOptions?: YDomainRange | null,
 ): ContinuousDomain {
-  const range = extent<any, number>(data, accessor);
+  const filteredData = domainOptions?.fit && scaleType === ScaleType.Log ? data.filter((d) => accessor(d) !== 0) : data;
```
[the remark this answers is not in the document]
nickofthyme: This PR doesn't impact the negative log domain. This is handled on the scale here...

https://github.com/elastic/elastic-charts/blob/c39d11388f74bbaef1dc2ed0b4febabd25b35241/src/scales/scale_continuous.ts#L57-L118

stories/utils/formatters.ts
nickofthyme: This is just a log formatter to make the log scale at large numbers easier to see in exponential notation. I think you are looking for https://github.com/elastic/elastic-charts/pull/1057#discussion_r604903788

The `sign` value here is the sign of the overall constant, _not_ the exponent. Example formatted values is shown below.

| value | formatted value |
| :-: | :-: |
| 0.001 | 1 x 10⁻³ |
| 0.01 | 1 x 10⁻² |
| 0.1 | 1 x 10⁻¹ |
| 1 | 1 x 10⁰  |
| 10 | 1 x 10¹  |
| 100 | 1 x 10²  |

stories/test_cases/2_filter_zero_values_log.tsx
nickofthyme: b30fa59

markov00 (approved):
I've checked again and it looks ok. We are already affecting the domain with the padding, so I'm also fine affecting it filtering out zero values in this specific case

Issue #767, opened 2020-07-29
Rename enum types to singular 


markov00:
**Is your feature request related to a problem? Please describe.**
Every enum types should be aligned to use singular name, instead of plural.

**Describe the solution you'd like**
Prefer singular over plural increase readability of code:
```
// bad
const chartType = ChartTypes.XYAxis

// good
const chartType = ChartType.XYAxis

```

PR #1064, opened 2021-03-08
refactor: rename enum types to singular

wayneseymour:
Simple search and replace...low hanging fruit.
Also, not sure if the markdown files should be included in this refactor?

fix https://github.com/elastic/elastic-charts/issues/767
## Summary

I was searching for an easy pr to start understanding this repo.

I think there are still more enums to fixup, such as:
https://github.com/elastic/elastic-charts/blob/dce139a1e72b2ad136cb09dc4c705dcfff1720e6/src/specs/constants.ts#L27

### DIFF
--- a/api/charts.api.md
+++ b/api/charts.api.md
@@ -44,13 +44,13 @@ export interface AngleFromTo {
 }
 
 // @public
-export type AnnotationDomainType = $Values<typeof AnnotationDomainTypes>;
-
-// @public
-export const AnnotationDomainTypes: Readonly<{
+export const AnnotationDomainType: Readonly<{
     XDomain: "xDomain";
     YDomain: "yDomain";
 }>;
+
+// @public
+export type AnnotationDomainType = $Values<typeof AnnotationDomainType>;
 
 // @public (undocumented)
 export type AnnotationId = string;
@@ -67,17 +67,17 @@ export type AnnotationSpec = LineAnnotationSpec | RectAnnotationSpec;
 // @public (undocumented)
 export type AnnotationTooltipFormatter = (details?: string) => JSX.Element | null;
 
-// @public (undocumented)
-export type AnnotationType = $Values<typeof AnnotationTypes>;
-
-// Warning: (ae-missing-release-tag) "AnnotationTypes" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
+// Warning: (ae-missing-release-tag) "AnnotationType" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
 //
 // @public (undocumented)
-export const AnnotationTypes: Readonly<{
+export const AnnotationType: Readonly<{
     Line: "line";
     Rectangle: "rectangle";
     Text: "text";
 }>;
+
+// @public (undocumented)
+export type AnnotationType = $Values<typeof AnnotationType>;
 
 // Warning: (ae-missing-release-tag) "ArcSeriesStyle" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
 //
@@ -107,7 +107,7 @@ export const AreaSeries: React_2.FunctionComponent<SpecRequiredProps & SpecOptio
 
 // @public
 export type AreaSeriesSpec = BasicSeriesSpec & HistogramConfig & Postfixes & {
-    seriesType: typeof SeriesTypes.Area;
+    seriesType: typeof SeriesType.Area;
     curve?: CurveType;
     areaSeriesStyle?: RecursivePartial<AreaSeriesStyle>;
     stackMode?: StackMode;
@@ -170,7 +170,7 @@ export type AxisId = string;
 // @public
 export interface AxisSpec extends Spec {
     // (undocumented)
-    chartType: typeof ChartTypes.XYAxis;
+    chartType: typeof ChartType.XYAxis;
     domain?: YDomainRange;
     gridLine?: Partial<GridLineStyle>;
     groupId: GroupId;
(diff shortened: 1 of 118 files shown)


monfera:
Thanks @wayneseymour for your PR! 

- The `api/charts.api.md` file can be regenerated with `yarn build:ts && yarn api:extract` and the output file (that gets output a different location) can be copied in place of `api/charts.api.md` and checked eg. via diffing if the changes are as you expect
- The PR itself looks good to me, wondering @markov00 what our policy is for approving/merging PRs that break the API without a linked requirement by the dependency, Kibana—I guess it's best to batch such changes to minimize disruption. But batching means, longer wait till it hits `master`, with inevitable merge conflicts, as a maintenance PR impacts many lines. Future additions of `ChartTypes` also won't be covered (TS checks should catch this). Ie. there's more work than usual on the part of the PR maker and whoever works on future merges.

Tre' one note is that sometimes we're assisting Kibana users in such API changes, eg. via small PRs there by the chart PR maker 😉 That's also a reason for, I think, merging a bunch of API breaking changes in one fell swoop, Marco what do you think, with 7.12 around the corner?

markov00:
Thank you Tre'!
This should definitely include the `BREAKING CHANGE` footer
We can merge this and update 7.13 branch, it will not be merged on 7.12 as we can't introduce breaking changes after FF

I think it's fine to include multiple changes here, some enums are not currently used much outside the library

nickofthyme:
Thanks @wayneseymour this issue is very near and dear to my ❤️ 

@monfera I think this is just a nice to have that will eventually get into `7.13` so I don't think we need to be too critical of batching other PRs nor waiting to get others merged before this as we could just backport any critical changes to `7.12` in the few remaining days. Also we end up having to merge the PRs so one of us should be sure to note the `BREAKING CHANGE`.

wayneseymour:
> Thanks @wayneseymour for your PR!
> 
> * The `api/charts.api.md` file can be regenerated with `yarn build:ts && yarn api:extract` and the output file (that gets output a different location) can be`cp tmp/charts.api.md api/charts.api.md`-ed in place of `api/charts.api.md` and checked eg. via diffing if the changes are as you expect
> * The source change looks good, wondering @markov00 what our policy is for approving/merging PRs that break the API without a linked requirement by the dependency, Kibana—I guess it's best to batch such changes to minimize the # of BREAKING. But batching means, longer wait till it hits `master`, with inevitable merge conflicts, as a maintenance PR impacts many lines. Future additions of `ChartTypes` also won't be covered (TS checks should catch this). Ie. there's more work than usual on the part of the PR maker and whoever works on future merges
> * It's good to run tests locally too, including, in such cases when many files are changed, the image tests, glad to help setting it up; besides firing up storybook and testing a variety of charts manually
> * Tre' a note for merge time, `elastic-charts` has commit message content conventions for breaking changes, see [here](https://github.com/elastic/elastic-charts/blob/master/CONTRIBUTING.md?rgh-link-date=2021-03-10T12%3A49%3A44Z#squash-and-merge-your-pr)
> 
> Tre' one note is that sometimes we're assisting Kibana users in such API changes, eg. via small PRs there by the chart PR maker 😉 That's also a reason for, I think, merging a bunch of API breaking changes in one fell swoop, Marco what do you think, with 7.12 around the corner?

Awesome, I'll get on this next week!  "Shut it down day" tomorrow in the US :)

wayneseymour:
> Thank you Tre'!
> This should definitely include the `BREAKING CHANGE` footer
> We can merge this and update 7.13 branch, it will not be merged on 7.12 as we can't introduce breaking changes after FF
> 
> I think it's fine to include multiple changes here, some enums are not currently used much outside the library

Perfect, I'll fixup asap!  Thanks for reviewing!

wayneseymour:
@monfera @nickofthyme Gents, I'm not sure how to proceed here.
The issue is with symbol shadowing.  In this case, the `object` and the `type` would have the same name if I refactored away the plurality.
Ideas?


Screen Shot 2021-03-26 at 1 19 55 PM

monfera:
@wayneseymour that's fine, please see eg. `LegendStrategy` or other similar symbols that get dual-bound (one value, one type) as part of our current approach of working around (or rather, avoid) TypeScript enums
```
export type LegendStrategy = Values<typeof LegendStrategy>;
```

wayneseymour:
Locally, I've run these build scripts and they passed:
 - `yarn build:ts && yarn api:extract`
 - `yarn lint:fix`
 - `yarn test`

markov00 (approved):
LGTM, scanned all the changes online and looks great!
Thanks for taking the time for that.
Just remember to add a breaking change line in the merge commit like the following:
```
your PR description

BREAKING CHANGE: `AnnotationDomainTypes`, `AnnotationType` and all the others (please list them) are renamed to ....
```

wayneseymour:
> LGTM, scanned all the changes online and looks great!
> Thanks for taking the time for that.
> Just remember to add a breaking change line in the merge commit like the following:
> 
> ```
> your PR description
> 
> BREAKING CHANGE: `AnnotationDomainTypes`, `AnnotationType` and all the others (please list them) are renamed to ....
> ```

Indeed I will!  Thanks @markov00

nickofthyme:
> Gents, I'm not sure how to proceed here.
The issue is with symbol shadowing. In this case, the object and the type would have the same name if I refactored away the plurality.
Ideas?

Hey @wayneseymour I was sick last week so I didn't see this till now and wanted to clear this up in case you ran into this in the future. I'm thinking this might be an issue with the linter because in typescript when exporting a type or value you create a separate entity for each, so the naming will _not_ clash when doing something like...

```ts
export const SeriesType = Object.freeze({
  Area: 'area' as const,
  Bar: 'bar' as const,
  Line: 'line' as const,
  Bubble: 'bubble' as const,
});
export type SeriesType = $Values<typeof SeriesType>;
```

Typescript calls this [`Declaration Merging`](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) whereby the type and value (and namespaces) are merged into one as `SeriesType` wherever it is imported. In this case, we call this a `const enum` as it is not a typescript `enum`, see https://github.com/elastic/kibana/pull/62957#discussion_r406385788 for a deeper explanation of why.

PR #1065, opened 2021-03-09
fix(tooltip): add boundary padding

### LINKED ISSUE #1062: Tooltip boundary padding
Simply applying a `boundary` is not enough to contain the tooltip positioning in kibana. The top headers in visualize library and dashboard are position fixed with a top page margin. This is fine until the boundary element can scroll, then I need to target the container of the scrolling element which has a margin that the positioning does not account for.

https://user-images.githubusercontent.com/19007109/110371512-b2a73000-8012-11eb-8a7b-54b83ba80beb.mp4

A new prop on the tooltip options for `boundaryPadding`. This would create an artificial padding around the boundary element. In the case of the example above. The `boundaryPadding` could be set to `{ top: 100 }` in which the boundary can account for the top margin to place the tooltip below the top navigation bar.

Selecting a different boundary to correctly place the tooltip, options here are limited and creating an invisible boundary element would be too verbose. 

Setting the `z-index` of the tooltip such that the tooltip renders above the header. This is undesirable UI.

**Kibana Cross Issues**
https://github.com/elastic/kibana/issues/93495

**Checklist**

nickofthyme:
## Summary

closes #1062 

Adds an artificial [`padding`](https://popper.js.org/docs/v2/utils/detect-overflow/#padding) around the specified `boundary` element to contain tooltip within boundaries with top margins.

https://user-images.githubusercontent.com/19007109/110491875-49c5c380-80b7-11eb-8949-31e3ee33131a.mp4

> **Note:** This does _not_ impact the tooltip offset for the achor.

### Checklist

### DIFF
--- a/api/charts.api.md
+++ b/api/charts.api.md
@@ -2221,6 +2221,8 @@ export interface TooltipInfo {
 // @public
 export interface TooltipPortalSettings<B = never> {
     boundary?: HTMLElement | B;
+    // Warning: (ae-forgotten-export) The symbol "Padding" needs to be exported by the entry point index.d.ts
+    boundaryPadding?: Partial<Padding_2> | number;
     fallbackPlacements?: Placement[];
     offset?: number;
     placement?: Placement;
--- a/api/charts.api.md
+++ b/api/charts.api.md
@@ -1444,6 +1444,12 @@ export interface OrderBy {
 // @public (undocumented)
 export type OrdinalDomain = (number | string)[];
 
+// Warning: (ae-forgotten-export) The symbol "PerSideDistance" needs to be exported by the entry point index.d.ts
+// Warning: (ae-missing-release-tag) "Padding" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
+//
+// @public (undocumented)
+export type Padding = PerSideDistance;
+
 // Warning: (ae-missing-release-tag) "PARENT_KEY" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
 //
 // @public (undocumented)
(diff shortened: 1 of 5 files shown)


api/charts.api.md
```diff
@@ -2221,6 +2221,8 @@ export interface TooltipInfo {
 // @public
 export interface TooltipPortalSettings<B = never> {
     boundary?: HTMLElement | B;
+    // Warning: (ae-forgotten-export) The symbol "Padding" needs to be exported by the entry point index.d.ts
```
markov00, line 2224: Do you want to expose also the `Padding` type?

src/components/portal/tooltip_portal.tsx
```diff
@@ -138,7 +154,7 @@ const TooltipPortalComponent = ({
             boundary,
             // checks main axis overflow before trying to flip
             altAxis: false,
-            padding: offset || 10,
+            padding: addToPadding(boundaryPadding, offset || 10),
```
markov00, line 157: with that `offset || 10` are we saying that we always want a padding of 10 and the user cannot set 0 here?

markov00 (approved):
LGTM, I've just added two comments

src/components/portal/tooltip_portal.tsx
nickofthyme: I'm not sure why I had `10` there. Maybe just to flip away `10px` from the edge.

Anyways I removed this and only add the `offset` value. See 8d9f3f6

monfera (approved):
LGTM!

src/components/portal/tooltip_portal.tsx
```diff
@@ -56,6 +57,20 @@ type PortalTooltipProps = {
   chartId: string;
 };
 
+function addToPadding(padding?: Partial<Padding> | number, extra: number = 0): Padding | number | undefined {
+  if (!padding) return undefined;
```
monfera, line 61: What if `padding` is 0, is it OK to return undefined?

nickofthyme: Good point, I don't think this is critical but it should return just the `extra` value when/if `padding` is `undefined`. I'll open a pr.

nickofthyme: @monfera fixed here https://github.com/elastic/elastic-charts/pull/1088

PR #1071, opened 2021-03-10
fix(brush): force brush tool per panel

### LINKED ISSUE #1070: Brushing on small multiples doesn't consider the multiples
**Describe the bug**
When brushing on a small multiples the following issues appear:
- in every small multiple and brush configuration, the brush is not limited to the single trellis, but can cover multiple charts. 
- this causes a wrongly returned set of scaled values, that covers more then the current trellis extent

**To Reproduce**
1. Goto https://codesandbox.io/s/quirky-flower-k1sst?file=/src/App.tsx
2. Brush the chart
3. When brushing more then one chart you start seeing the error reported on the timestamps

**Expected behaviour**
The brush should limit itself to the panel the brush event was started

 - Elastic Charts: since small multiple feature

**Errors in browser console**
n/a

**Kibana Cross Issues**
- [ ] https://github.com/elastic/kibana/issues/94252

markov00:
## Summary

This PR forced the brush tool to work per panel when using small multiples. This behavior is built to works independently from the presence or absence of a small multiples configuration.
It works by forcing the brushable area to be constraint between the boundaries of the panel where the brushing event started. The brush selection doesn't move between panels and the coordinate of the mouse down event determines the limiting panel.

I've cleaned up the code to compute the brush area and also updated the unit test by mocking the SmallMultiple specs

fixes #1070

https://user-images.githubusercontent.com/1421091/110787477-a66adf00-826d-11eb-9f4e-15f0b89c857a.mp4

### DIFF
--- a/src/chart_types/xy_chart/state/selectors/get_brush_area.ts
+++ b/src/chart_types/xy_chart/state/selectors/get_brush_area.ts
@@ -66,6 +66,7 @@ export const getBrushAreaSelector = createCachedSelector(
   },
 )(getChartIdSelector);
 
+/** @internal */
 export type PanelPoints = {
   start: Point;
   end: Point;
@@ -75,6 +76,7 @@ export type PanelPoints = {
   vPanelHeight: number;
 };
 
+/** @internal */
 export function getPointsConstraintToSinglePanel(
   startPlotPoint: Point,
   endPlotPoint: Point,
--- a/src/chart_types/xy_chart/state/selectors/get_brush_area.ts
+++ b/src/chart_types/xy_chart/state/selectors/get_brush_area.ts
@@ -89,11 +89,18 @@ export function getPointsConstraintToSinglePanel(
   const vPanelStart = vertical.scale(vPanel) ?? 0;
   const vPanelEnd = vPanelStart + vertical.bandwidth;
 
-  const x = clamp(endPlotPoint.x, hPanelStart, hPanelEnd);
-  const y = clamp(endPlotPoint.y, vPanelStart, vPanelEnd);
+  const start = {
+    x: clamp(startPlotPoint.x, hPanelStart, hPanelEnd),
+    y: clamp(startPlotPoint.y, vPanelStart, vPanelEnd),
+  };
+  const end = {
+    x: clamp(endPlotPoint.x, hPanelStart, hPanelEnd),
+    y: clamp(endPlotPoint.y, vPanelStart, vPanelEnd),
+  };
+
   return {
-    start: startPlotPoint,
-    end: { x, y },
+    start,
+    end,
     hPanelStart,
     hPanelWidth: horizontal.bandwidth,
     vPanelStart,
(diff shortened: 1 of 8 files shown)


src/chart_types/xy_chart/state/selectors/get_brush_area.ts
```diff
@@ -43,81 +44,115 @@ export const getBrushAreaSelector = createCachedSelector(
...
-    left: getLeftPoint(chartDimensions, start),
-    top: getTopPoint(chartDimensions, start),
-    height: getMinimalHeight(start, end),
-    width: getMinimalWidth(start, end),
+    left: rotated ? hPanelStart : start.x,
+    top: rotated ? start.y : vPanelStart,
+    height: rotated ? getMinSize(start.y, end.y) : vPanelHeight,
+    width: rotated ? hPanelWidth : getMinSize(start.x, end.x),
   };
 }
```
nickofthyme, line 125: These diffs are horribly misaligned 🙄 ....git 😞

src/chart_types/xy_chart/state/selectors/on_brush_end_caller.ts
```diff
@@ -93,55 +97,70 @@ export function createOnBrushEndCaller(): (state: GlobalChartState) => void {
...
             if (brushArea.x !== undefined || brushArea.y !== undefined) {
               onBrushEnd(brushArea);
             }
           }
           prevProps = nextProps;
         },
-      )({
-        keySelector: (state: GlobalChartState) => state.chartId,
-      });
+      )(getChartIdSelector);
```
nickofthyme, line 122: 👍🏼

stories/small_multiples/1_grid.tsx
```diff
@@ -91,6 +91,7 @@ export const Example = () => {
               },
             },
           }}
+          onBrushEnd={(d) => console.log(d)}
```
nickofthyme, line 94: action here instead maybe?

nickofthyme (approved):
Code LGTM, tested locally and fixes issue. I don't see any issues with brushing on non-small multiple charts.

stories/small_multiples/1_grid.tsx
markov00: will do thanks

src/chart_types/xy_chart/state/selectors/get_brush_area.ts
markov00: I know, I've refactored a bit the code cleaning up it a bit, but the diff is a bit messy, better to check both file side by side

stories/small_multiples/1_grid.tsx
markov00: done https://github.com/elastic/elastic-charts/pull/1071/commits/57c86bf4e88da627a8e35e76d5a140bc6acdd3da

src/chart_types/xy_chart/state/selectors/get_brush_area.ts
rshen91: This might just be me misunderstanding, but is the brush area adjusting for height less than the height of the panel?

small_multiples

nickofthyme (approved):
LGTM

PR #1077, opened 2021-03-11
fix(legend): stop legend from shifting in color picker story

### LINKED ISSUE #1075: Legend item jumps when series is disabled
**Describe the bug**
The legend dot jumps when using a color picker after series is disabled.

https://user-images.githubusercontent.com/19007109/110857106-0ec1bc80-827e-11eb-948f-13d0e350cbc4.mp4

**To Reproduce**
1. Go to https://elastic.github.io/elastic-charts/?path=/story/legend--color-picker
2. Click on any series to disable it
3. Notice the label width increases causing the 👁 icon to be shifted to the left.

**Expected behaviour**
No shifting

 - Elastic Charts: `master`

rshen91:
## Summary

Fixes #1075 

Added vrt showing the update from previous behavior (shown below)
Screen Shot 2021-03-11 at 3 57 15 PM

### Checklist


src/components/legend/_legend_item.scss
```diff
@@ -1,25 +1,23 @@
...
   color: $euiTextColor;
   display: flex;
   flex-wrap: nowrap;
   justify-content: space-between;
   align-items: center;
   position: relative;
-
+  > div:not(.background) {
+    // euiPopover adds a div with height of 19px
+    height: 18px; // prevents color dot from shifting
```
nickofthyme, line 11: Let's change this to match the `euiFontSizeXS` mixin

```suggestion
    // euiPopover adds a div with height of 19px otherwise
    height: #{$euiFontSizeXS * $euiLineHeight}; // prevents color dot from shifting
```

nickofthyme: https://github.com/elastic/eui/blob/527028606dc2ff601ea4baff3ee83555e05f3a54/src/global_styling/mixins/_typography.scss#L44-L47

rshen91: thanks! Updated in 6d5a240

nickofthyme:
@rshen91 I added a few more changes now that this is not expected to go into `7.12`. Namely changes:

- share margins for focus state style spacing, only single margin space between each direct child to allow room for focus state border
- set max width on action to square of height.
- make eui popover action full size of action, this allows for matching size of focus state.


https://user-images.githubusercontent.com/19007109/111207554-c6fda680-8597-11eb-967c-f0ceda84fa13.mp4


See 8e5dd88

markov00 (changes requested):
I've quickly tested locally and looks like the legend label max width is not computed perfectly and some of them are truncated

Screenshot 2021-03-16 at 17 01 02

rshen91:
> I've quickly tested locally and looks like the legend label max width is not computed perfectly and some of them are truncated
> 
> Screenshot 2021-03-16 at 17 01 02

Hey sorry I think I'm missing something. I'm having a hard time replicating from your screenshot: 

https://user-images.githubusercontent.com/20343860/111501536-b02b9100-870a-11eb-9403-5f0d4ef7c855.mp4

markov00 (changes requested):
I don't understand why we changed so many things in this particular case.
From the linked issue, there is a position change of the color/eyeClosed icon when using the issuing a colorPicker.

If I check the initial commits of this PR I already see the correct fix, but I don't understand why then more and more style was pushed causing a very long list of visual regressions changes.

I've tested locally and with this single change:
```css
&:not(&--hidden) {
    .echLegendItem__color--changable {
      cursor: pointer;
      //height: 18px;
      //margin-left: 5px;
    }
  }
```
basically removing the margin-left and the height when adding the colorPicker, it solves the issue and probably it just requires a very limited set of visual changes on our test.
The style of the legend without the color picker remains the same, and I've just fixed the style when the color picker is added

https://user-images.githubusercontent.com/1421091/112151274-54b54380-8be1-11eb-957b-d9e23360fd66.mp4

You can test it locally adding the following code to the legend color picker story:
```tsx
<Settings showLegend legendColorPicker={boolean('add color picker', true) ? CustomColorPicker : undefined} />
```


@nickofthyme @rshen91 what is the reason to change also the rest of the style?

nickofthyme:
@markov00 you are correct, the first commits fix the issue with the color picker flickering. But since this fix was not in a rush to get into kibana I figured it would be nice to clean up the legend focus and click states to prevent flickering and sizing differences during user interaction with the legend.

Notible changes/fixes include:
shared margins to create a more narrow footprint
Screen Recording 2021-03-24 at 09 09 32 AM

Prevent bouncing when color picker is opened and closed

### Before
ezgif com-gif-maker


### After
Screen Recording 2021-03-24 at 09 15 49 AM


Forced size of action popovers. This assumes that user will use the `EuiPopover` for the actions and forces the size to be full height and width to have uniform focus state across the color dot (w/ or w/o a color picker), the series name (w/ or w/o extra value) and the actions.

### Before
image
### After
image

I am happy to put the other changes in a separate PR but they are somewhat related to each other. But yes this will affect almost all vrt screenshots. I also removed the margin-right when there is no action so the extra values has the same zero margin as before.

nickofthyme:
Replace with #1101

PR #1079, opened 2021-03-15
test(partition): dictionary encoding and prop elimination

monfera:
## Summary

- file size went down from the earlier 2.6MB to 229kB, because
- switched to proper arrays in `layers`
- elements in `layers` got dictionary encoded, string count goes from 21061 to 1396
- dictionary encoding is frequency aware, ie. more frequently used strings get shorter index numbers (reverse frequency sort)
- removed `name` and `depth` properties that the story didn't use

No visual change.

### Steps

Just documenting the console steps, I don't think it's worth making a utility out of it, as it's a userland optimization w.r.t. `elastic-charts`.

```

temp1 = data;

// let's see count of strings: 21061
temp1.reduce((p,n) => p + Object.keys(n.layers).length ,0)

// let's see the count of distinct strings: 1396, more than 1 OOM fewer, naturally
Object.keys(temp1.reduce((p,n) => Object.assign(p, ...Object.values(n.layers).map(v => ({[v]: true}))), {})).length

// collecting names as keys with values as occurrence count here
a = {}

// populate a
temp1.forEach(n => Object.values(n.layers).forEach(name => a[name] ? a[name]++ : a[name] = 1))

// inverse frequency order of entries
ordered = Object.entries(a).sort(([_, a], [__, b]) => b - a)

// we just ultimately need the simple name array ie. string[]
dictionary = ordered.map(([name]) => name)

// let's make an object with keys as names and values as indices into dictionary
lookup = Object.fromEntries(Object.entries(dictionary).map(([numberString, name]) => [name, Number(numberString)]))

// let's now switch layers from string[] to integer[]
newData = temp1.map(n => ({...n, layers: Object.values(n.layers).map(name => lookup[name])}))
```
Then I put them in an object: `{ dictionary, facts: newData }` and that's what's in the `.json` now.

### Checklist


monfera:
jenkins test this please

nickofthyme (approved):
LGTM

dgieselaar (approved):

PR #1082, opened 2021-03-17
feat(tooltip): expose datum in the TooltipValue

### LINKED ISSUE #1042: Bubble chart tooltip customization capabilities
In APM we want to implement a bubble chart with tooltips (elastic/kibana#88852). The current capabilities of tooltips are not sufficient at this time. #615 exists to improve tooltips but @markov00 asked me to open a separate issue to describe our specific needs.

The instances latency distribution chart in APM is a bubble chart that plots points for instances running a service with throughput on the X-axis and latency on the Y-axis:

image

The data points look like this:

```json
{
  "serviceNodeName": "cdee148c7ee2cef54c1b1dbb6491b925d94f16e27c5ddaa1fa371e0bce3ee8ce",
    "throughput": {
    "value": 11.1875,
    },
  "latency": {
    "value": 623125.8212290503,
     },
}
```

We would like the tooltip to look something like:

```
+------------------------------------------------------------------+
| cdee148c7ee2cef54c1b1dbb6491b925d94f16e27c5ddaa1fa371e0bce3ee8ce |
+------------------------------------------------------------------+
| Latency 62 ms                                                    |
| Throughput 11 tpm                                                |
+------------------------------------------------------------------+
```

Since the tooltipProps.header does not receive a `datum` argument we are unable to easily get the `serviceNodeName`, and by default only the value of the throughput is shown in the body.

markov00:
## Summary

Adds the datum, when possible, to the `TooltipValue` object. 
Currently available for cartesian, gauge/goal, heatmap.
Disabled for partition charts as we cannot represent a single datum for a partition chart due to its aggregated and nested nature.

fix #1042


rshen91 (approved):
LGTM! Tested locally with playground - passing different types of test value(s) to the `datum` (when set to true) in tooltipOptions and then passing the options through tooltip in settings works 👍 . I like the option for users if they need it and setting it to default off the bat seems great.

darnautov (approved):
LGTM

PR #1084, opened 2021-03-22
feat(a11y): add basic aria-label to canvas element

rshen91:
## Summary

This PR adds `Chart` to the canvas element for Cartesian charts. Tested with VoiceOver and safari. Added a jest puppeteer test for the accessibility object model showing that the alternative text is indeed in the AOM. 

### Checklist


rshen91:
jenkins test this

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
```diff
@@ -168,6 +169,9 @@ class XYChartComponent extends React.Component<XYChartProps> {
           width,
           height,
         }}
+        aria-label="This is a chart"
```
myasonik, line 172: Can probably shorten this just to "chart" otherwise we could start every img alt text with "this is..." or something like it.
```suggestion
        aria-label="Chart"
```

myasonik (approved):
It's a start!

It doesn't _solve_ Charts' accessibility issues but it takes the first step.

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
rshen91: Thank you! Addressed in 36a71d8

markov00 (approved):
Good to merge after solving merge conflicts

PR #1088, opened 2021-03-24
chore: always use offset with flip modifier

nickofthyme:
## Summary

Adds `offset` back to `flip` modifier regardless of `padding`.

### Checklist


markov00 (approved):
LGTM

monfera (approved):
Looks good Nick!

PR #1090, opened 2021-03-25
chore: use LF as new line configuration on API-extractor

markov00:
## Summary

By default, `api-extractor` uses `CRLF` as line-ending configuration.
Since our code uses `LF`, as specified in the `.editorconf` file, it's better to align to that configuration for every file.
I'm using a configuration introduced here: https://github.com/microsoft/rushstack/pull/1626

I've noticed multiple situations where the generated export is interpreted in github/diff viewer as completely new due to this different line endings configuration. I hope with that to avoid completely this issue in the future

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -3,5 +3,6 @@
   "singleQuote": true,
   "trailingComma": "all",
   "printWidth": 120,
-  "arrowParens": "always"
+  "arrowParens": "always",
+  "endOfLine": "lf"
 }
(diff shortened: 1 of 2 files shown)


monfera (approved):
LGTM, I tested locally and worked.

Optional: do we also want to add 
```
"endOfLine":"lf" // prettier.io is down, didn't double-check
```
to `.prettierrc` too? I think that, without that, a Windows contributor ends up with CRLF, unless locally setting certain things I already forgot

PR #1094, opened 2021-03-29
feat(partitions): Small multiples legends

monfera:
## Summary

Flat legends:
image
Hierarchical legends:
image

### Checklist


markov00 (commented):
I've few doubts on what was changed here:
https://github.com/elastic/elastic-charts/blob/9aa1d202b6ccc55dc3169f2a100a3b9ba3943761/integration/tests/__image_snapshots__/legend-stories-test-ts-legend-stories-should-have-the-same-order-as-nested-with-no-indent-even-if-there-are-repeated-labels-1-snap.png

I've added the legendStrategy  to that story and it looks a bit wrong: it doesn't work correctly with all the strategies. Is that a wanted behaviour?

markov00 (approved):
LGTM, agreed on the new partition flat legend alphabetical order.
Lens uses the flat version with only the root layer or the hierarchical version with all layers.

PR #1102, opened 2021-03-31
fix(legend): action sizing ui and focus states

nickofthyme:
## Summary

Overall fixes for action UI to match rest of legend

https://user-images.githubusercontent.com/19007109/113190922-8a31e080-9222-11eb-8fff-7ab4a4113834.mp4

## Notible changes/fixes include:

shared margins to create a more narrow footprint
Screen Recording 2021-03-24 at 09 09 32 AM

Forced size of action popovers. This assumes that user will use the `EuiPopover` for the actions and forces the size to be full height and width to have uniform focus state across the color dot (w/ or w/o a color picker), the series name (w/ or w/o extra value) and the actions.

### Before
image
### After
image

### Checklist

- [x] This was checked for cross-browser compatibility -- `:focus` is problematic in safari https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#clicking_and_focus


markov00 (approved):
Visually checked all the screenshots!
LGTM

src/components/legend/legend.tsx
```diff
@@ -79,7 +79,8 @@ function LegendComponent(props: LegendStateProps & LegendDispatchProps) {
     containerDimensions,
     config,
   } = props;
-  if (items.length === 0) {
+
+  if (items.length === 0 || items.every(({ isItemHidden }) => isItemHidden)) {
     return null;
   }
```
nickofthyme, line 86: This does not free-up space for the chart when there is no legend rendered. We need to find a better way to set the size to zero in the `get_legend_size` selector without creating an infinite loop.

PR #1103, opened 2021-04-01
ci: fix release push to protected branches

nickofthyme:
## Summary

Attempt to fix pushing to protected branches in actions via https://github.com/semantic-release/github/issues/175#issuecomment-757448750


markov00 (approved):
Let's try

Issue #1100, opened 2021-03-31
[XY] negative bars are not clickable


markov00:
**Describe the bug**
Negative bars on a bar chart are not clickable

**To Reproduce**
1. Go to https://codesandbox.io/s/agitated-benz-indos?file=/src/App.tsx
2. Click on the negative bar
3. The console log is not triggered

**Expected behaviour**
The chart should trigger on element click independently if it's negative or positive value.

**Version (please complete the following information):**
 - Elastic Charts: master

**Errors in browser console**
n/a

**Kibana Cross Issues**
- [ ] Lens issue

PR #1109, opened 2021-04-06
fix(xy): negative highlight and click

markov00:
## Summary

fix https://github.com/elastic/elastic-charts/issues/1100

The logic to compute the intersection between the mouse click coordinates and the geometry doesn't cover bars with negative height.
I've fixed that making the actual geometry renders with positive values instead.
This also fixed the missing highlight of negative bars.


nickofthyme (approved):
Code changes LGTM. Tested locally on storybook.

Issue #1097, opened 2021-03-30
Allow passing description props to `<canvas>` element


myasonik:
## Describe the bug
There is no way to set `aria-describedby` on a chart.

## Notes on implementation
- With the end goal being something akin to [this demo](https://codepen.io/myasonik/pen/yLVXpKr?editors=0010), I'd recommend not consuming `aria-*` props directly because in the long-term we won't use many. Rather, I'd recommend using `describedby`. This, I hope, is both clear as a standalone prop name and hopefully hints at what their implementation might be/mean.
- I'd also recommend accepting `description` as a prop. [`aria-description`](https://w3c.github.io/aria/#aria-description) is currently a spec'ed property (that isn't implemented anywhere) so it serves to future proof a little and it's easy to shim in Charts to provide a nicer API for consumers. (If a user passes in a `description`, render their text to a visually hidden `<p>` with a random ID and add `aria-describedby` to the `<canvas>` pointing to that ID.

PR #1111, opened 2021-04-08
feat(a11y): allow user to pass custom description for screen readers

rshen91:
## Summary

Fixes https://github.com/elastic/elastic-charts/issues/1097
This PR allows the user of elastic charts to pass in a `description` via the `<Settings />` component (only for xy_charts) to be read by a screen reader. 

In the case where users descriptions are informational enough, they are able to disable the default generated series types through `useDefaultSummary`, which by default is true. 

### Checklist

- [x] Updated storybook to show functionality (`a11y_custom_description` within test_cases to test out with VoiceOver on Safari.)


api/charts.api.md
```diff
@@ -1732,9 +1732,11 @@ export interface SettingsSpec extends Spec, LegendSpec {
     animateData: boolean;
     baseTheme?: Theme;
     brushAxis?: BrushAxis;
+    customDescription?: string;
     debug: boolean;
     // @alpha
     debugState?: boolean;
+    disableGeneratedSeriesTypes: boolean;
```
myasonik, line 1739: Just reading this prop name, I'm not sure what it means... What do you think about something like `disableGeneratedDescription`?

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
```diff
@@ -78,6 +80,8 @@ export interface ReactiveChartStateProps {
   annotationSpecs: AnnotationSpec[];
   panelGeoms: PanelGeoms;
   seriesTypes: Set<SeriesType>;
+  customDescription: string | undefined;
```
myasonik, line 83: I think it's effectively equivalent but I think defining this as an optional prop communicates the intent a little better. 

```suggestion
  customDescription?: string;
```

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
```diff
@@ -178,11 +184,19 @@ class XYChartComponent extends React.Component<XYChartProps> {
           }}
           // eslint-disable-next-line jsx-a11y/no-interactive-element-to-noninteractive-role
           role="presentation"
+          aria-describedby="get_custom_description"
```
myasonik, line 187: You'll want to replace this with a randomly generated `id` – if a user renders two charts, each with a unique description, you'll end up getting only the first `id` read off for both charts.

You'll also want to conditionally add the attribute. An empty description will sometimes still get read out (something like "description empty") but probably just want to ignore it.

```suggestion
		  {...(customDescription ? {'aria-describedby': randomId} : {})}
```

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
myasonik, line 83: We should also accept something along the lines of a `descriptionId`. If the description is already rendered somewhere else on the page, we want users to be able to just point to it.

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
markov00: you can use the `chartId`, that is random and unique per chart, + some postfix

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
```diff
@@ -266,6 +282,8 @@ const mapStateToProps = (state: GlobalChartState): ReactiveChartStateProps => {
     annotationSpecs: getAnnotationSpecsSelector(state),
     panelGeoms: computePanelsSelectors(state),
     seriesTypes: getSeriesTypes(state),
+    customDescription: getCustomDescription(state),
+    disableGeneratedSeriesTypes: isDefaultDescriptionDisabled(state),
```
markov00, line 286: If we want to keep these props (`customDescription` and `disableGeneratedSeriesTypes`) in Settings there is no need to add 2 selectors, you already are calling the `getSettingsSpecSelector` on this `mapStateToProps` and we can reuse it directly

api/charts.api.md
markov00: I also really like the way you named both properties in the issue example:
`description` and `useDefaultSummary`
they are generic and simple.
the word `custom` doesn't add much to that property meaning

src/chart_types/xy_chart/state/chart_state.a11y.test.ts
```diff
@@ -0,0 +1,79 @@
...
+          data: [
+            { x: 1, y: 10 },
+            { x: 2, y: 5 },
+          ],
+        }),
+      ),
+    );
+    store.dispatch(upsertSpec(MockGlobalSpec.settings()));
+    store.dispatch(specParsed());
+    store.dispatch(updateParentDimensions({ width: 100, height: 100, top: 0, left: 0 }));
```
markov00, line 46: There are few Mocks that you can use instead of using directly redux for that.
Please check `MockStore` class that already has some utility methods to create the store and add/update specs

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
rshen91: Ah smart! thanks 3b13cc9

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
rshen91: Ok let me know if I'm following - 3b13cc9 thank you!

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
rshen91: Ok I am open to feedback about 3b13cc9 thank you!

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
rshen91: Smart! Thank you 3b13cc9

api/charts.api.md
rshen91: Thank you both please see 3b13cc9

myasonik: Looks like you missed this file for renaming these variables

rshen91: 😭  right, thanks! 76cc6b3

src/chart_types/xy_chart/state/chart_state.a11y.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      store,
+    );
+    MockStore.addSpecs([MockGlobalSpec.settings()], store);
+  });
+  it('should test defaults', () => {
+    const state = store.getState();
+    const descriptionValue = getSettingsSpecSelector(state).description;
+    const defaultGeneratedSeriesTypes = getSettingsSpecSelector(state).useDefaultSummary;
+    expect(descriptionValue).toBeUndefined();
+    expect(defaultGeneratedSeriesTypes).toBeTrue();
```
markov00, line 49: nit
```suggestion
    const {description, useDefaultSummary} = getSettingsSpecSelector(state);
    expect(description).toBeUndefined();
    expect(useDefaultSummary).toBeTrue();
```

src/chart_types/xy_chart/state/chart_state.a11y.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      [
+        MockGlobalSpec.settings({
+          description: 'This is sample Kibana data',
+        }),
+      ],
+      store,
+    );
+    const state = store.getState();
+    const descriptionValue = getSettingsSpecSelector(state).description;
+    expect(descriptionValue).toBe('This is sample Kibana data');
```
markov00, line 62: nit:
```suggestion
    const { description } = getSettingsSpecSelector(state);
    expect(description).toBe('This is sample Kibana data');
```

src/chart_types/xy_chart/state/chart_state.a11y.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      [
+        MockGlobalSpec.settings({
+          useDefaultSummary: false,
+        }),
+      ],
+      store,
+    );
+    const state = store.getState();
+    const disableGeneratedSeriesTypes = getSettingsSpecSelector(state).useDefaultSummary;
+    expect(disableGeneratedSeriesTypes).toBe(false);
```
markov00, line 75: nit:
```suggestion
    const { useDefaultSummary } = getSettingsSpecSelector(state);
    expect(useDefaultSummary).toBe(false);
```

src/chart_types/xy_chart/state/chart_state.a11y.test.ts
```diff
@@ -0,0 +1,77 @@
...
+        MockSeriesSpec.bar({
+          data: [
+            { x: 1, y: 10 },
+            { x: 2, y: 5 },
+          ],
+        }),
+      ],
+      store,
+    );
+    MockStore.addSpecs([MockGlobalSpec.settings()], store);
```
markov00, line 42: you can add the settings at the same time of the bar series:
```suggestion
    MockStore.addSpecs(
      [
        MockSeriesSpec.bar({
          data: [
            { x: 1, y: 10 },
            { x: 2, y: 5 },
          ],
        }),
        MockGlobalSpec.settings(),
      ],
      store,
    );
```

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
```diff
@@ -266,6 +283,9 @@ const mapStateToProps = (state: GlobalChartState): ReactiveChartStateProps => {
     annotationSpecs: getAnnotationSpecsSelector(state),
     panelGeoms: computePanelsSelectors(state),
     seriesTypes: getSeriesTypes(state),
+    description: getSettingsSpecSelector(state).description,
+    useDefaultSummary: getSettingsSpecSelector(state).useDefaultSummary,
```
markov00, line 287: you can directly just extract the settings before the return statement and extract and return what you need with a single call to the selector, something like
```ts
const { debug, description, useDefaultSummary }  = getSettingsSpecSelector(state);
return {
...
debug,
description,
useDefaultSummary,
};
```

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
```diff
@@ -178,11 +184,19 @@ class XYChartComponent extends React.Component<XYChartProps> {
...
           role="presentation"
+          {...(description ? { 'aria-describedby': `${chartId}--${chartSeriesTypes.length}` } : {})}
         >
-          <dl className="echScreen-reader">
-            <dt>Chart type</dt>
-            <dd>{chartSeriesTypes}</dd>
-          </dl>
+          {(description || useDefaultSummary) && (
+            <div className="echScreenReaderOnly">
+              {description && <p id={`${chartId}--${chartSeriesTypes.length}`}>{description}</p>}
```
markov00, line 191: This is unique, but not very descriptive in what this ID represents.
probably a simple thing like `${chartId}--description` would do the trick.
You can also extract this into a variable and reuse it in both places

rshen91: Thank you! Much cleaner - 75df86b

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
rshen91: Oh nice! I like it - please see 75df86b

src/chart_types/xy_chart/state/chart_state.a11y.test.ts
rshen91: 🙌 75df86b

src/chart_types/xy_chart/state/chart_state.a11y.test.ts
rshen91: thank you! 75df86b

src/chart_types/xy_chart/state/chart_state.a11y.test.ts
rshen91: thank you! 75df86b

src/chart_types/xy_chart/state/chart_state.a11y.test.ts
rshen91: Much cleaner thank you! 75df86b

.playground/playground.tsx
```diff
@@ -19,45 +19,22 @@
...
-import { Chart, AreaSeries, LineSeries, BarSeries, ScaleType } from '../src';
+import { Chart, AreaSeries, ScaleType, Settings } from '../src';
+import { KIBANA_METRICS } from '../src/utils/data_samples/test_dataset_kibana';
 
 export class Playground extends React.Component {
   render() {
     return (
       <div className="App">
         <Chart size={[500, 200]}>
+          <Settings customDescription="This is an area chart showing kibana sample data." />
```
markov00, line 30:
```suggestion
          <Settings description="This is an area chart showing kibana sample data." />
```

markov00 (approved):
LGTM all the changes are fine, just one last minor detail in the playground (you can also checkout the playground code from master so we don't introduce changes there at all)

Issue #1012, opened 2021-02-03
Piechart slice order


markov00:
Provide a way to specify an index for sorting slices in a pie or sunburst chart.

This should be useful in the following use cases:
- we want to treat one particular slice differently from the other ones, like the `Other` slice that we'd like to render  as the last slice, even if its value is larger than some of the category-assigned slices
- during animation (eg. time playthrough, or paging back/forth among category filters) for visual stability between data updates, avoiding slices rotating over one another
- to use in particular cases where the order is implied by the semantics of the categories, such as data that is ordinal in its _meaning_ (e.g. perfect, good, ok, bad) 

![image](https://user-images.githubusercontent.com/1548516/106773430-5337b800-6641-11eb-9205-cd924fcb83e6.png)

Issue #381, opened 2019-09-17
Pie/sunburst chart - discussion about small slices


monfera:
Hi @cchaos, I'm looking forward to your input and suggestions on this, textual and image / drawing / interaction alike 🙇

cchaos:
[Crudimentary prototype](https://www.figma.com/proto/vEMsDZGDmZdwGVBeL0mlTS/Untitled?node-id=3%3A2&scaling=min-zoom)

Essentially, all of the slices that cannot display a label would be have a single overlay hover state encompassing all slices. It would highlight that entire section and then onClick would display a tooltip of all the slices and their values.

**Static screenshot of after the click event:**
<img width="1766" alt="Screen Shot 2019-09-17 at 15 05 05 PM" src="https://user-images.githubusercontent.com/549577/65071513-f1e44b80-d95c-11e9-8dab-f0da0024f4f0.png">

nickofthyme:
@cchaos exactly what I was envisioning! Thanks for mocking up the visual.

I think for the sake of filtering/selection of a specific slice we should allow the tooltip to be interactable, which would be new, such that the user can hover and click on any of the "other" values and on hover the selected slice would be exploded or highlighted.

This is crude example with the limited tools on my mac 😞 (using @cchaos's as a base 😅) 

<img width="1224" alt="Screen Shot 2019-09-17 at 6 40 18 PM" src="https://user-images.githubusercontent.com/19007109/65087522-abeab000-d97b-11e9-82d0-7b35d50e7963.png">

monfera:
Looks good! Below are a few consequences:

1. The cutoff could/should be configurable, for example, anything smaller than x% (or arc length of n pixels) should be lumped into this "other" group - for example, here, the "other" group is quite large, much larger than the approx. minimum of what can be reasonably hovered over for a tooltip, while still respecting Fitts law. 

_Note: it may easily be the case that there are numerous slices which are still easy to hover over, but won't have static text labels associated with them due to their small size, ie. no room inside or outside to render text (I'm just trying to make it explicit that the "other" group is mostly a hover-convenience, as vanishingly small slices can't be hovered over. While it may be a point we all agree on, it looks like more than a coincidence that the cutoff on @cchaos's mock just happens to be past Netherlands, which is the last item with a static text label)_

[...]

PR #1112, opened 2021-04-09
feat(partition): order slices and sectors

monfera:
## Summary

Adds slice sorting for the use of pie charts and sunburst, which allows 
- placing a supplied "Other" slice to be placed at the end, even if, as in this example, the "Other" slice is larger than a regular slice
- slice sorting eg. when the partitioning is along an ordinal variable
- can be used for alphabetical sorting too, though it's almost always a practice to avoid

Closes https://github.com/elastic/elastic-charts/issues/1012
Contributes toward https://github.com/elastic/kibana/issues/90179
Partially addresses https://github.com/elastic/elastic-charts/issues/381

image

The newly introduced `sortPredicate` is for a sort predicate that compares order of any given two `ArrayEntry` objects. `ArrayEntry` has already been part of the public API, so we don't need to publish a new data type for this.

### Checklist


monfera:
jenkins please test this

stratoula:
@monfera it seems fine to me. What I want is:
- sort by the metric (numeric). I think that I can do it by just changing the `clockWise`option. 
- sort alphabetically if the user requests it (will use the new sortPredicate for that)

@flash1293 I think it is fine for Lens too. wdyt?

monfera:
@stratoula correct, and if you need special sorting, eg. you want to put the "Other" slice to the end, even if it's not the smallest slice, then you can rely on the `sortPredicate` too

markov00 (approved):
Tested locally with configurations of `specialFirstInnermostSector` , `clockwiseSectors` and legend. Works as expected

flash1293:
Didn't test the code, but the API looks good to me - a function will give full flexibility which can be use to look up the row index the data table for "user defined" sorting

monfera:
Thanks @markov00 for the review and for @stratoula and @flash1293 for looking at the API. I ended up merging relatively fast as this doesn't lock us in anyway, if there's post-merge or integration time feedback to act upon, and it clears the baseline for the upcoming partition related PRs (lower risk of merge conflict)

PR #1113, opened 2021-04-12
feat: mosaic plot

monfera:
## Summary

To support data visualization best practices, we've anticipated alternatives to pie/donut/sunburst and have been implementing them as opportunity arises: treemap, icicle, flame. 

Merging https://github.com/elastic/elastic-charts/pull/1112 ensured that things are in place for [mosaic plot](https://en.wikipedia.org/wiki/Mosaic_plot), a.k.a. Marimekko or Mekko chart, which is often better than the analogous two-layer pie (sunburst) and sometimes better than the two-layer treemap.

image

Mosaic plots are often preferred over full, two-layer sunburst such as [this](http://localhost:9001/?path=/story/sunburst--sunburst-with-two-layers). If the secondary partitioning is shared, eg. here, browsers, then sideways comparison is way more feasible than in the case of sunburst or treemap. In this case, the browser order should be shared too.

Mosaic plot is Lens friendly: it works with full-depth partitioning (Lens can't do heterogeneous depth such as [general sunburst](https://elastic.github.io/elastic-charts/?path=/story/sunburst--heterogeneous) or [icicle/flame](https://elastic.github.io/elastic-charts/?path=/story/flame-alpha--cpu-profile-flame-chart)). Mosaic plots are often a visual alternative to [crosstabs](https://en.wikipedia.org/wiki/Contingency_table).

Mosaic plot is activated by specifying `partitionLayout: PartitionLayout.mosaic` in the partition chart `config`, as usual.

Mosaic plots are great for 3+1 use case, as often, data is multidimensional and two dimensions can be simultaneously interesting.

"Other" slices can be also used, and placed on the top or bottom, irrespective of size, if we want to avoid too small areas:
image

Mosaic plots are a tool of choice in machine learning, statistics and analytics, as they're great for visualizing [contingency tables aka. cross tabulation](https://analyse-it.com/docs/user-guide/contingency/mosaic-plot) and they're sometimes good for showing a relationship between two variables:

image
([Article](https://novustat.com/statistik-blog/alternatives-standard-graphs-ggplot2-r.html) by Dr. Robert Grünwald)

Mosaic plots are like 
- treemaps and icicle/flame, as they partiton the space with rectangles, which are easier to gauge than disk slices and sectors
- icicle/flame, as they follow a specific 1D orientation per layer, rather than 2D tiling aiming at squarish shapes like treemap
- sunburst, as they work best with two discrete dimensions

There's room for future PRs, eg. 
- per layer specification of orientation (horizontal, vertical, squarifying) to meaningfully support arbitrary layer numbers (eg. 3)
- more configurability for in-shape text layouting, eg. cropping with ellipsis or vertical and horizontal text alignment (benefitting treemap too)
- code regularization for layout types, as the increasing number of options has made `viewModel.ts` etc. slightly more verbose and branching (ternaries) than ideal

### Checklist


markov00 (approved):
Absolutely great!
Is amazing to see the flexibility of this chart implementation.
Tested locally and works great

PR #1114, opened 2021-04-12
feat(xy): specify pixel width for bars

### LINKED ISSUE #497: Bar charts on linear scales with static bar width
Add the possibility to render bar-charts with static bar width along a linear/time scale to mimic the behaviour of various other tools like tableau.

I'd like to create a `<BarSeries>` component where the `minInterval` is 0 but where I can specify a bar width in pixel or percentage.
The consumer needs to know that the bars can overlaps each other.
The pointer detection algorithm needs to detect multiple bars if they overlaps, or should always detect the nearest one to the pointer x value.

n/a

**Additional context**
n/a

**Kibana Cross Issues**
n/a

**Checklist**

markov00:
## Summary

This PR adds the ability to specify the width of bars in pixel or as the ratio of the current computed width.
This option is exposed as part of the `BarSeriesStyle.rect` object

```
widthPixel?: Pixels;
widthRatio?: Ratio;
```

If `widthPixel` is used, then each bar width will use exactly that width in pixel. This width is limited to the max width available for the bars (that excludes the bar padding)
If `widthRatio` is used, then each bar width will use the ratio of the max computed width.
If both are expressed, then the `widthRatio` will express the max available size, where the `widthPixel` express the derired/min width.

fix #497 

### Checklist

Delete any items that do not apply to this PR.


src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -107,8 +107,27 @@ export function renderBars(
...
+    const seriesStyle = getBarStyleOverrides(datum, seriesIdentifier, sharedSeriesStyle, styleAccessor);
+
+    let x = xScaled + xScale.bandwidth * orderIndex;
+
+    const maxPixelWidth = clamp(seriesStyle.rect.widthRatio ?? 1, 0, 1) * xScale.bandwidth;
+    const minPixelWidth = clamp(seriesStyle.rect.widthPixel ?? 0, 0, maxPixelWidth);
+
+    const width = clamp(seriesStyle.rect.widthPixel ?? xScale.bandwidth, minPixelWidth, maxPixelWidth);
+
+    x = x + xScale.bandwidth / 2 - width / 2;
```
monfera, line 129: `x` isn't used before line 129, and `xScale.bandWidth` doesn't seem mutated either, so the `let` can be avoided by
```
const x = xScaled + xScale.bandwidth * orderIndex + xScale.bandwidth / 2 - width / 2;
```

src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -20,7 +20,7 @@
 import { Scale } from '../../../scales';
 import { ScaleType } from '../../../scales/constants';
 import { CanvasTextBBoxCalculator } from '../../../utils/bbox/canvas_text_bbox_calculator';
-import { Color, mergePartial } from '../../../utils/common';
+import { clamp, Color, mergePartial } from '../../../utils/common';
```
monfera, line 23: WebGL person spotted

src/utils/themes/theme.ts
```diff
@@ -398,6 +399,10 @@ export interface RectStyle {
   fill?: Color | ColorVariant;
   /** the opacity of each rect on the theme/series */
   opacity: number;
+  /** The width of the rect in pixel */
+  widthPixel?: Pixels;
+  /** The ratio of the width limited to [0,1] */
+  widthRatio?: Ratio;
```
monfera, line 405: Out of curiosity, if opacity is obligate, can't `width*` be that way too? Ofc there might be a reason

src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -107,8 +107,27 @@ export function renderBars(
...
+    };
+
+    const seriesStyle = getBarStyleOverrides(datum, seriesIdentifier, sharedSeriesStyle, styleAccessor);
+
+    let x = xScaled + xScale.bandwidth * orderIndex;
+
+    const maxPixelWidth = clamp(seriesStyle.rect.widthRatio ?? 1, 0, 1) * xScale.bandwidth;
+    const minPixelWidth = clamp(seriesStyle.rect.widthPixel ?? 0, 0, maxPixelWidth);
+
+    const width = clamp(seriesStyle.rect.widthPixel ?? xScale.bandwidth, minPixelWidth, maxPixelWidth);
```
monfera, line 127: nice!

src/utils/themes/theme.ts
markov00: maybe: `widthRatio` can be `1` by default, but not the `widthPixel`

monfera (approved):
Looks good, minor comments added

src/chart_types/xy_chart/rendering/bars.ts
markov00: fix in a0106fea

src/utils/themes/theme.ts
monfera: Yes, it can go either way for `widthPixel`, it'd basically act as a default bar width for those cases when the bin width is 100 pixels. Probably, even if the chart is very wide, and there are few bars, it's better to constrain the widths to a few 10s of pixels, otherwise the bars become dorian columns, then outright squares. I feel the name "bar chart" is at odds as they're no longer bars, they're quadrilaterals. It's a subjective call of course

stories/stylings/10_custom_bars.tsx
```diff
@@ -52,7 +55,7 @@ export const Example = () => {
     },
   };
 
-  const theme = {
+  const theme: RecursivePartial<Theme> = {
```
nickofthyme, line 58: nit

```suggestion
  const theme: PartialTheme = {
```

src/utils/themes/theme.ts
nickofthyme, line 405: Can we add a comment here about using both? 

> If both are expressed, then the widthRatio will express the max available size, where the widthPixel express the derired/min width.

nickofthyme (approved):
Looks good to me. Two minor comments

src/utils/themes/theme.ts
markov00: done 27f5971

stories/stylings/10_custom_bars.tsx
markov00: done 27f5971

PR #1116, opened 2021-04-12
feat(annotations): marker body with dynamic positioning

### LINKED ISSUE #1040: Re-position chart annotation when it is near the border
The date label of the event rate chart in the ML revert model snapshot flyout does not look ideal when it is near the border of the chart:
image

See codesandbox repro here https://codesandbox.io/s/sad-pond-n2pqh?file=/src/App.tsx

image

Leave the small arrow where it is but re-position the text such that it is better readable / does not contain so many line breaks.

**Additional context**
In a [slack discussion](https://elastic.slack.com/archives/CFK9BJ248/p1614614308011200?thread_ts=1614610942.010700&cid=CFK9BJ248), it was suggested to reuse tooltip functionality to do this.

### LINKED ISSUE #1125: Right annotation icon misaligned
**Describe the bug**
Using a `LineAnnotation` positioned on the right axis. The marker icon will be transformed in the wrong direction

before

**Expected behaviour**
after

nickofthyme:
## Summary

fixes #1040, fixes #1125

Adds a new render component `markerBody` on `LineAnnotationSpec` which is positioned using [`popper.js`](https://popper.js.org/) within the chart `<canvas>` area. `markerBody` and `marker` (i.e. the marker icon) are now able to render as `ReactNode` or a `ComponentType<LineAnnotationDatum>`. This allows for simple text or JSX components or a complex react component that is rendered with the `LineAnnotationDatum` as props. With these props one can render a marker icon or body based on values from the datum.

https://user-images.githubusercontent.com/19007109/115306796-ef943580-a12d-11eb-9d3f-c6a9ec1e1271.mp4

> Note: The `markerBody` is only rendered when a `marker` (aka icon) is provided.

Refactoring notes:
- Abstracted the marker logic from annotations into `LineMarker` component.
- Create a `eachRotation` helper to DRY-up code for `it.each<Rotation>` logic where `-90` in a png filename is converted to `90`, **unknowingly** causing duplicated png file names to be overridden.
- Add index to annotation `id` in the likely event there are multiple annotations with the same joined parameters as seen in the code below... https://github.com/elastic/elastic-charts/blob/4bf2ab19797ba29ea74ad3411ffc551420950131/src/chart_types/xy_chart/annotations/line/dimensions.ts#L407-L415

### Checklist


src/utils/common.tsx
```diff
@@ -541,7 +558,7 @@ export function isDefined<T>(value?: T): value is NonNullable<T> {
  *
  * @internal
  */
-export const isDefinedFrom = <T>(typeCheck: (value: RecursivePartial<T>) => boolean) => (
+export const isDefinedFrom = <T,>(typeCheck: (value: RecursivePartial<T>) => boolean) => (
```
nickofthyme, line 561: Converting this from `*.ts` to `*.tsx` file, the generic `const` function declaration has issues.

This is a convenient fix that does not affect the type of the generic.

See https://stackoverflow.com/questions/32308370/what-is-the-syntax-for-typescript-arrow-functions-with-generics

src/chart_types/xy_chart/renderer/canvas/styles/area.test.ts
```diff
@@ -20,11 +20,11 @@
 import { stringToRGB } from '../../../../../common/color_library_wrappers';
 import { Fill } from '../../../../../geoms/types';
 import { MockStyles } from '../../../../../mocks';
-import { getColorFromVariant } from '../../../../../utils/common';
+import * as common from '../../../../../utils/common';
 import { buildAreaStyles } from './area';
 
 jest.mock('../../../../../common/color_library_wrappers');
-jest.mock('../../../../../utils/common');
+jest.spyOn(common, 'getColorFromVariant');
```
nickofthyme, line 27: I'm not sure how or why my changes to `utils/common.ts(x)` affected this but this caused mocks to return `undefined`. This changes the logic to now only _spys on_ the `common#getColorFromVariant` utils module method for conducting mock returns in styles unit tests.

src/chart_types/xy_chart/renderer/dom/annotations/annotations.tsx
```diff
@@ -64,59 +63,32 @@ interface AnnotationsStateProps {
...
-          });
-        }}
-        onMouseLeave={onDOMElementLeave}
-        style={{ ...style, ...transform }}
-      >
-        {icon}
-      </div>,
+      <LineMarker
+        {...props}
+        key={`annotation-${props.specId}-${props.id}`}
```
markov00, line 86:
```suggestion
        key={`annotation-${props.id}`}
```
`id` already contains the specId.
We can also add the `annotation` part into the `getAnnotationLinePropsId` function

src/chart_types/xy_chart/renderer/dom/annotations/line_marker.tsx
```diff
@@ -0,0 +1,136 @@
...
+      setPopper();
+    }
+
+    return () => {
+      popper?.current?.destroy?.();
+      popper.current = null;
+    };
+  }, [setPopper, body]);
+
+  void popper?.current?.update?.();
```
markov00, line 116: 🤪

src/chart_types/xy_chart/renderer/dom/annotations/line_marker.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  return (
+    <div
+      className="echAnnotation"
+      key={`annotation-${specId}-${id}`}
```
markov00, line 121: as in the previous comment, we can just use the `id`

src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -757,8 +762,10 @@ export type LineAnnotationSpec = BaseAnnotationSpec<
   LineAnnotationStyle
 > & {
   domainType: AnnotationDomainType;
-  /** Custom marker */
-  marker?: JSX.Element;
+  /** Custom marker icon centered on data value */
+  marker?: ReactNode | ComponentWithAnnotationDatum;
+  /** Custom marker body, moves to be contained within chart area */
```
markov00, line 777:
```suggestion
  /** Optional marker body, always contained within chart area */
```

src/chart_types/xy_chart/renderer/dom/annotations/line_marker.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const style = {
+    color,
+    top: chartDimensions.top + position.top + panel.top,
+    left: chartDimensions.left + position.left + panel.left,
+  };
+  const transform = { transform: getMarkerCentredTransform(alignment, Boolean(dimension)) };
+
+  const setPopper = useCallback(() => {
+    if (!iconRef.current || !testRef.current) return;
+
```
markov00, line 77: On a different PR we should extract the `popper.js` to be a HOF or something similar to enhance the positioning of everything that we need

markov00 (approved):
LGTM. Tested locally and works fine.
I've few doubts about the "padding" comments on the story, but I think it's fine leaving it like that. Probably a different background for the annotation or pushing the annotation text and line after the axis labels can be useful

src/chart_types/xy_chart/renderer/dom/annotations/line_marker.tsx
nickofthyme: I know, I tried to not run this on every render but I couldn't get it to work correctly otherwise.

src/chart_types/xy_chart/renderer/dom/annotations/line_marker.tsx
nickofthyme: That's not a bad idea. I wonder if the logic is too bespoke for a HOC, I could see a class. Anyway, happy to discuss this in the future. I also secretly want to test popper on eui and see if it's better performance 🤪, we'll see if I ever have the time.

src/chart_types/xy_chart/renderer/dom/annotations/annotations.tsx
nickofthyme: e8f559e7cb4febeeb6c782e2a366b0559dec717d

src/chart_types/xy_chart/renderer/dom/annotations/line_marker.tsx
nickofthyme: e8f559e7cb4febeeb6c782e2a366b0559dec717d

src/chart_types/xy_chart/utils/specs.ts
nickofthyme: 070d031cf9a94d8448b42c723b43d9f71766b05a

PR #1117, opened 2021-04-13
feat: add debug state for partition charts

### LINKED ISSUE #917: Pie chart vislib replacement functional tests
As we are going to proceed on replacing the vislib pie chart with the elastic charts implementation, there is the need to retrieve some data in order the functional tests to work properly.

**Data that we need in order the vislib functional tests to run:**

- [ ]  Number of slices
- [ ] Slices Labels (array of strings)
- [ ] Slices Values (array of strings)
- [ ] Get slices colors (array of hex strings).
- [ ] Get all legends (and the nested ones) (array of strings)
- [ ] I want to be able to click on a specific slice (i.e. filterOnPieSlice('Other')) to test the click event (filtering)
- [x] I want to be able to click on a specific legend action(i.e. filterLegendAction('Other)) to test the action (filtering)
- [ ] I want to get the color of a specific slice (i.e. getPieSliceStyle('Other')
- [ ] Get the value of a specific slice

markov00:
## Summary

Add the debug state for partition charts with the following type signature:
```ts
type PartitionDebugState = {
  panelTitle: string; // the title of the small multiple, empty if not small multiple
  partitions: Array<{
    name: string; // the name of the slice, before applying the formatter (the string is the same as it was in the data array)
    depth: number; // the layer of the slice when using multilayered pies/treemap
    color: string; // usually in rgba, better to use a color utility function in FTR to convert this into what you really need. The format can change in the future, so better be decoupled from that
    value: number; // the actual value of the slice
    coords: [Pixels,Pixels]; // a clickable central coordinate of the slice/sector
  }>;
};
```

fix #917

### Checklist


stratoula:
@markov00 I think we are fine, I get all the data I need for all my test cases 🎉

markov00:
jenkins test this please

src/components/chart_status.tsx
```diff
@@ -66,13 +66,13 @@ class ChartStatusComponent extends React.Component<ChartStatusStateProps> {
...
 const mapStateToProps = (state: GlobalChartState): ChartStatusStateProps => {
-  const settings = getSettingsSpecSelector(state);
+  const { onRenderChange, debugState } = getSettingsSpecSelector(state);
 
   return {
     rendered: state.chartRendered,
     renderedCount: state.chartRenderedCount,
-    onRenderChange: settings.onRenderChange,
-    debugState: settings.debugState ? getDebugStateSelector(state) : null,
+    onRenderChange,
```
nickofthyme, line 74: is this method still bound correctly if needed?

src/chart_types/partition_chart/state/selectors/get_debug_state.test.ts
```diff
@@ -0,0 +1,139 @@
...
+import { getDebugStateSelector } from './get_debug_state';
+import { createOnElementClickCaller } from './on_element_click_caller';
+
+describe.each([
+  [PartitionLayout.sunburst, 9, 9],
+  [PartitionLayout.treemap, 9, 6],
+  [PartitionLayout.flame, 9, 6],
+  [PartitionLayout.icicle, 9, 6],
+  [PartitionLayout.mosaic, 9, 6],
+])('Partition - debug state %s', (partitionLayout, numberOfElements, numberOfCalls) => {
```
nickofthyme, line 44: 😄

nickofthyme (approved):
Nice work, seems appropriate to make a new `partition` property in the `DebugState` as there is not much crossover.

src/components/chart_status.tsx
markov00: I hope so, haven't checked

src/chart_types/partition_chart/state/selectors/get_debug_state.test.ts
monfera: TIL

src/chart_types/partition_chart/state/selectors/get_debug_state.ts
```diff
@@ -0,0 +1,72 @@
...
+  const X1 = x1 - TAU / 4;
+  const cr = y0px + (y1px - y0px) / 2;
+  const angle = X0 + (X1 - X0) / 2;
+  const x = Math.cos(angle) * cr + diskCenter.x;
+  const y = Math.sin(angle) * cr + diskCenter.y;
+  return [x, y];
+}
+
+function getCoordsForRectangle({ x0, x1, y1px, y0px }: QuadViewModel, diskCenter: PointObject): [Pixels, Pixels] {
+  const y = y0px + (y1px - y0px) / 2 + diskCenter.y;
```
monfera, line 69: Multilayer treemaps and mosaic have overlapping geoms for potential border/coloring around the larger group perimeter, so the mid Y activates a leaf node. It may be fine as the path still includes the non-leaf layers. If it's not desirable, an alternative would be, instead of computing the vertical center of the box, the `y` could be the top of the box plus a couple of pixels

monfera (approved):
Looks fantastic, thank you for having made this 🎉 

Worked great with `debugState`, probably worth committing it i place of or in addition to `debug`.

Due to special cases, I tested coordinates with:
- pie chart with big margins that push aside the chart
- regular sunburst
- sunburst with single 360 degree concentric disk / rings
- no slices at all
- sunburst small multiples
- two-layer treemap
- mosaic

src/chart_types/partition_chart/state/selectors/get_debug_state.ts
```diff
@@ -0,0 +1,72 @@
...
+  const x = x0 + (x1 - x0) / 2 + diskCenter.x;
+  return [x, y];
```
monfera, line 71: If we want to put short coords in the `data-ech-debug-state`, it could be rounded here

src/components/chart_status.tsx
markov00: it works fine

src/chart_types/partition_chart/state/selectors/get_debug_state.ts
markov00: done https://github.com/elastic/elastic-charts/pull/1117/commits/211e439fa86eb8af64012e249f394ad4f8caba64

Issue #1107, opened 2021-04-06
[a11y] add alt chart information for non-Cartesian charts


rshen91:
There have been improvements in alt text for xy_charts for screen readers and this functionality needs to be brought across all chart types. 

**Kibana Cross Issues**
Add any [Kibana](https://github.com/elastic/kibana) related issues/feature requests here.

**Checklist**

Delete any items that are not applicable to this feature request.

- [ ] this request is checked against already exist requests
- [ ] every related Kibana issue is listed under `Kibana Cross Issues` list
- [ ] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

PR #1118, opened 2021-04-13
feat(a11y): add alt text for all chart types 

rshen91:
## Summary
Fixes https://github.com/elastic/elastic-charts/issues/1107

There has been great alternative text improvements for cartesian charts and this PR aims to bridge the gap with the other chart types. 

---
## The new default with charts (in terms of a11y) 

This is a mixed chart (line, bar and area series). Nothing has been configured. In the video below, you can see that a dd element is generated with the following content. This is readable by a screen reader:
 ```
<dt>Chart type:</dt>
<dd>Mixed chart: area and line and bar chart</dd>
```

https://user-images.githubusercontent.com/20343860/116894234-0e470180-abef-11eb-8341-0a304f74adfa.mp4

If you have one type of chart, the generated text will be something like:
```
<dt>Chart type: </dt>
<dd>goal chart</dd> // or whatever type of series in the chart
```

## Adding a a11y description and disabling the default

Using the same chart as an example, you can add your own custom description for a screen reader user. Use the `ariaDescription` prop on the `Settings` component. 

In this example below, the autogenerated description is also disabled by setting the `ariaUseDefaultSummary` to false. The `ariaUseDefaultSummary` prop defaults to true, and you can always include your own description as well as the default one. 

``` 
<Settings 
ariaDescription="This chart has three different types of series. There is a bar, line and area series." 
ariaUseDefaultSummary={false} 
/>
```

https://user-images.githubusercontent.com/20343860/116895503-8235d980-abf0-11eb-9a1a-71afe3f31f34.mp4

### Setting the heading level of the screen reader description
You can also change the heading level (instead of `<p>` as the default seen in the above video) by passing the `ariaLabelHeadingLevel` to the `Settings` component. Valid options are "h1", "h2", "h3", "h4", "h5", "h6", or "p". 

<Settings
            ariaDescription="This description will be read by the screen reader as a heading 3"
            ariaLabelHeadingLevel="h3"
/>

## Setting your own aria-labels, aria-labelledby/aria-describedby properties on charts

Within the `Settings` component, you can add your own aria-label on the figure surrounding the chart element. Use the `ariaLabel` prop on the `Settings` component.  You can also set the `aria-describedby` with the `ariaDescribedBy` prop or the `aria-labeledby`  with the `ariaLabelledBy` prop to tie the charts to other accessible elements within your app. 

### Checklist

### CODE REVIEW

File: src/components/accessibility/types.tsx [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+
+import { A11ySettings } from '../../state/selectors/get_accessibility_config';
+
+interface SeriesTypesProps {
+  chartSeriesTypes: string;
+}
+
+/** @internal */
+export function Types(props: A11ySettings & SeriesTypesProps) {
+  if (!props.defaultSummaryId) return null;
```
nickofthyme, line 30: The prop type and the component names should match

src/components/accessibility/screen_reader_summary.tsx
```diff
@@ -0,0 +1,60 @@
...
+  seriesTypes: string;
+  a11ySettings: A11ySettings;
+}
+
+const ScreenReaderSummaryComponent = ({ seriesTypes, a11ySettings }: ScreenReaderSummaryProps) => {
+  return (
+    <div className="echScreenReaderOnly">
+      <Label {...a11ySettings} />
+      <Description {...a11ySettings} />
+      <Types {...a11ySettings} chartSeriesTypes={seriesTypes} />
```
nickofthyme, line 45: Just my opinion here but I think these components should be more specific because `Label` is very general. Maybe like `ScreenReaderLabel` or something.

src/components/accessibility/types.tsx
```diff
@@ -0,0 +1,37 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import React from 'react';
+
+import { A11ySettings } from '../../state/selectors/get_accessibility_config';
+
+interface SeriesTypesProps {
+  chartSeriesTypes: string;
```
nickofthyme, line 25: I think if this component name is more specific the prop name doesn't need to be this specific. Something like `chartTypeDescription`.

src/chart_types/heatmap/renderer/canvas/connected_component.tsx
```diff
@@ -34,6 +40,8 @@ interface ReactiveChartStateProps {
   initialized: boolean;
   geometries: ShapeViewModel;
   chartContainerDimensions: Dimensions;
+  a11ySettings: A11ySettings;
+  seriesTypes: string;
```
nickofthyme, line 44: I really don't like this variable name because it is not an array of series types as one would think. This is more like a chart type description or something similar. Same goes for `chartSeriesTypes`

src/components/accessibility/types.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+/** @internal */
+export function Types(props: A11ySettings & SeriesTypesProps) {
+  if (!props.defaultSummaryId) return null;
+  return (
+    <dl id={props.defaultSummaryId}>
+      <dt>Chart type:</dt>
+      <dd id={props.defaultSummaryId}>{props.chartSeriesTypes}</dd>
```
markov00, line 34: Why are you adding two times the same ID? That's not allowed.
This was the previous code:
```tsx
{a11ySettings.defaultSummaryId && (
  <dl id={a11ySettings.defaultSummaryId}>
    <dt>Chart type</dt>
    <dd>{chartSeriesTypes}</dd>
    </dl>
)}
```

src/components/accessibility/screen_reader_summary.tsx
```diff
@@ -0,0 +1,60 @@
...
+  if (getInternalIsInitializedSelector(state) !== InitStatus.Initialized) {
+    return DEFAULT_A11Y_SETTINGS;
+  }
+  return {
+    seriesTypes: getSeriesTypesSelector(state),
+    a11ySettings: getA11ySettingsSelector(state),
+  };
+};
+/** @internal */
+export const ScreenReaderSummary = memo(connect(mapStateToProps)(ScreenReaderSummaryComponent));
```
markov00, line 60: not sure we need `memo` here, @nickofthyme ?

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
```diff
@@ -188,16 +184,7 @@ class XYChartComponent extends React.Component<XYChartProps> {
...
-            <Label {...a11ySettings} />
-            <Description {...a11ySettings} />
-            {a11ySettings.defaultSummaryId && (
-              <dl id={a11ySettings.defaultSummaryId}>
-                <dt>Chart type</dt>
-                <dd>{chartSeriesTypes}</dd>
-              </dl>
-            )}
-          </div>
+          <ScreenReaderSummary a11ySettings={a11ySettings} seriesTypes={seriesTypes} />
```
markov00, line 187: `<ScreenReaderSummary>` is already connected to redux store, so you don't have to pass any props

src/chart_types/wordcloud/renderer/svg/connected_component.tsx
```diff
@@ -119,8 +119,12 @@ function layoutMaker(config: Configs, data: Datum[]) {
     .fontSize((d: Word) => getFontSize(d));
 }
 
+const uniqeWordCloudId = `test--description`;
 const View = ({ words, conf }: { words: Word[]; conf: Configs }) => (
-  <svg width={getWidth(conf)} height={getHeight(conf)}>
+  <svg width={getWidth(conf)} height={getHeight(conf)} role="img">
+    <title className="echScreenReaderOnly" id={uniqeWordCloudId}>
+      Chart type: Word cloud chart
+    </title>
```
markov00, line 127: I think we should reuse the way we are doing that in the other charts with the `figure` + the ScreenReaderSummary

src/chart_types/partition_chart/state/selectors/get_series_types.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+import createCachedSelector from 're-reselect';
+
+import { getChartIdSelector } from '../../../../state/selectors/get_chart_id';
+import { getPartitionSpec } from './partition_spec';
+
+/** @internal */
+export const getSeriesTypesSelector = createCachedSelector([getPartitionSpec], (partitionSpec): string => {
+  return `${partitionSpec?.chartType} chart` ?? 'Partition chart';
```
markov00, line 27: In this case we should be return the `particionSpec.config.partitionLayout` instead, that provides the exact chart type we are using

markov00 (commented):
Looks good, I have added few comments, in particular you should remove all props from `          <ScreenReaderSummary />` as all the required props are injected by redux.

src/components/accessibility/types.tsx
rshen91: Sorry for the confusion adding this due to https://github.com/elastic/elastic-charts/pull/1121#discussion_r618524462

src/components/accessibility/screen_reader_summary.tsx
nickofthyme: I don't think it's doing a whole lot but this is basically like using `PureComponent` which does a _shallow prop and state comparison_. I tend to just use them by default when creating a class `Component`.

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
rshen91: Thank you!! ea91feb

src/components/accessibility/types.tsx
rshen91: Thank you! ea91feb

src/components/accessibility/screen_reader_summary.tsx
rshen91: Thank you!! ea91feb

src/components/accessibility/types.tsx
rshen91: Good point 😅 ea91feb

src/components/accessibility/types.tsx
myasonik: The `id` needs to only be on the `<dd>` because this text is being used in an `aria-describedby` which does not understand structured content (e.g., the markup within a `<dl>` so it needs to a point to a simple string).

```
{a11ySettings.defaultSummaryId && (
  <dl>
    <dt>Chart type</dt>
    <dd id={a11ySettings.defaultSummaryId}>{chartSeriesTypes}</dd>
    </dl>
)}


```

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
markov00: Good, but you there are some injected properties not used now, like seriesTypes. If a component doesn't use those, let's not add them to the mapStateToProps

src/chart_types/wordcloud/renderer/svg/connected_component.tsx
rshen91: Ok I wasnt sure because of the svg nature of it https://www.deque.com/blog/creating-accessible-svgs/

rshen91: Please see 93a93a1 thanks

src/chart_types/heatmap/renderer/canvas/connected_component.tsx
rshen91: Ok sweet I think I found them all and have changed in this commit e967711

src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
rshen91: 🤯 fa3ddcd thank you

nickofthyme (approved):

src/chart_types/wordcloud/renderer/svg/connected_component.tsx
```diff
@@ -119,8 +124,12 @@ function layoutMaker(config: Configs, data: Datum[]) {
     .fontSize((d: Word) => getFontSize(d));
 }
 
+const uniqeWordCloudId = `test--description`;
```
myasonik, line 127: `test--description`? Was this supposed to be temporary or should this be passed into `a11ySettings` somehow?

src/components/accessibility/types.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+interface ScreenReaderTypesProps {
+  chartTypeDescription: string;
+}
+
+/** @internal */
+export function ScreenReaderTypes(props: A11ySettings & ScreenReaderTypesProps) {
+  if (!props.defaultSummaryId) return null;
+  return (
+    <dl id={props.defaultSummaryId}>
```
myasonik, line 32:
```suggestion
    <dl>
```

The `id` is used below.

myasonik (approved):

src/chart_types/wordcloud/renderer/svg/connected_component.tsx
rshen91: Good catch - since word cloud doesn't have a chart ID the same way as the cartesian charts I'm pulling the id from the default a11y settings `id={`${DEFAULT_A11Y_SETTINGS.descriptionId}--description`}>` 843b955

src/components/accessibility/types.tsx
nickofthyme: @rshen91 is the resolved? Can you add the commit.

src/chart_types/heatmap/renderer/canvas/connected_component.tsx
```diff
@@ -148,6 +166,8 @@ const mapStateToProps = (state: GlobalChartState): ReactiveChartStateProps => {
     initialized: true,
     geometries: geometries(state),
     chartContainerDimensions: getHeatmapContainerSizeSelector(state),
+    a11ySettings: getA11ySettingsSelector(state),
+    seriesTypes: 'heatmap chart',
```
markov00, line 170: this prop doesn't seems to be used anywhere in the code, let's remove it from the `ReactiveChartStateProps`

src/chart_types/wordcloud/renderer/svg/connected_component.tsx
```diff
@@ -120,7 +125,10 @@ function layoutMaker(config: Configs, data: Datum[]) {
 }
 
 const View = ({ words, conf }: { words: Word[]; conf: Configs }) => (
-  <svg width={getWidth(conf)} height={getHeight(conf)}>
+  <svg width={getWidth(conf)} height={getHeight(conf)} role="img">
+    <title className="echScreenReaderOnly" id={`${DEFAULT_A11Y_SETTINGS.descriptionId}--description`}>
+      Chart type: Word cloud chart
+    </title>
```
markov00, line 131: We should use the same `ScreenReaderSummary` componenent used elsewhere to avoid duplicate code.

markov00 (changes requested):
Few missing things that needs to be fixed

src/chart_types/wordcloud/renderer/svg/connected_component.tsx
markov00: and anyway we can't use an id like this `id={`${DEFAULT_A11Y_SETTINGS.descriptionId}--description`}` because it can conflicts with other world cloud charts

src/chart_types/heatmap/renderer/canvas/connected_component.tsx
rshen91: Sorry I missed this yesterday - please see e9c5255

src/chart_types/wordcloud/renderer/svg/connected_component.tsx
rshen91: Please see: e9c5255 Thank you

markov00 (approved):
LGTM, I've pushed a slight change on the wordcloud to move the ScreenReaderSummary out of the SVG and changed the `svg` `role` to `presentation` as done in the other canvases

PR #1087, opened 2021-03-23
feat: nice ticks everywhere

markov00:
## Summary

This PR adds the nicing to the scales domain.
As the first step, I've just introduced an optional `nice` prop per axis, off by default to avoid breaking change now.

```ts

export interface SeriesScales {
    ...
    xScaleType: XScaleType;
    xNice?: boolean;
    yScaleType: ScaleContinuousType;
    yNice?: boolean;
}

```

The following improvements and refactoring are implemented:
- the scale default parameters are now stored into `scale_defaults.ts`. 
- we have now a selector that merges and computes the final "configuration layer" where we collect, merge, cast all the scales related configuration from the current API (merging configurations from the `SettingsSpec`, `AxisSpec`, `SeriesSpec` and the all relative default values).

This is the first step toward refactoring the scale-related configuration into its own component.

fix #397

PR #1119, opened 2021-04-14
fix(xy): consider `useDefaultGroupDomain` on scale config

markov00:
## Summary

The `useDefaultGroupDomain` was introduced to allow clustering of multiple series with a common Y domain.

The #1087 PR introduced a regression where the `useDefaultGroupDomain` was not considered when defined as the actual `groupId` for a specific group of series. This issue throws an error when a series use `useDefaultGroupDomain` with a groupId that doesn't exist in the chart configuration. This is an edge case but a legit one.

### Checklist


nickofthyme (approved):
LGTM, tested locally.

stories/bar/3_with_axis.tsx
```diff
@@ -37,6 +37,8 @@ export const Example = () => {
         yScaleType={ScaleType.Linear}
         xAccessor="x"
         yAccessors={['y']}
+        groupId="IDgroupId"
+        useDefaultGroupDomain="IDUseDefaultGroupDomain"
```
nickofthyme, line 41: Was this intentional? If so need to update vrt to fix ci errors.

```
yarn test:integration:local --testNamePattern='With Axis' -u
```

markov00: you are right, not intentional, I've rolled back the changes on that story: https://github.com/elastic/elastic-charts/pull/1119/commits/c0f9650f29f080e3401c1987d077b276b43ae4bb

markov00: I've also checked with a previous commit if that was the existing behaviour (removing the axis) and it looks like is correct.

nickofthyme (approved):
👍🏼

PR #1122, opened 2021-04-15
feat(tooltip): anchored/fixed position

### LINKED ISSUE #921: Tooltip: fixed position instead of following mouse and improved line indicator
APM is in the process of migrating from react-vis to Elastic Charts, and while Elastic Charts is getting a lot of things right, I think two small changes could improve the UX:
 
- [ ]  The tooltip should be fixed on the y-axis and only move along along the x-axis. Having it move in both direction is distracting and doesn't add anything
- [ ] As a user I expect to see the point on the line highlighted as I move long the x-axis. Unless it hit the exact spot wrt to both the y-axis and x-axis where the line is, it is not highlighted.

**Old charts (react-vis)**
apm-old-charts

**New charts (Elastic Charts)**
new-charts

markov00:
## Summary

Adding the prop `stickTo` to the tooltip settings to specify where to stick the tooltip respective to the cursor.

https://user-images.githubusercontent.com/1421091/114895723-05dd8100-9e10-11eb-871e-1b146c75a043.mp4

fix #921

### Checklist


stories/area/2_with_time.tsx
```diff
@@ -19,14 +19,15 @@
...
-import { AreaSeries, Axis, Chart, Position, ScaleType, timeFormatter } from '../../src';
+import { AreaSeries, Axis, Chart, Placement, Position, ScaleType, Settings, timeFormatter } from '../../src';
 import { KIBANA_METRICS } from '../../src/utils/data_samples/test_dataset_kibana';
 import { SB_SOURCE_PANEL } from '../utils/storybook';
 
 const dateFormatter = timeFormatter('HH:mm');
 
 export const Example = () => (
   <Chart className="story-chart">
+    <Settings tooltip={{ stickTo: Position.Top, placement: Placement.RightStart, offset: 5 }} rotation={0} />
```
nickofthyme, line 30: Can we add knobs here?

```suggestion
    <Settings
      tooltip={{
        stickTo: getPositionKnob('stickTo'),
        placement: getPlacementKnob(),
        offset: number('offset', 5),
      }}
      rotation={0}
    />
```

src/components/portal/types.ts
```diff
@@ -104,7 +104,7 @@ export interface TooltipPortalSettings<B = never> {
   boundary?: HTMLElement | B;
   /**
    * Boundary element padding.
-   * Used to reduce extents of boundary placement when magins or paddings are used on boundary
+   * Used to reduce extents of boundary placement when margins or paddings are used on boundary
```
nickofthyme, line 107: 🐝 🏆

nickofthyme (approved):
LGTM. Tested locally and works great.

One thing that may be confusing is that the `offset` is relative to the placement and not `stickTo`. So if `stickTo = 'top'` and `placement = 'rightStart'` with `offset = 50` you get this...

Screen Recording 2021-04-22 at 04 53 12 PM

Where there is no offset from the top. I think this is the correct way of doing it but people might get confused.

stories/area/2_with_time.tsx
markov00: done in https://github.com/elastic/elastic-charts/pull/1122/commits/5067c16b18b7fed03ed99d07b52fb964723a24ab

markov00:
>One thing that may be confusing is that the offset is relative to the placement and not stickTo. So if stickTo = 'top' and placement = 'rightStart' with offset = 50 you get this...

@nickofthyme  you are right, probably we should prefix the offset prop with something like `placementOffset` to be more clear

stories/area/2_with_time.tsx
```diff
@@ -17,17 +17,28 @@
...
-    <Settings tooltip={{ stickTo: Position.Top, placement: Placement.RightStart, offset: 5 }} rotation={0} />
+    <Settings
+      tooltip={{
+        stickTo: getPositionKnob('stickTo', Position.Top),
+        placement: getPlacementKnob('placement', Placement.RightEnd),
+        fallbackPlacements: [getPlacementKnob('fallback placement', Placement.LeftStart)].filter(isDefined),
+        offset: number('placement offset', 5),
+      }}
+      rotation={0}
+    />{' '}
```
nickofthyme, line 41: 🙃 ❓

markov00: 🤦

nickofthyme (approved):

PR #1126, opened 2021-04-20
fix(xy): negative bar highlight and click (#1109) [25.0.x]

nickofthyme:
Backports the following commits to 25.0.x:
 - fix(xy): negative bar highlight and click (#1109)


src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -77,20 +76,16 @@ export function renderBars(
...
 
     if (y === null || y0Scaled === null) {
       return;
     }
-    let height = y0Scaled - y;
 
-    // handle minBarHeight adjustment
+    const absMinHeight = Math.abs(minBarHeight);
+    let height = y0Scaled - y;
     if (absMinHeight !== undefined && height !== 0 && Math.abs(height) < absMinHeight) {
```
markov00, line 89:
```suggestion
    let height = y0Scaled - y;
    if (height !== 0 && Math.abs(height) < absMinHeight) {
```

markov00 (approved):
Good for me, tested locally and I don't see any issues

src/chart_types/xy_chart/rendering/bars.ts
markov00: nit: absMinHeight is always defined

PR #1128, opened 2021-04-21
build(deps): update dev dependencies

markov00:
## Summary

Updating a bunch of dev-dependencies to fix various security issues

### Checklist

Delete any items that are not applicable to this PR.

- [ ] Any consumer-facing exports were added to `src/index.ts` (and stories only import from `../src` except for test data & storybook)
- [ ] This was checked for cross-browser compatibility
- [ ] Proper documentation or storybook story was added for features that require explanation or tutorials
- [ ] Unit tests were updated or added to match the most common scenarios

PR #1130, opened 2021-04-21
ci: fix flaky loading on visual test runner

nickofthyme:
## Summary

Changes include:

- Preloads all EUI icons used in our storybook for vrt renderer per eui [docs](https://github.com/elastic/eui/blob/b2ffddee61913202224f2967599436ca95265879/src-docs/src/views/guidelines/getting_started.md#failing-icon-imports).
- Rename legacy `.dot` files to simplify typings and configs.
- Add global types for `process.env` variables.
- Add flag (`STORYBOOK_VRT`) to perform tasks needed for consistent screenshots, of which being disable animations and preload icons.

Fixes:
- Flakiness with color picker dot focus state. See https://github.com/elastic/elastic-charts/pull/1128#issuecomment-824192399
- Flakiness due to all animations and removes costly `delays`.
- Flakiness due to icon dynamic `imports`

### Checklist


rshen91 (approved):
Tested locally LGTM

.eslintignore
```diff
@@ -1,7 +1,4 @@
 # Force include
-!.storybook
-!.storybook-docs
-!.playground
```
rshen91, line 4: Yay 🎉  does this mean I'll stop including the playground file in my PRs 😅

markov00 (approved):
All fine for me, just please check the screenshot with the color picker to understand why the legend item is no more highlighted

rshen91 (approved):

.eslintignore
nickofthyme: lol no but good thought, this was needed for the `.dot` files cuz eslint ignores all of them by default.

integration/tests/legend_stories.test.ts
```diff
@@ -50,6 +50,7 @@ describe('Legend stories', () => {
   it('should render color picker on mouse click', async () => {
     const action = async () => {
       await common.clickMouseRelativeToDOMElement({ left: 0, top: 0 }, '.echLegendItem__color');
+      await common.moveMouseRelativeToDOMElement({ left: 30, top: 10 }, '.echLegendItem');
```
markov00, line 53: @nickofthyme  isn't better to first move and then click?

nickofthyme: No that is what causes the issue because the move causes the mouse to hover elsewhere (i.e. not over the legend item). Swapping them would give this...

legend-stories-test-ts-legend-stories-should-render-color-picker-on-mouse-click-1-diff

What do you say we just leave it without the mouse move and update the screenshot. I don't know how removing the `disableAnimations` call did was a `page.evaluate` (see below). But regardless of why the mouse being over the legend item is irrelevant to this test.

https://github.com/elastic/elastic-charts/blob/920e5856d4c52416c0ea394e5605bb756266b178/integration/page_objects/common.ts#L449-L453

nickofthyme: 0f761c8

PR #1131, opened 2021-04-22
fix(interaction): remove unnecessary elements

### LINKED ISSUE #1074: Interactive elements render when chart is not interactive
**Describe the bug**
I'm rendering charts at https://ci-stats.kibana.dev which are not interactive, but the `.echCrosshair__crossLine` and `.echCrosshair__cursor` elements are still rendered, and don't seem to be positioned correctly, causing them to be rendered off to the right of the chart (probably based on the rules of the container or something) requiring that I use `display: none;` to hide these elements.

**To Reproduce**
1. Go to https://ci-stats.kibana.dev/
2. Inspect one of the sparkline charts
3. Look for the `.echCrosshair__crossLine` element and disable the `display: none;` CSS rule in the inspector

**Expected behaviour**
These element probably shouldn't be rendered if the chart isn't interactive, or they should at least be rendered within the container of the chart.

**Screenshots**
image

 - OS: any
 - Browser: Chrome
 - Elastic Charts: 25.1.1

markov00:
## Summary

Remove DOM element when not needed.

fix #1074


src/components/tooltip/tooltip.tsx
```diff
@@ -204,6 +204,9 @@ const TooltipComponent = ({
     };
   }, [settings, chartRef, rotation]);
 
+  if (!visible) {
+    return null;
+  }
```
nickofthyme, line 209: Are you sure this works the same? I thought that the call to `createPortal` should be done regardless to prevent multiple creating a new portal each time. This seems to be the case in `tooltip_portal.tsx` as there is no other return than `return createPortal(...);`

markov00: We are recreating the portal each time the mouse is over the chart and removing if then it is outside.
I've seen dashboards with a lot of portals because a lot of charts were used...
It looks like quick enough to me to create a new portal everytime you exit and reenter the chart area

markov00: I've changed the behavior a bit. The portal is always removed when the cursor leaves the chart element and recreated whenever enter the area again. It doesn't look like a problem as you are not recreating the portal on mouse move but just on entering/exiting.
I've seen the drawback of keeping the portal open with dashboards where a lot of charts were used and the DOM was populated with unused portals.

nickofthyme (approved):

PR #1132, opened 2021-04-22
fix(partition):  fix safari highlight bug on single slice

### LINKED ISSUE #1085: Single slice pie chart rendering issue in Safari
**Describe the bug**
In Safari, when a single slice pie chart has hovered, the highlighter renders an white cross artefact, probably caused by the SVG configuration of that specific Circle highlight.

**To Reproduce**
1. Go to https://elastic.github.io/elastic-charts/?path=/story/sunburst--single-slice with Safari
2. See the white cross on hover

**Expected behaviour**
A clear circle with the right fill color

**Screenshots**
Screenshot 2021-03-23 at 11 31 42

 - Browser: Safari
 - Elastic Charts: master

**Additional context**
The highlighter is built as:
```
return <circle key={key} r={(y0px + y1px) / 2} {...props} fill="none" strokeWidth={y1px - y0px} />;
```
we can probably fix by using the fill color and fixing the radius to the correct one.

**Errors in browser console**
N/A

**Kibana Cross Issues**
- [ ] https://github.com/elastic/kibana/issues/92194

markov00:
## Summary

It looks like safari has a rendering issue on circles if rendered entirely applying a strokeWidth instead of the fill.
I've updated the code for the case where we have a single slice at the root level.

fix #1085


src/chart_types/partition_chart/renderer/dom/highlighter.tsx
```diff
@@ -98,8 +98,15 @@ function renderRectangles(
...
-    return <circle key={key} r={(y0px + y1px) / 2} {...props} fill="none" strokeWidth={y1px - y0px} />;
+    const isRoot = y0px === 0;
+    const props = {
+      r: isRoot ? y1px : (y0px + y1px) / 2,
+      fill: isRoot ? style.color : 'none',
+      stroke: !isRoot ? style.color : 'none',
+      className: style.color ? undefined : isRoot ? style.fillClassName : style.strokeClassName,
+      strokeWidth: isRoot ? undefined : y1px - y0px,
+    };
+    return <circle key={key} {...props} />;
```
nickofthyme, line 109: nit: might me clean with 2 separate objects but just my thought

nickofthyme (approved):
LGTM Confirm fixes issue.

image

src/chart_types/partition_chart/renderer/dom/highlighter.tsx
markov00: I've cleaned this up a bit in https://github.com/elastic/elastic-charts/pull/1132/commits/1639199b40a9b2051ffc07293d6f3161f3365b06

nickofthyme: Yeah that's exactly what I was thinking, cleaner to read for me thanks!

PR #1134, opened 2021-04-22
fix(legend): disable handleLabelClick for one legend item

### LINKED ISSUE #1055: Ability to disable toggle visiblity behavior on legend items.
We have charts in Observability that only ever plot 1 or 2 pieces of information. Users are able to click on a legend item to toggle the visibility of data. While this feature is useful when there are many pieces of data being shown (it allows the user to focus), it doesn't seem like it provides any value when the chart only has a single legend item.

A way for developers to disable the toggle capability when it's not desired.

Consistency / expected behavior might be a concern. Some charts would allow toggling visibility (current default behavior), whereas other may not. How would a user know when this toggle behavior is available?

A simple solution could be to just remove this capability when there is only a single legend item. This behavior could apply to all chart types. If more than 1 legend item exists, then the toggle visibility feature would become available.

**Additional context**

<details>
<summary>See screenshots</summary>

image
image

</details>

**Checklist**

rshen91:
## Summary
Fixes #1055 

If there is only one legend item, the label will not be able to be clicked to be visible or not. That functionality still works for charts with multiple legend items. 

### Checklist


nickofthyme (changes requested):
Looks like the `ItemLabel` is still toggleable. We need to update that in addition to the click handler.

Screen Recording 2021-04-22 at 05 27 27 PM

rshen91:
> Looks like the `ItemLabel` is still toggleable. We need to update that in addition to the click handler.
> 
> Screen Recording 2021-04-22 at 05 27 27 PM

ah good catch - please see a34bf4a thank you!

src/components/legend/legend_item.tsx
```diff
@@ -125,7 +125,14 @@ export class LegendListItem extends Component<LegendItemProps, LegendItemState>
...
   handleLabelClick = (legendItemId: SeriesIdentifier[]): MouseEventHandler | undefined => {
-    const { item, onClick, toggleDeselectSeriesAction } = this.props;
+    const { item, onClick, toggleDeselectSeriesAction, totalItems } = this.props;
+    if (totalItems <= 1) {
+      return;
+    }
+
+    if (legendItemId.length === 0) {
+      return;
+    }
```
nickofthyme, line 135: Why is this logic here and not in the condition for the `isToggleable` prop?

rshen91: Oh man so I thought this was in master but it isn't so I removed it in 8e57477. Also changed the accessibility integration tests to test the HTML since the accessibility tree functionality in puppetteer is spotty and unreliable. I think we might be moving to playwright eventually so I'm just loading the HTML for now.

nickofthyme: Is the spotty accessibility tree logic failing just this PR? If not I would say move that to another PR.

So to be clear the `legendItemId.length` had no impact on the toggle ability of the legend items, correct?

src/components/legend/legend_item.tsx
```diff
@@ -211,7 +214,7 @@ export class LegendListItem extends Component<LegendItemProps, LegendItemState>
           />
           <ItemLabel
             label={label}
-            isToggleable={item.isToggleable}
+            isToggleable={totalItems > 1 ? item.isToggleable : false}
```
nickofthyme, line 217:
```suggestion
            isToggleable={totalItems > 1 && item.isToggleable}
```

rshen91: Much cleaner! 586adbf

src/chart_types/xy_chart/state/chart_state.accessibility.test.ts
```diff
@@ -52,7 +52,7 @@ describe('test accessibility prop defaults', () => {
       ariaLabelledBy,
     } = getSettingsSpecSelector(state);
     expect(ariaDescription).toBeUndefined();
-    expect(ariaUseDefaultSummary).toBeTrue();
+    expect(ariaUseDefaultSummary).toBeTruthy();
```
nickofthyme, line 55: Isn't this a `boolean` why the looser assertion?

rshen91: 🤔 I'm facing this error in my IDE `any Property 'toBeTrue' does not exist on type 'JestMatchers<boolean>'.ts(2339)`

src/components/legend/legend_item.tsx
```diff
@@ -125,7 +125,10 @@ export class LegendListItem extends Component<LegendItemProps, LegendItemState>
...
    */
   handleLabelClick = (legendItemId: SeriesIdentifier[]): MouseEventHandler | undefined => {
-    const { item, onClick, toggleDeselectSeriesAction } = this.props;
+    const { item, onClick, toggleDeselectSeriesAction, totalItems } = this.props;
+    if (totalItems <= 1) {
+      return;
+    }
 
     if (!item.isToggleable && !onClick) {
       return;
```
nickofthyme, line 134: Let's merge these branches

src/chart_types/xy_chart/state/chart_state.accessibility.test.ts
nickofthyme: UGH!! 🙄 

This should fix that https://github.com/elastic/elastic-charts/pull/1136. It's a vscode thing not a type issue. You should be able to revert this and just save the file ignoring the error. Test it by running...

```
yarn typecheck:all
```

src/components/legend/legend_item.tsx
rshen91: Thank you - please see 2da6e3d

src/components/legend/legend_item.tsx
rshen91: Yep! Ok I can open a separate PR for the accessibility tree logic failing. 👍🏻 And the legendItemId.length should not impact the toggle ability of the legend items.

rshen91: removed test changes in 390c73d and opened a new PR to address the accessibility integration test changes. https://github.com/elastic/elastic-charts/pull/1137

integration/tests/accessibility.test.ts
```diff
@@ -20,17 +20,17 @@
...
+  //   const tree = await common.testAccessibilityTree(
+  //     'http://localhost:9001/iframe.html?id=annotations-lines--x-continuous-domain',
+  //     '.echCanvasRenderer',
+  //   );
+  //   // the legend has bars and lines as value.descriptions not value.name
+  //   const hasTextOfChartTypes = tree.children.filter((value) => {
+  //     return value.name === 'bar chart';
+  //   });
+  //   expect(hasTextOfChartTypes[0].name).toBe('bar chart');
+  // });
```
nickofthyme, line 33: `xit` (aka `it.skip` or `xtest`) is much cleaner way to skip tests, less code line diff. One exception is if the test has a type error. See https://jestjs.io/docs/api#testskipname-fn

```suggestion
  xit('should include the series types if one type of series', async () => {
    const tree = await common.testAccessibilityTree(
      'http://localhost:9001/iframe.html?id=annotations-lines--x-continuous-domain',
      '.echCanvasRenderer',
    );
    // the legend has bars and lines as value.descriptions not value.name
    const hasTextOfChartTypes = tree.children.filter((value) => {
      return value.name === 'bar chart';
    });
    expect(hasTextOfChartTypes[0].name).toBe('bar chart');
  });
```

integration/tests/legend_stories.test.ts
```diff
@@ -150,6 +150,34 @@ describe('Legend stories', () => {
...
+      const labels = page.evaluate(() =>
+        Array.from(document.getElementsByClassName('echLegendItem'), (e) => e.outerHTML),
+      );
+      (await labels).forEach((label, index) => {
+        if (label.includes('Activate to show series')) {
+          hiddenResults.push(index);
+        }
+      });
+      expect(hiddenResults).toEqual([]);
+    });
```
nickofthyme, line 180: Seeing as this test does not expect a screenshot. I think it could be done in an enzyme/jest unit test with much less overhead and achieve the same goal. What do you think?

integration/tests/accessibility.test.ts
rshen91: Thank you!! The linter changed xit -> it.skip 22bfe35

integration/tests/legend_stories.test.ts
rshen91: Yasss please see 7dce80e

src/components/legend/legend.test.tsx
```diff
@@ -288,4 +288,31 @@ describe('Legend', () => {
       });
     });
   });
+  describe('disable toggle and click for one legend item', () => {
+    it('should not be able to click or focus if there is only one legend item in total legend items', () => {
+      const onLegendItemClick = jest.fn();
```
nickofthyme, line 293: Exactly!

src/components/legend/legend.test.tsx
```diff
@@ -288,4 +288,31 @@ describe('Legend', () => {
...
+      const data = [{ x: 2, y: 5 }];
+      const wrapper = mount(
+        <Chart>
+          <Settings showLegend showLegendExtra onLegendItemClick={onLegendItemClick} />
+          <BarSeries
+            id="areas"
+            xScaleType={ScaleType.Linear}
+            yScaleType={ScaleType.Linear}
+            xAccessor="x"
+            yAccessors={['y']}
```
nickofthyme, line 303: These are default so could remove.

```suggestion
```

src/components/legend/legend.test.tsx
```diff
@@ -288,4 +288,31 @@ describe('Legend', () => {
...
+            data={data}
+          />
+        </Chart>,
+      );
+      const legendWrapper = wrapper.find(Legend);
+      const legendItems = legendWrapper.find(LegendListItem);
```
nickofthyme, line 309: This is an unnecessary step. You just need to get the `LegendListItem`

```suggestion
      const legendItems = wrapper.find(LegendListItem);
```

nickofthyme (approved):
LGTM, just a few minor comments

src/components/legend/legend.test.tsx
rshen91: 02b5f97

PR #1137, opened 2021-04-23
test(a11y): fix flaky accessibility integration tests

rshen91:
## Summary

Moving some test changes from #1134 into its own PR. Instead of using the puppetteer's accessibility tree, which is sometimes including certain roles and names in the accessiblity tree implementation, these test changes test the HTML of the charts. I think we might be moving to playwright eventually so I'm just loading the HTML for now.

### Checklist

- [x] tests were updated or added to match the most common scenarios


integration/tests/accessibility.test.ts
```diff
@@ -19,27 +19,17 @@
...
-      '.echCanvasRenderer',
-    );
-    // the legend has bars and lines as value.descriptions not value.name
-    const hasTextOfChartTypes = tree.children.filter((value) => {
-      return value.name === 'Mixed chart: bar and line chart';
-    });
-    expect(hasTextOfChartTypes[0].name).toBe('Mixed chart: bar and line chart');
+    const html = await common.getElementHTML('http://localhost:9001/iframe.html?id=mixed-charts--bars-and-lines');
+    const startIndex = html.indexOf('<dd>');
+    expect(html.slice(startIndex, startIndex + 30)).toBe('<dd>Mixed chart: bar and line ');
```
nickofthyme, line 33: I think there is a much easier way of doing this with the Puppeteer [`Page`](https://pptr.dev/#?product=Puppeteer&version=v3.2.0&show=api-class-page) class methods. Let's jump on a zoom next week and see what the final output is that you want.

rshen91: Thank you! I cleaned up the function - let me know what you think when you have a chance - 9455998

integration/tests/accessibility.test.ts
```diff
@@ -21,15 +21,17 @@ import { common } from '../page_objects/common';
...
-    const html = await common.getElementHTML('http://localhost:9001/iframe.html?id=mixed-charts--bars-and-lines');
-    const startIndex = html.indexOf('<dd>');
-    expect(html.slice(startIndex, startIndex + 30)).toBe('<dd>Mixed chart: bar and line ');
+    const html = await common.getSelectorHTML(
+      'http://localhost:9001/iframe.html?id=mixed-charts--bars-and-lines',
+      'dd',
+    );
+    expect(html[0].innerHTML).toBe('Mixed chart: bar and line chart');
   });
 });
```
nickofthyme: I still think we move this test to jest like you did here https://github.com/elastic/elastic-charts/commit/7dce80e596134f122abaa16832007443b88695b2.

Unless you see a difference in the tested assertions

rshen91: Ok cool! 3d0f4aa

src/components/accessibility/accessibility.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+    const wrapper = mount(
+      <Chart size={[100, 100]} id="chart1">
+        <Settings debug rendering="svg" showLegend />
+        <BarSeries id="test" data={[{ x: 0, y: 2 }]} />
+      </Chart>,
+    );
+    const html = wrapper.html();
+    const ddTagStartIndex = html.indexOf('<dd>');
+    const ddTag = html.slice(ddTagStartIndex, ddTagStartIndex + 18);
+    expect(ddTag).toBe('<dd>bar chart</dd>');
```
nickofthyme, line 37: Can we use [`find`](https://enzymejs.github.io/enzyme/docs/api/ReactWrapper/find.html), [`first`](https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/first.html) and [`text`](https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/text.html) to simplify all this?

```suggestion
    expect(wrapper.find('dd').first().text()).toBe('bar chart');
```

nickofthyme (approved):

PR #1145, opened 2021-05-05
feat(axis): allow pixel domain padding for y axes 

### LINKED ISSUE #1144: Allow specifying axis domain padding as pixel
Axis domain padding can be specified as percentage or total value related to the scale itself. However as this setting is often used for aesthetics, it would be useful to specify it as a pixel value.

elastic-charts should take a setting like this:
```
{ padding:  '10px' }
```

Then it does the calculation of values to pixels and applies the appropriate padding

Theoretically the consumer could do this calculation, but it's cumbersome because the height of the chart has to be tracked to do so.

**Additional context**
This came up because the comment of the `padding` property actually specifies it can be a pixel value which is not the case

**Checklist**

nickofthyme:
## Summary

Adds support for pixel domain padding in the vertical/y axis, added to the existing padding options of domain value and domain ratio value.

BREAKING CHANGES: 
`domain.padding` now only takes a `number` value. If you are using a percent-based padding such as `'10%'` please set `domain.padding` to `0.1` and `domain.paddingUnit` to `DomainPaddingUnit.DomainRatio`.

## Details

This pr adds the ability to set the y domain padding to a fixed pixel value in the spatial or screen space. Currently, the domain is allowed to be set to a domain value or domain percentage value with the misaligned documentation.

Adds `paddingUnit` as type `DomainPaddingUnit` to domain options rather than parsing units from a string.

```ts
export const DomainPaddingUnit: Readonly<{
    Domain: "domain";
    Pixel: "pixel";
    DomainRatio: "domainRatio";
}>;
```

Adds `domainPixelPadding` and `constrainDomainPadding` as options to continuous scale, but does not apply to time scales.

> Note: All domain padding is applied _before_ and domain nicing.

## Connected issues

Closes #1144

### Checklist


src/scales/scale_continuous.ts
```diff
@@ -310,6 +328,38 @@ export class ScaleContinuous implements Scale {
...
+        const newEnd = scaleCopy.invert(r2 > r1 ? r2 + delta : r2 - delta);
+
+        newDomain = constrainPadding(newDomain[0], newDomain[1], newStart, newEnd, constrainDomainPadding);
+        scaleCopy.domain(newDomain);
+
+        const lowerDelta = Math.abs(r1 - scaleCopy(this.domain[0]));
+        const upperDelta = Math.abs(r2 - scaleCopy(this.domain[1]));
+        // Update delta with error in numerical solution
+        delta = Math.abs(Math.max(lowerDelta, upperDelta) - domainPixelPadding);
+      }
```
markov00, line 352: I'm pretty sure you can use the `screenspaceMarkerScaleCompressor` fn to compute the scale compression when you need to limit by pixels an existing scale

markov00 (commented):
I'm a bit confused about when/if the padding is applied to the lower part of the domain.
It looks like it always gets applied to the higher part, and the lower one always gets limited at 0 or at the specified domain min.

src/scales/scale_continuous.ts
nickofthyme: Great catch! See 14c792b

nickofthyme:
> It looks like it always gets applied to the higher part, and the lower one always gets limited at 0 or at the specified domain min.

This is controlled by the `constrainPadding` option.

src/scales/scale_continuous.ts
monfera: @markov00 good call here— with Nick, we discussed the way to invoke the `screenspaceMarkerScaleCompressor` such that it returns the needed value, whether padding is needed on the top, bottom, or both (eg. bar chart with both + and - values). So far so good.

@nickofthyme what we discussed yesterday is still good, as in, you're guaranteed to get sufficient padding, but it's possible to get _too much_ padding that way.

- nicing the raw domain may already ensure sufficient padding, while, if we add the padding and then nice, then it may lead to a much taller pad than would be needed otherwise, so one would think, it's worth attempting to nice and check the padding first (screenspace distance between raw domain max and last niced tick domain value)
- if the nicing of the raw domain doesn't lead to sufficient padding, even then is still possible that adding the padding first (with the little function drawn up yesterday) then nicing leads to a too tall padding—maybe adding a smaller bit to the domain, then do the nicing may yield good results

There is another problem too: sometimes it's not sensible (legal) to extend the domain. Eg. if we show some ratio, and the user knows it can't ever be more than 100%, and our value is 95% and padding would make it, say, 120% with nicing, then it'll be a nonsensical axis.

So, this is what could be done:
- compute the niced domain min/max from the raw domain min/max and the desired tick count
- observe the difference between the raw domain max and niced domain max, I'll call it `domainGrowth` (min is the same, so I won't be repetitive with that)
- this `domainGrowth` will have a screenspace height in pixels, and that should be _subtracted_ from the required `domainPaddingPx`
- this requires a small linear solver (basically a couple of lines with the four basic operators) which I'm happy to write, we should just first clarify the requirements

So I guess, what I'm proposing is that the aesthetically pleasing top pad has two parts:
- the effect of nicing, as it already often provides some clearing; this is also the Y axis domain
- the remaining, visually necessary padding, so the nicing plus this together gives us the required pixel padding

A key point is that the axis is extended only due to nicing, but not for the remainder of the padding. Why?
- it avoids most of the illegal axes (eg. nicing up 95% to 100% is OK)
- it avoids the issue of double nicing (if we were going to extend the axis past the niced domain max, then it means that another nicing may be needed) 

=============

tl;dr if I clearly expressed the issue here, and everyone is OK with this approach (we include the nicing `domainGrowth` as partial fulfilment toward the full `domainPaddingPx`; and we only extend the actual axis to the nicing of the original domain, but not into the rest of the padding) then pls. give the go ahead and I can add couple-liner calculation that
- takes as input the raw domain max, and the niced domain max(*)
- full height in pixels
- required `domainPaddingPx`
And returns the scale, and the actual (residual) domain padding in pixels, ie. the part beyond the nicing.

=============

(*) this assumes that the addition of pixel padding  doesn't materially change the height of the Y axis. Reason: as you vary the nicing, it varies the leftover pixel padding. Which in turn varies the actual cartesian projection area height. Which in turn influences how many Y ticks fit. Which in turn modifies the desired tick counts. Which in turn impacts the nicing, since nicing is a function of the increment, therefore the tick count. If we want to solve this, it's a lovely problem and it needs that we make `screenspaceMarkerScaleCompressor` nice-aware. Ie. it would compute not just the pairwise scales but the pairwise _niced_ scales. I'd love to work on that, and generally, nicing needs to be integral to the solving process, not a post-processing step

nickofthyme: You make some good points @monfera most of which I agree with. I think these changes are best left for a future PR, especially considering none of these changes are applies to the current domain padding logic.

I get limiting user's ability to pad the axes to avoid nonsensical axes but I think this is beyond the scope. I could see some option that limits the upper/lower domain to prevent and even specify lower and upper padding.

As far as the nicing, I think the idea of nicing is to nicely round out the domain. Thus the simplest combination of this and domain padding would be to apply the padding then apply nicing. This could be refined in the future the better account for edge cases but I don't think this is a critical feature.

monfera: Yes, it's more of a future refinement, the goals of this PR represents a significant improvement already 🥇

src/scales/scale_continuous.ts
```diff
@@ -118,6 +121,45 @@ export function limitLogScaleDomain([min, max]: ContinuousDomain, logMinLimit?:
...
+    } else if (paddedDomainHigh > 0 && orderedDomain[1] <= 0) {
+      const { scaleMultiplier } = screenspaceMarkerScaleCompressor(
+        [orderedDomain[0], 0],
+        [2 * desiredPixelPadding, 0],
+        chartHeight,
+      );
+      paddedDomainLo = orderedDomain[0] - desiredPixelPadding / scaleMultiplier;
+      paddedDomainHigh = 0;
+    }
+  }
```
monfera, line 158: Nice solution for adding both a top and bottom padding _except_ when the padding would be the thing that extends the domain over/under the zero axis line 🎉 

We might have a situation in the future where the reference line is not the zero line, but something else; for such a case, it would be eventually useful to DRY up and extract the reference value, eg. `const intercept = 0;` so it's later easy to make configurable. It's not needed in this PR, do you think @markov00 it'd be useful at some point?

monfera (commented):
Great PR 🚀 
Please add VRT pics for the full variety of cases, with top and bottom padding applied:
- domain values are both positive and negative
- positive-only and negative-only; both with and without the padding breaching the zero intercept

This also exercises all the code paths where the scale compressor is used

src/scales/scale_continuous.ts
markov00: Yep, is definitely a nice to have!

src/scales/scale_continuous.ts
nickofthyme: Created separate issue for this #1179

src/scales/scale_continuous.ts
nickofthyme: Is this what you had in mind @monfera 0577a2c?

nickofthyme:
> Great PR Nick 🚀
> 
> Please add VRT pics for the full variety of cases, with top and bottom padding applied:
> 
> * domain values are both positive and negative
> * positive-only and negative-only; both with and without the padding breaching the zero intercept
> 
> This also exercises all the code paths where the scale compressor is used

❤️ Good idea, see 7511666

markov00 (approved):
LGTM, thanks Nick for all the changes

markov00:
jenkins test this please

nickofthyme:
jenkins test this please

PR #1159, opened 2021-05-14
feat(legend): render multiple columns on float config

### LINKED ISSUE #1158: Multi column legend
Add the ability to display the floating legend with multiple columns

markov00:
## Summary

When rendering a floating legend inside the chart (`Settings.legendPosition.floating:true`), the added `Settings.legendPosition.floatingColumns` property allows you to specify the number of columns used to distributed the legend items.
Screenshot 2021-05-14 at 19 46 54

## Connected issues
This feature was added as part of the Kibana Timelion chart library replacement. 

fix #1158

### Checklist


src/components/legend/style_utils.ts
```diff
@@ -68,18 +69,22 @@ export function getLegendListStyle(
...
  */
-export function getLegendStyle(direction: LegendPositionConfig['direction'], size: BBox, margin: number): LegendStyle {
+export function getLegendStyle({ direction, floating }: LegendPositionConfig, size: BBox, margin: number): LegendStyle {
   if (direction === LayoutDirection.Vertical) {
     const width = `${size.width}px`;
     return {
-      width,
-      maxWidth: width,
+      width: floating ? undefined : width,
+      maxWidth: floating ? undefined : width,
```
nickofthyme, line 87: I think we should set this to chart width. This is probably not easy to get perfect but the `max-width` should help.

https://user-images.githubusercontent.com/19007109/118983909-90029300-b942-11eb-9cb7-84e9a3530e59.mp4

src/components/legend/_legend.scss
```diff
@@ -9,12 +9,6 @@
     }
   }
```
nickofthyme, line 11: Can we change the `.echLegendList` styles to include grid by default now?

```scss
  .echLegendList {
    display: grid;
  }

  &--horizontal {
    .echLegendList {
      grid-column-gap: $echLegendColumnGap;
      grid-row-gap: $echLegendRowGap;
      margin-top: $echLegendRowGap;
      margin-bottom: $echLegendRowGap;
    }
  }
```

src/components/legend/style_utils.ts
```diff
@@ -68,18 +69,22 @@ export function getLegendListStyle(
   return {
     paddingTop,
     paddingBottom,
+    ...(floating && {
+      display: 'grid',
```
nickofthyme, line 73: Make default, see previous comment

```suggestion
```

src/components/legend/style_utils.ts
```diff
@@ -68,18 +69,22 @@ export function getLegendListStyle(
...
+      gridTemplateColumns: `repeat(${clamp(floatingColumns ?? 1, 1, totalItems)}, 1fr`,
```
nickofthyme, line 74: `1fr` will cause the largest string to control the width for all columns like this....

image

If we use `auto` instead of `1fr` then each column width will be independent. Not sure which you think is desirable but I like the independent widths.

image

```suggestion
      gridTemplateColumns: `repeat(${clamp(floatingColumns ?? 1, 1, totalItems)}, auto`,
```

nickofthyme (commented):
LGTM, tested locally.

I think we should try to constrain the width to within the chart because this could get unwieldy pretty quickly. Constraining the overall width will start to truncate the longest strings to satisfy the constrain.

src/components/legend/style_utils.ts
markov00: @nickofthyme adding the `width` and/or `maxWidth` together with the grid and a fixed number of columns doesn't work, it just reduce the grid space but overflows to the other side.
Unfortunately I'm not aware of an option to specify the wanted column or auto-fill if overflows.
It could be some convoluted mix of auto-fit + minmax + auto but I've tried a bit and doesn't work

nickofthyme (approved):
LGTM just one comment.

src/components/legend/style_utils.ts
```diff
@@ -62,16 +62,15 @@ export function getLegendListStyle(
     return {
       paddingLeft,
       paddingRight,
-      gridTemplateColumns: `repeat(auto-fill, minmax(${legendStyle.verticalWidth}px, 1fr))`,
+      gridTemplateColumns: `repeat(auto-fill, minmax(${legendStyle.verticalWidth}px, auto))`,
```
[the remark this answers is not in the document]
markov00: done here thanks https://github.com/elastic/elastic-charts/pull/1159/commits/8facfebee89f8c405789978d052b5d84382bf798

src/components/legend/style_utils.ts
markov00: I'm not going to solve this issue right now, this can be part of the legend redesign #580 task

src/components/legend/style_utils.ts
markov00: done in 3b1cfc4

src/components/legend/style_utils.ts
markov00: fixed in 3b1cfc4

PR #1164, opened 2021-05-20
build: enable linking charts library to kibana

nickofthyme:
## Summary

Adds a new command line aid for linking charts library to kibana.

Important features:
- Single command runs everything
- Cleanup is simple
- Both charts and kibana are restored to original state
- Handles many non-happy paths, but not all 😢 

### Entry command:

```bash
yarn link:kibana [--debug]
```

You will be prompted to `Link`, `Unlink` or `Watch mode`.

`Link` - Starts linking process and ends in watch mode to update built files
`Unlink` - Cleans up all changes related to the linking process
`Watch mode` - Skips linking and just starts watching. Requires previous link. Used if something really bad happens and the watchers give up, or you accidentally quit the watch mode.

> The `--debug` flag will log all `stdout`, `stderr` and `errors` to the console in `chalk.blue`.

### Linking

Linking requires you to bootstrap and start kibana before linking.

The only other prompt is for the relative path from `@elastic/charts` directory to `kibana` directory.

The script does all of the following:
- Builds `@elastic/charts` is watch mode
- Replaces all `linkedPackages` require statements with relative paths (`require("react")` becomes `require("../path/to/kibana/react")`)
- Builds `@kbn/ui-shared-deps` in watch mode

### The struggle 😫

The `@kbn/ui-shared-deps` package is used, to my best understanding, as a poor-mans [`yarn resolution`](https://classic.yarnpkg.com/en/docs/package-json/#toc-resolutions) creating a single export for singlton packages and packages needing consistent versions. Most important is `react` as hooks require the use of the same react instance. Linking charts to kibana would enable kibana to use the linked react but that local charts would use the local `react` (i.e. `elastic-charts/node_modules_react`).

If we then attempt to link `react` in charts to point at react in kibana, it would use the correct react (i.e. `path/to/kibana/node_modules/react` **BUT** the way that webpack bundles `@kbn/ui-shared-deps` this will not be the same instance of react and thus still throwing the dreaded [invalid hook call warning](https://reactjs.org/warnings/invalid-hook-call-warning.html) so symlinking `react` does not work.

The solution is to replace all the paths in the built chart `dist` output to thier relative path to the kibana `react`. This will use the webpack exported instance of `react` from `@kbn/ui-shared-deps`. This however is not going to work just replacing the `react` since other packages like `react-redux` point to react internally and since that is the `react-redux` local to charts it will be wrong yet again. So the solution is to replace `react`, `react-dom`, `react-redux` and `redux`. So every change that happens in charts needs to be built, swap the links and build `@kbn/ui-shared-deps` to be reflected in kibana.

The good news is that there are no changes to src in either kibana nor charts for linking to take place. This makes it easy to cleanup the linked files once no longer needed.

### But how do I know it's linked and what charts directory is linked?

I added a console.log to clearly indicate when charts is linked and was chart directory is linked.

image

### Limitations

- You can only link one charts directory to one kibana directory
- Not all errors are handled though many are passed through command line interface to resolve.
- Many get in a stuck state as it's not tested in the wild...yet!
- Only rebuilds `.ts` files, not `scss`
- Cannot run `nvm install` to update kibana on the fly, requires user to update node manually.


packages/link_kibana/process_utils.js
```diff
@@ -0,0 +1,152 @@
...
+
+        if (stdout && stdout.includes(errorStr)) {
+          errors.push(stdout);
+        }
+
+        if (errors.length > 0) {
+          spinner.fail();
+          printErrorMsg();
+          errors.forEach(printError);
+          return;
```
markov00, line 72: isn't better to reject the promise on error?

packages/link_kibana/utils.js
```diff
@@ -0,0 +1,166 @@
...
+const { promisify } = require('util');
+
+const chalk = require('chalk');
+const { snakeCase } = require('change-case');
+
+const { exec } = require('./process_utils');
+
+const exists = async (somePath) => {
+  try {
+    await promisify(fs.access)(somePath, fs.constants.F_OK);
```
markov00, line 33: IN think you can use 
```
const fs = require('fs/promises');
```
for a fs promise

nickofthyme: I don't think there is a method for `access` but I'll replace the others and check.

packages/link_kibana/process_utils.js
nickofthyme: Yes I would agree in most cases but In this case I don't want it to reject because I am allowing the user to fix the errors. Calling reject would go onto the next call which I don't want.

markov00 (approved):
Tested locally and I love how fast we can now test the changes to the library directly within Kibana!
Thanks Nick, that's awesome!!!

PR #1167, opened 2021-05-26
fix: annotation tooltip display when remounting specs

### LINKED ISSUE #1063: Annotation tooltip not being displayed.
**Describe the bug**
When I move the mouse cursor from the data to the annotation, it will not display the tooltip for the annotation, but if I move my mouse out of the chart then back in on the annotation from the top, then it does display the tooltip.

**Expected behaviour**
The tooltip should be shown no matter where the mouse is coming from.

**Screenshots**

https://user-images.githubusercontent.com/55978943/112521586-0fc11680-8d73-11eb-949f-591a5b8557b3.mov

**Kibana Cross Issues**
https://github.com/elastic/kibana/pull/91339#issuecomment-792970360

nickofthyme:
## Summary

fix #1063

Fixes issue where annotation tooltip would not render when hovering annotation marker when using `useContext`.

## Details

The main issue here is that apm uses `useContext` on the `onPointerUpdate` listener callback. This causes the chart to rerender and re-upsert the chart specs. Doing this changes the object reference between `hoveredDOMElement.datum` and the datum in `annotationDimensions` map.

This was a check to ensure the correct tooltip is rendered but since the is of all specs must be unique, this check can be removed without concern.

This is a demo of the linked charts library with the changes included in this PR. Notice the annotation tooltip being shown in all hover cases.

https://user-images.githubusercontent.com/19007109/119690360-3c46ec80-be0f-11eb-9616-520ab8e2e513.mp4

### Checklist


markov00 (approved):
Perfectly agree, thanks Nick for the investigation on this.
The `id` is used here comes from  `getAnnotationLinePropsId` that already take in consideration the following properties:
```
[specId, verticalValue, horizontalValue, datum.header, datum.details, index]
```
and can be considered unique

Issue #1169, opened 2021-05-13
[APM] Profiling flamegraph text uses node id instead of label


trentm:
**Versions**
Kibana: master

**The issue:**

![Screen Shot 2021-05-13 at 1 45 48 PM](https://user-images.githubusercontent.com/46866/118185577-8bddef00-b3f1-11eb-84da-32dc7b702c84.png)

Those box labels in the flamegraph are using `.id` values rather than `.label` values. The tooltips are getting the correct values.  When it was working it looked like this:

![Screen Shot 2021-05-13 at 1 49 25 PM](https://user-images.githubusercontent.com/46866/118186005-0d358180-b3f2-11eb-8568-7be4457f39e1.png)

**Some details:**

In https://github.com/elastic/kibana/pull/91818, @dgieselaar added a "Profiling" section to the APM app with a flamegraph, using an icicle chart from `@elastic/charts`. At the time of that commit Kibana master was using `@elastic/charts@25.1.0`.

Kibana commit https://github.com/elastic/kibana/commit/2bfba7b4602 (#94505) broke it. That change updated to `@elastic/charts@25.3.0`.

Playing directly with charts version, v25.1.1 works, but v25.2.0 fails. That means the breaking change is in https://github.com/elastic/elastic-charts/compare/v25.1.1...v25.2.0  From those commits, **I assume it is https://github.com/elastic/elastic-charts/pull/1041 that changed behaviour to break this flamegraph in Kibana**. I don't know if it is a bug in elastic-charts or an incompatible change that Kibana's code need to adjust to. Both of these comments from that PR seem possibly relevant:

> - as the drilldown logic is different, the recently added groupByRollup and tree.ts additions for drilldown became unnecessary, they got removed
> - optimizes the clipped text render path for icicle/flame for speed, by adding a specialized render function with reduced API and function call count, and by avoiding multirow text fill calculation in viewmodel.ts

Here is Kibana's code creating this chart: https://github.com/elastic/kibana/blob/9d9dfe4bbf6b37243b509a99375067a8781e32a7/x-pack/plugins/apm/public/components/app/service_profiling/service_profiling_flamegraph.tsx#L123-L337

/cc @dgieselaar @monfera @nickofthyme

sqren:
Transfered to elastic-charts repo.

PR #1173, opened 2021-05-27
fix(flamegraph): render nodeLabel formatted text into the nodes

monfera:
## Summary

Fixes https://github.com/elastic/elastic-charts/issues/1169 which is indeed a regression introduced with the addition of drilldown/clipText fast path renderer.

The original test case used the id strings for labels; now the `nodeLabel` formatter adds a closing slash `/` to make them different, to test this aspect wing to wing:

image

### Checklist

- [X] Chart tests were updated or added to match the most common scenarios


nickofthyme (approved):
LGTM

PR #1174, opened 2021-05-27
feat(a11y): accessible goal and gauge chart

### LINKED ISSUE #1160: Accessible goal chart (numbers and strings)
Goals have few data points: 

- min
- max
- target
- value
- label major
- label minor

min, max, target, and value we should incorporate into the `<dl>`

The major label should be incorporated into the accessible labels for the chart

rshen91:
## Summary
Goal chart are more accessible! By default, users of assistive technologies will be able to learn more about goal charts such as the minimum, maximum, target, current value, and label major and minor. 

## Details
Fixes #1160 

### Checklist


packages/charts/src/components/accessibility/label.tsx
```diff
@@ -19,11 +19,30 @@
...
+  const Heading = labelHeadingLevel;
+  const goalChartLabelsSection = !goalChartLabels?.majorLabel
+    ? null
+    : `Goal chart label: ${goalChartLabels.majorLabel} ${goalChartLabels.minorLabel}`;
+
+  return (
+    <Heading id={labelId}>
+      {label}
+      {goalChartLabelsSection}
+    </Heading>
```
myasonik, line 46: I think this whole "while label to render" thing is, unfortunately, going to be more complicated...

So, first, the easy change:
The `minorLabel` we can just render as a `<p>` under the heading as there isn't a great semantic way to to signify a subheading but also shoving it all into a heading seems incorrect. So the DOM will look something like this:

```
return <>
	{unifedLabel && 
	      <Heading id={labelId}>
	        {unifiedLabel}
	      </Heading>
	}
    {goalChartLabels.minorLabel && <p>{goalChartLabels.minorLabel}</p>}
</>;
```

Now we've got the more difficult change that I hid under a new variable I made up called `unifiedLabel`.

1. Goal charts are kind of great in that they're one of the few that have a visible label built in. I imagine a separate accessible label doesn't need to actually be passed in very often so let's hide that bit of indirection from users.

2. If both an accessible label and a chart label are passed in, just from what I see users doing a lot, I'd bet they will often be identical so let's check for that too before printing both.

3. Finally, only if both are supplied and are different should we print both and make clear what's what.

```
let unifiedLabel = '';

if (!label && goalChartLabels?.majorLabel) unifiedLabel = goalChartLabels?.majorLabel;
else if (label && !goalChartLabels?.majorLabel) unifiedLabel = label;
else if (label && goalChartLabels?.majorLabel && label !== goalChartLabels.majorLabel) {
	unifiedLabel = `${label}; Chart visible label: ${goalChartLabels.majorLabel}`;
}
```

👆 I also changed the string a little bit. Because we don't know what the user might be passing in, and it seems a little strange to pass in both, I'm being super explicit about it but also trying not to duplicate info (we elsewhere say it's a goal chart).

Hope this all makes sense! Happy to zoom about it too!

myasonik (commented):
👏 👏 👏

packages/charts/src/components/accessibility/label.tsx
rshen91: I think I'm following!! 44b9eb5 for changes, thank you!

packages/charts/src/components/accessibility/label.tsx
```diff
@@ -19,11 +19,30 @@
...
-  if (!props.label) return null;
-  const Heading = props.labelHeadingLevel;
-  return <Heading id={props.labelId}>{props.label}</Heading>;
+export function ScreenReaderLabel({
+  label,
+  labelHeadingLevel,
+  labelId,
+  goalChartLabels,
+}: A11ySettings & ScreenReaderLabelProps) {
+  if (!label && !goalChartLabels?.majorLabel) return null;
```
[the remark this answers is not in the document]
rshen91: Yas - 44b9eb5 thanks!!

packages/charts/src/components/accessibility/types.tsx
```diff
@@ -19,19 +19,33 @@
...
+    chartTypeDescription === 'verticalBullet chart';
   return (
     <dl>
       <dt>Chart type:</dt>
-      <dd id={props.defaultSummaryId}>{props.chartTypeDescription}</dd>
+      <dd id={defaultSummaryId}>{chartTypeDescription}</dd>
+      {validGoalChart && goalChartData && <dd>{`Minimum: ${goalChartData.minimum}`}</dd>}
+      {validGoalChart && goalChartData && <dd>{`Maximum: ${goalChartData.maximum}`}</dd>}
+      {validGoalChart && goalChartData && <dd>{`Target: ${goalChartData.target}`}</dd>}
+      {validGoalChart && goalChartData && <dd>{`Value: ${goalChartData.value}`}</dd>}
```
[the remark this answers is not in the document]
rshen91: Thank you, this makes sense! 44b9eb5

packages/charts/src/components/accessibility/types.tsx
```diff
@@ -19,19 +19,41 @@
...
+    chartTypeDescription === 'horizontalBullet chart' ||
+    chartTypeDescription === 'verticalBullet chart';
   return (
     <dl>
       <dt>Chart type:</dt>
-      <dd id={props.defaultSummaryId}>{props.chartTypeDescription}</dd>
+      <dd id={defaultSummaryId}>{chartTypeDescription}</dd>
+      {validGoalChart && goalChartData ? (
+        <>
+          <dt>{'Minimum: '}</dt>
```
myasonik, line 47: Do we need the extra space at the end here? I don't think screen readers need it and I think this is hidden visually.

myasonik (commented):
Code LGTM, will test this next week before giving it a green light

packages/charts/src/components/accessibility/types.tsx
rshen91: Ah good catch thank you 4f8e86c

packages/charts/src/components/accessibility/label.tsx
```diff
@@ -19,11 +19,42 @@
...
+    unifiedLabel = label;
+  } else if (label && goalChartLabels?.majorLabel && label !== goalChartLabels?.majorLabel) {
+    unifiedLabel = `${label}; Chart visible label: ${goalChartLabels?.majorLabel}`;
+  }
+
+  return (
+    <>
+      {unifiedLabel && (
+        <Heading id={labelId}>
+          {label}
```
myasonik, line 53:
```suggestion
```

Can delete, the `unifiedLabel` includes it

myasonik (approved):
LGTM! 🚀

packages/charts/src/components/accessibility/label.tsx
rshen91: Thanks!!! Good catch 7ad9893

nickofthyme (approved):
LGTM

Issue #378, opened 2019-09-13
Convert to yarn workspaces


nickofthyme:
**Is your feature request related to a problem? Please describe.**
Switch to use [yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/).

This would enable to following:
- Cleaner dependencies in bundled library
- Separate dependencies for app and library
- Remove the dependency on `node@>10` #359 
- All your project dependencies will be installed together, giving Yarn more latitude to better optimize them

Issue #359, opened 2019-08-28
package.json "engine" is too restrictive


chandlerprall:
The `node` version in the package.json's `engines` field is too restrictive to include `elastic-charts` in other projects. For example, EUI's CI breaks when trying to add elastic charts as we are not on that exact `10.15.2` node version. Even if we were to update and track this version requirement, we'd be forcing all consumers of EUI to use that specific node version, even in non-Kibana use cases.

markov00:
Hey @chandlerprall 
I think we need to change the .nvmrc config instead. 
Quoting https://docs.npmjs.com/files/package.json#engines 
> Unless the user has set the engine-strict config flag, this field is advisory only and will only produce warnings when your package is installed as a dependency.

So I think it's our `.nvmrc` that create issues with the CI

markov00:
...and rethinking about my last comment, it's also more strange because the `.nvmrc` is not inside the distributed package, but only on the repo

markov00:
ok so actually that `note` is only for `npm` and doesn't apply to `yarn`.
On yarn the that thing is only applied when passing `--ignore-engines`
Let me check to what version we can downgrade to satisfy our packages and I will update the library

markov00:
So I've investigated a bit more the issue, here my findings:
- as part of our devDependency we are using `commitizen` that has a strict engine require of node >=10 https://github.com/commitizen/cz-cli
- when you execute `yarn` on the CI, `yarn` download the linked dependency of a direct dependency (in your case the dependencies of `@elastic/charts`) using `yarn --production`
- the issue is: `yarn` resolved the devDependencies to create deterministic build between prod and dev builds https://github.com/yarnpkg/yarn/issues/3630 
- this goes in hand with this issue: https://github.com/yarnpkg/yarn/issues/5796 where `yarn --production` does the engines check on dev dependency

So the problem now is: it doesn't  matter if we remove the `engines` configuration on the release `package.json`, we also have to get remove of the `commitizen` dependency at least. This, btw, will not completely create a deterministic build because there can maybe be shared dependencies between the existing packages.

Maybe the solution is completely remove the `commitizen` CLI from the package. @nickofthyme what do you think?

jbudz:
Hmm - [related](https://github.com/elastic/elastic-charts/pull/361), I'm trying to upgrade kibana's node version now and it's being blocked on this.  It looks like CI is installing kibana w/ a circular depdency on node version mismathes now :).  LMK if I can help

[...]

PR #1178, opened 2021-05-30
build(workspaces): implement src as workspace package

nickofthyme:
## Summary

Implement `@elastic/charts/src` as [yarn workspace](https://classic.yarnpkg.com/en/docs/workspaces/). Benefits of this are mostly internal but could improve productivity. These changes will have no effect on the library consumers.

## Details

Step towards closing #378. Ensured all npm scripts function as before, some even better 😏 🎉 

### List of benefits

- Cleaner library `package.json`.
- Greater control over app and library build process
- Cleaner dependencies in bundled library or libraries in the future
- Separate dependencies for app and library, isolation between apps and library (allows for #359)
- Applications can use built library instead of `src` files.
- Dependencies can be linked together, which means that your workspaces can depend on one another while always using the most up-to-date code available.
- All your project dependencies will be installed together, giving Yarn more latitude to better optimize them.
- Yarn will use a single lockfile rather than a different one for each project, which means fewer conflicts and easier reviews.

### List of key changes

- move `src/**` to `packages/elastic-charts/**`
- correct all paths due to file structure changes
- fix bad storybook-docs paths and config options
- create top-level mono package, move lib package
- simplify lib `tsconfig.json` files
- fix errors in `link_kibana` package, cleanup path logic
- fix linting config to reflect file structure changes
- update semantic release to point at new lib package
- add [`lerna`](https://github.com/lerna/lerna) and [`@lerna/run`](https://github.com/lerna/lerna/tree/main/commands/run#readme) to facilitate running workspace/package scripts

### File Structure

```bash
# before
elastic-charts/
├── src/
└── package.json

# after
elastic-charts/
├── packages/
│   ├── elastic-charts/
│   │   ├── src/
│   │   └── package.json
└── package.json
```

As you may have noticed, there are now two `package.json` files. The main top-level `package.json` (named `elastic-charts-mono`) is needed to control the repo and all its apps and libraries via `workspaces`. The other one at `packages/elastic-charts/package.json` is the package for the library itself, moved from the top-level.

One of the beauties of this new approach is offloading all of the `devDependencies` to the top level `package.json` that can be used across all workspaces.

> A more complete example of the workspace file structure is react, see https://github.com/facebook/react

### Limitations

Currently semantic release assumes a `1:1` relation between repo and packages (see https://github.com/semantic-release/semantic-release/issues/193). This is fine for now with only one external package, but this could be addressed when we have more packages to release.

### Checklist


packages/elastic-charts/src/reset_dark.scss
```diff
@@ -0,0 +1,6 @@
+@import '../../../node_modules/@elastic/eui/src/themes/eui/eui_colors_dark';
+@import 'style/themes/colors_dark';
+@import '../../../node_modules/@elastic/eui/src/global_styling/functions/index';
+@import '../../../node_modules/@elastic/eui/src/global_styling/variables/index';
+@import '../../../node_modules/@elastic/eui/src/global_styling/mixins/index';
+@import '../../../node_modules/@elastic/eui/src/global_styling/reset/index';
```
nickofthyme, line 6: git could not track this rename, but this is the same file as `src/reset_dark.scss`. Only the path changed to point at top `node_modules` folder.

packages/link_kibana/replace_modules.js
```diff
@@ -50,33 +50,45 @@ console.log(
...
+ * require('../react'); // points at kibana/node_modules/react
+ * ```
+ */
+const replaceModules = async (libDir, linkedPackages, packageName) => {
   if (linkedPackages.length === 0) return;
 
   const spinner = ora(`Replacing package links ${chalk.dim(`(${linkedPackages.join(',')})`)}`).start();
+  const cwd = path.join(libDir, 'dist');
+  const nodeModulesDir = path.join(libDir, 'node_modules');
+  const re = new RegExp(`require\\("(?:.*?\\/)*(${linkedPackages.join('|')})"\\)`, 'g');
```
nickofthyme, line 73: No simply converts to relative path. Then when linked elsewhere the dist files will use the relative `node_modules` depending on where the package is linked.

src/_eui_imports.scss
```diff
@@ -1,3 +0,0 @@
-@import '../node_modules/@elastic/eui/src/global_styling/functions/index';
-@import '../node_modules/@elastic/eui/src/global_styling/variables/index';
-@import '../node_modules/@elastic/eui/src/global_styling/mixins/index';
```
nickofthyme, line 3: Not delete only moved to `packages/elastic-charts/src`

tslint.yaml
```diff
@@ -1,15 +0,0 @@
...
-  # - node_modules/tslint-microsoft-contrib
-
-rules:
-  object-literal-sort-keys: false
-  interface-name: false
-  no-default-export: true
-  no-irregular-whitespace: true
-  no-unused-expression: true
-  quotemark: [true, 'single', 'jsx-double']
-  member-access: [true, 'no-public']
```
nickofthyme, line 15: We have not used tslint since I can recall

tsconfig.json
```diff
@@ -21,5 +20,5 @@
     "resolveJsonModule": true,
     "typeRoots": ["./node_modules/@types", "./**/*.d.ts", "./scripts/custom_matchers.ts"]
   },
-  "exclude": ["**/tmp"]
+  "exclude": ["**/tmp", "**/dist"]
```
nickofthyme, line 23: prevents IDE from attempting to auto import from `dist` directory

tsconfig.json
```diff
@@ -2,7 +2,6 @@
   "compilerOptions": {
     "declaration": true,
     "declarationMap": true,
-    "outDir": "./dist/",
```
nickofthyme, line 5: moved to `packages/elastic-charts/tsconfig.json`

storybook-docs/webpack.config.js
```diff
@@ -45,7 +45,7 @@ module.exports = async ({ config }) => {
       {
         loader: 'style-loader',
         options: {
-          attrs: {
+          attributes: {
```
nickofthyme, line 48: fixes broken config options that got stale

packages/elastic-charts/package.json
```diff
@@ -0,0 +1,66 @@
+{
+  "name": "@elastic/charts",
+  "description": "Elastic-Charts data visualization library",
+  "version": "29.2.0",
```
nickofthyme, line 4: moved from top-level to here. Removed all `devDependencies`.

packages/elastic-charts/package.json
```diff
@@ -0,0 +1,66 @@
...
+    "build:ts": "yarn build:clean && yarn build:compile && yarn build:check",
+    "build:css": "yarn build:sass && yarn autoprefix:css && yarn concat:sass",
+    "build:clean": "echo 'Cleaning dist...' && rm -rf ./dist",
+    "build:compile": "echo 'Compiling...' && tsc -p ./tsconfig.json && tsc -p ./tsconfig.nocomments.json",
+    "build:sass": "echo 'Building sass...' && node-sass src/theme_light.scss dist/theme_light.css --output-style compressed && node-sass src/theme_dark.scss dist/theme_dark.css --output-style compressed && node-sass src/theme_only_light.scss dist/theme_only_light.css --output-style compressed && node-sass src/theme_only_dark.scss dist/theme_only_dark.css --output-style compressed",
+    "build:check": "echo 'Type checking dist...' && tsc -p ./tsconfig.check.json",
+    "build:watch": "echo 'Watching build...' && yarn build:clean && yarn build:css && tsc -p ./tsconfig.json -w",
+    "concat:sass": "echo 'Concat SASS...' && node scripts/concat_sass.js",
+    "semantic-release": "semantic-release",
+    "typecheck": "tsc -p ./tsconfig.json --noEmit && tsc -p ./tsconfig.nocomments.json --noEmit"
```
nickofthyme, line 32: Moved non-library specific scripts to top level `package.json`

package.json
```diff
@@ -142,7 +135,7 @@
     "@typescript-eslint/eslint-plugin": "^4.12.0",
     "@typescript-eslint/parser": "^4.12.0",
     "@webpack-cli/serve": "^1.3.1",
-    "autoprefixer": "^10.2.5",
+    "autoprefixer": "^9.0.0",
```
nickofthyme, line 138: caused and error with postcss 🤔  See https://stackoverflow.com/a/64075012/6943587

package.json
```diff
@@ -212,24 +206,13 @@
...
   "config": {
     "commitizen": {
       "path": "./node_modules/cz-conventional-changelog"
     }
-  },
-  "resolutions": {
-    "conventional-changelog-conventionalcommits": "^4.3.0",
-    "refractor": "~3.1.0",
-    "postcss": "^8.2.10"
-  },
```
nickofthyme, line 230: These were to forcibly fix moderate security vulnerabilities. Removing to wait on deps to update themselves.

package.json
```diff
@@ -212,24 +206,13 @@
...
-  "browserslist": [
-    "last 2 versions",
-    "ie 11"
-  ]
```
nickofthyme, line 234: Moved to `packages/elastic-charts/package.json`

package.json
```diff
@@ -1,40 +1,33 @@
...
   ],
   "scripts": {
-    "autoprefix:css": "echo 'Autoprefixing...' && yarn postcss dist/*.css --no-map --use autoprefixer -d dist",
-    "api:check": "yarn build:ts && yarn api:extract",
-    "api:check:local": "yarn api:check --local",
-    "api:extract": "yarn api-extractor run --verbose",
+    "autoprefix:css": "lerna run --loglevel=silent --scope @elastic/charts autoprefix:css",
+    "api:check": "lerna run --loglevel=silent --scope @elastic/charts api:check",
+    "api:check:local": "lerna run --loglevel=silent --scope @elastic/charts api:check:local",
+    "api:extract": "lerna run --loglevel=silent --scope @elastic/charts api:extract",
```
nickofthyme, line 15: Kept all previous scripts, except for `watch` which I never used as it's faster to just saw `yarn test -w`.

Now repo-wide scripts are run from here and library scripts are run using [`@lerna/run`](https://github.com/lerna/lerna/tree/main/commands/run#readme)

I think we can cleanup these scripts more in the future or even have a cli much like `yarn kbn <action>`

package.json
```diff
@@ -1,40 +1,33 @@
...
-  "types": "dist/index.d.ts",
+  "private": "true",
   "repository": "git@github.com:elastic/elastic-charts.git",
-  "publishConfig": {
-    "access": "public"
-  },
-  "files": [
-    "dist/**/*"
+  "workspaces": [
+    "packages/*"
```
nickofthyme, line 9: must be private to add `workspaces`

package.json
```diff
@@ -1,40 +1,33 @@
 {
-  "name": "@elastic/charts",
-  "description": "Elastic-Charts data visualization library",
-  "version": "29.2.0",
+  "name": "elastic-charts-mono",
+  "description": "Mono repo for @elastic/charts",
```
nickofthyme, line 3: Open to a better name or description her just cannot be `@elastic/charts`

.releaserc.js
```diff
@@ -0,0 +1,36 @@
...
+    ...(isDryRun
+      ? []
+      : [
+          'semantic-release-slack-bot',
+          {
+            notifyOnSuccess: true,
+            notifyOnFail: true,
+            markdownReleaseNotes: true,
+          },
+        ]),
```
nickofthyme, line 34: This always fails when running in `dry-run` mode

.releaserc.js
```diff
@@ -0,0 +1,36 @@
...
+    '@semantic-release/release-notes-generator',
+    '@semantic-release/changelog',
+    '@semantic-release/github',
+    [
+      '@semantic-release/npm',
+      {
+        // must point to the child package
+        pkgRoot: 'packages/elastic-charts',
+      },
+    ],
```
nickofthyme, line 23: I'm pretty sure this will work and is the only change needed to point to the now nested library. 🤞🏼 

If this does not work, we can move this config to `packages/elastic-charts`.

We could update the `git` and `github` assets to only include the `dist` but I didn't see other orgs like `react` doing this on github.

nickofthyme (commented):
A few code comments to explain some of the key changes.

monfera:
What's the reason for putting `src` into somewhere deep? Also I'm not sure how I feel about `elastic-charts/packages/elastic-charts/src/` or similar paths with duplicated directory names. The outer `elastic-charts` refers to the containing directory, though its name is automatically determined by cloning the repo.

If I understand, `elastic-charts` is a single thing, and this file structuring change feels like we're turning it into some monorepo, what's the reason for this? Also, in this case, shouldn't the outer or inner `elastic-charts` be named differently? Eg. `elastic-charts/packages/charts/src/`, with the idea that maybe some of the logic that isn't charts dependent moves into eg. `elastic-charts/packages/charts-tooltip/src/` (assuming that our tooltip is made available to other users, without them having to use the entire charts library)

I'll put it on today's agenda to learn more

tsconfig.json
markov00: isn't that excluded with .gitignore?

storybook-docs/webpack.config.js
markov00: `storybook-docs` is still not much in synch with the rest of the storybook

markov00:
> If I understand, elastic-charts is a single thing, and this file structuring change feels like we're turning it into some monorepo, what's the reason for this? 

@monfera most of the benefits are described clearly in the PR description and in the 2 linked issues and can be resumed in a cleaner set of dependencies for the library and a clear distinction between library and the docs+storybook infrastructure.

This monorepo can be exploited in the future to exposing multiple independent packages within the same repo. One example could be: split the React Chart API from the Chart implementation and rendering whenever we have defined a clean JSON, JSON like API, so that one can just import the chart implementation and use the JSON API, or someone wants to use the react API bridge to build charts.

You are probably right about the folder naming: calling them `charts` is probably way better, and also reflect a bit more the current npm package name: `@elastic/charts`

monfera:
Thanks @markov00! 

> This monorepo can be exploited in the future to exposing multiple independent packages within the same repo

That's great, if it is a monorepo indeed. We've been talking about a decomposition anyway, so at some point, `charts` proper might only represent the (equivalent of the) current API. Would be possible to add other builds  too, eg. Cartesian-only etc., or share a common time axis logic and/or axis rendering with other teams, as a random example 😄

tsconfig.json
nickofthyme: I kept getting this `dist` import suggestions, maybe but I think dist was already gitignored

storybook-docs/webpack.config.js
nickofthyme: Yup I think we could simplify the build process between these two, cause keeping two virtually identical builds separate is tough.

nickofthyme:
> What's the reason for putting src into somewhere deep? Also I'm not sure how I feel about elastic-charts/packages/elastic-charts/src/ or similar paths with duplicated directory names. The outer elastic-charts refers to the containing directory, though its name is automatically determined by cloning the repo.

@monfera You are right It is a little deeper but the idea is that we have now the ability to have independent packages that are either coupled with `@elastic/charts` or their own independent package.

But the beauty about yarn workspaces is that they automatically symlink all dependent packages across the repo. Thus if I am in `stories/*` or any other top-level directory I can simply import from `@elastic/charts` instead of `../../src/*`. I have not done this yet just to focus on moving the charts library into packages with minimal code diff.

> If I understand, elastic-charts is a single thing, and this file structuring change feels like we're turning it into some monorepo, what's the reason for this? Also, in this case, shouldn't the outer or inner elastic-charts be named differently? Eg. elastic-charts/packages/charts/src/, with the idea that maybe some of the logic that isn't charts dependent moves into eg. elastic-charts/packages/charts-tooltip/src/ (assuming that our tooltip is made available to other users, without them having to use the entire charts library)

Yes it is a single thing, at least for the moment, but this not only allows for adding more packages in the future but also clearly defining boundaries of the library and the other apps. This includes moving all our application code such as `stories/`, `storybook[-docs]/`, `playground/`  into their own isolated workspace.

> This monorepo can be exploited in the future to exposing multiple independent packages within the same repo. One example could be: split the React Chart API from the Chart implementation and rendering whenever we have defined a clean JSON, JSON like API, so that one can just import the chart implementation and use the JSON API, or someone wants to use the react API bridge to build charts.

Exactly! Similarly, we could even have separate rending packages for `canvas`, `svg` and `webGL`.

markov00:
> Thus if I am in stories/* or any other top-level directory I can simply import from @elastic/charts instead of ../../src/*.

@nickofthyme  this is a great thing, it helps a lot with the testing as we isolate the lib from Storybook and help us identifying missing exposed types easily.

markov00:
@nickofthyme I've pull down this PR and run `yarn` to update the packages but got this error:
```
 yarn
yarn install v1.22.5
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
[3/5] 🚚  Fetching packages...
[4/5] 🔗  Linking dependencies...
warning " > @elastic/eui@32.1.0" has unmet peer dependency "@types/react@^16.9.34".
warning "@storybook/addon-actions > @storybook/api@5.3.21" has unmet peer dependency "regenerator-runtime@*".
warning "@storybook/addon-docs > @egoist/vue-to-react@1.1.0" has unmet peer dependency "vue@^2.6.10".
warning "@storybook/addon-docs > vue-docgen-api > @vue/compiler-sfc@3.0.0-rc.10" has unmet peer dependency "vue@3.0.0-rc.10".
[5/5] 🔨  Building fresh packages...
$ echo 'This package os not published, see pacakges/*' && exit 1
This package os not published, see pacakges/*
error Command failed with exit code 1.

```
Why is the `prepublish` script called here?

nickofthyme:
> @nickofthyme I've pull down this PR and run `yarn` to update the packages but got this error:
> 
> ```
>  yarn
> yarn install v1.22.5
> [1/5] 🔍  Validating package.json...
> [2/5] 🔍  Resolving packages...
> [3/5] 🚚  Fetching packages...
> [4/5] 🔗  Linking dependencies...
> warning " > @elastic/eui@32.1.0" has unmet peer dependency "@types/react@^16.9.34".
> warning "@storybook/addon-actions > @storybook/api@5.3.21" has unmet peer dependency "regenerator-runtime@*".
> warning "@storybook/addon-docs > @egoist/vue-to-react@1.1.0" has unmet peer dependency "vue@^2.6.10".
> warning "@storybook/addon-docs > vue-docgen-api > @vue/compiler-sfc@3.0.0-rc.10" has unmet peer dependency "vue@3.0.0-rc.10".
> [5/5] 🔨  Building fresh packages...
> $ echo 'This package os not published, see pacakges/*' && exit 1
> This package os not published, see pacakges/*
> error Command failed with exit code 1.
> ```
> 
> Why is the `prepublish` script called here?

Interesting, might be related to https://github.com/yarnpkg/yarn/issues/3209. I'll use `prepack` for now. The idea is to block that package from being published.

PR #1182, opened 2021-06-01
fix(value_labels): zero as valid value for textBorder and borderWidth

markov00:
## Summary

We recently introduced a readability enhancement that involves dropping a background opaque border around the value labels. The logic is now fixed to allow the user to hide this border using zero as a valid value in the `textBorder` or the `borderWidth` property of the `Theme.barSeriesStyle.displayValue.fill` theme style. 

BREAKING CHANGE: the `textBorder` of `ValueFillDefinition` is now optional or a number only

Screen Recording 2021-06-03 at 09 28 33 AM

### Checklist


rshen91 (approved):
Thanks for adding the new snapshot tests LGTM

src/chart_types/xy_chart/renderer/canvas/primitives/text.ts
```diff
@@ -56,13 +56,14 @@ export function renderText(
       }
       ctx.translate(origin.x, origin.y);
       ctx.scale(scale, scale);
-      if (font.shadow) {
+      const shadowSize = font.shadowSize ?? 0;
+      if (font.shadow && shadowSize > 0) {
+        ctx.save();
```
dej611, line 61: Is there any particular reason to prefer `save/restore` to the previous implementation?
Both commands are usually a huge performance hit

markov00: You are right, but if we want to restore the previous, we should save all the changing styles: like the `lineJoin`, the `lineWidth` and the `strokeStyle`.
I think we can directly remove the `save/restore` and just update the style directly without saving any state because we are already within the `withContext` that do a `save/restore`
There was a specific reason on why you have just saved/restored that single style about the `lineWidth`?

nickofthyme (approved):
LTGM, tested locally

src/chart_types/xy_chart/renderer/canvas/primitives/text.ts
dej611: I wasn't aware at the time of the internal `save/restore` thing with the `withContext` function.
I guess both implementation can just be removed then

markov00: @dej611 I've removed the save/restore https://github.com/elastic/elastic-charts/pull/1182/commits/2bb8901f7de13abd6dc9e8b8f298effb4823c35c

PR #1194, opened 2021-06-07
feat(xy): add onPointerUpdate debounce and trigger options

### LINKED ISSUE #906: Ability to determine precise y-axis mouse position on the stacked bar chart and represent it in the crosshair label
We are using the Elastic Stacked Bar Chart in Uptime to generate a waterfall chart for Synthetic monitors.  We are doing this by rotating the standard chart by 90 degrees, so it becomes a horizontal bar chart.

We have a requirement to provide time based information to the user as they move their mouse across the chart.  This is based on their mouse position horizontally, which relates to the y-axis on the chart because we have rotated in by 90 degrees.

It is not currently possible to get the precise y-axis position that the user’s mouse is on, which is what we are requesting in this issue.

In addition, we want to label the position of the precise position of the mouse in the crosshair label (https://elastic.github.io/elastic-charts/?path=/story/interactions--crosshair-with-time-axis).

Further details

```onPointerUpdate``` provides this:```{chartId: "51881c65-149a-487c-872a-72de61222410", type: "Over", unit: undefined, scale: "linear", value: CLOSEST_X_VALUE}``` so the closest we can get is the closest ```x``` value

```onElementOver``` provides:
```
[
  [
    {
      "x": 0,
      "y": 165.225,
      "mark": null,
      "accessor": "y1",
      "datum": {
        "x": 0,
        "y0": 31.366999999999994,
        "y": 165.225,
        "config": {
          "colour": "#edc5a2",
          "tooltipValue": "ssl: 133.858ms"
        }
      }
    },
    {
      "key": "spec{waterfallItems}yAccessor{y}splitAccessors{}",
      "specId": "waterfallItems",
      "yAccessor": "y",
      "splitAccessors": {},
      "seriesKeys": [
        "y"
      ]
    }
  ]
]
```
Which gets us the ```x```,```y0``` and ```y``` of the closest bar, but not the cursor position.

The props passed to the `customTooltip` component also provide information pertaining to the relevant bar, but not the current cursor position.

### PR DESCRIPTION
**Pull request by nickofthyme**
## Summary

Adds `onProjectionUpdate` listener to `Settings` spec to listen for changes to projected cursor values.

### BREAKING CHANGE
`PointerOverEvent` type now extends `ProjectedValues` and drops `value`. This effectively replaces `value` with `x`, `y`, `smVerticalValue` and `smHorizontalValue`

## Issues

Adds feature desired by logs to listen to cursor updated with projected values
closes #906

### Checklist

### CODE REVIEW

File: src/chart_types/xy_chart/state/selectors/on_pointer_move_caller.ts [hunk 1]
```diff
@@ -125,8 +137,14 @@ export function createOnPointerMoveCaller(): (state: GlobalChartState) => void {
...
-          if (settings && settings.onPointerUpdate && hasPointerEventChanged(tempPrev, nextPointerEvent)) {
-            settings.onPointerUpdate(nextPointerEvent);
+          if (onPointerUpdate && hasPointerEventChanged(tempPrev, nextPointerEvent, true))
+            onPointerUpdate(nextPointerEvent);
+
+          if (onProjectionUpdate && values && hasPointerEventChanged(tempPrev, nextPointerEvent, false)) {
+            const oldYValues = yValuesString;
+            yValuesString = values.y.map(({ value }) => value).join(',');
+
+            if (oldYValues !== yValuesString) onProjectionUpdate(values);
```
nickofthyme, line 147: @markov00 I'm not sure about this logic. I want to only run this listener when the y values change but the `PointerEvent` only includes the x values. Should I add the `yValues` as optional to the `PointerEvent` or do you think this is ok?

packages/charts/src/chart_types/xy_chart/state/selectors/on_pointer_move_caller.ts
```diff
@@ -125,8 +137,14 @@ export function createOnPointerMoveCaller(): (state: GlobalChartState) => void {
...
-          if (settings && settings.onPointerUpdate && hasPointerEventChanged(tempPrev, nextPointerEvent)) {
-            settings.onPointerUpdate(nextPointerEvent);
+          if (onPointerUpdate && hasPointerEventChanged(tempPrev, nextPointerEvent, true))
+            onPointerUpdate(nextPointerEvent);
+
+          if (onProjectionUpdate && values && hasPointerEventChanged(tempPrev, nextPointerEvent, false)) {
+            const oldYValues = yValuesString;
+            yValuesString = values.y.map(({ value }) => value).join(',');
+
+            if (oldYValues !== yValuesString) onProjectionUpdate(values);
```
nickofthyme, line 147: @markov00 I'm not sure about this logic. I want to only run this listener when the y values change but the `PointerEvent` only includes the x values. Should I add the `yValues` as optional to the `PointerEvent` or do you think this is ok?

src/chart_types/xy_chart/state/selectors/on_pointer_move_caller.ts
nickofthyme: See new comments on rebase code

packages/charts/src/chart_types/xy_chart/state/selectors/on_pointer_move_caller.ts
nickofthyme: @markov00?

markov00: Yep, I think we could just reuse the `onPointerUpdate?: PointerUpdateListener;`

You can merge the definitions of `PointerOverEvent` and `ProjectedValues` to have a global and more intuitive pointer position like:
```ts
interface PointerOverEvent extends BasePointerEvent, ProjectedValues{
  type: typeof PointerEventType.Over;
  scale: ScaleContinuousType | ScaleOrdinalType;
  /**
   * Unit for event (i.e. `time`, `feet`, `count`, etc.) Not currently used/implemented
   * @alpha
   */
  unit?: string;
// value removed in favour of ProjectedValues x
};
```

nickofthyme: Ok thanks, Do you think I should debouce the listener? I'm not really sure how given how the listener is called in the selector.

markov00: I haven't used a debounce for the current onPointerUpdate used to sync charts together.
we can try but adding a way to disable/configure the debounce time

markov00: let default to 16ms debounce to have 1 event per frame max

nickofthyme: See 232f929 for updates the pointer logic. I'll add the debounce in another commit.

packages/charts/src/chart_types/xy_chart/state/selectors/get_projected_pointer_position.ts
```diff
@@ -92,10 +92,10 @@ function getPosRelativeToPanel(panelScale: ScaleBand, pos: number): { pos: numbe
     if (relativePos > panelScale.bandwidth) {
       return { pos: -1, value: null };
     }
-    return { pos: relativePos, value: panelScale.domain[relativePosIndex] };
+    return { pos: relativePos, value: panelScale.domain[relativePosIndex] ?? null };
   }
   return {
     pos: posWOInitialOuterPadding - panelScale.step * (numOfDomainSteps - 1),
-    value: panelScale.domain[numOfDomainSteps - 1],
+    value: panelScale.domain[numOfDomainSteps - 1] ?? null,
```
nickofthyme, line 99: These values could be undefined which does not match the expected `PrimativeType`

packages/charts/src/chart_types/xy_chart/state/selectors/on_pointer_move_caller.ts
nickofthyme: Ok debouncing changes are done here 5f085b5

markov00 (commented):
I don't see much difference between the `onPointerUpdate` and the `onProjectionUpdate`
if not the logic of checking the `y` values.
I think we can easily get rid of the `onProjectionUpdate`, keeping only the `onPointerUpdate`.
We can probably improve the code if the user can specify what to listen to:
- changes to the x position
- changes to the y position
- both
In this way, if the users are interested in the x movements, they receive updates only when the x position changed (in particular useful when using large bars where the x only change between bars and large movements). If someone is just interested in the Y space, then they can listen just to y changes.
Finally they can listen for both changes and use the debounce to limit the amount of data received

PR #1198, opened 2021-06-08
build(workspaces): implement src as workspace package (#1178)

nickofthyme:
Backports the following commits to master:
 - build(workspaces): implement src as workspace package (#1178)


nickofthyme:
Verified release functions as normal. Installed released alpha version in kibana without issue.

I noticed some top-level `.txt` and `.md` files were not being included so for the time being (i.e. before moving to esbuild, etc) I added a copy step to the build process.

PR #1199, opened 2021-06-08
fix(partition): getLegendItemsExtra no longer assumes a singleton

monfera:
## Summary

The legend extras (values) now show up in the legend for slices of all panels of a partition small multiple. In the past, the collection of values assumed the presence of one partition only.

image

### BREAKING CHANGE
The `LegendPath` array in event callbacks gained a new element for partition charts, which represents the positional `index` and stringified `value` of the `groupBy` categorical breakdown, if any; `NULL_SMALL_MULTIPLES_KEY` if not applicable, ie. it's not part of a small multiples esemble.

## Details

Go to the http://localhost:9001/?path=/story/small-multiples-alpha--sunbursts knobs to activate Show legend extra. All values should show up

## Issues

The issue was identified via searching the code for todo comments left in the partition small multiples, referencing singletons.

### Checklist


packages/charts/src/chart_types/partition_chart/partition.test.tsx
```diff
@@ -130,6 +130,7 @@ describe('Retain hierarchy even with arbitrary names', () => {
           childId: 'A',
           color: 'rgba(128, 0, 0, 0.5)',
           path: [
+            { index: 0, value: '' },
             { index: 0, value: HIERARCHY_ROOT_KEY },
```
nickofthyme, line 134: What's the significance of this new path?

nickofthyme (approved):
Changes LGTM, tested locally. How do the legend values represent values across partitions? Are they all just lumps together under the same legend path?

packages/charts/src/chart_types/partition_chart/partition.test.tsx
monfera: Thanks Nick. Re this question and also 

> How do the legend values represent values across partitions? Are they all just lumps together under the same legend path?

`path` had been responsible for carrying, well, the path of the hierarchical breakdown in partitioning charts, which are currently all hierarchical (tree-like). However, with the introduction of small multiples, the very first breakdown is the small multiples breakdown, which now got added. The rest of the breakdown, which used to start with `HIERARCHY_ROOT_KEY` so that there's a single-node root for a pie/treemap/etc was left unchanged. (The reason for a single node as a root is that it's conceptually a tree with a single root, even if eg. a pie chart or sunburst doesn't show it, as the first visible, rendered layer is already the breakdown into several slices)

nickofthyme: Nice, makes sense 👍🏼

markov00 (commented):
I tested it locally and compared it to the current version and we can finally see all the legend extra values on small multiples too.

The is a major thing here to note: this introduces a BREAKING CHANGE: the `LegendPath` is actually exposed as API, the API is not changed itself, but the actual value now includes one more depth level (the actual small multiple one).

Are we sure we want to add an unclear `{index: 0, value: ""}` at level 0 when we are not using small multiples?
In any case this still a BREAKING CHANGE

packages/charts/src/chart_types/partition_chart/state/selectors/get_legend_items_extra.test.ts
```diff
@@ -97,7 +97,7 @@ describe('Partition - Legend item extra values', () => {
     MockStore.addSpecs([settings, spec], store);
 
     const extraValues = getLegendItemsExtra(store.getState());
-    expect([...extraValues.keys()]).toEqual(['0', '0__0', '0__1']);
+    expect([...extraValues.keys()]).toEqual(['0__0', '0__0__0', '0__0__1']);
```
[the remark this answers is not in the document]
monfera: > the API is not changed itself, but the actual value now includes one more depth level 

I agree, I haven't considered it as the type signature hasn't changed. 

> Are we sure we want to add an unclear {index: 0, value: ""} at level 0 when we are not using small multiples?

Looks possible to avoid this. I very briefly though about it while changing the code and brushed it aside as it also creates a special case that the user code needs to handle with conditional branching in their event handler depending on whether or not it's a small multiples ensemble. When there's no clear benefit either way, I'm leaning toward avoiding corner cases (of eg. singleton chart) as it complicates code paths, and sometimes forces types into union types in our code (not in this case) or in the userland, which, all things being equal, is good to avoid.

Nonetheless I'm open to conditionally adding the first layer if you have an argument for it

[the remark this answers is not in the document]
markov00: I agree with you about not having corner cases.
I'm wondering if useful to have a more descriptive value there, a `value: null` o `value:__no_sm__` can help understand better what that path value is actually about? if not we can probably document the `LegendPath` type describing that the first 2 elements are always `[sm category, an empty root node, .....]`

rshen91 (approved):
LGTM tested locally!

monfera:
jenkins test this please

Issue #1148, opened 2021-05-06
Memory leak in elastic-charts


timroes:
Elastic charts currently (in the version on Kibana master 29.1.0) has a memory leak in it's caching layer, that keeps a huge amount of detached DOM nodes around. We're using `re-reselect` in some place, that by default uses a `FlatObjectCache` to cache selectors. That will lead to those selectors simply be "cached forever". We use this via the `getSeriesSpecsSelector` funtion and our state that we cache actually has references to DOM nodes around, that might later be detached, but then kept alive by this cache, preventing the browser from cleaning up the DOM node (and a potential large DOM tree attached to them).

We should ideally make sure that the state that we cache there, does not keep strong references to DOM elements around, that will prevent them from being garbage collected. Alternatively (or better in addition) we can tell `re-reselect` what cache object it should use (https://github.com/toomuchdesign/re-reselect#cacheobject) and might want to consider one that at least limits the cache size.

This problem is causing problems (high memory consumption, huge detached DOM node count) in placed like Discover where we exchange the chart, e.g. when switching index patterns. If you continue switching index patterns you will see the amount of DOM nodes and JS heap size growing and never be reduced, due to elastic-charts keeping them alive via this cache.

nickofthyme:
Thanks for the detailed description Tim, we'll get on it right away.

nickofthyme:
More findings:

I am seeing a retained reference the the `.echTooltip` element which is a child of the `createPortal` call. Could be related to https://github.com/facebook/react/issues/18790, also could be a dev-only memory issue.

PR #1201, opened 2021-06-10
fix: memory leak related to re-reselect cache

nickofthyme:
## Summary

Fixes memory leak related to re-reselect cache holding onto old chart store references.

## Details

Now selectors are creeated using a centralized method `createCustomCachedSelector` to facilitate `keySelector` and `objectCache` for all selectors in one place.

```diff
- selector = createCachedSelector(
+ selector = createCustomCachedSelector(
  [getSpecOrNull, getPickedShapesLayerValues, getSettingsSpecSelector],
  getOnElementOverSelector(prev),
- )({
-  keySelector: getChartIdSelector,
-});
+);

// Or similarly 
- selector = createCachedSelector(
+ selector = createCustomCachedSelector(
  [getSpecOrNull, getPickedShapesLayerValues, getSettingsSpecSelector],
  getOnElementOverSelector(prev),
- )((state) => state.chartId);
+);
```

> There is an added lint rule to prevent importing `createCachedSelector` from `re-reselect`.

The types for the new `createCustomCachedSelector` method are `@ts-ignored` in order to facilitate this central method for creating selectors as the [`createCachedSelector` type](https://github.com/toomuchdesign/re-reselect/blob/6e870b7352fae16f470f1d77358e91158e0f4609/src/index.d.ts#L75-L4252) uses over 90 overrides with generics to type the function.

### Before

Detached nodes in memory heap snapshot point to chart id in `FlatObjectCache` in `re-reselect`. The `FlatObjectCache` keeps around chart states based on ids that have since been randomized and recreated, thus the `_cache` grows proportional to the count of remounted charts.👇🏼 

https://user-images.githubusercontent.com/19007109/121589848-2b949a00-c9fd-11eb-8ffb-509557d5cf2d.mp4

Dom elements are much higher and continue to grow after charts are mounted and unmounted repeatedly, even after forced garbage collection. 👇🏼 

https://user-images.githubusercontent.com/19007109/121589503-c17bf500-c9fc-11eb-9087-8115dab8f74c.mp4

### After

Detached nodes in memory heap snapshot no longer point to chart id in `FlatObjectCache` in `re-reselect`. 👇🏼 

https://user-images.githubusercontent.com/19007109/121588010-043ccd80-c9fb-11eb-9571-7bed2abc9293.mp4

Dom elements are now being garbage collected after charts are mounted and unmounted repeatedly. 👇🏼 

https://user-images.githubusercontent.com/19007109/121588218-47973c00-c9fb-11eb-9a74-32e1c772fa30.mp4

## Issues

Related to #1148, still need to address `tooltip_portal` memory leaks

### Checklist

### DIFF
--- a/packages/charts/src/components/chart_status.tsx
+++ b/packages/charts/src/components/chart_status.tsx
@@ -35,9 +35,7 @@ interface ChartStatusStateProps {
   debugState: DebugState | null;
 }
 
-type Props = ChartStatusStateProps;
-
-class ChartStatusComponent extends React.Component<Props> {
+class ChartStatusComponent extends React.Component<ChartStatusStateProps> {
   componentDidMount() {
     this.dispatchRenderChange();
   }
--- a/packages/charts/src/components/chart_status.tsx
+++ b/packages/charts/src/components/chart_status.tsx
@@ -22,23 +22,32 @@ import { connect } from 'react-redux';
 
 import { RenderChangeListener } from '../specs';
 import { GlobalChartState } from '../state/chart_state';
+import { globalSelectorCache } from '../state/create_selector';
 import { getDebugStateSelector } from '../state/selectors/get_debug_state';
 import { getSettingsSpecSelector } from '../state/selectors/get_settings_specs';
 import { DebugState } from '../state/types';
 
 interface ChartStatusStateProps {
+  chartId: string;
   rendered: boolean;
   renderedCount: number;
   onRenderChange?: RenderChangeListener;
   debugState: DebugState | null;
 }
-class ChartStatusComponent extends React.Component<ChartStatusStateProps> {
+
+type Props = ChartStatusStateProps;
+
+class ChartStatusComponent extends React.Component<Props> {
   componentDidMount() {
     this.dispatchRenderChange();
   }
 
   componentDidUpdate() {
     this.dispatchRenderChange();
+  }
+
+  componentWillUnmount() {
+    globalSelectorCache.removeKeyFromAll(this.props.chartId);
   }
 
   dispatchRenderChange = () => {
@@ -69,6 +78,7 @@ const mapStateToProps = (state: GlobalChartState): ChartStatusStateProps => {
   const { onRenderChange, debugState } = getSettingsSpecSelector(state);
 
   return {
+    chartId: state.chartId,
     rendered: state.chartRendered,
     renderedCount: state.chartRenderedCount,
     onRenderChange,
--- a/packages/charts/src/state/create_selector.ts
+++ b/packages/charts/src/state/create_selector.ts
@@ -28,12 +28,7 @@ import { GlobalChartState } from './chart_state';
  * see https://github.com/toomuchdesign/re-reselect/tree/master/src/cache#write-your-custom-cache-object
  */
 class CustomMapCache implements ICacheObject {
-  debug: boolean;
   private cache: any = {};
-
-  constructor(debug: boolean = false) {
-    this.debug = debug;
-  }
 
   set(key: string, selectorFn: () => any) {
     this.cache[key] = selectorFn;
@@ -67,10 +62,10 @@ class GlobalSelectorCache {
     return chartId;
   }
 
-  getNewOptions(debug?: boolean): Options<GlobalChartState, unknown, unknown> {
+  getNewOptions(): Options<GlobalChartState, unknown, unknown> {
     return {
       keySelector: GlobalSelectorCache.keySelector,
-      cacheObject: this.getCacheObject(debug),
+      cacheObject: this.getCacheObject(),
     };
   }
 
@@ -80,8 +75,8 @@ class GlobalSelectorCache {
     });
   }
 
-  private getCacheObject(debug?: boolean): CustomMapCache {
-    const cache = new CustomMapCache(debug);
+  private getCacheObject(): CustomMapCache {
+    const cache = new CustomMapCache();
     this.selectorCaches.push(cache);
 
     return cache;
--- a/packages/charts/src/state/create_selector.ts
+++ b/packages/charts/src/state/create_selector.ts
@@ -1,0 +1,112 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// eslint-disable-next-line no-restricted-imports
+import createCachedSelector, { ICacheObject, Options } from 're-reselect';
+import type { createSelector } from 'reselect';
+
+import { GlobalChartState } from './chart_state';
+
+/**
+ * Custom object cache
+ * see https://github.com/toomuchdesign/re-reselect/tree/master/src/cache#write-your-custom-cache-object
+ */
+class CustomMapCache implements ICacheObject {
+  debug: boolean;
+  private cache: any = {};
+
+  constructor(debug: boolean = false) {
+    this.debug = debug;
+  }
+
+  set(key: string, selectorFn: () => any) {
+    this.cache[key] = selectorFn;
+  }
+
+  get(key: string) {
+    return this.cache[key];
+  }
+
+  remove(key: string) {
+    delete this.cache[key];
+  }
+
+  clear() {
+    this.cache = {};
+  }
+
+  isEmpty() {
+    return Object.keys(this.cache).length === 0;
+  }
+
+  isValidCacheKey(key: string) {
+    return typeof key === 'string';
+  }
+}
+
+class GlobalSelectorCache {
+  private selectorCaches: CustomMapCache[] = [];
+
+  static keySelector({ chartId }: GlobalChartState) {
+    return chartId;
+  }
+
+  getNewOptions(debug?: boolean): Options<GlobalChartState, unknown, unknown> {
+    return {
+      keySelector: GlobalSelectorCache.keySelector,
+      cacheObject: this.getCacheObject(debug),
+    };
+  }
+
+  removeKeyFromAll(key: string) {
+    this.selectorCaches.forEach((cache) => {
+      cache.remove(key);
+    });
+  }
+
+  private getCacheObject(debug?: boolean): CustomMapCache {
+    const cache = new CustomMapCache(debug);
+    this.selectorCaches.push(cache);
+
+    return cache;
+  }
+}
+
+/**
+ * Global singleton to manage state of selector caches
+ *
+ * @internal
+ */
+export const globalSelectorCache = new GlobalSelectorCache();
+
+/**
+ * Wrapper around `createCachedSelector` to provide `keySelector` and `cacheObject`
+ * for all selector instances in on place. This should be used in place of `createCachedSelector`.
+ *
+ * The types defining `createCachedSelector` are very complex and essentially hardcoded overloads for having any
+ * number of selector inputs up to about 20 with genetic types. Thus the types are extremely hard to duplciate.
+ * To fix this I used the type of `createSelector` which is what is the same as that of `createCachedSelector`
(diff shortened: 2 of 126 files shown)


packages/charts/src/state/create_selector.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+class GlobalSelectorCache {
+  private selectorCaches: CustomMapCache[] = [];
+
+  static keySelector({ chartId }: GlobalChartState) {
+    return chartId;
+  }
+
+  getNewOptions(debug?: boolean): Options<GlobalChartState, unknown, unknown> {
```
markov00, line 70: Is there a specific reason for the `debug` argument?

markov00 (commented):
Overall looks good to me.
I've left few comments that we should discuss before final approval

packages/charts/src/state/create_selector.ts
```diff
@@ -0,0 +1,112 @@
...
+
+import { GlobalChartState } from './chart_state';
+
+/**
+ * Custom object cache
+ * see https://github.com/toomuchdesign/re-reselect/tree/master/src/cache#write-your-custom-cache-object
+ */
+class CustomMapCache implements ICacheObject {
+  debug: boolean;
+  private cache: any = {};
```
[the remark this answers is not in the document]
nickofthyme: I don't know how `WeakMap` can work here since the key must be an object. I don't know what that object would be that would persist across changes since redux returns a new object for every reducer. I tried to use a weak map on the chart element and the redux store and both caused an infinite loop of updating.

packages/charts/src/state/create_selector.ts
```diff
@@ -0,0 +1,112 @@
...
+    return {
+      keySelector: GlobalSelectorCache.keySelector,
+      cacheObject: this.getCacheObject(debug),
+    };
+  }
+
+  removeKeyFromAll(key: string) {
+    this.selectorCaches.forEach((cache) => {
+      cache.remove(key);
+    });
```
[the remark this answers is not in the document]
nickofthyme: Good point. However, it looks like ~30 new selectors are added to this array each time the chart is created. I don't see a way to cleanup the old/unused selectors. Any thoughts?

packages/charts/src/state/create_selector.ts
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * The types defining `createCachedSelector` are very complex and essentially hardcoded overloads for having any
+ * number of selector inputs up to about 20 with genetic types. Thus the types are extremely hard to duplciate.
+ * To fix this I used the type of `createSelector` which is what is the same as that of `createCachedSelector`
+ * method with the added curring for the cached options which this wrapper handles.
+ *
+ * @internal
+ */
+export const createCustomCachedSelector: typeof createSelector = (...args: any[]) => {
+  // @ts-ignore - forced types to simplify usage. All types align correctly
```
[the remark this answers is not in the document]
nickofthyme: Yeah I avoided doing this before because of the nasty types but I think this is the safest and DRYest solution.

packages/charts/src/state/create_selector.ts
nickofthyme: Nope, removed here c07f279

packages/charts/src/components/chart_status.tsx
```diff
@@ -22,17 +22,22 @@ import { connect } from 'react-redux';
...
   rendered: boolean;
   renderedCount: number;
   onRenderChange?: RenderChangeListener;
   debugState: DebugState | null;
 }
-class ChartStatusComponent extends React.Component<ChartStatusStateProps> {
+
+type Props = ChartStatusStateProps;
+
+class ChartStatusComponent extends React.Component<Props> {
```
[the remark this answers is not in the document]
nickofthyme: d9903a4

packages/charts/src/state/create_selector.ts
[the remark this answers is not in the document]
nickofthyme: Also if we use `WeakMap` then wouldn't it defeat the purpose of `removing` keys from the cache when the chart is destroyed since the reference is now weak?

packages/charts/src/state/create_selector.ts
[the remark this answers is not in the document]
nickofthyme: Discussed this over zoom, non-issue

packages/charts/src/state/create_selector.ts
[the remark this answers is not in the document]
nickofthyme: Discussed over zoom, could use `WeakMap` in the future.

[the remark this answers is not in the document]
nickofthyme: @markov00 are you saying to use a `Map` here instead of an `object`?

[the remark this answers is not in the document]
markov00: We can also use `Map` actually, just a suggestion not a request

packages/charts/src/components/chart_status.tsx
```diff
@@ -69,6 +76,7 @@ const mapStateToProps = (state: GlobalChartState): ChartStatusStateProps => {
   const { onRenderChange, debugState } = getSettingsSpecSelector(state);
 
   return {
+    chartId: state.chartId,
```
markov00, line 79: The test fails due to this added property, you can easily update the snapshot and everything else looks fine

packages/charts/src/state/create_selector.ts
```diff
@@ -0,0 +1,107 @@
...
+export const globalSelectorCache = new GlobalSelectorCache();
+
+/**
+ * Wrapper around `createCachedSelector` to provide `keySelector` and `cacheObject`
+ * for all selector instances in on place. This should be used in place of `createCachedSelector`.
+ *
+ * The types defining `createCachedSelector` are very complex and essentially hardcoded overloads for having any
+ * number of selector inputs up to about 20 with genetic types. Thus the types are extremely hard to duplciate.
+ * To fix this I used the type of `createSelector` which is what is the same as that of `createCachedSelector`
+ * method with the added curring for the cached options which this wrapper handles.
```
monfera, line 100: That's a fantastic TS approach, I wish I thought of it! So, basically, you can wrap `re-reselect` around, without having to duplicate its dozen or to signatures for variadic-ness. 

Maybe not for this PR: since before joining the team, we've been talking with @markov00  about using a commonly used, FP-style function signature for a `createCachedSelector` wrapper. So, instead of 
```
const plainFun = (a1, a2, a3) => { whatever }
const newSelector = createCustomCachedSelector([arg1, arg2, arg3), plainFun)
```
it'd look like 
```
const plainFun = (a1, a2, a3) => { whatever }
const newSelector = select(plainFun)(arg1, arg2, arg3)
```
Advantages:
- no need for an array, which is a bit of anti-pattern, as these are arguments
- currying order is sensible: partial application would start with the function argument, not the list of inputs
- we don't bake in some specific library's awkward signature convention and overly specific function name, therefore we can easily switch from `re-reselect` to eg. data flow libraries like rxjs, flyd or anything we devise (liteFields uses `crosslink` though maybe not yet wrapped in a `select`)
- we already use this, FP-wise common place signature in Canvas, all over the place [here](https://github.com/elastic/kibana/blob/4a541883557bcee83baf09b2c0ddab702f780e45/x-pack/plugins/canvas/public/lib/aeroelastic/layout.js#L170-L175)
- lastly, name (just `select`): shorter is better, and doesn't encode how it works; there's not a big variety

So I wonder if maybe this could be done in this PR, as it touches so many files anyway. Could of course be a separate one but then it again touches so many files.

FP note: this kind of `select` operation is called _lift_ operation, because it lifts a normal, plain old function such as `(a, b) => a + b` into the realm of a higher order approach (streams, transducers, whatever), eg. our current practice of composing with selectors. Often, it's called, appropriately, a `lift` operation, eg. [for flyd](https://www.npmjs.com/package/flyd-lift). 

One real world application to such plumbing is that it's possible to (eventually) be smarter about executing data pipelines. For example, recognizing that some functions have one input and one user, you can identify if you have some stretches of pipelineable operators, and you can just execute `a(b(c(...args)))` ie. compose, instead of caching or whatnot at every single function boundary, and it's only the shallowest optimization opportunity.

monfera: Oh, wanted to list the drawbacks I know of, too:

[end of list]
😺

packages/charts/src/state/selectors/get_accessibility_config.ts
```diff
@@ -81,7 +80,7 @@ export const getA11ySettingsSelector = createCachedSelector(
       tableCaption: ariaTableCaption,
     };
   },
-)(getChartIdSelector);
+);
```
monfera, line 83: Great to have disappeared this last arg!

nickofthyme: Yeah given it was different all over the place 😅, it's nice to finally be gone

packages/charts/src/state/create_selector.ts
nickofthyme: I think this could be a good idea, I am generally fine with this change. The one issue I see is that the types as they are now, the `inputSelectors` arguments passed to `createSelector` implicitly create the type of the `resultFunc` very nicely. In your new approach we would need to define the type of the `resultFunc` for all selectors. Thus an example would be...

```ts
// fake selectors
const getString = (): string => 'hello';
const getNum = (): number => 21;
const getArray = (): number[] => [1, 2];

// current
const newSelector = createCustomCachedSelector([getString, getNum, getArray], (s, n, a) => {
  console.log(s); // knows it's a string
  console.log(n); // knows it's a number
  console.log(a); // knows it's an array
})

// new approach
const newSelector = createCustomCachedSelector((s, n, a) => {
  console.log(s); // assumes any
  console.log(n); // assumes any
  console.log(a); // assumes any
})([getString, getNum, getArray])
```

Thus this change would add a lot more duplicated (i.e. non-derived) typings, than the current approach.

packages/charts/src/state/create_selector.ts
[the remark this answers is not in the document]
markov00: let's keep as it is for now!

packages/charts/src/state/create_selector.ts
monfera: Thanks @nickofthyme! The 2nd example would not use the array, ie.
```
// new approach
const newSelector = select((s, n, a) => {
  console.log(s); // assumes any
  console.log(n); // assumes any
  console.log(a); // assumes any
})(getString, getNum, getArray)
```
Probably it doesn't make a difference regardig your point. Would be great if we could "pick" bits and pieces of an existing function signature. Likely not possible though

nickofthyme: Thanks I updated my comment but yeah that doesn't change my point.

monfera: What would happen, if instead of today's signature (btw should it be `unknown` instead of `any`?)
```
export const selector: typeof createSelector = (...args: any[]) => {
  return createCachedSelector(...args)(globalSelectorCache.getNewOptions());
};
```
it'd be
```
declare function select<R, S, A, B>(fun: (a: A, b: B) => R): (a: S => A, b: S => B) => (S => R)
```
where R is the result type, S is some opaque state object, and A and B are the two params of the plain function that we lift.

If this is repeated for all options `re-reselect` already supports, ie. 12 params, then we have 12 + 1 lines, like
```
declare function select<R, S>(fun: () => R): () => (S => R) // 0 args
declare function select<R, S, A>(fun: (a: A) => R): (a: S => A) => (S => R) // 1 arg
declare function select<R, S, A, B>(fun: (a: A, b: B) => R): (a: S => A, b: S => B) => (S => R) // 2 args
declare function select<R, S, A, B, C>(fun: (a: A, b: B, c: C) => R): (a: S => A, b: S => B, c: S => C) => (S => R) // 3 args
...
```
at least I had approx. this mental model, but haven't tried, is it something that can be ruled out?

nickofthyme: Do you mean...?

```ts
declare function select<R, S, A, B>(fun: (a: A, b: B) => R): (f1: (a: S) => A, f2: (b: S) => B) => (c: S) => R; // 2 args
```

Could you use `select` in an example as you describe above?

monfera: Yes, totally haven't added the arg names 👯 

An example would be something like:
```
// opaque state object
const s = {
  one: 14,
  other: 32,
  somethingElse: 9
}

// primitive selectors
const getOne = s => s.one
const getTheOther = s => s.other
const getSomethingElse = s => s.somethingElse

// plain functions we're gonna lift
const add = (a, b) => a + b
const mul = (a, b) => a * b
const sub = (a, b) => a - b

// ze selector maker
const select = f => (...inputSelectors) => s => f(...inputSelectors.map(sel => sel(s)))

// selectors
const getSum = select(add)(getOne, getTheOther)
const getMul = select(mul)(getTheOther, getSomethingElse)
const getSub = select(sub)(getMul, getSum)

// the test
console.log(getSub(s))
// => 242 // === (32 * 9) - (14 + 32)
```
Did you mean an example like this?

PR #1207, opened 2021-06-16
feat(annotations): option to render rect annotations outside chart

### LINKED ISSUE #928: RectAnnotation markers or annotations outside axis
Some use cases require the annotation to be positioned outside the chart, to improve the chart readability without cluttering with overlapping elements and colors.

We should allow rendering rectangular annotations outside the chart area, for example right below the axis.

Current:
Screenshot 2021-04-07 at 22 18 20

Desired:
Screenshot 2021-04-07 at 22 19 48

For the implementation we can consider multiple alternatives or go in steps:
1. allow the user to specify a `marker` for `RectAnnotation`s: the marker can be placed below the annotation in the center, like for a line annotation. A `size` parameter representing the width of the annotation in pixel, can be passed to the marker component to size it.
- Add a configuration to allow specifying a `RectAnnotation` outside the "linked" axis (below, above the horizontal axis, or at the right/ left of the vertical axis), but this can subsequentially be a worm hole of requests/options and configurations.
- consider the annotation as a composable chart

**Additional context**

https://codesandbox.io/s/elastic-charts-playground-qs1jr?file=/src/timeseries.tsx

### LINKED ISSUE #1205: RectAnnotation with groupId with chart rotation
**Describe the bug**
When using a `RectAnnotation` with `groupId` and a `chartRotation` of `90` or `-90`, where only the `x` dimensions are specified, creates an incorrect annotation. The annotation attempts to fallback to the height rather than the width in this given case. Given a rotation of `0` or `180` the height default is expected.

**To Reproduce**
1. Go to https://codesandbox.io/s/gracious-taussig-vetcd?file=/src/App.tsx
2. Resize the rendered chart
3. Notice the width of the chart maintains the height of the chart and not the width.

**Expected behaviour**
Annotation should default to full width when not specified

**Screenshots**

https://user-images.githubusercontent.com/19007109/122129498-c9240b00-cdfb-11eb-8cf9-339f016dd73d.mp4

https://user-images.githubusercontent.com/19007109/122129664-0be5e300-cdfc-11eb-8bad-81b807747196.mp4

 - Browser: chrome
 - Elastic Charts: `latest`

nickofthyme:
## Summary

The `outside` prop is now available in the `RectAnnotation` component to allow rendering annotations outside of the chart within the axis tick gutter.

image

## Details

Using `RectAnnotation` with `outside` prop regardless of `rotation` or `groupId` correctly renders the annotation within the `tickLine.size` gutter. 

```tsx
<RectAnnotation
  {...props}
  outside={true}
  outsideDimension={5}
/>
```

Limitations:
- Only renders outside for full-height or full height annotations such that only `x` coordinates or only `y` coordinates are specified by the user.
- Only renders with `tickLine` theme is visible and non-zero `strokeWidth`. This the outside annotation height/width is set by the `outsideDimension` prop which defaults to the `tickLine.size` for the respective axis.
- Outside annotations will **ALWAYS** render even when not suitable, such as when all axes are hidden. It is up the the user to configure the axis and annotation in a suitable way.

### Demo

https://user-images.githubusercontent.com/19007109/122256640-e0630700-ce94-11eb-875a-c81d9f1e6637.mp4

## Issues

closes #928 allowing outside rect annotations
fix #1205 causing issues with `groupId` when using the default height/width

### Checklist


packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts
```diff
@@ -76,14 +82,28 @@ export function computeRectAnnotationDimensions(
...
     }
+
     if (!yScale) {
       if (!isDefined(initialY0) && !isDefined(initialY1)) {
+        const isLeftSide =
+          (chartRotation === 0 && xAxis?.position === Position.Bottom) ||
+          (chartRotation === 180 && xAxis?.position === Position.Top) ||
+          (chartRotation === -90 && yAxis?.position === Position.Right) ||
+          (chartRotation === 90 && yAxis?.position === Position.Left);
+        const orthoDimension = chartRotation === 0 || chartRotation === 180 ? panelSize.height : panelSize.width;
```
markov00, line 93: You can use `isHorizontal` for that

packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts
```diff
@@ -76,14 +82,28 @@ export function computeRectAnnotationDimensions(
...
+        const outsideDim = getOutsideDimension(getAxisStyle(xAxis?.id ?? yAxis?.id), outside);
```
markov00, line 94: I will not force this behavior in code. If the users specify `outside` they need to know that this, currently, depends on the tickline size.

packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts
```diff
@@ -242,3 +274,11 @@ function getMin(min: number, value?: number | string | null) {
   }
   return value;
 }
+
+function getOutsideDimension(style: AxisStyle, outside?: boolean): number {
+  if (!outside) return 0;
+
+  const { visible, size, strokeWidth } = style.tickLine;
+
+  return visible && size > 0 && strokeWidth > 0 ? size : 0;
```
markov00, line 283: independently from my previous comment, I'm not even sure we need to bind the dimension to the strokeWidth. It shouldn't be an issue.

markov00: we should also think already if we want to connect the height of the annotation to a specific passed parameter instead of the `tickLine`. It can be easier to migrate instead of relying on the theme

packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts
nickofthyme: I'm not sure what you are asking here

packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts
nickofthyme: Sure so having its own separate dimension would be fine with me. But do you think we should limit the dimension to the `tickLine.size`? Otherwise it would render arbitrarily within the axis regardless of the space.

packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts
markov00: You have introduced an edge case behavior that is not aligned with the current configuration.
If the user specifies `outside` and then, for some reason, the axis style is configured such as no space is available outside the chart, we force the logic to render the axis inside the chart.
We should respect the user input and not introduce an edge case. If there isn't enough space, just don't render the annotation outside the chart, the editor library should take handle that case

packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts
markov00: `tickLine.size` is controlled by the user, and the same happens for such annotation size. So it's up to the user and the editor to make such case.

packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts
nickofthyme: See db22537

packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts
nickofthyme: I added a separate prop called `outsideDimension` to set the width/height of the outside dimension. I used the `tickLine.size` as the default dimension if none is provided. See 8886aeb

packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts
nickofthyme: Gotcha, so I refactored it to always render the annotation no matter if there is space or not. See 8886aeb

nickofthyme: https://user-images.githubusercontent.com/19007109/123119130-deef8c80-d408-11eb-8519-0693f495bcac.mp4

markov00 (approved):
It looks good to me, the with the latest changes we have a clear way to configure the dimension outside, independently from external factors

Issue #1187, opened 2021-06-03
[Bullet] Tooltips shouldn't appear on the entire canvas


ghudgins:
**Is your feature request related to a problem? Please describe.**
Distracting to have the tooltip appear when you hover over white space in the chart canvas

**Describe the solution you'd like**
Consider only letting the "Actual" value show when hovering on the bullet 

**Kibana Cross Issues**
related to this integration: https://github.com/elastic/kibana/issues/89859

**Checklist**

Delete any items that are not applicable to this feature request.

- [x] this request is checked against already exist requests
- [x] every related Kibana issue is listed under `Kibana Cross Issues` list
- [x] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

monfera:
Documenting some design considerations here:

- It's possible, and sometimes desirable to style a horizontal or vertical bullet such that it's thin (eg. for a clean look) eg. [this](https://elastic.github.io/elastic-charts/?path=/story/goal-alpha--horizontal-bar) or thinner
- Fitts law prompts us to have a decent sized mouse capture zone for the tooltip to activate; while in Lens config, the bullet chart may be in a huge area with hundreds of pixels of empty space around it, in a dashboard the entire thing may be a few tens of pixels horizontally or vertically; so it's useful to take the maximum of the computed width/height and some value at which the tooltip-on-background is not disturbing (something like 100px CSS)
- In case of circly gauge/goal (should they become requested by customers) it'd be esp. annoying to have to precisely hit a potentially thin circle (or tick label)
- due to tick placement, other annotations, ticks etc. the chart has no definite area

So I'm planning with something like this:

The width of the capture zone would be the largest of
- 100 CSS pixels (Fitts law driven)
- 50% of the chart width (so a decent fraction of the full chart area is tooltip sensitive)
- the width of the bounding box of anything that gets rendered

Same for height.

@markov00 @ghudgins  please bring up alternatives

markov00:
@monfera I think this is a sensible selection of params, so I'm ok with your suggestion.
One alternative could be just the bounding box of the bullet chart without the label

markov00:
or the background area of the quantitative scale independently from the size and weight of the bar

ghudgins:
[...]

PR #1218, opened 2021-06-23
feat(a11y): allow user to add optional semantic meaning to goal/gauge charts

### LINKED ISSUE #1161: Accessible goal chart (colors)
Goals have color-focused range definitions, but what's a good way to make those accessible?

Currently, the range definition might look something like: 
```
const colorMap: { [k: number]: Color } = {
  '-200': `rgb(${q1},${q1},${q1})`,
  '-250': `rgb(${q2},${q2},${q2})`,
  '-300': `rgb(${q3},${q3},${q3})`,
};
```

My idea is to extend that to include semantic information as well:
```
const colorMap: { [k: number]: [Color, string] } = {
  '-200': [`rgb(${q1},${q1},${q1})`, 'cold',
  '-250': `rgb(${q2},${q2},${q2})`, 'really cold'
  '-300': `rgb(${q3},${q3},${q3})`, 'freezing'
};
```

This way, the author is setting up one map and has everything in one place.

And then when rendering, this info can be added to the accessible graph summary.

rshen91:
For reviewers, please review https://github.com/elastic/elastic-charts/pull/1174 first!

## Summary

The `bandLabels` prop is now added to `Goal` charts. This is an optional prop. 

For example, users can pass band labels with semantic values associated with each band: 
`['chilly', 'brisk', 'warm']`.

## Details

The idea of creating a separate prop on the Goal component was to not interfere with the rendering logic that the `bands` prop handles. 

## Issues
Closes #1161 

### Checklist


packages/charts/src/components/accessibility/types.tsx
```diff
@@ -19,19 +19,33 @@
...
+    chartTypeDescription === 'verticalBullet chart';
   return (
     <dl>
       <dt>Chart type:</dt>
-      <dd id={props.defaultSummaryId}>{props.chartTypeDescription}</dd>
+      <dd id={defaultSummaryId}>{chartTypeDescription}</dd>
+      {validGoalChart && goalChartData && <dd>{`Minimum: ${goalChartData.minimum}`}</dd>}
+      {validGoalChart && goalChartData && <dd>{`Maximum: ${goalChartData.maximum}`}</dd>}
+      {validGoalChart && goalChartData && <dd>{`Target: ${goalChartData.target}`}</dd>}
+      {validGoalChart && goalChartData && <dd>{`Value: ${goalChartData.value}`}</dd>}
```
nickofthyme, line 48:
```suggestion
      {validGoalChart && goalChartData ? (
        <dd>{`Minimum: ${goalChartData.minimum}`}</dd>
        <dd>{`Maximum: ${goalChartData.maximum}`}</dd>
        <dd>{`Target: ${goalChartData.target}`}</dd>
        <dd>{`Value: ${goalChartData.value}`}</dd>
      ) : null}
```

rshen91: I love this suggestion - so much cleaner - moving it to https://github.com/elastic/elastic-charts/pull/1174 where this change is most relevant 👍🏻

rshen91: Sweet! much cleaner - 7c4e432 thanks!

nickofthyme: I guess you could also do `<dd>Minimum: {goalChartData.minimum}</dd>` but didn't see that earlier.

packages/charts/src/chart_types/goal_chart/renderer/canvas/connected_component.tsx
```diff
@@ -133,6 +135,7 @@ class Component extends React.Component<Props> {
           role="presentation"
         >
           <ScreenReaderSummary />
+          {semanticValues && semanticValues.length > 1 && <GoalSemanticDescription semanticValues={semanticValues} />}
```
myasonik, line 138: Will this be the same thing?

```suggestion
          {semanticValues?.length > 1 && <GoalSemanticDescription semanticValues={semanticValues} />}
```

myasonik: Actually, should we hide all of this logic in `GoalSemanticDescription`? (have that component do the conditional check so that we have a cleaner structure here)

packages/charts/src/components/accessibility/goal_semantic_description.tsx
```diff
@@ -0,0 +1,43 @@
...
+        const prevValue = semanticValues[index - 1];
+        return nextValue !== undefined ? (
+          <dd key={index}>{`values ${value} - ${nextValue[0]}: ${semantic}`}</dd>
+        ) : prevValue[0] < value ? (
+          <dd key={index}>{`values above ${value}: ${semantic}`}</dd>
+        ) : (
+          <dd key={index}>{`values below ${value}: ${semantic}`}</dd>
+        );
+      })}
+    </div>
```
myasonik, line 41: It looks like you're trying to build out a description list here, is that right?

So we need the whole thing wrapped in a `<dl></dl>` (can replace the `<div>` with this) and then the key/value pairs ("`value`: `semantic`" in this case) are broken up into separate `<dt>` and `<dd>` elements.

So bringing that all together, the (heavily simplified) final markup should be something like:

```
<dl>
	{semanticValues.map(([value, semantic], index) => {
		return <>
			<dt>{value}</dt>
			<dd>{semantic}</dd>
		</>;
	})}
</dl>
```

stories/goal/25_goal_semantic.tsx
```diff
@@ -0,0 +1,77 @@
...
+        ticks={[0, 50, 100, 150, 200, 250, 300]}
+        tickValueFormatter={({ value }: BandFillColorAccessorInput) => String(value)}
+        bandFillColor={({ value }: BandFillColorAccessorInput) => semanticBandFillColor(value)}
+        labelMajor="Revenue 2020 YTD  "
+        labelMinor="(thousand USD)  "
+        centralMajor="170"
+        centralMinor=""
+        config={{ angleStart: Math.PI, angleEnd: 0 }}
+        semanticValues={bandsWithSemantics}
+      />
```
myasonik, line 74: I wasn't sure where to leave this comment but I figured here is as good a place as any.

If a user is passing in semantic data, I don't think they should also have to send in the old `bands` prop. Here it looks like you're able to deduce it programmatically so it'd be nice to do that for the consumer. 

So I'm not sure if this would be best as a new prop like you have it, or whether to allow a different shape of data to be passed into `bands` so that you can pass in either type of data (semantic or not) into one prop. (I would love to see it all in one prop but that's just a personal preference.)

packages/charts/src/components/accessibility/goal_semantic_description.tsx
rshen91: thank you! 250ab2c

packages/charts/src/chart_types/goal_chart/renderer/canvas/connected_component.tsx
rshen91: thank you! 250ab2c

packages/charts/api/charts.api.md
```diff
@@ -801,6 +801,8 @@ export interface GoalSpec extends Spec {
     // (undocumented)
     labelMinor: string | BandFillColorAccessor;
     // (undocumented)
+    semanticValues?: Array<(number | string)[]>;
```
monfera, line 804: If my other comments are observed, there won't be a need for this data type, as `bandLabels` will be a flat `string[]`.

Still, worth mentioning that in this current version of the branch, the proper type isn't an array of `number | string`, for these reasons:
- Array<number | string> admits an array with zero, one, two or more values, whereas you want exactly two
- also, it's not true that the two elements can be any of numbers or strings, eg. `['hello', 5]` shouldn't be admitted

The tight type here would be `[number, string][]`, but again, this would be not needed, as a flat array is preferable

monfera (commented):
The nature of the change itself looks good to me, only asking for a trimming of the API, pls. see my comments

stories/goal/25_goal_semantic.tsx
```diff
@@ -0,0 +1,77 @@
...
+        bands={bandsWithSemantics.flat().filter((val) => typeof val === 'number')}
+        ticks={[0, 50, 100, 150, 200, 250, 300]}
+        tickValueFormatter={({ value }: BandFillColorAccessorInput) => String(value)}
+        bandFillColor={({ value }: BandFillColorAccessorInput) => semanticBandFillColor(value)}
+        labelMajor="Revenue 2020 YTD  "
+        labelMinor="(thousand USD)  "
+        centralMajor="170"
+        centralMinor=""
+        config={{ angleStart: Math.PI, angleEnd: 0 }}
+        semanticValues={bandsWithSemantics}
```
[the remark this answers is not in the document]
myasonik: Do you think two arrays is better than allowing a more complicated type for `bands`? 

`bands` type could be `number[]|SomethingElse` and then internally a type check would let you know what to do.

(I don't have the perfect answer for what `SomethingElse` could be but there are a lot of options. Rachel's tuple (`[number, string][]`) is pretty good but could also create an object of some sort or an array of objects.)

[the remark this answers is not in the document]
monfera: Union types (which we use here and there) have negatives, they need to be tested in runtime, and so far we haven't reached the rigor of normalizing data right behind the API into canonical, uniform structures, therefore union types cause coalescing ripples deep into the code. Also, API users are a few developers mostly in Kibana, so the pressure to make the API flexible for possibly light-coder end users doesn't apply. 

It is still possible to do it this way, but it's nice to avoid union types.

Also, we could just go with the `SomethingElse` eg. Rachel's structure and expect the developer who uses the API to fill in empty strings, if none applies. Rachel mentioned about attempting to not introduce a breaking change, and that has value too.

We've been discussing switching to column arrays in our API, which has a bunch of benefits; most of them (performance) don't apply here, because of low count, yet the use of column arrays is natural.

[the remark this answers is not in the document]
myasonik: Thanks for the explanation – souonds like all good reasoning to me.

rshen91:
The prop bandLabels isnt being picked up in b5a1364. I need to investigate why that's happening but also welcome any insights.

rshen91:
jenkins test this

packages/charts/api/charts.api.md
rshen91: Thank you!! 22ef1a6 please for changes

stories/goal/25_goal_semantic.tsx
```diff
@@ -0,0 +1,77 @@
...
+    <Chart className="story-chart">
+      <Goal
+        id="spec_1"
+        subtype={subtype}
+        base={0}
+        target={260}
+        actual={170}
+        // doesn't mess with canvas_renderers.ts
+        // @ts-ignore
+        bands={bandsWithSemantics.flat().filter((val) => typeof val === 'number')}
```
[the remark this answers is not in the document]
rshen91: Changed in 22ef1a6

monfera:
> The prop bandLabels isnt being picked up in [b5a1364](https://github.com/elastic/elastic-charts/commit/b5a13641dbcbd63d37b89f42c3e9c1749ec6ae79). I need to investigate why that's happening but also welcome any insights.

As it's not a review comment, we can't set it to "resolve", so doing it via a reply (we looked at it together)

stories/goal/25_goal_semantic.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+import { Chart, Goal } from '../../packages/charts/src';
+import { BandFillColorAccessorInput } from '../../packages/charts/src/chart_types/goal_chart/specs';
+import { GoalSubtype } from '../../packages/charts/src/chart_types/goal_chart/specs/constants';
+import { Color } from '../../packages/charts/src/utils/common';
+
+const subtype = GoalSubtype.Goal;
+
+export const Example = () => {
+  const bandsWithSemantics = ['freezing', 'chilly', 'brisk'];
```
monfera, line 30: Thanks for the changes! Nit: the variable could be renamed to `bandLabels`, thought it's not significant for using electricity for a new test run

monfera:
@rshen91 could you please add the purpose of the PR into the description, and the driver for that, eg. text in the description or link to an issue? Besides code aspects, the PR needs to be reviewed in light of the objectives which you are most familar with

monfera (commented):
Wondering if instead of 
```
<dt>200</dt><dd>freezing</dd>
<dt>250</dt><dd>chilly</dd>
<dt>300</dt><dd>brisk</dd>
```
it'd be better to use
```
<dt>0 - 200</dt><dd>freezing</dd>
<dt>200 - 250</dt><dd>chilly</dd>
<dt>250 - 300</dt><dd>brisk</dd>
```
as the bands/colors represent ranges. The "from" value is the previous band value, except for the lowest band value, where it's the lowest value in the `ticks` array. 

The visualization itself does not show band label texts, while the colors represent intervals, not single values. What do you think @rshen91 @markov00 @myasonik?

myasonik:
> it'd be better to use
> ```
> <dt>0 - 200</dt><dd>freezing</dd>
> <dt>200 - 250</dt><dd>chilly</dd>
> <dt>250 - 300</dt><dd>brisk</dd>
> ```
> as the bands/colors render as ranges. The "from" value is the previous band value, except for the lowest band value, where it's the lowest value in the ticks array.

++ I like it!

rshen91:
> Wondering if instead of
> 
> ```
> <dt>200</dt><dd>freezing</dd>
> <dt>250</dt><dd>chilly</dd>
> <dt>300</dt><dd>brisk</dd>
> ```
> 
> it'd be better to use
> 
> ```
> <dt>0 - 200</dt><dd>freezing</dd>
> <dt>200 - 250</dt><dd>chilly</dd>
> <dt>250 - 300</dt><dd>brisk</dd>
> ```
> 
> as the bands/colors render as ranges. The "from" value is the previous band value, except for the lowest band value, where it's the lowest value in the `ticks` array.
> 
> The visualization itself does not show band label texts, while the colors represent intervals, not single values. What do you think @rshen91 @markov00 @myasonik?

I also think this is fine! 0bfdec4 for changes - thanks!

myasonik:
In an effort to provide value to visual and screen reader users alike, I wonder if it makes sense to expose the semantic meaning in a tooltip over the ranges as well?

Color understanding isn't universal so having it explicitly defined for many users may be helpful.

rshen91:
> Thanks Rachel, looks good! I don't remember if there's a guarantee that the first value is the lowest, though in practice that's probably always the case (we could add it to the documentation for `ticks`). Or safer, just use
> 
> ```
> return geoms.bulletViewModel.ticks.reduce((min: number, {value}: TickViewModel) => Math.min(min, value), Infinity);
> ```

Actually I'm seeing that the bulletViewModel includes a lowestValue. Is this value good for this case? Thanks!

monfera:
> Actually I'm seeing that the bulletViewModel includes a lowestValue. Is this value good for this case? Thanks!

Yes @rshen91 it's better than my suggestions, good call!

monfera:
I agree @myasonik with band info in a visual channel. It'd need some design, eg. 
- Should the tooltip show all bands all the time, or only the currently hovered band? (in case of the latter, there's work needed toward capturing the specific band under the mouse; I'm working on a somewhat related PR which would be prerequisite for this)
- Usually, it's worth allowing display without requiring the user to perform interactions, esp. if it's compact and there is space for a legend
- Is the tooltip still useful if there's also legend support? I think yes, because a legend is not mandatory, and it also helps if some users have difficulty correlating colors (a hover gives a direct band name in the tooltip)

So, 
- as there are some choices, and 
- it has some blockers, yet this PR is a prerequisite of ongoing bullet tooltip feature work, due to big merge conflict from some refactoring (addressing https://github.com/elastic/elastic-charts/issues/1187), and
- there's no drawback to merging this PR as it is, compared to adding this task,

the tooltip with tbe band names could be a separate PR. Pls. share if it has some negatives

monfera (approved):
LGTM!

PR #1220, opened 2021-06-28
test: replace mocks with snapshots

markov00:
## Summary

We use mocks for checking the equality of very big objects.
The PR refactored that code using snapshot instead: they provide the same functionality with less burden.

## Details

This can simplify the refactoring in the following cases:
- a field is removed or slightly changed
- a completely new complex field is added. In particular, it is helpful when the new field is derived from others part of the data structure
- The mocks are themself code with some logic. They are not a guarantee to be bug-free.

If we are testing an object structure, we probably can benefit directly from snapshots instead of manually writing the object.
The only caveat is to threat snapshots as code, checking them during the review as they where written inline in the test


rshen91 (approved):
LGTM - This changing to snapshots makes sense to me especially in terms of the context of changing the data logic for the null values in tooltips PR. I think it should be manageable as a team for us to check any snapshot changes in a PR in future reviews (to make sure snapshots aren't being changed for bad reasons).

packages/charts/src/chart_types/xy_chart/rendering/rendering.bands.test.ts
```diff
@@ -70,84 +68,11 @@ describe('Rendering bands - areas', () => {
...
-            x: 25,
-            y: 0,
-          },
-        }),
-      );
+      expect(points).toHaveLength(4);
+      expect(points[0]).toMatchSnapshot();
+      expect(points[1]).toMatchSnapshot();
+      expect(points[2]).toMatchSnapshot();
+      expect(points[3]).toMatchSnapshot();
```
nickofthyme, line 75: Can't we simplify this? I don't see a reason for individual snapshots.

```suggestion
      expect(points).toMatchSnapshot();
```

nickofthyme (approved):
LGTM

packages/charts/src/chart_types/xy_chart/rendering/rendering.bands.test.ts
markov00: done in https://github.com/elastic/elastic-charts/pull/1220/commits/46aa1c830ba048346a9e4a4fb1e36d38512ef59f

PR #1206, opened 2021-06-16
feat: expose NaN values to tooltips

markov00:
### Checklist

- [ ] The proper chart type label was added (e.g. :xy, :partition) if the PR involves a specific chart type
- [ ] The proper feature label was added (e.g. :interactions, :axis) if the PR involves a specific chart feature
- [ ] Whenever possible, please check if the closing issue is connected to a running GH project
- [ ] Any consumer-facing exports were added to `packages/charts/src/index.ts` (and stories only import from `../src` except for test data & storybook)
- [ ] This was checked for cross-browser compatibility
- [ ] Proper documentation or storybook story was added for features that require explanation or tutorials
- [ ] Unit tests were updated or added to match the most common scenarios

PR #1226, opened 2021-06-29
feat(xy): add null bars to geometry indexing

### LINKED ISSUE #950: Add option to show tooltip for empty buckets
# Summary

Related to https://github.com/elastic/kibana/issues/85793

It would be beneficial to have an option in the charts to always tooltip for instances where multiple charts are synced, but an individual chart instance cannot show a value because of an empty bucket in the data. In this case we'd like to show a custom tooltip message like "Metric: N/A" or similar.

image

<details>
<summary>Example of current tooltip behaviour</summary>

Kapture 2020-12-15 at 10 14 27

</details>

markov00:
## Summary

The `showNullValues` prop is now available in the `Settings.tooltip` to allow displaying `null` Y values in the tooltip.
This should goes hand in hand with the right tooltip formatter to catch and correctly format these `null` values.

Jul-06-2021 12-37-30

## Details

This PR replace: https://github.com/elastic/elastic-charts/pull/1206 reducing the scope to only show `null` values, keeping the current typings intact. 

The PR consist of removing the null values checks in the geometry generation phase, moving every geometry (visible or not) within the geometry index map (the map used by the tooltip to detect the cursor-geometry collisions). The geometries with null values are not pushed within the arrays of renderable geometries.

## Issues

close #950

### Checklist


packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -81,12 +77,15 @@ export function renderBars(
       y0Scaled = y0 === null ? yScale.scale(0) : yScale.scale(y0);
     }
 
-    if (y === null || y0Scaled === null) {
-      return;
+    const absMinHeight = Math.abs(minBarHeight);
+
+    // safeguard against null y values
+    let height = isNil(y0Scaled) || isNil(y) ? 0 : y0Scaled - y;
```
monfera, line 83: Minor: this could be substituted with `height = y0Scaled - y || 0`; using `isNil` doesn't add a lot of documentational value as the types already reveal they can be undefined or null. Still, it's a subjective call.

Also, as the row below sets `y` to zero if either is nil (in row 86) maybe reordering helps simplify a bit (eg. setting y and y0Scaled to 0 before; in which case the subtraction will result zero)

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -125,7 +124,7 @@ export function renderBars(
     const width = clamp(seriesStyle.rect.widthPixel ?? xScale.bandwidth, minPixelWidth, maxPixelWidth);
     const x = xScaled + xScale.bandwidth * orderIndex + xScale.bandwidth / 2 - width / 2;
 
-    const originalY1Value = stackMode === StackMode.Percentage ? y1 - (y0 ?? 0) : initialY1;
+    const originalY1Value = stackMode === StackMode.Percentage ? (isNil(y1) ? null : y1 - (y0 ?? 0)) : initialY1;
```
monfera, line 127: May be tangential to this PR, but `GeometryValue` has `x` and `y` as `any`, and maybe it could be tightened up. If it's not possible to do it globally, it could start by perhaps mapping to `NaN` instead of `null` so that eventually, `y` in `GeometryValue` is always a number

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -195,7 +194,10 @@ export function renderBars(
       panel,
     };
     indexedGeometryMap.set(barGeometry);
-    barGeometries.push(barGeometry);
+
+    if (y1 !== null && initialY1 !== null && filled?.y1 === undefined) {
+      barGeometries.push(barGeometry);
+    }
```
monfera, line 200: Should `indexedGeometryMap.set(barGeometry)` still be called unconditionally? If that were conditionalized, this predicate could go to the top of the `forEach`.

Btw. this function is almost 200 lines long with a bunch of `let` and co-dependent mutations and non-obvious type narrowings to `number`, and I don't comprehend it enough to do a good review, would be great to refactor before, or as part of feature work otherwise it'll be more and more complicated

packages/charts/src/specs/settings.tsx
```diff
@@ -744,6 +751,23 @@ export function getTooltipType(settings: SettingsSpec, externalTooltip = false):
   return defaultType;
 }
 
+/**
+ * Return snowNullValues or the default
+ * @internal
+ */
+export function getShowNullValues(settings: SettingsSpec): TooltipProps['showNullValues'] {
+  const { tooltip } = settings;
```
nickofthyme, line 759: nit

```suggestion
export function getShowNullValues({ tooltip }: SettingsSpec): TooltipProps['showNullValues'] {
```

stories/interactions/18_null_values.tsx
```diff
@@ -0,0 +1,179 @@
...
+    if (ref2.current) {
+      ref2.current.dispatchExternalPointerEvent(event);
+    }
+  };
+  const { data } = KIBANA_METRICS.metrics.kibana_os_load[0];
+  const data1 = KIBANA_METRICS.metrics.kibana_os_load[0].data;
+  const data2 = KIBANA_METRICS.metrics.kibana_os_load[1].data;
+
+  const TopSeries = getSeriesKnob();
+  const BottomSeries = getSeriesKnob();
```
nickofthyme, line 81: 😄 😏

stories/interactions/interactions.stories.tsx
```diff
@@ -46,4 +46,5 @@ export { Example as brushDisabledOnOrdinalXAxis } from './13_brush_disabled_ordi
 export { Example as crosshairWithTimeAxis } from './14_crosshair_time';
 export { Example as renderChangeAction } from './15_render_change';
 export { Example as cursorUpdateAction } from './16_cursor_update_action';
+export { Example as interactionWithNullValues } from './18_null_values';
```
nickofthyme, line 49: Side note not related to this PR.

Do you know why we use this style rather than setting the name in a `<name>.story.tsx` file directly?

nickofthyme (approved):
Changes LGTM. This is a really clean solution! 🎉

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -81,12 +77,15 @@ export function renderBars(
...
+    // safeguard against null y values
+    let height = isNil(y0Scaled) || isNil(y) ? 0 : y0Scaled - y;
+
+    if (isNil(y0Scaled) || isNil(y)) {
+      y = 0;
     }
 
-    const absMinHeight = Math.abs(minBarHeight);
-    let height = y0Scaled - y;
     if (absMinHeight !== undefined && height !== 0 && Math.abs(height) < absMinHeight) {
```
monfera, line 89: `absMinHeight` is always a `number`, so this part of the condition can be removed.

Also, using a sign rather than retaining the `if(height < 0)` means that the `height !== 0` check is easy to make redundant:
```
    const heightDelta = absMinHeight - Math.abs(height);
    if (heightDelta > 0) {
      const heightSign = Math.sign(height);
      height = heightSign * absMinHeight;
      y -= heightSign * heightDelta;
    }
```

monfera, line 89: `absMinHeight` is always a `number`, so this part of the condition can be removed.

Also, using a sign rather than retaining the `if(height < 0)` means that the `height !== 0` check is easy to make redundant:
```
    const heightDelta = absMinHeight - Math.abs(height);
    if (heightDelta > 0) {
      const heightSign = Math.sign(height);
      height = heightSign * absMinHeight;
      y -= heightSign * heightDelta;
    }
```

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -81,12 +77,15 @@ export function renderBars(
       y0Scaled = y0 === null ? yScale.scale(0) : yScale.scale(y0);
```
monfera, line 77: This whole `let` / `if` imperative construct may be replaced with a single assignment:
```
    // prettier-ignore
    const y0Scaled = yScale.type === ScaleType.Log
      ? (y0 === 0 || y0 === null ? yScale.range[yScale.isInverted ? 1 : 0] : yScale.scale(y0))
      : (y0 === null ? yScale.scale(0) : yScale.scale(y0));
```
It's not super pretty, there may be ways to more deeply simplify though

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -57,10 +57,6 @@ export function renderBars(
 
   dataSeries.data.forEach((datum) => {
     const { y0, y1, initialY1, filled } = datum;
-    // don't create a bar if the initialY1 value is null.
-    if (y1 === null || initialY1 === null || (filled && filled.y1 !== undefined)) {
-      return;
-    }
     // don't create a bar if not within the xScale domain
     if (!xScale.isValueInDomain(datum.x)) {
       return;
```
monfera, line 62: Wondering if, probably outside this PR, something like this can be done with the y scale, ie. bail early if `yScale.scale(someY)` would now return `null`

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
monfera: Btw. if `y0 === 0` then does it not follow that  `yScale.scale(y0) === yScale.range[yScale.isInverted ? 1 : 0]` ?

monfera (approved):
LGTM, my comments are only about how the render function could be simplified or at least made easier to understand

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
markov00: Yes agree, we can push this into a separate PR I think

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
markov00: `indexedGeometryMap.set(barGeometry)` needs to be called unconditionally to take the advantage of the current picking logic.
Definitely a refactoring here is required!

stories/interactions/interactions.stories.tsx
markov00: I pretty sure there was a reason for that. Is probably related to the need of having subfolders and the need of having a single story with the subfolder title with a default export.
But probably things have now changed, not 100% sure

markov00:
> LGTM, my comments are only about how the render function could be simplified or at least made easier to understand

Thanks @monfera I will consider the simplification in a subsequent PR

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
markov00: I can't replace it with `height = y0Scaled - y || 0` because y0Scaled and y can be null and TS rise a warning/error here

PR #1227, opened 2021-06-29
feat(legend): add point shape styles to legend item

### LINKED ISSUE #1081: [Legends] Align the legend item to the series shape
# Summary

When using the Bubble chart you're able to change the shape for each series, but the legend item does not change from the dot.

Would the ideal to align the legend item dot to the shape.

Screenshot 2021-03-16 at 09 13 08

Here's an example of an already implemented chart in APM where it would be beneficial to have the legends be differentiable.

Screenshot 2021-06-23 at 11 25 59

rshen91:
## Summary
By enabling `alignLegendPointStyles` as false, legend item icons now show the appropriate shape of the markers for line, bubble, and area series. For instance, if a user specifies a `plus` shape marker on a line series and has a legend, the legend item will have the plus icon. 

image

## Issues
Closes #1081

### Checklist


packages/charts/src/chart_types/xy_chart/legend/legend.ts
```diff
@@ -139,6 +147,12 @@ export function computeLegend(
     const lastValue = lastValues.get(key);
     const seriesIdentifier = getSeriesIdentifierFromDataSeries(series);
     const xScaleType = getXScaleTypeFromSpec(spec.xScaleType);
+
+    let shape = 'dot' as IconType;
+    if (spec.seriesType === 'bubble') {
+      const { bubbleSeriesStyle } = spec as BubbleSeriesSpec;
+      shape = bubbleSeriesStyle?.point?.shape ?? 'dot';
+    }
```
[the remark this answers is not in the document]
monfera: An explanation for the code change, part 2 (types). The `spec` may or may not have a `bubbleSeriesStyle` in part because the `spec` may or may not be a bubble series spec (we wouldn't need the `if` if we knew in advance). The original snippet solves it with a _type assertion_ (`as BubbleSeriesSpec`), which removes typing power, therefore a bit brittle (iow we try to avoid `as`, except in occasional literal assignments, as TS is kinda dumb). While this assertion is simple and trivial now, a future code change may violate this, and it wouldn't be caught, because `as` basically tells TypeScript, "ignore this type check". (So it's not an assertion like in other languages, where it _increases_ rather than _undercuts_ safety).

The current solution for these is to use a _type guard_. In this case, as the predicate (`seriesType === 'bubble')` is, and can now only be, a runtime predicate (meaning, it's not possible to know if it's true or false in code editing time), it must be a function. This is always the case if the subtyping depends on some property value. Luckily there is a function for this already: `isBubbleSeriesSpec`. These are called _type predicates_. Upon code reading, we can recognize type predicatie functions after the `someValue **is** SomeType` instead of the usual `boolean`, ie. the `is`. Worth reading up on type predicates, and type guards in general [in the doc](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates).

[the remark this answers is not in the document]
monfera: Here, the specific benefit of using a type predicate function is that you get both the correct value (true/false) even if, in the future, the logic of determining how you identify a bubble series changes; and TypeScript grokking that in the `true` branch of the ternary, the spec will be of type `BubbleSeriesSpec`. It's good to not repeat these in the code, so `spec.seriesType === 'bubble'`, and similar, are ideally never found in the code, except in the single source of truth that `function isBubbleSeriesSpec...` in `packages/charts/src/chart_types/xy_chart/utils/specs.ts` is.

packages/charts/src/components/icons/icon.tsx
```diff
@@ -37,13 +42,19 @@ const typeToIconMap = {
...
+  square: SquareIcon,
+  triangle: TriangleIcon,
 };
 
 /** @internal */
 export type IconColor = string;
 
 /** @internal */
-export type IconType = keyof typeof typeToIconMap;
+export type IconType = keyof typeof typeToIconMap & string;
```
monfera, line 57: I think the `& string` is unnecessary.  

The type intersection operator `&` intersect the type on the left and the type on the right, ie. both the left and the right type must simultaneously apply. `IconType` needs to be both a key of `typeToIconMap` and a `string`. 

But `keyof typeof typeToIconMap` is only ever a `string`, because  each key of the `typeToIconMap` object is a string. JavaScript objects can _have_ key types of `string` or `symbol` only, but this object has no `symbol`s. (It's possible to _specify_ other object keys eg. numbers, but they'll be coerced into `string` first)

monfera (commented):
I haven't runtime tested; please consider some code changes, see the comments

packages/charts/src/chart_types/xy_chart/legend/legend.ts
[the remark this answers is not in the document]
monfera: (my local TS is acting up so if you see an issue, might worth checking with Nick or Marco until my setup gets fixed; hope the above code has no red squigglies)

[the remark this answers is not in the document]
rshen91: Hey thanks for the review - In a68797c I'm actually trying to account for any pointstyles in area and lines too. I'll keep thinking about the ternary format but I'm not sure if it will get too convoluted with these other two types also included?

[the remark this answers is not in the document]
monfera: Whether it's `if` or ternary is secondary (haha). You can still avoid tests like `spec.seriesType === 'line'` in favor of calling the types of functions shown above, eg. `isBubbleSeriesSpec` and similar (reasons above).

Also, https://github.com/elastic/elastic-charts/commit/a68797c61aca5f7e074e7283263ca47d11596cbc does two things within each type branch:
1. recovering the style object (linked with the `spec.xxxxSeriesStyle` property)
2. assignment is the same for each: `shape = seriesStyle?.point?.shape ?? 'dot'`
Maybe it's nicer to separate the two parts, so that you make a function outside the already long `getLegendExtra` function, which returns a seriesStyle object, and then you just take the `seriesStyle?.point?.shape ?? 'dot'` value.

Something like this (haven't tested and my TS is in shambles):

```
function getSeriesStyle(spec): BubbleSeriesStyle | undefined {
      return isBubbleSeriesSpec(spec)
        ? spec.bubbleSeriesStyle
        : isLineSeriesSpec(spec)
        ? spec.lineSeriesStyle
        : isAreaSeriesSpec(spec)
        ? spec.areaSeriesStyle
        : undefined;
}

...

    const shape: IconType = getSeriesStyle(spec)?.point?.shape ?? 'dot';
```

I used `BubbleSeriesStyle` in the function return type above, because that has the `point` and that's all we need, though you can DRY up the types if needed, and use some newly created interface `PointSeriesStyle` which has `point`:
```
/** @public */
export interface PointSeriesStyle {
  point: PointStyle;
}

/** @public */
export type BubbleSeriesStyle = PointSeriesStyle;

/** @public */
export interface LineSeriesStyle extends PointSeriesStyle {
  line: LineStyle;
}

/** @public */
export interface AreaSeriesStyle extends PointSeriesStyle {
  area: AreaStyle;
  line: LineStyle;
}
```

@markov00 @nickofthyme do we have objects which have more than one `xxxSeriesStyle`? Can't we just rename these props to `seriesStyle` for uniformity across the series types? Looks like, at least across bubble, line and area, it'd be enough, and there'd be no need for such weird getters.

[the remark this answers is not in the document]
nickofthyme: hey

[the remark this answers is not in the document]
nickofthyme: > do we have objects which have more than one xxxSeriesStyle

No @monfera, I agree with you, I think the naming of the `xxxSeriesStyle` is a little unorthodox and redundant to say something like...

```tsx
<LineSeries lineSeiesStyles={{}} />
```

That being said all of these `xxxSeriesStyle` props are overrides for the matching name in the chart `Theme` object so `LineSeries.lineSeriesStyle` overrides `Theme.lineSeriesStyle` so there is nice consistency in naming. Also, these `xxxSeriesStyle` props are `Partials` of their respective `Theme` styles so the types also have a consistency. I don't think this change is warranted in this PR, but happy to discuss in a separate issue.

Merging the various types to extend `PointSeriesStyle` is a good improvement to enforce consistency, but the `getSeriesStyle` function you show above would work either way without using `as`.

As for the type guards you mentioned, these are only necessary because we `extend` the `BasicSeriesSpec` to form the bubble, line and area specs and use the `BasicSeriesSpec` as a shorthand type rather than a union of `BubbleSeriesSpec|AreaSeriesSpec|LineSeriesSpec|...`. We could discuss changing this to improve performance but this is a common thing across charts.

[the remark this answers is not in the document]
nickofthyme: Also we need to provide a fallback from the `Theme` for the respective point styles so extracting the style logic into separate function might be a little redundant.

```ts
function getSeriesStyle(spec): PointSeriesStyle | undefined {
  return isBubbleSeriesSpec(spec)
    ? spec.bubbleSeriesStyle
    : isLineSeriesSpec(spec)
    ? spec.lineSeriesStyle
    : isAreaSeriesSpec(spec)
    ? spec.areaSeriesStyle
    : undefined;
}

function getDefaultSeriesStyle(spec, theme): PointSeriesStyle | undefined {
  return isBubbleSeriesSpec(spec)
    ? theme.bubbleSeriesStyle
    : isLineSeriesSpec(spec)
    ? theme.lineSeriesStyle
    : isAreaSeriesSpec(spec)
    ? theme.areaSeriesStyle
    : undefined;
}

const pointStyles = (getSeriesStyle(spec) ?? getDefaultSeriesStyle(spec, theme)).point;
```

rather than just using the one round...

```ts
function getPointSeriesStyle(spec, theme): PointStyle | undefined {
  return isBubbleSeriesSpec(spec)
    ? spec.bubbleSeriesStyle?.point ?? theme.bubbleSeriesStyle.point
    : isLineSeriesSpec(spec)
    ? spec.lineSeriesStyle?.point ?? theme.lineSeriesStyle.point
    : isAreaSeriesSpec(spec)
    ? spec.areaSeriesStyle?.point ?? theme.areaSeriesStyle.point
    : undefined;
}

const pointStyles = getPointSeriesStyle(spec, theme);
```

markov00 (commented):
I took a look at the point shapes example and looks like that square doesn't support the fill property as specified in the chart. We should fully align the shapes styles to the one in the chart. If the square has a red border and a yellow fill, we should reflect this in the legend

rshen91:
> I took a look at the point shapes example and looks like that square doesn't support the fill property as specified in the chart. We should fully align the shapes styles to the one in the chart. If the square has a red border and a yellow fill, we should reflect this in the legend

This makes a lot of sense to me - I have removed the fill for the square icon in this commit 37cf9c1

packages/charts/src/components/icons/icon.tsx
```diff
@@ -37,13 +42,19 @@ const typeToIconMap = {
   eyeClosed: EyeClosedIcon,
   list: ListIcon,
   questionInCircle: QuestionInCircle,
+  circle: DotIcon,
+  x: XIcon,
+  plus: PlusIcon,
+  diamond: DiamondIcon,
+  square: SquareIcon,
+  triangle: TriangleIcon,
```
nickofthyme, line 50: Is there a reason you are not using the `PointShapes` here?

I think this map here should only be for actual icons not shapes.

We have a map called `ShapeRendererFn` of all `PointShapes` used for rendering the points on the chart and for textures. I think we should reuse these to be consistent and keep the code DRY.

https://github.com/elastic/elastic-charts/blob/87629e23e7cc8aef125a8b6d2b87092085abe2bd/packages/charts/src/chart_types/xy_chart/renderer/shapes_paths.ts#L28-L63

Which might be a little harder to use here than just a simple `svg` cuz you need to account for the rotation. Here is a usage example.

https://github.com/elastic/elastic-charts/blob/87629e23e7cc8aef125a8b6d2b87092085abe2bd/packages/charts/src/chart_types/xy_chart/renderer/dom/highlighter.tsx#L53-L61

So I think you should say does it have a shape then render a the pointShape, otherwise render the `DotIcon`, thus they are treated different.

packages/charts/src/components/icons/icon.tsx
```diff
@@ -37,13 +42,19 @@ const typeToIconMap = {
...
 };
 
 /** @internal */
 export type IconColor = string;
 
 /** @internal */
-export type IconType = keyof typeof typeToIconMap;
+export type IconType = keyof typeof typeToIconMap | undefined;
```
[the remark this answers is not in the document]
rshen91: I get the following error in the getPointShape() function (```return spec.bubbleSeriesStyle?.point?.shape ?? theme.bubbleSeriesStyle.point.shape;``` and for the line and area series return statements too)(typescript issue) when I make this not undefined. But I don't think it matters because we are returning the PointShape.Circle if there is no point.shape or theme.point shape: 
```Type 'PointShape | undefined' is not assignable to type '"alert" | "dot" | "empty" | "eye" | "eyeClosed" | "list" | "questionInCircle" | "circle" | "x" | "plus" | "diamond" | "square" | "triangle"'.
  Type 'undefined' is not assignable to type '"alert" | "dot" | "empty" | "eye" | "eyeClosed" | "list" | "questionInCircle" | "circle" | "x" | "plus" | "diamond" | "square" | "triangle"'.
```

Would it make sense to get rid of the type | undefined as suggested and then use @ts-ignore or something in the getPointShape() function? Thanks!

packages/charts/src/common/legend.ts
```diff
@@ -47,6 +48,7 @@ export type LegendItem = {
   // TODO: Remove when partition layers are toggleable
   isToggleable?: boolean;
   keys: Array<string | number>;
+  shape?: IconType;
```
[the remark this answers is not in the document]
nickofthyme: Yeah this maybe optional now once you separate the shapes from the icons. So when this is undefined use the `DotIcon`.

packages/charts/src/chart_types/xy_chart/state/selectors/compute_legend.ts
```diff
@@ -63,6 +63,7 @@ export const computeLegendSelector = createCustomCachedSelector(
       settings.showLegendExtra,
       siDataSeriesMap,
       deselectedDataSeries,
+      chartTheme,
       // @ts-ignore
       settings.sortSeriesBy,
```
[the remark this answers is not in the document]
rshen91: Thank you! Commit 957d817 for changes

packages/charts/src/chart_types/xy_chart/renderer/dom/highlighter.tsx
```diff
@@ -50,7 +50,8 @@ function getTransformForPanel(panel: Dimensions, rotation: Rotation, { left, top
   return `translate(${left + panel.left + x}, ${top + panel.top + y}) rotate(${rotation})`;
 }
 
-function renderPath(geom: PointGeometry) {
+/** @internal */
+export function renderPath(geom: PointGeometry) {
```
markov00, line 54: it doesn't look like this is used externally, we can remove the export

packages/charts/src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -62,6 +62,7 @@ export const SeriesType = Object.freeze({
   Bar: 'bar' as const,
   Line: 'line' as const,
   Bubble: 'bubble' as const,
+  NonXY: 'non-xy' as const,
```
markov00, line 65: I don't see where this change is used, actually. This type is also very specific to XY charts so it doesn't have sense to have a non XY here

packages/charts/src/components/legend/color.tsx
```diff
@@ -18,24 +18,43 @@
...
+  pointStyle?: PointStyle;
   onClick?: MouseEventHandler;
 }
+const MARKER_SIZE = 16;
+const getCustomization = (pointStyle: PointStyle, seriesType: SeriesType): boolean | undefined =>
+  // bubble charts will always have the circle icon vs dot
+  seriesType === 'bubble' ||
+  (pointStyle.shape !== PointShape.Circle && seriesType === 'area') ||
+  (pointStyle.shape !== PointShape.Circle && seriesType === 'line') ||
+  pointStyle.fill !== 'blue'
```
markov00, line 43: I don't get the reason why checking against a color like the blue.
Can you explain a bit better what this `custominzation` represent?
is not clear what we are checking: a color, a specific radius (bigger or smaller radius then 2 is ok?)

packages/charts/src/components/legend/color.tsx
```diff
@@ -45,6 +64,30 @@ export const Color = memo(
...
+          <svg height={MARKER_SIZE} width={MARKER_SIZE}>
+            <g
+              transform={`
+                translate(${MARKER_SIZE / 2}, ${MARKER_SIZE / 2})
+                rotate(${rotation})`}
+            >
+              <path
+                d={shapeFn(
+                  shape === 'triangle' || shape === 'plus' || shape === 'x' ? adjustedSize / 2 : adjustedSize / 3,
+                )}
```
markov00, line 80: I think we should clean this up a bit more: checking the VRT screenshots you can see that the difference between the original dot and the markers are really big, we should align the size of the geometry to the one of the original dot.
Screenshot 2021-07-12 at 16 38 09

Screenshot 2021-07-12 at 16 37 59

packages/charts/src/components/legend/color.tsx
```diff
@@ -61,7 +104,12 @@ export const Color = memo(
 
       return (
         <div className="echLegendItem__color" title="series color">
-          <Icon type="dot" color={color} aria-label={`series color: ${color}`} />
+          {pointStyle && getCustomization(pointStyle, seriesType) ? (
+            // @ts-ignore
```
markov00, line 108: can we try to advoid ignore here, what was the reason?

packages/charts/src/components/legend/color.tsx
markov00: You are returning a boolean (better if we avoid completely the `undefined`), so it's more common to name a function like that as `isCustomized` or `isCustom`

packages/charts/src/components/legend/color.tsx
```diff
@@ -61,7 +104,12 @@ export const Color = memo(
...
+            renderShape(pointStyle)
+          ) : (
+            <Icon type="dot" color={color} aria-label={`series color: ${color}`} />
```
markov00, line 111: can we instead of this, always render through `renderShape` and render a filled dot if the pointstyle is not specified (the default case)?
We can get rid of a condition in the code and probably we can get rid of the `getCustomization` function

packages/charts/src/components/legend/legend_item.tsx
```diff
@@ -207,6 +209,8 @@ export class LegendListItem extends Component<LegendItemProps, LegendItemState>
             isSeriesHidden={isSeriesHidden}
             hasColorPicker={hasColorPicker}
             onClick={this.handleColorClick(hasColorPicker)}
+            pointStyle={pointStyle}
+            seriesType={seriesType ?? 'non-xy'}
```
markov00, line 213: if that was the reason for adding the `SeriesType.NonXY` you can mark the `seriesType` as optional, or just use the pointStyle to identify whenever use a specific share or the default dot.

packages/charts/src/components/legend/legend_icon.tsx
```diff
@@ -0,0 +1,47 @@
...
+interface LegendIconProps {
+  pointStyle?: PointStyle;
+  color: Color;
+}
+
+const MARKER_SIZE = 16;
+/** @internal */
+export const LegendIcon = ({ pointStyle, color }: LegendIconProps) => {
+  if (!pointStyle || !pointStyle.shape) {
+    return <Icon type="dot" color={color} aria-label={`series color: ${color}`} />;
```
nickofthyme, line 25: Do you know why the dot size got bigger? I don't see from these changes where that happened.

Image 2021-07-13 at 10 58 31 AM

I think we should keep the old size for now unless the shape is specified because in that case a bigger icon is better to see the details of the marker.

packages/charts/src/components/legend/legend_icon.tsx
```diff
@@ -0,0 +1,47 @@
...
+  }
+  const { shape, fill, stroke, strokeWidth, opacity } = pointStyle;
+  const [shapeFn, rotation] = ShapeRendererFn[shape];
+  const adjustedSize = MARKER_SIZE - strokeWidth;
```
nickofthyme, line 29: What happened to the `strokeWidth` ratio and conditional adjusted size we talked about?

nickofthyme (changes requested):

packages/charts/src/components/legend/legend_icon.tsx
rshen91: The larger size is when the shapeRendererFn is called for Circle, but the dot.tsx asset is the smaller icon (the default pre this PR). I think the refactoring in b2df204 is to change the default to the shapeRendererFn?

packages/charts/src/components/legend/legend_icon.tsx
rshen91: I think 2bc10ec has the changes thanks!

packages/charts/src/chart_types/xy_chart/renderer/dom/highlighter.tsx
rshen91: refactored in 4197004 thanks!

packages/charts/src/chart_types/xy_chart/utils/specs.ts
rshen91: Good call - major refactoring in 4197004

packages/charts/src/components/legend/color.tsx
rshen91: Refactored! 4197004

packages/charts/src/components/legend/legend_item.tsx
rshen91: Good call - 4197004

packages/charts/src/chart_types/xy_chart/legend/legend.ts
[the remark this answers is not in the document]
nickofthyme: This code has changed to reflect comments, so just gonna resolve this one.

packages/charts/src/chart_types/heatmap/state/selectors/scenegraph.ts
```diff
@@ -37,7 +37,7 @@ export function render(
   const { width, height } = chartDimensions;
   const { config: specConfig } = spec;
   const partialConfig: RecursivePartial<Config> = { ...specConfig, width, height };
-  const config = mergePartial<Config>(defaultConfig, partialConfig);
+  const config = mergePartial<Config>(defaultConfig, partialConfig, { mergeOptionalPartialValues: true });
```
nickofthyme, line 40: Thanks 👍🏼 

@markov00 We have noticed that in several places where `mergePartial` is use we have have issues where there are optional parameters on the base object that when `mergeOptionalPartialValues` is `false` will not look for such values on the partial object.

What do you think we make this option default to `true`?

packages/charts/src/chart_types/xy_chart/utils/texture.ts
```diff
@@ -18,19 +18,19 @@ const getSpacing = ({ spacing }: TexturedStyles): Point => ({
   y: typeof spacing === 'number' ? spacing : spacing?.y ?? 0,
 });
 
-const getPath = (textureStyle: TexturedStyles, size: number, stokeWith: number): [path: Path2D, rotation: number] => {
+const getPath = (textureStyle: TexturedStyles, size: number, strokeWidth: number): [path: Path2D, rotation: number] => {
```
nickofthyme, line 21: 😅

packages/charts/src/components/legend/legend_icon.tsx
```diff
@@ -9,36 +9,43 @@
...
-  const adjustedSize = MARKER_SIZE - strokeWidth;
+  const { shape, fill, stroke, strokeWidth = 1, opacity, radius } = pointStyle ?? { radius: 4 };
+
+  const adjustedFill = shape ? fill ?? color : color;
+  const adjustedStrokeWidth = getRatio(radius, strokeWidth);
+  const [shapeFn, rotation] = ShapeRendererFn[shape ?? PointShape.Circle];
+  const adjustedSize = MARKER_SIZE - adjustedStrokeWidth;
   return (
-    <svg height={MARKER_SIZE} width={MARKER_SIZE} aria-label={`series color: ${color}`}>
+    <svg width={MARKER_SIZE * 2} height={MARKER_SIZE * 2} viewBox="0 0 16 16" aria-label={`series color: ${color}`}>
```
nickofthyme, line 38: is `viewBox` necessary here?

packages/charts/src/components/legend/legend_icon.tsx
```diff
@@ -9,36 +9,43 @@
...
 }
 
-const MARKER_SIZE = 16;
+const MARKER_SIZE = 8;
+
+// to limit size, set a min and max
+const getRatio = (radius: number | undefined, strokeWidth: number) => {
+  const adjustedRadius = isNil(radius) ? strokeWidth : ((MARKER_SIZE / 2) * strokeWidth) / (radius * 2);
+  if (1.5 > adjustedRadius) return 1.5;
+  return Math.min(adjustedRadius, (radius ?? strokeWidth) * 2);
```
nickofthyme, line 26: I think we could just set a static min of `1px` and max of the `MARKER_SIZE / 2`

```suggestion
  const adjustedRadius = isNil(radius) ? strokeWidth : ((MARKER_SIZE / 2) * strokeWidth) / (radius * 2);

  return Math.max(Math.min(adjustedRadius, MARKER_SIZE / 2), 1);
```

packages/charts/src/components/legend/legend_icon.tsx
```diff
@@ -9,36 +9,43 @@
...
 interface LegendIconProps {
   pointStyle?: PointStyle;
   color: Color;
 }
 
-const MARKER_SIZE = 16;
+const MARKER_SIZE = 8;
+
+// to limit size, set a min and max
+const getRatio = (radius: number | undefined, strokeWidth: number) => {
```
nickofthyme, line 23: This should be `getAdjustedRadius` no?

```suggestion
const getAdjustedRadius = (radius: number | undefined, strokeWidth: number) => {
```

packages/charts/src/components/legend/legend_icon.tsx
```diff
@@ -9,36 +9,43 @@
...
 
 interface LegendIconProps {
   pointStyle?: PointStyle;
   color: Color;
 }
 
-const MARKER_SIZE = 16;
+const MARKER_SIZE = 8;
+
+// to limit size, set a min and max
```
nickofthyme, line 22: Let's make this a tsdocs comment

```suggestion
/** to limit size, set a min and max */
```

packages/charts/src/components/legend/legend_icon.tsx
```diff
@@ -9,36 +9,43 @@
...
 export const LegendIcon = ({ pointStyle, color }: LegendIconProps) => {
-  if (!pointStyle || !pointStyle.shape) {
-    return <Icon type="dot" color={color} aria-label={`series color: ${color}`} />;
-  }
-  const { shape, fill, stroke, strokeWidth, opacity } = pointStyle;
-  const [shapeFn, rotation] = ShapeRendererFn[shape];
-  const adjustedSize = MARKER_SIZE - strokeWidth;
+  const { shape, fill, stroke, strokeWidth = 1, opacity, radius } = pointStyle ?? { radius: 4 };
+
+  const adjustedFill = shape ? fill ?? color : color;
```
nickofthyme, line 33:
```suggestion
  const { shape, fill = color, stroke, strokeWidth = 1, opacity, radius } = pointStyle?.shape ? 
    pointStyle ?? { } : {
      fill: color,
    };
```

packages/charts/src/components/legend/legend_icon.tsx
```diff
@@ -9,36 +9,43 @@
...
+    <svg width={MARKER_SIZE * 2} height={MARKER_SIZE * 2} viewBox="0 0 16 16" aria-label={`series color: ${color}`}>
       <g
         transform={`
-            translate(${MARKER_SIZE / 2}, ${MARKER_SIZE / 2})
+            translate(${MARKER_SIZE}, ${MARKER_SIZE})
             rotate(${rotation})`}
       >
         <path
-          d={shapeFn(shape === 'triangle' || shape === 'plus' || shape === 'x' ? adjustedSize / 2 : adjustedSize / 3)}
+          d={pointStyle?.shape === PointShape.Diamond ? shapeFn(adjustedSize / 3) : shapeFn(adjustedSize / 2)}
```
nickofthyme, line 45: What is this for? We shouldn't divide by `3` cuz that's just arbitrary. Some shapes may appear bigger than others but they are all still the same size.

```suggestion
          d={shapeFn(adjustedSize / 2)}
```

packages/charts/src/components/legend/legend_icon.tsx
```diff
@@ -9,36 +9,43 @@
...
           stroke={stroke ?? color}
```
nickofthyme, line 46: This should be assigned only when `shape` is defined

```suggestion
          stroke={shape ? stroke ?? color : undefined}
```

packages/charts/src/components/legend/legend_icon.tsx
```diff
@@ -9,36 +9,43 @@
...
-          strokeWidth={strokeWidth}
-          fill={color || fill}
+          strokeWidth={adjustedStrokeWidth}
+          fill={adjustedFill}
```
nickofthyme, line 48: Now applied in destructuring

```suggestion
          fill={fill}
```

nickofthyme (commented):
Looking good. Just a few comments to cleanup the color icon logic.

packages/charts/src/components/legend/legend_icon.tsx
rshen91: The diamonds only have cutoff but the other shapes seem fine with `/2`

packages/charts/src/components/legend/color.tsx
nickofthyme: done, not sure what commit though

packages/charts/src/components/legend/color.tsx
nickofthyme: now match `8` previous size for all shaped. See latest changes.

packages/charts/src/components/legend/legend_icon.tsx
nickofthyme: can you show me a pic of what you mean?

packages/charts/src/components/legend/legend_icon.tsx
rshen91: 80c4cea thank you!

packages/charts/src/components/legend/legend_icon.tsx
rshen91: 80c4cea thanks!

packages/charts/src/components/legend/legend_icon.tsx
rshen91: Sweet! I think I am following - I did something similar in 80c4cea LMK what you think

packages/charts/src/components/legend/legend_icon.tsx
rshen91: 80c4cea thanks!

packages/charts/src/components/legend/legend_icon.tsx
rshen91: 80c4cea thank you!!!

packages/charts/src/components/legend/legend_icon.tsx
rshen91: Woah! Cool 80c4cea

packages/charts/src/components/legend/legend_icon.tsx
rshen91: No it is not!  80c4cea

rshen91:
jenkins test this please

packages/charts/src/chart_types/xy_chart/legend/legend.ts
```diff
@@ -139,6 +150,15 @@ export function computeLegend(
...
     const xScaleType = getXScaleTypeFromSpec(spec.xScaleType);
+
+    let shape = PointShape.Circle as IconType;
+    if (isBubbleSeriesSpec(spec)) {
+      shape = spec.bubbleSeriesStyle?.point?.shape ?? theme.bubbleSeriesStyle.point.shape;
+    } else if (isLineSeriesSpec(spec)) {
+      shape = spec.lineSeriesStyle?.point?.shape ?? theme.lineSeriesStyle.point.shape;
+    } else if (isAreaSeriesSpec(spec)) {
+      shape = spec.areaSeriesStyle?.point?.shape ?? theme.areaSeriesStyle.point.shape;
+    }
```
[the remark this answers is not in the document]
monfera: I agree it'd be nicer in a separate function, with a slight twist. Instead of a serial set of boolean predicates, how about something like
```
switch(spec.seriesType) {
    case SeriesType.Bubble: 
        return spec.bubbleSeriesStyle?.point?.shape ?? theme.bubbleSeriesStyle.point.shape;
    case SeriesType.Line: 
        return spec.lineSeriesStyle?.point?.shape ?? theme.lineSeriesStyle.point.shape;
    case SeriesType.Area: 
        return spec.areaSeriesStyle?.point?.shape ?? theme.areaSeriesStyle.point.shape;
    default: 
        return PointShape.Circle;
}
```

OK this wouldn't work as these tests don't act as type guards, while the `is...` functions do.

What do y'all think about the `switch(true)` pattern? Something like
```
switch(true) {
    case isBubbleSeriesSpec(spec): 
        return spec.bubbleSeriesStyle?.point?.shape ?? theme.bubbleSeriesStyle.point.shape;
    case isLineSeriesSpec(spec): 
        return spec.lineSeriesStyle?.point?.shape ?? theme.lineSeriesStyle.point.shape;
    case isAreaSeriesSpec(spec): 
        return spec.areaSeriesStyle?.point?.shape ?? theme.areaSeriesStyle.point.shape;
    default: 
        return PointShape.Circle;
}
```
It has a minimal learning curve and especially with TypeScript's need for boolean type guards, it's a closer abstraction than an `if` chain. It is also perfectly legit JS, no worse than other `switch` statements. Probably more controversial in webapp code than in library code eg. our use case. Here's [more](https://medium.com/trabe/using-switch-true-in-javascript-986e8ad8ae4f) on `switch(true)`.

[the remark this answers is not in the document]
monfera: A related term [yoda condition](https://eslint.org/docs/rules/yoda) is.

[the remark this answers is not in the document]
nickofthyme: wow 🤩 🤯

packages/charts/src/components/legend/legend_icon.tsx
nickofthyme: Ugh sorry I'm dumb this should be `getAdjustedStrokeWidth`

packages/charts/src/components/legend/legend_icon.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return Math.max(Math.min(adjustedRadius, (radius ?? strokeWidth) * 2), 1);
+};
+
+/** @internal */
+export const LegendIcon = ({ pointStyle, color }: LegendIconProps) => {
+  const { radius = 4, fill, strokeWidth = 1, stroke, shape = PointShape.Circle, opacity = 1 } = pointStyle?.shape
+    ? pointStyle ?? {}
+    : {
+        fill: color,
+        stroke: color,
```
nickofthyme, line 35: This `stroke` is never used cuz of line 50...

```tsx
stroke={shape ? stroke ?? color : undefined}
```

```suggestion
```

packages/charts/src/components/legend/legend_icon.tsx
nickofthyme: Yeah that looks fine. I just made one comment on the stroke default. https://github.com/elastic/elastic-charts/pull/1227#discussion_r671288361

packages/charts/src/components/legend/legend_icon.tsx
rshen91: 13e576d for changes, thank you!

packages/charts/src/components/icons/icon.tsx
```diff
@@ -40,6 +40,7 @@ interface IconProps {
   'data-test-subj'?: string;
   type?: IconType;
   color?: IconColor;
+  fillColor?: IconColor;
```
markov00, line 43: `fillColor`is not a valid attribute of an `svg` element and it seems not be used anyway by our code

markov00 (approved):
It looks great, works as expected.
Just a minor comment: in the legend we can ignore the `strokeWidth` if greater than 1 to avoid having a big stroke that cover the entire icon.
We can also ignore `opacity` that is less than 0.5 if we don't want to see the icon disappear

packages/charts/src/components/legend/legend_icon.tsx
```diff
@@ -28,11 +28,17 @@ const getAdjustedRadius = (radius: number | undefined, strokeWidth: number) => {
...
-  const { radius = 4, fill, strokeWidth = 1, stroke, shape = PointShape.Circle, opacity = 1 } = pointStyle?.shape
+  const {
+    radius = 4,
+    fill,
+    strokeWidth = 1,
+    stroke = color,
+    shape = PointShape.Circle,
+    opacity = 1,
+  } = pointStyle?.shape
     ? pointStyle ?? {}
```
nickofthyme: The idea behind checking the `pointStyle.shape` and using `{ fill: color }` is to avoid applying default styles to the default dot. Otherwise you could have the default dot with an opacity (etc.) that is not desirable. I think we pull this into a separate function and provide defaults easier.

Something like this but figure out the typings
```ts
const getStyles = (color: string, {
  radius,
  fill,
  strokeWidth,
  stroke,
  shape,
  opacity,
}: PointStyle = {}) => (shape ? {
  radius,
  fill,
  strokeWidth: strokeWidth ?? 1,
  stroke: stroke ?? color,
  shape: shape ?? PointShape.Circle,
  opacity: opacity ?? 1,
} : {
  fill: color,
  shape: shape ?? PointShape.Circle,
});
```

packages/charts/src/components/legend/legend_icon.tsx
```diff
@@ -28,11 +28,17 @@ const getAdjustedRadius = (radius: number | undefined, strokeWidth: number) => {
 
 /** @internal */
 export const LegendIcon = ({ pointStyle, color }: LegendIconProps) => {
-  const { radius = 4, fill, strokeWidth = 1, stroke, shape = PointShape.Circle, opacity = 1 } = pointStyle?.shape
+  const {
+    radius = 4,
```
nickofthyme, line 32: This is not needed, this default is handled in the logic below.

```suggestion
    radius,
```

packages/charts/src/components/legend/legend_icon.tsx
rshen91: Lmk if you think this is ok! 57d1684 thanks!

packages/charts/src/components/legend/legend_icon.tsx
rshen91: 57d1684 thank you!

packages/charts/src/components/icons/icon.tsx
rshen91: this makes sense to me, thanks for the catch of the invalid svg prop - let me know what you think about changes in 150b3d6 thanks

stories/stylings/stylings.stories.tsx
```diff
@@ -40,3 +40,4 @@ export { Example as partitionLabels } from './21_partition_labels';
 export { Example as darkTheme } from './22_dark_theme';
 export { Example as withTexture } from './23_with_texture';
 export { Example as textureMultipleSeries } from './24_texture_multiple_series';
+export { Example as shapeLegendMultipleSeries } from './25_mixed_chart_shapes';
```
nickofthyme, line 43: Generally we match these so they are easy to find

```suggestion
export { Example as mixedChartShapes } from './25_mixed_chart_shapes';
```

packages/charts/src/components/legend/legend_icon.tsx
nickofthyme: Ok I see what you mean here. I think a better solution would be to scale down the diamond to fit inside of the "viewport". Like described [here](https://math.stackexchange.com/a/60722/952008). 

image

### Before
Image 2021-07-21 at 12 57 13 PM

### After
Image 2021-07-21 at 1 16 54 PM

> Notice the diamond is now contained inside the square. See code changes in 8637423b929d3ee55f2e7d3ba8b556f4a605d45a.

packages/charts/src/components/legend/color.tsx
nickofthyme: gone

packages/charts/src/components/legend/color.tsx
nickofthyme: Fixed in latest changes...

image

packages/charts/src/chart_types/xy_chart/renderer/shapes_paths.ts
```diff
@@ -14,30 +15,32 @@ export type SVGPath = string;
...
+/**
+ * Returns shape function based on rotation of square in degrees
+ */
+const square = (rotation = 0): SVGPathFn => (r: number) => {
+  const d = getRadians(rotation);
+  const s = Math.cos(d) + Math.sin(d);
+  // scaled r to account for rotation;
+  const sr = s > 0 ? r / s : r;
+  return `M ${-sr} ${-sr} L ${-sr} ${sr} L ${sr} ${sr} L ${sr} ${-sr} Z`;
 };
```
nickofthyme, line 37: See https://github.com/elastic/elastic-charts/pull/1227#discussion_r674236783 and rational [here](https://math.stackexchange.com/a/60722/952008)

packages/charts/src/components/legend/legend_item.tsx
```diff
@@ -196,6 +206,7 @@ export class LegendListItem extends Component<LegendItemProps, LegendItemState>
             isSeriesHidden={isSeriesHidden}
             hasColorPicker={hasColorPicker}
             onClick={this.handleColorClick(hasColorPicker)}
+            pointStyle={alignLegendPointStyles ? pointStyle : undefined}
```
nickofthyme, line 209: Used to explicitly control legend point style alignment by user. We could infer when to apply the styles as a fallback by comparing the local `[area/line/bubble]SeriesStyles.point` to the respective value on the `Theme`. This could be something like...

```ts
deepEqual(areaSeriesStyles.point, theme.area.point, true); // true for partial comparison
```

But that seems a bit much for now.

packages/charts/src/components/legend/legend.tsx
```diff
@@ -109,7 +112,7 @@ function LegendComponent(props: LegendStateProps & LegendDispatchProps) {
     <div className={legendClasses} style={positionStyle}>
       <div style={containerStyle} className="echLegendListContainer">
         <ul style={listStyle} className="echLegendList">
-          {items.map((item, index) => renderLegendItem(item, itemProps, items.length, index))}
+          {items.map((item, index) => renderLegendItem(item, itemProps, index))}
```
nickofthyme, line 115: removed unused argument

nickofthyme (approved):
Ok sorry for all the back and forth. I think this is good to merge now. We can discuss my changes tomorrow and modify them as you see fit.

packages/charts/src/components/legend/color.tsx
```diff
@@ -46,14 +46,14 @@ export const Color = memo(
             title="change series color"
             ref={ref}
           >
-            <Icon type="dot" color={color} aria-label={`Change series color, currently ${color}`} />
+            <LegendIcon pointStyle={pointStyle} color={color} ariaLabel={`Change series color, currently ${color}`} />
```
nickofthyme, line 49: I'm not sure about this in the future. When the point styles take over this could mean the series color being selected/set would not be reflected in the color icon. Should revisit at a later time.

rshen91:
@markov00 @nickofthyme updated the vrts in 46ab4e4. The large file change is because of the styling for the pointStyles are taken from the theme by default with the empty fill. Let me know if this looks good to you thanks!

rshen91:
@elastic/eui-design do you mind providing a review on this PR please?

rshen91:
jenkins test this

rshen91:
jenkins test this please

nickofthyme:
@miukimiu the basic changes here is that for any area, line or bubble chart that has visible points we use the point styles on the legend icon. This would be all area styles now have hollow points with white fill. 👇🏼 

image

If this is not desirable we could only apply the styles to the legend icon with they differ from the main them styles.

monfera (changes requested):
tl;dr using solid colored shapes would help readability and accessibility, for when there's no solid/holey shape ambiguity on a given chart. WCAG guidelines don't sufficiently address, yet still acknowledge the impact of shape size and aspect ratio when it comes to distinguishing colors or contrast. An example: given the same bounding box for the marker in the legend, the full paint of that bounding box will be more ink than a solid disk in that bounding box, which is still better than, say, an empty circle with a thin border.

-------------------------------

_Thanks for letting in a late comment after the Kibana testing week and PTO. I'd still like to squeeze in a question relating to this as a significant BREAKING change, specifically, the mass changes to mocks due to solid to hollow markers, as those can't be solved well with a follow-up PR._

If there's a variety of markers, it's indeed great to give the user a marker shape cue, as in this change, which is good for cross-referencing and accessibility:

image

Yet most charts, thankfully, don't have a plethora of marker types. In fact it's almost always a bad idea to mix marker types, as
- on line charts, it's redundant, and the markers themselves introduce noise (see the above example for this as well - it's exaggerated, but real line charts will turn to such barbed wire charts easily)
- on scatterplots, each marker type has different characteristics (ink surface area; ink bounding box; occlusion effects; crowding effects; perceptibility, saliency etc.) so a medium or high overplotting scatterplot will be a bit of a mess
- etc

The upshot is that most of the charts won't need legend disambiguation for marker shape.

image

Using holey markers (circle instead of disk etc.) has the negative that the legend ink area is diminished to a fraction of it, which reduces the ability to find the right legend item, discern colors etc. and is thus an accessibility issue. Even if the shape is holey on the line, it's easier to discern the color of a long line than of a small circle. Even if we disregard it, it's useful if at least the legend colors have significant "meat", for easier cross-referencing.

Proposal: 
- use solid markers in the legend if there's no ambiguity about holey vs solid markers on a given chart
- even if the user opts for some other marker, eg. `x` everywhere on a chart, use solid colors in the legend (*)
- if it's needed, please point a well discussed use case (eg. issue comments, PR comments here and conclusions) and add an option for enforcing the behavior currently in this PR commit; eg. which user team would need holey legends on non-ambiguous charts, and why 

(*) Comments for where it might be heading, post this PR:

Currently, the legends area is a singleton on the chart. Therefore the encoded information of the markers are conflated. For example, it's currently not possible to have three distinct legends for color, marker type and size. This impairs readability of each encoding channel. So the long term thing is to treat legends as data visualizations in and of themselves (a recurring theme) so that a legend is not a singleton on the chart, and also, so that multiple charts can share some legend(s).

image

In light of such future, we can restate my request this way: when there's no marker ambiguity, let's NOT EVEN try to use the legend swatch to convey anything about marker shape; it's just a color swatch.

Also, having decoupled the marker shape from the color assignment for non-ambiguous cases, we can do better and switch to squares or rectangles for the color swatch (with or without rounded corner) to further increase the color ink area.

packages/charts/src/chart_types/xy_chart/legend/legend.ts
```diff
@@ -79,20 +88,33 @@ export function getLegendExtra(
...
+function getPointStyle(spec: BasicSeriesSpec, theme: Theme): PointStyle | undefined {
+  const mergeOptions: MergeOptions = { mergeOptionalPartialValues: true };
+  if (isBubbleSeriesSpec(spec)) {
+    return mergePartial(theme.bubbleSeriesStyle.point, spec.bubbleSeriesStyle?.point, mergeOptions);
+  } else if (isLineSeriesSpec(spec)) {
+    return mergePartial(theme.lineSeriesStyle.point, spec.lineSeriesStyle?.point, mergeOptions);
+  } else if (isAreaSeriesSpec(spec)) {
+    return mergePartial(theme.areaSeriesStyle.point, spec.areaSeriesStyle?.point, mergeOptions);
+  }
+}
```
monfera, line 101: It's not a PR comment, yet a great example for how TS leads to hoops we have to jump through (in the code, as well as runtime) to satisfy its type verification operations. TS requires a specific, and sometimes artificial coding style. Maybe we could eventually move the series type to distinct classes, as class instances have both static and runtime type, and something like this `if` cascade would become a simple method call

monfera (approved):

markov00 (approved):
Changes looks good to me!

PR #1233, opened 2021-07-07
feat(xy): hide labels that protrude the bar geometry

### LINKED ISSUE #1234: Add option to hide labels when they protrude the respective bar
The current option `hideClippedValue` only hides labels that flow outside the projection area.
As requested by #1123, we should also add the ability to hide labels that cannot be rendered fully within the bar geometry.

markov00:
## Summary
The `overflowConstraints` prop now replaces the existing `BarSeries.displayValueSettings.hideClippedValue` prop extending the existing feature with the ability to hide labels that protrude the bar geometry.

Screenshot 2021-07-08 at 15 03 31

BREAKING CHANGE: an API change is introduced: `hideClippedValue` is removed in favour of `overflowConstraints?: Array<LabelOverflowConstraint>;`. The array can contain one or multiple overflow constraints identified in `LabelOverflowConstraint`

## Details

I've added a different prop because both visibility filters are valid and useful.

## Issues

fix #1234

### Checklist


dej611 (commented):
@dej611 Thanks for the review

> 1. the new API prop is not super nice.
>  * The previous one was singular (`*ClippingValue`) while this has plural name `*ClippingValues`.
>  * It's a long name but at the same time it has shortenings in it (`*Geom*`). I would rather have a full name >`hideGeometryClippedValue`.
>  * As alternative I would consider also adding an object value to the `hideClippedValue` as:
>  ```ts
>  hideClippedValue?: boolean | { edges?: boolean, geometries?: boolean }
>  ```
>  The default `boolean` value should be converted to `{ edges?: boolean }` internally

I've discussed and agree with @monfera  to use a slight cleaner approach here https://github.com/elastic/elastic-charts/pull/1233/commits/97fd23b1b1ee180fa1b852602ba9117a417b2881
the `hideClippedValue` and `hideGeomClippedValues` are replaced by:
```ts
const LabelOverflowConstraint = Object.freeze({
  BarGeometry: 'barGeometry' as const,
  ChartEdges: 'chartEdges' as const,
});
...
hideIfOverflows?: Array<LabelOverflowConstraint>;
```
This allows us and the users to: combine easily the constraints. The API is also scalable and can accommodate other constraints along the way.


> 2. I find a bit confusing the way this flag works combined with the `contain value label within bar element`
>  * I get that probably they won't get combined that often, but while playing in the storybook example I found that enabling >both leads to something I would not expect.
>  * If there's a plan to remove the `contain value label within bar element` flag, it does not matter probably

You are right, I've deprecated the API prop and we will remove it soon.

Also, can you add the same flag to the advanced value labels example?

nickofthyme (approved):
LGTM, so if `barGeometry` is not specified in the `hideIfOverflow` prop, then this value below would be visible correct?

Image 2021-07-12 at 11 04 37 AM

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -154,21 +149,29 @@ export function renderBars(
...
 
-    const hideClippedValue = displayValueSettings ? displayValueSettings.hideClippedValue : undefined;
+    const hideIfOverflowsChartEdges = hideIfOverflows.has(LabelOverflowConstraint.ChartEdges);
+    const hideIfOverflowsBarGeometry = hideIfOverflows.has(LabelOverflowConstraint.BarGeometry);
     // Based on rotation scale the width of the text box
     const bboxWidthFactor = isHorizontalRotation ? textScalingFactor : 1;
 
-    const displayValue =
-      displayValueSettings && displayValueSettings.showValueLabel
+    const displayValue: BarGeometry['displayValue'] | undefined =
```
monfera, line 164: Does it need to also be `undefined`, or could it just be an empty string? There are then many ways to not render an empty string, eg. just render it (nothing will show up) or test for text lengt === 0 or test for text truthiness (`""` is falsey)

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -154,21 +149,29 @@ export function renderBars(
...
+      displayValueText && displayValueSettings?.showValueLabel
         ? {
             fontScale: textScalingFactor,
             fontSize: fixedFontScale,
             text: displayValueText,
             width: bboxWidthFactor * displayValueWidth,
             height: textScalingFactor * fixedFontScale,
-            hideClippedValue,
-            isValueContainedInElement: displayValueSettings.isValueContainedInElement,
+            hideIfOverflowsChartEdges,
```
monfera, line 172: Let's suppose, there's yet another constraint, would it need a new parameter here? If it's likely, maybe it's easier to pass around the Set, and check for `has` at the place of application

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -154,21 +149,29 @@ export function renderBars(
       fixedFontScale,
       fontSize,
     );
+    const hideIfOverflows: Set<LabelOverflowConstraint> = new Set(
+      displayValueSettings?.hideIfOverflows ?? [
+        LabelOverflowConstraint.ChartEdges,
+        LabelOverflowConstraint.BarGeometry,
+      ],
+    );
```
monfera, line 157: Still pondering: the users of the library are not end users but coders, so a `Set<LabelOverflowConstraint>` might be a clearer API receptacle as the input is a set, rather than a bag (=multiset ie. possible duplicates) or array (ordered multiset). At the expense of trivial verbosity. From your earlier example:
```
overflowConstraints={['barGeometry', 'chartEdges']}
overflowConstraints={new Set(['barGeometry', 'chartEdges'])}
```

monfera (approved):
LGTM, the comments are just for consideration, some of them fitting in the old theme of trying to avoid/remove union types esp. where the non-nil type can accomodate a sensible "nil" value eg. NaN for numbers or `''` for strings

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
dej611: As a consumer of the API I think the `Set` is an implementation detail that should not be exposed. Unless there's another API usage of the `Set` container I see no reason to expose it.

package.json
```diff
@@ -41,7 +41,7 @@
     "test:tz-ny": "TZ=America/New_York jest --verbose --config=jest.tz.config.js",
     "test:tz-jp": "TZ=Asia/Tokyo jest --verbose --config=jest.tz.config.js",
     "test:integration": "./scripts/vrt.sh",
-    "test:integration:local": "LOCAL_VRT_SERVER=true PORT=9001 yarn test:integration",
+    "test:integration:local": "LOCAL_VRT_SERVER=true PORT=9002 yarn test:integration",
```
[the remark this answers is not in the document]
markov00: I'm still using the local server when I need to update snapshots locally. What do you use instead? Do you run just the integration that create a server everytime you run it?

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
markov00: this object has multiple properties, is not just the text string. IMHO I don't see how this can be better than having or not having an object.

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
markov00: I agree with @dej611 here, adding a `Set` doesn't prevent the user adding duplicates in the array `overflowConstraints={new Set(['barGeometry', 'chartEdges','barGeometry'])}`
The cardinality is relatively low that it's easy to spot that manually. Internally we just get rid of duplicates

package.json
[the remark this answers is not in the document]
markov00: reverted in https://github.com/elastic/elastic-charts/pull/1233/commits/4d2a1c44ca8f25136129aceb1cfc55022af16e66

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
markov00: fixed in https://github.com/elastic/elastic-charts/pull/1233/commits/5f0b2eee8dd340a992bc5dea99f621d85f41c51c

packages/charts/api/charts.api.md
```diff
@@ -645,8 +645,9 @@ export type Direction = $Values<typeof Direction>;
 
 // @public (undocumented)
 export interface DisplayValueSpec {
-    hideClippedValue?: boolean;
+    hideIfOverflows?: Array<LabelOverflowConstraint>;
```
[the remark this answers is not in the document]
markov00: done in https://github.com/elastic/elastic-charts/pull/1233/commits/5f0b2eee8dd340a992bc5dea99f621d85f41c51c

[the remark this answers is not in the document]
markov00: completed in https://github.com/elastic/elastic-charts/pull/1233/commits/c8943287952641146465f4125b9a447f15ada2bf

markov00:
@nickofthyme 

> LGTM, so if `barGeometry` is not specified in the `hideIfOverflow` prop, then this value below would be visible correct?

Yes, you are right

package.json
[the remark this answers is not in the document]
nickofthyme: Yeah I just run the `test:integration` the `test:integration:local` was giving me issues from the start and I never investigated, I just moved on from local 😅

nickofthyme (approved):
Updated changes LGTM!

Issue #1213, opened 2021-06-22
Update library license 


nickofthyme:
Change the license to match Kibana as ELv2 and SSPL.

PR #1242, opened 2021-07-09
chore(license): elastic license 2.0

ghudgins:
Resolves https://github.com/elastic/elastic-charts/issues/1213


markov00:
thanks @ghudgins for taking this step.
It looks good , you just have to update the API reporter doc (used to verify any breaking change in the API) using:
```
yarn api:check:local
```
and commit the changes if any

LICENSE.txt
```diff
@@ -1,202 +1,7 @@
...
-   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-   See the License for the specific language governing permissions and
-   limitations under the License.
+Source code in this repository is covered by (i) a dual license under the Server
+Side Public License, v 1 and the Elastic License 2.0 or (ii) an Apache License
+2.0 compatible license or (iii) solely under the Elastic License 2.0, in each
+case, as noted in the applicable header. The default throughout the repository
+is a dual license under the Server Side Public License, v 1 and the Elastic
+License 2.0, unless the header specifies another license. Code that is licensed
+solely under the Elastic License 2.0 is found only in the x-pack folder.
```
nickofthyme, line 7: x-pack?

nickofthyme (commented):
Nice!!! :🎉 

Code changes look good to me.

@markov00 do we need any changes to `NOTICE.txt`? It looks to only pertain to 3rd-party licenses but just wanted to check. 👇🏼 

https://github.com/elastic/elastic-charts/blob/87629e23e7cc8aef125a8b6d2b87092085abe2bd/NOTICE.txt#L1-L42

ghudgins:
> Nice!!! :🎉
> 
> Code changes look good to me.
> 
> @markov00 do we need any changes to `NOTICE.txt`? It looks to only pertain to 3rd-party licenses but just wanted to check. 👇🏼
> 
> https://github.com/elastic/elastic-charts/blob/87629e23e7cc8aef125a8b6d2b87092085abe2bd/NOTICE.txt#L1-L42

I believe this is for _third party_ OSS disclosures and is unaffected by this change

markov00:
Yep, the NOTICE.txt is only for third-party libraries.
@ghudgins to correctly merge this as breaking change let's do that together if you want. The final squash commit should contain the body with a breaking change footer like:
```
This commit changes the library license to align with .....

Resolves #1213

BREAKING CHANGE: the library is no more XXX but YYY
```

PR #1250, opened 2021-07-13
ci: fix semantic-release slack-bot config

markov00:
## Summary

The current spread operator wrongly spread the `semantic-release-slack-bot` plugin definition as there were two different plugin.
This PR correct that spread, keeping the plugin name and the configuration within an array as requested


nickofthyme (approved):
Ugh that's my fault. I noticed it wasn't triggering and was going to look into it so thanks for fixing it 🎉.

PR #1253, opened 2021-07-14
fix: heatmap snap domain to interval

### LINKED ISSUE #1165: Heatmap time buckets should use value from timeScale
**Describe the bug**
When using the `Heatmap` spec with a time scale, the bucket values are trivial computed and don't always align with the values on the data. This causes charts to render without the cells that should appear the in given domain.

**Expected behavior**
Min and max domain values should be extended to include the full extent of partial buckets. This could be done using the `timeScale.rawTcks`.

**Screenshots**
image

 - Elastic Charts: `@latest`

This is also related to handling partial time buckets in heatmap. Now lens is just increasing the domain to include the partial buckets as if they were full buckets which is not ideal. See https://github.com/elastic/kibana/pull/97978/files/6283fb0d54cfa1c8e482ffad4b249e294750f02a..356109c161583af3ec60a69be57c2e96e1521cc5#diff-ecf0195846e85c9002680a385d9afe155d27ad5de1c696bdf5f08101201a440fR88-R100

**Kibana Cross Issues**
https://github.com/elastic/kibana/pull/97978#issuecomment-845280585

markov00:
## Summary

Heatmaps with a time scale on the X-axis now adjust the rendered time range to fully cover the edges when a custom domain is used.
We also took this opportunity to clean and abstract the code used for computing and handling date and time.
The library is now able to abstract from the underlying implementation library (moment or luxon at the moment), allowing us to experiment and work with diverse libraries removing some tech debt.

https://user-images.githubusercontent.com/1421091/126339475-fc429a5b-3cb4-4c12-9dd0-b3893b7557b9.mp4

## Details

I've abstracted some time logic as asked by @monfera taking inspiration(copying) from what was done in the timeslip code.
I've also introduced some specific Elasticsearch types like the CalendarInterval and FixedInterval. I've made them ES specific as the concept and the snapping logic applies the same strategy as the one described in https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html

## Issues

fix #1165

### Checklist


packages/charts/src/utils/data/date_time.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    const initialDate = DateTime.fromISO('2020-01-03T07:00:01Z');
+    const snappedDate = snapDateToInterval(initialDate.toMillis(), 'fixed', '30m', 'start', 'UTC');
+    expect(DateTime.fromMillis(snappedDate, { zone: 'utc' }).toISO()).toBe('2020-01-03T07:00:00.000Z');
+  });
+
+  it('should snap to end of fixed interval', () => {
+    const initialDate = DateTime.fromISO('2020-01-03T07:00:01Z');
+    const snappedDate = snapDateToInterval(initialDate.toMillis(), 'fixed', '30m', 'end', 'UTC');
+    expect(DateTime.fromMillis(snappedDate, { zone: 'utc' }).toISO()).toBe('2020-01-03T07:29:59.999Z');
+  });
```
nickofthyme, line 36: Can you add a test for non-es time unit case?

```ts
  it('should handle bad unit gracefully - calander', () => {
    const initialDate = DateTime.fromISO('2020-01-03T07:00:01Z');
    const snappedDate = snapDateToInterval(initialDate.toMillis(), 'calendar', '30z', 'end', 'UTC');
    expect(DateTime.fromMillis(snappedDate, { zone: 'utc' }).toISO()).toBe(initialDate.toMills());
  });

 it('should handle bad unit gracefully - fixed', () => {
    const initialDate = DateTime.fromISO('2020-01-03T07:00:01Z');
    const snappedDate = snapDateToInterval(initialDate.toMillis(), 'fixed', '30z', 'end', 'UTC');
    expect(DateTime.fromMillis(snappedDate, { zone: 'utc' }).toISO()).toBe(initialDate.toMills());
  });
```

packages/charts/src/utils/data/date_time.ts
```diff
@@ -18,3 +19,85 @@ export function getMomentWithTz(date: number | Date, timeZone?: string) {
...
+  if (unit === 'month' || unit === '1M') {
+    return 'month';
+  }
+  if (unit === 'quarter' || unit === '1q') {
+    return 'quarter';
+  }
+  if (unit === 'year' || unit === '1y') {
+    return 'year';
+  }
+}
```
nickofthyme, line 103: Just a thought 💭 ...

```suggestion
const esMomentIntervalMapping = {
  minute: 'minutes', 
  '1m': 'minutes',
  hour: 'hour',
  '1h': 'hour',
  day: 'day',
  '1d': 'day',
  week: 'week',
  '1w': 'week',
  month: 'month',
  '1M': 'month',
  quarter: 'quarter',
  '1q': 'quarter',
  year: 'year',
  '1y': 'year',
}

function esCalendarIntervalsToMoment(unit: string): unitOfTime.StartOf | undefined {
  esMomentIntervalMapping[unit]
}
```

stories/heatmap/3_time.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+import { Chart, Heatmap, RecursivePartial, ScaleType, Settings } from '../../packages/charts/src';
+import { Config } from '../../packages/charts/src/chart_types/heatmap/layout/types/config_types';
+import { getRandomNumberGenerator } from '../../packages/charts/src/mocks/utils';
+
+const rng = getRandomNumberGenerator();
+const start = DateTime.fromISO('2021-03-27T20:00:00', { zone: 'CET' });
+const end = DateTime.fromISO('2021-03-28T11:00:00', { zone: 'CET' });
+const data = new Array(14).fill(0).reduce((acc, d, i) => {
+  acc.push([start.plus({ hour: i }).toMillis(), 'cat A', rng(0, 10)]);
```
nickofthyme, line 21: 😼

stories/heatmap/3_time.tsx
```diff
@@ -0,0 +1,109 @@
...
+ * Side Public License, v 1.
+ */
+
+import { number } from '@storybook/addon-knobs';
+import { DateTime } from 'luxon';
+import React, { useMemo } from 'react';
+
+import { Chart, Heatmap, RecursivePartial, ScaleType, Settings } from '../../packages/charts/src';
+import { Config } from '../../packages/charts/src/chart_types/heatmap/layout/types/config_types';
+import { getRandomNumberGenerator } from '../../packages/charts/src/mocks/utils';
```
nickofthyme, line 15: I don't see `snapDateToInterval` used here. Am I missing something?

nickofthyme (commented):
Looking good to me. I left a few comments below.

packages/charts/src/utils/data/date_time.ts
monfera: Short, less code noise (curlies), no branching and O(1) 👍 

Would it be possible to collect time related utils in the monorepo but outside `charts`? Eg. under some `time-utils`. It's one of those things on which various bits of Kibana need to depend, and now, everybody implements their own thing.

packages/charts/src/utils/data/date_time.ts
```diff
@@ -18,3 +19,85 @@ export function getMomentWithTz(date: number | Date, timeZone?: string) {
...
 }
+
+/** @internal */
+export type UnixTimestamp = number;
+
+/** @internal */
+export function snapDateToInterval(
+  date: number | Date,
+  interval: 'calendar' | 'fixed',
+  unit: string,
```
monfera, line 30: At some point in the data flow, there could be a type guard, something like `isValidUnit`, which would ensure that later down the road, it'll be a proper key to the `esMomentIntervalMapping` lookup object. The type guard function itself could simply query `esMomentIntervalMapping` and if it returns something, it's legit. Doing so would remove the need for the permissive `string`, and the avoidable union type `unitOfTime.StartOf | undefined`

packages/charts/src/utils/data/date_time.ts
monfera: Also, the `unitOfTime.StartOf | undefined` union type may be avoided, see my other note

stories/heatmap/3_time.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+const rng = getRandomNumberGenerator();
+const start = DateTime.fromISO('2021-03-27T20:00:00', { zone: 'CET' });
+const end = DateTime.fromISO('2021-03-28T11:00:00', { zone: 'CET' });
+const data = new Array(14).fill(0).reduce((acc, d, i) => {
```
monfera, line 20: Minor: `[...new Array(14)].reduce` is good too

stories/heatmap/3_time.tsx
```diff
@@ -0,0 +1,109 @@
...
+  acc.push([start.plus({ hour: i }).toMillis(), 'cat A', rng(0, 10)]);
+  acc.push([start.plus({ hour: i }).toMillis(), 'cat B', rng(0, 10)]);
+  acc.push([start.plus({ hour: i }).toMillis(), 'cat C', rng(0, 10)]);
+  acc.push([start.plus({ hour: i }).toMillis(), 'cat D', rng(0, 10)]);
+  acc.push([start.plus({ hour: i }).toMillis(), 'cat E', rng(0, 10)]);
```
monfera, line 25: Minor: `[].push()` takes an arbitrary number of arguments to push, though, as efficiency isn't a concern here, a `[...new Array(14)].flatMap()` works too

packages/charts/src/utils/data/date_time.ts
```diff
@@ -18,3 +19,85 @@ export function getMomentWithTz(date: number | Date, timeZone?: string) {
...
+  if (isNaN(unitMultiplier)) {
+    return date;
+  }
+
+  const roundedDate = Math.floor(date.valueOf() / unitMultiplier) * unitMultiplier;
+  return snapTo === 'start' ? roundedDate : roundedDate + unitMultiplier - 1;
+}
+
+function esFixedIntervalsToMomentUnit(unit: string): number {
+  if (unit.endsWith('ms')) {
```
monfera, line 59: A `switch(true)` would be clearer and more compact, similar to [here](https://github.com/elastic/elastic-charts/pull/1227#discussion_r670995307).

Though a better, deeper solution here is, splitting the `unit`(*) to a numerical part and a unit string part as upstream as possible, and using a lookup object similar to Nick's suggestion for `esCalendarIntervalsToMoment`

In general, all repeated `if`s are very strong candidates for considering its alternatives.

(*) The `unit` is now a bit of a misnomer, because it's a tuple of `{quantity, unit}`. So something like `bin` or `interval` or some such would be a more truthful name.

packages/charts/src/utils/data/date_time.ts
```diff
@@ -18,3 +19,85 @@ export function getMomentWithTz(date: number | Date, timeZone?: string) {
...
+  if (unit.endsWith('m')) {
+    return getNumericalUnit(unit, 'm') * 1000 * 60;
+  }
+  if (unit.endsWith('h')) {
+    return getNumericalUnit(unit, 'h') * 1000 * 60 * 60;
+  }
+  if (unit.endsWith('d')) {
+    return getNumericalUnit(unit, 'd') * 1000 * 60 * 60 * 24;
+  }
+  return NaN;
```
monfera, line 74: Nice preservation of the `number` type

packages/charts/src/utils/data/date_time.ts
```diff
@@ -18,3 +19,85 @@ export function getMomentWithTz(date: number | Date, timeZone?: string) {
...
+function calendarIntervalSnap(date: moment.Moment, unit: string, snapTo: 'start' | 'end') {
+  const momentUnitName = esCalendarIntervalsToMoment(unit);
+  if (!momentUnitName) {
+    return date;
+  }
+  return snapTo === 'start' ? date.startOf(momentUnitName) : date.endOf(momentUnitName);
+}
+function fixedIntervalSnap(date: moment.Moment, unit: string, snapTo: 'start' | 'end') {
+  const unitMultiplier = esFixedIntervalsToMomentUnit(unit);
+  if (isNaN(unitMultiplier)) {
```
monfera, line 50: `isNaN` here is OK because TS rules out other types like strings (eg. what will `isNaN("3")`, `isNaN(null)`, `isNaN("")` be?) and `Number.isNaN` is good for a closer check, or (not possible here) if the number can be infinite, `!Number.isFinite`. I used `Number.isNaN` so far, though maybe worth shortening to `isNaN` where TS narrows it to `number` (though I'm less committal to TS; doing the good thing via a runtime guarantee has value too). Might be worth entertaining a code convention for the NaN-ness or finiteness check

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
```diff
@@ -122,10 +123,11 @@ export function shapeViewModel(
...
     const result = [];
-    let [timePoint] = xValues;
-    while (timePoint < xValues[1]) {
-      result.push(timePoint);
-      timePoint += xDomain.minInterval;
+    const [timePoint] = xValues;
+    let startDomainPoint = snapDateToInterval(timePoint, 'fixed', `${xDomain.minInterval}ms`, 'start');
+    while (startDomainPoint < xValues[1]) {
+      result.push(startDomainPoint);
+      startDomainPoint += xDomain.minInterval;
```
monfera, line 130: This would also work as it's a fixed cadence numerical iteration:

```javascript
const start = snapDateToInterval(timePoint, 'fixed', `${xDomain.minInterval}ms`, 'start');;
for(let s = start; s < xValues[1]; s += xDomain.minInterval) {
  result.push(s)
}
```
Maybe the reason for  a separate `xValues = result;` is the clarity about the purpose of each `if` branch, though the cost is that jumping to assignment places of `xValues` lands you on a noninformative row (`xValues = result;`) rather than the actual place where it's filled (in the loop) so maybe it's slightly better to directly push into it

packages/charts/src/utils/data/date_time.ts
```diff
@@ -18,3 +19,85 @@ export function getMomentWithTz(date: number | Date, timeZone?: string) {
   }
   return moment.tz(date, timeZone);
 }
+
+/** @internal */
+export type UnixTimestamp = number;
```
monfera, line 24: 👍

packages/charts/src/utils/data/date_time.ts
```diff
@@ -18,3 +19,85 @@ export function getMomentWithTz(date: number | Date, timeZone?: string) {
...
+  timeZone?: string,
+): UnixTimestamp {
+  const momentDate = getMomentWithTz(date, timeZone);
+
+  return interval === 'calendar'
+    ? calendarIntervalSnap(momentDate, unit, snapTo).valueOf()
+    : fixedIntervalSnap(momentDate, unit, snapTo).valueOf();
+}
+
+function calendarIntervalSnap(date: moment.Moment, unit: string, snapTo: 'start' | 'end') {
```
monfera, line 41: It may be the good tradeoff in this PR, we're adding a bit of technical debt, specifically, tight coupling with a specific date library. With more work, it'd be possible to separate date library wrapper functions from logic, such that you can change the date library with one import change in the entire code base (or part of it) by doing something similar to this in timeslip:
```js
import {
  addTimeToObj, timeObjFromCalendarObj, timeObjFromEpochSeconds, timeObjToSeconds, timeObjToWeekday, timeObjToYear,
}
  from './chrono-luxon/chrono-luxon.js'
// from './chrono-moment/chrono-moment.js'
// from './chrono-proposal-temporal/chrono-proposal-temporal.js'
``` 
Similarly, the opaque date type, `moment.Moment` here, can be abstracted out, so eg. it's called `Chrono` and is equally imported from a set of alternative time library wrappers.

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
monfera: As the above loop is just an implementation of the `range(from, to, increment)` function (our own util, or [d3.range](https://observablehq.com/@d3/d3-range))  which is customarily done left-closed, right-open like our use case here, it could be extracted into a utility function, so it becomes something like
```js
const xValues = timeScale 
  ? range(
        snapDateToInterval(timePoint, 'fixed', `${xDomain.minInterval}ms`, 'start'), 
        xValues[1],
        xDomain.minInterval
     )
  : [];
```
Two `let`s avoided 🙏

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
```diff
@@ -122,10 +123,11 @@ export function shapeViewModel(
 
   if (timeScale) {
     const result = [];
-    let [timePoint] = xValues;
-    while (timePoint < xValues[1]) {
-      result.push(timePoint);
-      timePoint += xDomain.minInterval;
+    const [timePoint] = xValues;
+    let startDomainPoint = snapDateToInterval(timePoint, 'fixed', `${xDomain.minInterval}ms`, 'start');
+    while (startDomainPoint < xValues[1]) {
```
monfera, line 128: The problem description in the issue doesn't seem to rule out snapping with temporal narrowing rather than broadening. Narrowing has the benefit that each heatmap cell is guaranteed to be complete. With something like sum or count, a partial bin can be quite misleading, while some other metrics like average might, depending on the data, be more resilient, but still a strong compromise (there may be very few, one or no data points in the partial bin)

packages/charts/src/utils/data/date_time.ts
monfera: Eelsewhere we have `export type TimeMs = number;` (that's for duration, but it encodes the unit, ms). So a `S` or `Ms` postfix may be useful.

It might be worth putting both this and the existing `TimeMs` into a common time type file (even outside `/charts`). The latter could perhaps be renamed `DurationMs`.

stories/heatmap/3_time.tsx
markov00: Is used internally in the Heatmap implementation

packages/charts/src/utils/data/date_time.ts
markov00: @monfera
>Would it be possible to collect time related utils in the monorepo but outside charts? Eg. under some time-utils. It's one of those things on which various bits of Kibana need to depend, and now, everybody implements their own thing.

we could but we can't at the moment publish multiple packages easily due to the semantic-release tool.

nickofthyme: > @monfera
> 
> > Would it be possible to collect time related utils in the monorepo but outside charts? Eg. under some time-utils. It's one of those things on which various bits of Kibana need to depend, and now, everybody implements their own thing.
> 
> we could but we can't at the moment publish multiple packages easily due to the semantic-release tool.

Yeah I think we need to figure out how this would work or either as parallel packages published as one or multiple published packages consumed by charts. I do think though that we can publish libraries that are only development util functions like the random number generator that eui and others use, in this case the release notes of that package are not really important.

Or we could do it as react, storybook and others and just publish all packages under the same version and release notes.

stories/heatmap/3_time.tsx
nickofthyme: Ah ok, sorry I thought the idea was to export this for users to apply to their data.

packages/charts/src/utils/data/date_time.test.ts
markov00: I've refactored a bit the code and you can't enter invalid interval right now: https://github.com/elastic/elastic-charts/pull/1253/commits/c93defff89c45bad4b0ceb5f6255e70f797224b2

packages/charts/src/utils/data/date_time.ts
markov00: I've refactored the code [here](https://github.com/elastic/elastic-charts/pull/1253/commits/c93defff89c45bad4b0ceb5f6255e70f797224b2), using a switch in one case, and an object map in another

packages/charts/src/utils/data/date_time.ts
markov00: I've revisited the interval describing it as the following:
```ts
export type CalendarInterval = {
  type: 'calendar';
  unit: CalendarIntervalUnit;
  quantity: number;
};
/** @internal */
export type FixedInterval = {
  type: 'fixed';
  unit: FixedIntervalUnit;
  quantity: number;
};
```

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
markov00: done https://github.com/elastic/elastic-charts/pull/1253/commits/c93defff89c45bad4b0ceb5f6255e70f797224b2

packages/charts/src/utils/data/date_time.ts
markov00: done [here](https://github.com/elastic/elastic-charts/pull/1253/commits/c93defff89c45bad4b0ceb5f6255e70f797224b2), no more nan or branching let's push the correctness check at the consumer level

packages/charts/src/utils/data/date_time.ts
markov00: refactored the code [here](https://github.com/elastic/elastic-charts/pull/1253/commits/c93defff89c45bad4b0ceb5f6255e70f797224b2)

stories/heatmap/3_time.tsx
markov00: done

stories/heatmap/3_time.tsx
markov00: done in https://github.com/elastic/elastic-charts/pull/1253/commits/c93defff89c45bad4b0ceb5f6255e70f797224b2

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
markov00: I haven't understood your statement here, let's discuss it on a call

packages/charts/src/utils/data/date_time.ts
markov00: I've extracted the time/date logic as done in the timeslip with some small adjustments
https://github.com/elastic/elastic-charts/pull/1253/commits/32f33c7229bb3c435eaae4d6ffb384a957121295
We can get rid of the `moment` specific code in few other parts of the library, but I'd like to do that in a different PR to avoid conflicts and regressions

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
monfera: Glad to! The gist of it is, snapping is a discretization, and in-between values can fall either way, depending on the strategy, and the current strategy seems to be obligate temporal broadening. Maybe that's the desirable or the only sensible option, it'd just be nice to see some words on why we aren't entertaining temporal narrowing

packages/charts/src/utils/chrono/elasticsearch.ts
```diff
@@ -0,0 +1,119 @@
...
+  | 'day'
+  | 'd'
+  | 'week'
+  | 'w'
+  | 'month'
+  | 'M'
+  | 'quarter'
+  | 'q'
+  | 'year'
+  | 'y';
```
monfera, line 28: @delvedor can you please take a look at the extraction here? Not sure if there's a feasible way for simple, lightweight type unification for the small overlap between `elasticsearch.js` and `elastic-charts` types, but maybe it's possible to approximate it. What would be the types in this file analogous in `elasticsearch.js` that could perhaps be copied or depended on?

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
markov00: After a discussion with the team, we are going to keep the current implementation, broadening the interval instead of narrowing due to the way data and time range is handled in Kibana and its editors

packages/charts/src/utils/chrono/elasticsearch.ts
markov00: @monfera I've took a look at the DateHistogram aggregation types in `elasticsearch-js` but I've found that they are currently using a different loose type:
```ts
export enum DateInterval {
  second = 0,
  minute = 1,
  hour = 2,
  day = 3,
  week = 4,
  month = 5,
  quarter = 6,
  year = 7
}

type Time = string | integer

interface AggregationsDateHistogramAggregation extends AggregationsBucketAggregationBase {
fixed_interval?: DateInterval | Time
calendar_interval?: DateInterval | Time
...
}
```
the `interval-unit` like `1d` `30d` etc is covered by the `Time` definition simply as `string`
I can create a PR to fix that in their specification but for now I think we can go on with the current implementation

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
```diff
@@ -300,8 +296,12 @@ export function shapeViewModel(
...
 
     // find X coordinated based on the time range
-    const leftIndex = typeof startValue === 'number' ? bisectLeft(xValues, startValue) : xValues.indexOf(startValue);
-    const rightIndex = typeof endValue === 'number' ? bisectLeft(xValues, endValue) : xValues.indexOf(endValue) + 1;
+    // the xValues array is casted as any because the slight incompatible type of d3 bisect.
+    // it works anyway without problems also if the data is a mix of string and numbers
+    const leftIndex =
+      typeof startValue === 'number' ? bisectLeft(xValues as any, startValue) : xValues.indexOf(startValue);
+    const rightIndex =
+      typeof endValue === 'number' ? bisectLeft(xValues as any, endValue) : xValues.indexOf(endValue) + 1;
```
monfera, line 304: The type of the 1st and 2nd params of `bisectLeft` needs to match, and we're typeguarding the 2nd param to `number`, so it's slightly tighter to say `as number[]`. Though it's not truthful, maybe we could brainstorm so I can learn about whether `OrdinalDomain` needs to be `(number | string)[]` and can't be `number[] | string[]` eg. by coercing numbers to strings if at least one element is a string. Though there's still the lack of type coherence between `xValues` and `startValue` - probably they correlate strongly (otherwise we can't do `as any` or `as number[]` here) but this seems to be lost on TS

monfera: If the type of `bisectLeft` is incorrect or too restrictive (we can't overrule it as it's beyond what the D3 TS API guarantees, but iirc these are 3rd party post hoc libs, not canonical Bostock contract) then maybe it'd be better to factor it out into our `bisectLeft` utility with the desired types, so we can easily change from D3 if needed. I think we could already use the also logarithmically bisecting `monotonicHillClimb` instead of `d3.bisectLeft`

monfera (commented):
LGTM in general, there's an as yet unexpected mock change, and maybe we could do the bisecting differently. Glad to add a `bisectLeft`-like thin wrapper of the lower level `monotonicHillClimb` function we already have for logarithmic search in an array

packages/charts/src/utils/chrono/elasticsearch.ts
monfera: Thanks for checking! I assumed it'd be of a similar granularity

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
monfera: nicely done!

packages/charts/src/utils/data/date_time.ts
monfera: Very nice, thanks! I'm positively surprised there was no need for the library-specific function exports to use an opaque date type like the `timeObj...` things in [timeslip](https://github.com/elastic/elastic-charts-gists/blob/2e3241f0b1d7580fbd63e284b0326c8879149476/examples/02_timeslip/timeslip/chrono/chrono-luxon/chrono-luxon.js), so there was also no need to export an opaqe type object like `TimeObj` per library. So it's even cleaner. Still, if runtime demands it, it'd be OK to pass around such library-specific time objects as long as we only ever use them via `chrono.ts` exposed functions

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
markov00: There is one main reason for having `(number | string)[]` as `OrdinalDomain`: we like to preserve the original type, coming from the user data, to return the original values on the interaction callbacks (brushing or clicking)
We can probably force it to be `number[] | string[]` but I don't have strong opinion here

monfera: Some more info may strengthen the case for `string[] | number[]`: 

D3 `bisectLeft`, and generally, logarithmic search requires [total order](https://en.wikipedia.org/wiki/Total_order), which we likely ensure by sorting the values beforehand. The mechanism doesn't matter, eg. `[].sort` also relies on a predicate plus array which together guarantee a total order, otherwise the order is ill defined (EcmaScript [requires](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.sort) transitivity for predictable results: `If a <CF b and b <CF c, then a <CF c (transitivity of <CF)`)

Out of the properties of total order, transitivity is most important: if a <= b and b <= c then a <= c must be true (it's not necessarily true for [partial orders](https://en.wikipedia.org/wiki/Partially_ordered_set), eg. when comparing a number and a string is defined non-comparable).

Sorting `(number | string)[]` with a `<`/`>` predicate doesn't lead to total order. Example:

```
a = '2'
b = 2.5
c = '10'

a <= b: true
b <= c: true
a <= c: false
```
So a sensible option is to string-compare all elements if at least one element in the array is not a number, which is the default `sort` predicate (and number-compare for all-numeric arrays, which is not the default `sort` predicate). Instead of checking for this at the place of sorting, we may as well come clean and convert upfront. As you suggested Marco, the original value can still be retained. The effect of the conversion though is that a user-supplied `[5, '5']` will map to `['5', '5']` so an equally good method is, documenting in the API, perhaps even enforcing via TS, that the domain values be homogeneous.

There can of course be other sorting rules, eg. all strings are moved to the end (tiebreaker rule when string and number are compared), and within both the strings and numbers, there's full order

packages/charts/src/utils/data/date_time.ts
```diff
@@ -21,83 +22,96 @@ export function getMomentWithTz(date: number | Date, timeZone?: string) {
...
+      return 'week';
+    case 'month':
+    case 'M':
+      return 'month';
+    case 'quarter':
+    case 'q':
+      return 'quarter';
+    case 'year':
+    case 'y':
+      return 'year';
```
monfera: As it's quite a long switch, requiring up to 14 comparisons, it'd be a bit nicer if these were in a map, eg.
```
const esCalendarIntervalsToMoment = { 
  minurte: 'minutes',
  m: 'minutes',
  hour: 'hour',
  h: 'hour',
  ...
}

...

const whatever = esCalendarIntervalsToMoment[interval.unit];
```

packages/charts/src/utils/data/date_time.ts
monfera: Thanks, added comments there, so I'm setting this specific comment resolved

packages/charts/src/utils/chrono/elasticsearch.ts
```diff
@@ -0,0 +1,119 @@
...
+      return 'month';
+    case 'quarter':
+    case 'q':
+      return 'quarter';
+    case 'year':
+    case 'y':
+      return 'year';
+    default:
+      return 'hour';
+  }
```
nickofthyme, line 118: nit

```suggestion
    case 'hour':
    case 'h':
    default:
      return 'hour';
  }
```

packages/charts/src/utils/chrono/chrono.ts
```diff
@@ -0,0 +1,89 @@
...
+  addTimeToObj,
+  timeObjToUnixTimestamp,
+  startTimeOfObj,
+  endTimeOfObj,
+  timeObjFromAny,
+  timeObjToUTCOffset,
+  subtractTimeToObj,
+  formatTimeObj,
+  diffTimeObjs,
+} from './moment';
```
nickofthyme, line 20: I'm really not sure why this is needed as this is not configurable from outside of charts.

As it is we use `moment` across the charts code so are you thinking of being able to swap time libraries entirely? In that case we could have optional dependencies on `moment` and `luxon` so the user can switch between them.

If not is this to only be used as a utility to handle different time types?

nickofthyme (approved):
LGTM, still not entirely sure what the end game is having multiple time libraries.

packages/charts/src/utils/chrono/chrono.ts
markov00: This was done in the hope we can switch soon and easily to `luxon` or to `Temporal` https://tc39.es/proposal-temporal/docs/index.html
It really depends on Kibana right now, so extracting everything like that can help us migrating easily

packages/charts/src/utils/chrono/elasticsearch.ts
markov00: I've update the logic to use a map instead of a switch as proposed by robert

packages/charts/src/utils/chrono/chrono.ts
markov00: This was done in the hope we can switch soon and easily to `luxon` or to `Temporal` https://tc39.es/proposal-temporal/docs/index.html
It really depends on Kibana right now, so extracting everything like that can help us migrating easily

packages/charts/src/utils/data/date_time.ts
markov00: done in https://github.com/elastic/elastic-charts/pull/1253/commits/f471fd0ce4c82a05b5571cf2c4518d7ec54b99a9

markov00:
Heatmaps with a time scale on the X-axis now adjust the rendered time range to fully cover the edges when a custom domain is used. We also took this opportunity to clean and abstract the code used for computing and handling date and time. The library is now able to abstract from the underlying implementation library (moment or luxon at the moment), allowing us to experiment and work with diverse libraries removing some tech debt.

fix #1165

PR #1255, opened 2021-07-14
feat(partition): waffle chart

monfera:
## Summary

Adds waffle chart to the set of partition charts. It shows a parts to whole relationship, like pie charts or treemaps, but sometimes a better option than these. It is distinct from the gridded charts in APM that show specific units and were titled _waffle map_.

image

The waffle chart is a great alternative to single-layer part-to-whole charts such as the pie chart. It may also be preferable to a single layer treemap as well, for example, if many such charts can share a single color assignment to the categories, or if there's no room for in-shape text labels on the treemap, or the simplicity, data order consistency or aesthetics of this chart type is preferred.

## Details

The waffle chart is a lower precision format, because the data is often continuous, while the waffle cells necessitate rounding, so each cell represents a specific ratio (percentage).

At the current alpha level, there's very limited configurability, and it ought to stay this way until more demanding requirements are registered. For example, it's currently always a 10x10 grid (each cell represents 1% and each row or column, 10%).

This PR alters the legend item order such that the waffle colors and legend items match in sequence. Both follow a decreasing value order.

The waffle is responsive, eg. the separating borders get smaller as the chart gets smaller, though not in linear proportion. The `config.sectorLineWidth` is settable for more or less separation, though other aspects eg. cell corner radius aren't currently configurable (and they may not to be)

Known requirements for non-alpha status:
- hovering over cells highlights the specific cell; the desired behavior is, highlighting of the entire band of a given category, like with legend hover
- solving this will also solve the accessibility issue that each cell gets repeated in the a11y table, and fix the non-flattened legend, and events may arise on a per cell basis rather than per band (the root cause of all these is the currently per-cell geometry that needs to shift to a per-band representation)

### Checklist


packages/charts/src/chart_types/partition_chart/layout/utils/legend.ts
```diff
@@ -71,7 +83,15 @@ export function getLegendItems(
...
-  items.sort(flatLegend ? compareNames : compareTreePaths);
+  items.sort(
+    partitionLayout === PartitionLayout.waffle // waffle has inherent top to bottom descending order
+      ? descendingValues
+      : partitionLayout === PartitionLayout.mosaic // mosaic has inherent top to bottom ascending order
+      ? compareIndices
+      : flatLegend
+      ? compareNames
+      : compareTreePaths,
+  );
```
nickofthyme, line 94: 🙈

packages/charts/src/chart_types/partition_chart/renderer/canvas/canvas_wrapped_renderers.ts
```diff
@@ -0,0 +1,114 @@
...
+import { ChartId } from '../../../../state/chart_state';
+import { ShapeViewModel } from '../../layout/types/viewmodel_types';
+import { ContinuousDomainFocus } from './partition';
+
+const linear = (x: number) => x;
+const easeInOut = (alpha: number) => (x: number) => x ** alpha / (x ** alpha + (1 - x) ** alpha);
+
+const MAX_PADDING_RATIO = 0.25;
+
+const latestRafs: Map<ChartId, number> = new Map();
```
nickofthyme, line 18: Could this cause a memory leak as charts are created and destroyed?

packages/charts/src/chart_types/partition_chart/state/selectors/compute_legend.ts
```diff
@@ -16,8 +16,9 @@ import { getPartitionSpecs } from './get_partition_specs';
 /** @internal */
 export const computeLegendSelector = createCustomCachedSelector(
   [getPartitionSpecs, getLegendConfigSelector, partitionMultiGeometries],
-  (specs, { flatLegend, legendMaxDepth, legendPosition }, geometries): LegendItem[] =>
+  (specs, lelele, geometries): LegendItem[] =>
     specs.flatMap((partitionSpec, i) => {
+      const { flatLegend, legendMaxDepth, legendPosition } = lelele;
```
nickofthyme, line 21: ??? 🤔 

```suggestion
  (specs, { flatLegend, legendMaxDepth, legendPosition }, geometries): LegendItem[] =>
    specs.flatMap((partitionSpec, i) => {
```

stories/waffle/1_simple.tsx
```diff
@@ -0,0 +1,45 @@
...
+        valueFormatter={(d: number) => `$${config.fillLabel.valueFormatter(Math.round(d / 1000000000))}\u00A0Bn`}
+        layers={[
+          {
+            groupByRollup: (d: Datum) => d.sitc1,
+            nodeLabel: (d: Datum) => productLookup[d].name,
+            fillLabel: { textInvertible: true },
+            shape: {
+              fillColor: (d: ShapeTreeNode) => discreteColor(colorBrewerCategoricalStark9.slice(1))(d.sortIndex),
+            },
+          },
```
nickofthyme, line 36: Do we prohibit nested partition configs for waffles or just ignore them?

nickofthyme (commented):
LGTM, made a few non-critical comments. I'll let @markov00 have a look before approving.

packages/charts/src/chart_types/partition_chart/layout/utils/legend.ts
monfera: Yea there's a bit of work to be done here 👍

packages/charts/src/chart_types/partition_chart/state/selectors/compute_legend.ts
monfera: thanks for catching this debug destructuring 😆

stories/waffle/1_simple.tsx
monfera: It's advertised as a single-layer chart, though the structure is kept identical with the other part to whole charts, which sensibly work with more layers. 

This makes me think about nesting waffles within waffles or worse, little pie charts within waffles 😄 

At some point, we may want to shed the chart-global `config.partitionLayout` and allow layer by layer specification. Though there's no current pressure toward icicle-within-treemap and such combinations

nickofthyme: Ok cool 😎. But if someone configures a nested partition waffle it won't throw and error?

packages/charts/src/chart_types/partition_chart/renderer/canvas/canvas_linear_renderers.ts
```diff
@@ -86,7 +86,9 @@ export function renderLinearPartitionCanvas2d(
 
       if (fx1 < 0 || fx0 > width) return;
 
-      const formatter = layers[depth]?.nodeLabel ?? String;
+      const layer = layers[depth - 1]; // depth === 0 corresponds to root layer (above API `layers`)
```
markov00, line 89: why picking the `depth -1` instead of `depth` as in the previous code?

packages/charts/src/chart_types/partition_chart/renderer/canvas/canvas_wrapped_renderers.ts
markov00: Yep, I agree with Nick, and this needs to be destroyed or cleaned on unmount

packages/charts/src/chart_types/partition_chart/renderer/canvas/canvas_linear_renderers.ts
monfera: It was an off-by-one bug that hasn't manifested yet in icicle/flame as it almost always has the same formatter for all layers. I noticed in context of the waffle implementation and fixed it here too

packages/charts/src/chart_types/partition_chart/renderer/canvas/canvas_wrapped_renderers.ts
monfera: It was a primitive map (one string and one number per chart), so this leak was unlikely to be ever detectable without millions of instantiation. Still this was useful feedback, as giving it another look, there's no need for the map to begin with.

39f477afc885ba00bfc5c34f5f35991c77ecfca0 simply persists the raf id in the new `animationState` property of the component object, so it goes away upon garbage collection. I think it's almost always better to link memory cleanup to scoped variables/properties, because there's no need for error prone explicit cleanup.

There'll be some DRY up work in some future commit/PR too for animation

monfera: f27c20231fac261de213916133e92941806ddcb6 removes `chartId` from the component. Using such an id, external to the component was a bit of an antipattern anyway

monfera: fe06920081f79c2f8c9690fa152108bac733998b removes animation code from the wrapped renderer as it's not yet utilized

stories/waffle/1_simple.tsx
monfera: No, it'll just be ignored currently. I haven't planned to introduce another input format to partitions, not just to avoid a needless union type or a new property in the API, but because of currently vague plans to allow per layer combinations. Random example: a treemap outer layer and a waffle inner layer. So it'd be the layer rather than, as now, the chart that'd be tagged as treemap, waffle etc.

packages/charts/src/chart_types/partition_chart/layout/utils/legend.ts
monfera: Removed a couple of lines so it's bordering on tolerable ee30f2c3a1593b233886763088c3e2fae015f90f

packages/charts/src/chart_types/partition_chart/renderer/canvas/canvas_wrapped_renderers.ts
markov00: All good for me, we should remember this PR when bringing back animation

markov00 (approved):
Code changes looks good to me!

packages/charts/src/chart_types/partition_chart/renderer/canvas/canvas_wrapped_renderers.ts
monfera: Thanks @markov00! Just to clarify, the animation that existed before (icicle/flame) remained in place (and changed so that there's no `rAF` `Map` anymore). The waffle renderer was patterned after the icicle/flame renderer but the animation wasn't (yet) adapted to waffles, so there's no loss with the removal of that bit

nickofthyme (approved):
LGTM thanks for the changes and explanations!

PR #1258, opened 2021-07-19
fix: interaction in linear/log x scales

markov00:
## Summary

The interaction with a log scale across the X axis is now fixed and works correctly. A bug that wrongly computes the inverse value from a projected mouse coordinate is fixed.

Before:

https://user-images.githubusercontent.com/1421091/126178638-e20d20db-92bf-4915-90e6-11d640643eaa.mp4

After:

https://user-images.githubusercontent.com/1421091/126178666-b01713da-d0a2-4d6b-a2c1-6ede35c78101.mp4

## Details

The `inverseWithStep` function takes care of inverting a set of screen coordinates to data coordinates.
We have two main scales: a continuous and a band/categorical one. There is a third scale type: a continuous scale that is actually formed by consecutive bands (used with time bar charts for example). This third scale type works like a continuous scale but the internal logic make use of  _bandwidth_ and _minInterval_ to correctly compute the bar width.
This is true only for bar charts, for line/area charts there is no _bandwidth_. The code was checking for the wrong parameter, instead of _bandwidth_ it was checking for the _minInterval_ that is always present and > 0.

### Checklist


rshen91 (approved):
Nice! LGTM - tested locally, thanks for the story addition 👍🏻

nickofthyme (approved):
Nice!! One liner magic! 🎱

PR #1265, opened 2021-07-21
fix: persisted color via color picker

### LINKED ISSUE #1264: persisted color is broken
**Describe the bug**
Changes in https://github.com/elastic/elastic-charts/pull/999/files#diff-b28b3197319ccdad650a47b8132e31a481b51374bc2eb6da6e0b1db3f881ecba broke colors with non-stateful color picker.

https://github.com/elastic/elastic-charts/blob/f3f8d35be850298311d01884f986aacf58b0a53b/packages/charts/src/state/chart_state.ts#L384-L393

Notice `action.color` is never passed used to update the state in the reducer above.

**Expected behaviour**
Color is persisted in state

**Screenshots**
Screen Recording 2021-07-21 at 03 52 58 PM

> Notice the color is temporarily set then removed after confirming selection

nickofthyme:
## Summary

Fixes a regression that disabled `persisted` colors when using the `legendColorPicker` prop.

## Details

This setting is more for internal use as all users of charts should persist the colors outside of memory but still a regession.

## Issues

fixes #1264

### Checklist


markov00 (approved):
Tested locally and works fine!

PR #1268, opened 2021-07-26
fix: type error with TextMetrics

nickofthyme:
## Summary

Fix bad types related to `TextMetrics`

## Details

Fixes issue in `integrations`

```node
 FAIL  integration/tests/interactions.test.ts
  ● Test suite failed to run

    packages/charts/src/common/text_utils.ts:172:18 - error TS2339: Property 'emHeightAscent' does not exist on type 'TextMetrics'.

    172   const { width, emHeightAscent, emHeightDescent } = measure(fontSize, [{ ...font, text }])[0];
                         ~~~~~~~~~~~~~~
    packages/charts/src/common/text_utils.ts:172:34 - error TS2339: Property 'emHeightDescent' does not exist on type 'TextMetrics'.

    172   const { width, emHeightAscent, emHeightDescent } = measure(fontSize, [{ ...font, text }])[0];

```

### Checklist

### DIFF
--- a/packages/charts/src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
+++ b/packages/charts/src/chart_types/partition_chart/layout/viewmodel/fill_text_layout.ts
@@ -307,9 +307,9 @@ function tryFontSize<C>(
     // model text pieces, obtaining their width at the current font size
     const measurements = measure(fontSize, boxes);
     const allMeasuredBoxes: RowBox[] = measurements.map(
-      ({ width, emHeightDescent, emHeightAscent }: TextMetrics, i: number) => ({
+      ({ width, actualBoundingBoxDescent, actualBoundingBoxAscent }: TextMetrics, i: number) => ({
         width,
-        verticalOffset: -(emHeightDescent + emHeightAscent) / 2, // meaning, `middle`,
+        verticalOffset: -(actualBoundingBoxDescent + actualBoundingBoxAscent) / 2, // meaning, `middle`,
         wordBeginning: NaN,
         ...boxes[i],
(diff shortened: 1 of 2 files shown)


monfera (commented):
Would be great to understand the nature of the [change](https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1029) esp. why some TextMetrics properties were removed while others were retained, and maybe roll in the new stuff, instead of changing our code to work around a TS evolution issue

monfera:
This PR didn't need to update mocks, though the definition text of `actualBoundingBoxDescent` differs from that of  `emHeightDescent`:

Definition: The distance in CSS pixels from the horizontal line indicated by the textBaseline attribute... 
- actualBoundingBoxDescent: ...to the bottom of the bounding rectangle of the given text; positive numbers indicating a distance going down from the given baseline.
- emHeightDescent: ...to the lowest bottom of the em squares in the inline box; positive numbers indicating that the given baseline is above the bottom of that em square. (Zero if the given baseline is the bottom of that em square.)

Possible explanations:
1. we don't have an image mock that exercises this code (combination of baselines and offsetting needs)
2. we do, but the two values are the same if there's no text stroke, only text fill (though `strokeText` uses `lineWidth` which I guess stays at the default 1; and `measureText` takes no `lineWidth` or `hasStroke` arguments so it can't know if there'll be stroke or not) 
3. the browser environment of the CI has an incompliant `measureText` implementation
4. they're guaranteed to be the same

nickofthyme:
I love your thought process @monfera.

I'll open another PR with type fixes that look at keeping the existing with `@types/web` for now.

PR #1278, opened 2021-07-30
feat(bullet): the tooltip shows up around the drawn part of the chart only

monfera:
## Summary

An overall bounding box of all graphical elements is computed, with some padding. The tooltip will only activate within this padded, rectangular capture zone.

tooltip

Closes https://github.com/elastic/elastic-charts/issues/1187

### Checklist


packages/charts/src/chart_types/goal_chart/layout/viewmodel/geoms.ts
```diff
@@ -61,8 +77,8 @@ export class Arc implements Mark {
   protected readonly x: number;
   protected readonly y: number;
   protected readonly radius: number;
-  protected readonly startAngle: number;
-  protected readonly endAngle: number;
+  protected readonly startAngle: Radian;
```
nickofthyme, line 80: 👍🏼

nickofthyme (commented):
Code changes LTGM, much more organized!

packages/charts/src/chart_types/goal_chart/renderer/canvas/connected_component.tsx
```diff
@@ -96,9 +97,12 @@ class Component extends React.Component<Props> {
...
     const { chartCenter } = geometries;
-    const x = e.clientX - box.left - chartCenter.x;
-    const y = e.clientY - box.top - chartCenter.y;
-    return picker(x, y);
+    const x = e.clientX - box.left;
+    const y = e.clientY - box.top;
+    const capture = fullBoundingBox(this.ctx, this.props.geoms);
+    if (capture.x0 <= x && x <= capture.x1 && capture.y0 <= y && y <= capture.y1) {
+      return picker(x - chartCenter.x, y - chartCenter.y);
+    }
```
markov00, line 105: This will execute the `fullBoundingBox` call at 60FPS, recalculating everytime the same capture rectangle.
Can we just reuse the one coming from the `getCaptureBoundingBox` selector?

packages/charts/src/chart_types/goal_chart/layout/viewmodel/geoms.ts
```diff
@@ -124,11 +177,30 @@ export class Text implements Mark {
...
     ctx.textAlign = this.textAlign;
     ctx.textBaseline = this.textBaseline;
     ctx.font = cssFontShorthand(this.fontShape, this.fontSize);
+  }
+
+  boundingBoxes(ctx: CanvasRenderingContext2D) {
+    if (this.text.length === 0) return [];
+
+    this.setCanvasTextState(ctx);
+    const box = ctx.measureText(this.text);
```
markov00, line 190: Can we reuse `measureText` that also brings in the font, size  etc?

packages/charts/src/chart_types/goal_chart/layout/viewmodel/geoms.ts
```diff
@@ -87,6 +106,40 @@ export class Arc implements Mark {
...
+    if (this.lineWidth === 0) return [];
+
+    const box = { x0: Infinity, y0: Infinity, x1: -Infinity, y1: -Infinity };
+
+    // instead of an analytical solution, we approximate with a GC-free grid sampler
+
+    const angleFrom: Radian = Math.floor(this.startAngle / arcBoxSamplePitch) * arcBoxSamplePitch;
+    const angleTo: Radian = Math.ceil(this.endAngle / arcBoxSamplePitch) * arcBoxSamplePitch;
+
+    for (let angle: Radian = angleFrom; angle <= angleTo; angle += arcBoxSamplePitch) {
```
markov00, line 119: We need to be sure that `angleFrom <= angleTo` if not this will end up  in a infinite loop

packages/charts/src/chart_types/goal_chart/layout/viewmodel/geoms.ts
monfera: I thought about it, but these bounding box methods are mirrors of the actual rendering code (`this.setCanvasTextState(ctx)` and precise input is fully shared), so it's already as close to font, size etc. as possible. The function `measureText` in `src/common` assumes text `Box` objects which are only used in most partition charts, due to the multirow layout, where its purpose is, adaptive font sizing and line breaks, so it's not a close fit currently. 

It'd be a good idea to eventually migrate bullet graphs to that though: the user would specify the bounding box into which the bullet/goal titles must fit. It'd be a larger, API-involving task. Also, then the bounding box would be given, because that's what the user wants to fill.

If you meant `CanvasTextBBoxCalculator`, it's a heavier weight machinery that attaches/detaches to the DOM tree, stateful, allocates resources to dispose etc., it didn't seem necessary here. Also, the method in the PR is safe when introducing some other text property (say, we add `fontStyle` or some such), because the same code is called for text styling, while the `CanvasTextBBoxCalculator` needs explicit passing of properties one by one. I'm interested in learning about the circumstances where `CanvasTextBBoxCalculator` works but a simpler approach wouldn't.

packages/charts/src/chart_types/goal_chart/layout/viewmodel/geoms.ts
monfera: Thanks, good catch!

monfera: I added aa80f8accd4fe927849cdbe5034e06411c13dd6a which provides for these:
- accurate circle traversal in light of normal or reversed `angleFrom`->`angleTo relation and `counterclockwise`, as there are two alternative directions to take for arc drawing, given two points on a circle (we compute what `ctx.arc` also computes)
- angleTo < angleFrom (as @markov's PR feedback)
- some or both angles are negative (could mess with the nice incrementing of the `for` loop)

packages/charts/src/chart_types/goal_chart/renderer/canvas/connected_component.tsx
monfera: Thanks, good catch, done in a54735c1ad22a6dbccd53f13f084d8e3e87b31e0

rshen91 (approved):
LGTM! Nice refactoring and story addition, tested locally.

nickofthyme (approved):
LGTM!

PR #1281, opened 2021-08-02
fix(bar): filter zero-height bars from rendering

### LINKED ISSUE #1279: A Zero bar with rectBorder renders a small bar
**Describe the bug**
When rendering a bar chart with a zero data point and the rectBorder is set to visible with a configured width, the bar is shown.

**To Reproduce**
1. Go to https://codesandbox.io/s/cold-glitter-y6k5i
2. See the bar at category B should be 0 and instead shows up

**Expected behaviour**
Independently from the bar border size, if the border should not protrude out of the current bar and renders outside the bar size, in particular with bar with zero value.

Screenshot 2021-08-02 at 12 52 38

 - Elastic Charts: master

rshen91:
## Summary
The rect border does not render for bar charts with values smaller than the strokeWidth (including zero y-values). 

## Details
- [x] does not render a bar if the height is 0 (f425d13)
- [x] Independent from the bar border size, the border should not protrude out of the current bar and renders outside the bar size. In the example below, the border around `{ x: 'B', y: 20, val: 1222 },` should not exist since the y value is so small (d42a3f2)
```
<Chart renderer="canvas">
      <Settings theme={useBaseTheme()} />
      <Axis id="count" title="count" position={Position.Left} />
      <Axis id="x" title="goods" position={Position.Bottom} />
      <BarSeries
        id="bars"
        name="amount"
        xAccessor="x"
        yAccessors={['y']}
        barSeriesStyle={{
          rectBorder: {
            visible: true,
            strokeWidth: 10,
            stroke: 'black',
          },
        }}
        data={[
          { x: 'A', y: 0, val: 1222 },
          { x: 'B', y: 20, val: 1222 },
          { x: 'C', y: 750, val: 1222 },
          { x: 'D', y: 854, val: 1222 },
        ]}
      />
    </Chart>
```

## Issues
Fixes #1279

### Checklist


packages/charts/src/chart_types/xy_chart/renderer/canvas/styles/bar.ts
```diff
@@ -47,7 +48,10 @@ export function buildBarStyles(
   const strokeColor = stringToRGB(getColorFromVariant(baseColor, themeRectBorderStyle.stroke), strokeOpacity);
   const stroke: Stroke = {
     color: strokeColor,
-    width: themeRectBorderStyle.visible ? themeRectBorderStyle.strokeWidth : 0,
+    width:
+      themeRectBorderStyle.visible && rect.height > themeRectBorderStyle.strokeWidth
+        ? themeRectBorderStyle.strokeWidth
+        : 0,
```
nickofthyme, line 54: This is a tough choice, I would think to be consistent with the border something like...

image

But then the color is completely unknown for that bar so this...

image

would be better in that case. I think this is fine just wanted to make a mention of it. In any case this is hopefully a extreme edge case.

nickofthyme (approved):
LGTM, tested locally.

packages/charts/src/chart_types/xy_chart/renderer/canvas/styles/bar.ts
rshen91: I agree - hopefully this is an extreme edge case.

Issue #1245, opened 2021-07-09
[Legend] Improved readability on long legend values


ghudgins:
_Problem with long values_
**When** my data is long and varies mostly at the end of the value
**I need** to be able to see the ends of long values
**So I can** see the differences in my legend values from the front _and_ the back of the values

**Describe the solution you'd like**
Many options:
 - ~Option that lets you choose to middle `...` instead of at the end (as it is today)?~
 - Selected approach: Option multi-line legends

There are other solutions we shouldn't look into in the scope of this issue: https://github.com/elastic/elastic-charts/issues/963 Focus on a responsive solution (we should still do the resize issue too as that satisfies interactivity cases...just not as a solution to this issue)

**Describe alternatives you've considered**
There are none, makes lines with multiple series hard to read

**Additional context**
Screenshots on this issue: https://github.com/elastic/kibana/issues/41418
TSVB related issue: https://github.com/elastic/kibana/issues/102327
Related problem on dropdowns: https://github.com/elastic/kibana/issues/103701

**Checklist**

Delete any items that are not applicable to this feature request.

- [x] this request is checked against already exist requests
- [x] every related Kibana issue is listed under `Kibana Cross Issues` list
- [x] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

ghudgins:
we selected multi-line legends as the approach because it 1) works really well with resizing #963 2) it works very well on horizontal legends (right / left) and 3) it enables future table legends. We also found issues with middle `...` - the most notable was the copy/paste not working correctly which was an original requirement

nickofthyme:
Testing for center `...` solution...

Option 1 - [demo](https://codesandbox.io/s/legend-center-truncate-lhtbi)

separates legend labels at an index

![Screen Recording 2021-07-14 at 01 15 54 PM](https://user-images.githubusercontent.com/19007109/125845850-cc873f42-1e63-463b-b786-aea1cdfd7692.gif)

Causes line breaks when copying.

```
cloudfla
re.com - direct-cdn - y2
cloudfla
re.com - indirect-cdn - y1
```

Option 2 - [demo](https://codesandbox.io/s/legend-center-truncate-coping-6dgcg?file=/src/styles.css)

The same as option 1 with separating legend labels at an arbitrary index and uses `user-select: none` on split strings with a separate adjacent full string element. Thus the copied text is the full invisible text.

[...]

PR #1285, opened 2021-08-04
feat(legend): multiline labels

nickofthyme:
## Summary

The `labelOptions.maxLines` property is now available in on `Theme.legend` type to set the max number of lines a legend label can take-up before it is truncated.

Screen Recording 2021-08-05 at 11 42 16 AM

## Details

When enabled the multiline mode will trigger items to be align from the top (i.e. `flex-start`). These style changes have no affect on existing single line legend labels.

https://user-images.githubusercontent.com/19007109/128388710-6852229b-d48c-4e7e-833d-efb81d27c0b3.mp4

## Issues

#1245

### Checklist


packages/charts/src/components/legend/_legend_item.scss
```diff
@@ -43,6 +48,7 @@ $legendItemHeight: #{$euiFontSizeXS * $euiLineHeight};
     display: flex;
     justify-content: center;
     align-items: center;
+    height: $legendItemHeight;
```
nickofthyme, line 51: makes the focus size of the action the same as the color dot

packages/charts/src/components/legend/label.tsx
```diff
@@ -9,26 +9,30 @@
...
     'echLegendItem__label--clickable': Boolean(onClick),
+    'echLegendItem__label--multiline': options?.multiline,
   });
 
   return isToggleable ? (
     <button
       type="button"
       className={labelClassNames}
-      title={label}
+      title={options?.multiline ? '' : label} // full text already visible
```
nickofthyme, line 35: I don't think the title is necessary if the full label is visible

nickofthyme:
Improves inside legends ability to adapt to small widths. The inside legend styles, still needs some help (not in this pr).

image

monfera (commented):
Looks great so far! Should we breakup long words? Or maybe there's some upper limit to row count at play here:

image

nickofthyme:
> Should we breakup long words?

We could could add another option for [`overflow-wrap`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap)

> Or maybe there's some upper limit to row count at play here

We could add a [`maxLines`](https://api.flutter.dev/flutter/widgets/Text/maxLines.html) option like flutter does. I'm not sure if that's needed for the first pass, what do you think?

packages/charts/api/charts.api.md
```diff
@@ -1147,6 +1147,11 @@ export interface LegendColorPickerProps {
 // @public (undocumented)
 export type LegendItemListener = (series: SeriesIdentifier[]) => void;
 
+// @public (undocumented)
+export interface LegendLabelOptions {
+    multiline: boolean;
```
monfera, line 1152: In partition chart, not for the legend, we use a `maxRowCount` property, which might be useful here and be a slightly more uniform API. So, a value of 1 would be the default, and it could be whatever number incl. Infinity if unbounded. While it's not integral to the task, it feels useful to set some sensible row count, and with a `maxRowCount` there's no need to break the API in the future, or introduce an additional property, which is only applicable with a `multiline: true` value. If a row count is deemed useful, we can preemptively cut down on ifs, ternaries and union types 😸

nickofthyme: Yeah so you are thinking change `multiline` option to `maxRowCount`?

packages/charts/src/components/legend/legend_item.tsx
```diff
@@ -45,6 +46,7 @@ export interface LegendItemProps {
   positionConfig: LegendPositionConfig;
   extraValues: Map<string, LegendItemExtraValues>;
   showExtra: boolean;
+  labelOptions?: LegendLabelOptions;
```
monfera, line 49: Nit: could it be relatively easily solved such that it's only the user facing config API where this is optional, but we can rely on its presence everywhere downstream? Ie. one question mark only

nickofthyme: You saying `labelOptions: LegendLabelOptions;`?

packages/charts/src/components/legend/label.tsx
```diff
@@ -9,26 +9,30 @@
...
 import React, { MouseEventHandler } from 'react';
 
+import { LegendLabelOptions } from '../../utils/themes/theme';
+
 interface LabelProps {
   label: string;
   isSeriesHidden?: boolean;
   isToggleable?: boolean;
   onClick?: MouseEventHandler;
+  options?: LegendLabelOptions;
```
monfera, line 19: Minor: same here, maybe all such optionality like can go away except for the single public API enty point. I admit to having done plenty of this in code I wrote, more recently thinking that it'd be more economical to reify user input right away, and then avoid having to worry about partials, optionals etc. 

We can think about removing question marks from non-`@public` types, TS is great for safely allowing such a transition, in the meantime we can consider avoiding them in new code

integration/tests/legend_stories.test.ts
```diff
@@ -160,25 +172,40 @@ describe('Legend stories', () => {
       },
     );
   });
+
   describe('Legend inside chart', () => {
+    const getPositionalUrl = (p1: string, p2: string, others: string = '') =>
+      `http://localhost:9001/?path=/story/legend--inside-chart&knob-vAlign_Legend=${p1}&knob-hAlign_Legend=${p2}${others}`;
```
nickofthyme, line 178: fixes knob to align with correct `vAlign` and `hAlign` properties.

monfera (approved):
Good improvement 🎉  Approving it as there's no blocker.

Optional notes:
- if possible, tweak the CSS such that what's hidden now in case of very long, unbroken words would show up, but no need to invest too much time into it, as the user provided formatter function can detect and apply delimiters eg. at every 10 characters (so the user can solve cornercase need)
- prefer using `maxRowCount` as it's numerical and reuses a name with the like meaning instead of the boolean, even if we initially support `1` and some other value eg. `Infinity` (if CSS reasonably supports a row count, we can use it now or in some future PR, or JS if really needed in the future)
- prefer nullish coalescing of new API properties right at the entrance, so that the entire implementation has a fixed type to work with

packages/charts/src/components/legend/legend_item.tsx
monfera: Yes, it's an internal type, by this time the reification (augmentation with defaults etc.) of the user input should've happened (if we have shared buy-in of how it often simplifies the implementation and prevents us from possibly handling the non-specified nullish case at multiple places, so also a DRY principle)

nickofthyme:
Replaced `multiline` option with `maxLines` option in d8be979.

- `0` - true multiline, unconstrained.
- `1` - as is today simple `text-overflow`, `nowrap`, etc.
- `>1` - limit lines of text to a finite value.

https://user-images.githubusercontent.com/19007109/128429557-0fd6e736-613d-4da9-b005-270bc73c498d.mp4

packages/charts/src/components/legend/label.tsx
nickofthyme: Done here d8be979

packages/charts/src/components/legend/legend_item.tsx
nickofthyme: Done here d8be979

packages/charts/api/charts.api.md
nickofthyme: Done here d8be979

integration/tests/interactions.test.ts
```diff
@@ -366,6 +366,7 @@ describe('Interactions', () => {
         { left: 282, top: 80 },
         {
           screenshotSelector: '#story-root',
+          delay: 1000,
```
nickofthyme, line 369: This story was misbehaving, need to look into more later.

packages/charts/src/components/legend/_legend_item.scss
```diff
@@ -6,12 +6,17 @@ $legendItemHeight: #{$euiFontSizeXS * $euiLineHeight};
   display: flex;
   flex-wrap: nowrap;
   justify-content: space-between;
-  align-items: center;
+  align-items: flex-start;
```
nickofthyme, line 9: This caused a slight diff in the inside legends but I think it's acceptable.

packages/charts/src/components/legend/label.tsx
```diff
@@ -7,37 +7,60 @@
...
+    // https://stackoverflow.com/questions/68673034/webkit-line-clamp-does-not-apply-to-buttons
+    <div
+      role="button"
+      tabIndex={0}
       className={labelClassNames}
-      title={label}
+      title={title}
       onClick={onClick}
+      onKeyDown={onKeyDown}
+      aria-pressed={isSeriesHidden}
```
nickofthyme, line 54: Converted to button due to [this](https://stackoverflow.com/questions/68673034/webkit-line-clamp-does-not-apply-to-buttons), added all required ARIA props to make it seem like a button.

packages/charts/src/components/legend/legend_item.tsx
```diff
@@ -189,19 +200,22 @@ export class LegendListItem extends Component<LegendItemProps, LegendItemState>
...
+            <ItemColor
+              ref={this.colorRef}
+              color={color}
+              seriesName={label}
+              isSeriesHidden={isSeriesHidden}
+              hasColorPicker={hasColorPicker}
+              onClick={this.handleColorClick(hasColorPicker)}
+              pointStyle={pointStyle}
+            />
+          </div>
```
nickofthyme, line 213: This is to prevent jitter when using `EuiWrappingPopover` to display the color picker. Before this was picked using `:first-of-type` but this is more strict though a little more verbose.

nickofthyme:
Linting job hung up for 30 minutes for some reason, not cancelable. Ran locally with no errors so gonna merge.

Issue #1202, opened 2021-06-14
Improve texture coloring on bar/area


markov00:
**Is your feature request related to a problem? Please describe.**
A texture or a pattern should always start from the same point on the geometry.
When resizing, the texture/pattern should not move around.  

![Jun-14-2021 10-34-29](https://user-images.githubusercontent.com/1421091/121863706-6c8df680-ccfc-11eb-8d53-0ba5bad60ee1.gif)

**Describe the solution you'd like**
The texture should probably always start from the baseline, or anyway always start at the same point in the geometry.
In this case, for example, the dot pattern on every bar starts at a different x position. This can reduce the readability because the pattern effect introduces artifacts on how the bars are rendered.
<img width="266" alt="Screenshot 2021-06-14 at 10 41 14" src="https://user-images.githubusercontent.com/1421091/121864322-12d9fc00-ccfd-11eb-9e8a-209b3b782e8d.png">

We can also discuss if the texture should use:
-  screen pixels: this means that the texture doesn't rescale with the chart size, the texture is always the same but the image can be clipped
- projection size: the pattern is applied based on one dimension of the geometry (height or width) and is rescaled based on that with the chart size.

**Describe alternatives you've considered**
n/a

**Additional context**
Needs as follow up of #1138  

**Kibana Cross Issues**
Not yet used in Kibana

monfera:
While this issue auto-closed due to mention of the partial solution brought by https://github.com/elastic/elastic-charts/pull/1286 there is still much opportunity to refine the size responsiveness of pattern fill, referenced in the further improvements [meta issue #1303](https://github.com/elastic/elastic-charts/issues/1303). Once work continues, this issue can be reopened or a new one can be created. I've some preference for the latter, as, during the review process, we discussed specific approaches toward future improvements, such as options for integral (full, unclipped) tile counts for consistent aesthetic of the bar thickness coverage, and roughly constant ratios (eg. approx. tile pixel size) to go with it; and the attention we need to give to negative bars

nickofthyme:
I think the idea of this issue was to make the patterns independent such that changing the size doesn't shift the values, as if it is masking a picture in the background, but that each follows their respective bar.

PR #1286, opened 2021-08-05
refactor(cartesian): timeslip preparation

monfera:
## Summary

Non-functional code cleanup, code removal and type fixes. Also removes the large pixel dimensions with texture use. 

### BREAKING CHANGE
-  `TextStyle.fontStyle` is no longer a `string`, it's the more specific `FontStyle`
- For symmetry, `fontStyle` in word cloud is also switching from `string` to `FontStyle`
- Certain text configurations included both `fill` and `textColor` for the text color; `fill` got removed, because `textColor` is part of the public `Font` type, and because `textColor` has clearer meaning than `fill`. Yet, some of the code used the `fill` property and/or made the `fill` property also mandatory. So, userland code needs to remove some `fill` property, and might need to ensure that the correct value is going into `textColor`
- `getRadians` got unpublished

## Details

Reviewers,
- commit 848e188d866be40c79bc545dbf4cac81c8f347ce contains the breaking change of switching to `FontStyle`
- commit 80c743fa879003a986a2e78d0e297229e712edf2 contains the breaking change of phasing out `fill`
- commit b544c0c9364b1e14c3e4816a6de120f92ba35e8f unpublishes `getRadian` (no apparent change to API MD)

### Checklist


monfera:
Jenkins test please

monfera:
jenkins test this

monfera:
jenkins test this

packages/charts/src/chart_types/xy_chart/renderer/canvas/panels/panel_title.ts
```diff
@@ -114,20 +113,14 @@ function renderHorizontalTitle(ctx: CanvasRenderingContext2D, props: PanelTitleP
...
-  fontSize,
-}: AxisStyle['axisTitle'] | AxisStyle['axisPanelTitle']): TextFont {
+export function getFontStyle({ fontFamily, fontStyle, fill, fontSize }: TextStyle): TextFont {
   return {
     fontFamily,
     fontVariant: 'normal',
-    fontStyle: fontStyle ? (fontStyle as FontStyle) : 'normal',
-    fontWeight: 'normal',
+    fontStyle: fontStyle ?? 'normal',
+    fontWeight: 'bold',
```
monfera, line 121: We could introduce an API option for font weight instead, then this `bold` can go back into the theme

packages/charts/src/utils/themes/theme.test.ts
```diff
@@ -49,10 +49,10 @@ describe('Theme', () => {
         opacity: 1,
       };
 
-      const customDetailsConfig = {
+      const customDetailsConfig: TextStyle = {
         fontSize: 50,
         fontFamily: 'custom-font-family',
-        fontStyle: 'custom-font-style',
+        fontStyle: 'italic',
```
monfera, line 55: If there's a real need for `'custom-font-style'` we could introduce a `fontShorthand` or `fontOverride` property

packages/charts/src/utils/common.tsx
```diff
@@ -178,12 +179,8 @@ export function getColorFromVariant(seriesColor: Color, color?: Color | ColorVar
...
 }
 
-/**
- * Converts degree to radians
- * @param angle - in degrees
- * @public
- */
-export const getRadians = (angle: number) => (angle * Math.PI) / 180;
+/** @internal */
+export const degToRad = (angle: Degrees): Radian => (angle / 180) * Math.PI;
```
monfera, line 183: Unpublishing it as we don't currently expose pure utility functions in our API, and it wouldn't be a good fit for our library because the related areas take in angles in degrees, so the user needs not convert to radians

packages/charts/src/renderers/canvas/index.ts
```diff
@@ -98,19 +96,3 @@ export function withClipRanges(
...
-  origin: Point,
-  rotation: number = 0,
-  fn: (ctx: CanvasRenderingContext2D) => void,
-) {
-  withContext(ctx, (ctx) => {
-    const { x, y } = origin;
-    ctx.translate(x, y);
-    ctx.rotate(getRadians(rotation));
-    ctx.translate(-x, -y);
-    fn(ctx);
```
monfera, line 114: Removed this function, b/c a single place used it, and it immediately reverted the last (negative) translate here

packages/charts/src/chart_types/xy_chart/renderer/canvas/primitives/rect.ts
```diff
@@ -7,84 +7,34 @@
...
-        const { offset } = texture;
-
-        if (offset && offset.global) ctx.translate(offset?.x ?? 0, offset?.y ?? 0);
-        if (rotation) ctx.rotate(rotation);
-        if (offset && !offset.global) ctx.translate(offset?.x ?? 0, offset?.y ?? 0);
-
-        ctx.fillStyle = texture.pattern;
-
-        // Use oversized rect to fill rotation/offset beyond path
-        const rotationRectFillSize = ctx.canvas.clientWidth * ctx.canvas.clientHeight;
```
monfera, line 53: > const rotationRectFillSize = ctx.canvas.clientWidth * ctx.canvas.clientHeight;
> ctx.translate(-rotationRectFillSize / 2, -rotationRectFillSize / 2);
> ctx.fillRect(0, 0, rotationRectFillSize, rotationRectFillSize);

This could lead to `ctx.translate` and `ctx.fillRect` operating with multimillion by multimillion pixel sizes here and in area `path.ts`, so I pre-transformed the pattern itself upon its single creation. 

DRY principle too 😇 

It's more efficient to have moved the transforms upstream too, about a dozen steps _per geom_ could be erased, including an expensive `withContext`

The mocks changed but that's OK: the rather random previous offsetting (due to placing the start a million pixels away, rotating etc.) is now different, and more predictable.

packages/charts/src/chart_types/xy_chart/renderer/canvas/primitives/rect.ts
```diff
@@ -7,84 +7,34 @@
...
-
-  if (stroke && stroke.width > 0.001) {
-    const borderOffset = !disableBoardOffset && stroke && stroke.width > 0.001 ? stroke.width / 2 : 0;
-    const x = rect.x + borderOffset;
-    const y = rect.y + borderOffset;
-    const width = rect.width - borderOffset * 2;
-    const height = rect.height - borderOffset * 2;
-
+  const borderWidth = !disableBorderOffset && stroke.width >= MIN_STROKE_WIDTH ? stroke.width : 0;
+  if (stroke.width >= MIN_STROKE_WIDTH) {
```
monfera, line 22: A bunch of 0.001's got replaced with MIN_STROKE_WIDTH

packages/charts/src/chart_types/xy_chart/renderer/canvas/primitives/rect.ts
```diff
@@ -7,84 +7,34 @@
...
+  const borderOffset = !disableBorderOffset && stroke.width >= MIN_STROKE_WIDTH ? stroke.width : 0;
+  if (stroke.width >= MIN_STROKE_WIDTH) {
     ctx.strokeStyle = RGBtoString(stroke.color);
     ctx.lineWidth = stroke.width;
-    drawRect(ctx, { x, y, width, height });
-    if (stroke.dash) {
-      ctx.setLineDash(stroke.dash);
-    } else {
-      // Setting linecap with dash causes solid line
-      ctx.lineCap = 'square';
```
monfera, line 74: I need to check and perhaps restore the `lineCap` setting, the `square` linecap may have had a purpose, and may not have been set globally or higher up in the call tree. The `round` linecap has some advantages, eg. corners with higher line widths look nicer, more continuous. What was the motivation for the `square`?

packages/charts/src/chart_types/xy_chart/state/utils/utils.ts
```diff
@@ -520,31 +511,10 @@ function renderGeometries(
...
-      x: width,
-      y: height,
-      rotate: 180,
-    };
-  }
   return {
-    x: 0,
-    y: 0,
-    rotate: 0,
+    x: chartRotation === 90 || chartRotation === 180 ? width : 0,
```
rshen91, line 515: Nice clean up!

rshen91 (commented):
Thanks for doing this clean up. Tested locally and changes look great!

rshen91 (approved):

packages/charts/api/charts.api.md
```diff
@@ -745,6 +745,12 @@ export type FitConfig = {
     endValue?: number | 'nearest';
 };
 
+// @public (undocumented)
+export const FONT_STYLES: readonly ["normal", "italic", "oblique", "inherit", "initial", "unset"];
+
+// @public (undocumented)
+export type FontStyle = typeof FONT_STYLES[number];
+
```
nickofthyme, line 753: nit: Can we make this a true enum if we are going to export it?

For example...

```ts
/**
 * Available chart types
 * @public
 */
export const ChartType = Object.freeze({
  Global: 'global' as const,
  Goal: 'goal' as const,
  Partition: 'partition' as const,
  XYAxis: 'xy_axis' as const,
  Heatmap: 'heatmap' as const,
  Wordcloud: 'wordcloud' as const,
});
/** @public */
export type ChartType = $Values<typeof ChartType>;
```

packages/charts/src/chart_types/xy_chart/renderer/canvas/primitives/rect.ts
nickofthyme: How will this affect our ability to do this ➡️ https://github.com/elastic/elastic-charts/issues/1202?

packages/charts/src/chart_types/xy_chart/renderer/canvas/primitives/text.ts
```diff
@@ -26,30 +27,28 @@ export function renderText(
   origin: Point,
   text: string,
   font: TextFont,
-  degree: number = 0,
-  translation?: Partial<Point>,
+  angle: Degrees = 0,
```
nickofthyme, line 30: 👍🏼

packages/charts/src/chart_types/xy_chart/state/utils/utils.ts
nickofthyme: `O(n)` 🎉

packages/charts/src/chart_types/xy_chart/utils/texture.ts
```diff
@@ -76,7 +76,16 @@ function createPattern(
     if (fill) pCtx.fill(path);
   }
 
-  return ctx.createPattern(patternCanvas, 'repeat') ?? undefined;
+  const pattern = ctx.createPattern(patternCanvas, 'repeat')!; // HTMLCanvasElement always yields a CanvasPattern anyway
```
nickofthyme, line 79: So this will never ever `throw`?

nickofthyme (commented):
LGTM, the vr tests do not like your changes 🤣 . Idk what that annotation story always triggers a small diff for all.

packages/charts/src/chart_types/xy_chart/renderer/canvas/primitives/rect.ts
monfera: Great cross-reference! I missed this issue. 

This PR improves the situation in that the old method uses an offset of an almost infinitely big and unpredictable `clientWidth * clientHeight` possibly with rotation and/or scaling at the point of pattern application, while the new method performs the scaling/rotation at the pattern level, which is therefore stable, and the pattern application deterministically pastes the pattern, guaranteed by the standard:

> Place a copy of the image on the bitmap, anchored such that its top left corner is at the origin of the coordinate space, with one coordinate space unit per CSS pixel of the image, then place repeated copies of this image horizontally to the left and right, if the repetition behavior is "repeat-x", or vertically up and down, if the repetition behavior is "repeat-y", or in all four directions all over the bitmap, if the repetition behavior is "repeat".

So, for the cases when a top left anchoring of the pattern to the origin of the Cartesian area is sufficient, I think it should be a solution (it may be, for area charts and maybe most histograms due to no gap).  If it's not sufficient, eg. the zero base of each bar should be held fixed (whether the bar is positive or negative), then one of these is needed: 

1. a new `CanvasPattern` object should be instantiated possibly for each shape with its own `setTransfer`s (necessitating the propagation of the `HTMLCanvasElement` object into the rect/area renderers)
2. the existing `CanvasPattern` object's transform matrix needs to change (likely more efficient) per element, so some of the `setTransform`s then need to move back to the shape renderer and done element per element
3. (my fave, if we can pull it off) the shape space transforms would guarantee the proper anchor point, eg. positive vertical bars would undergo a `ctx.scale(1, -1)` so that the zero base of the bar, rather than the top of the bar is considered origin for the purpose of applying the pattern (though there are some wrinkles)

Currently, the area chart in this PR is doing something that (to me) looks sensible:
area

The bar chart, due to the simplification (reduction) of per bar `transforms`, is doing something similar, which may or may not be the best, depending on the nature of the pattern:
bar

- If the pattern width is _small_ relative to the vertical bar thickness, then I think it's OK that each vertical bar doesn't have the same pixels on their 1st, 2nd etc. pixel columns, but it'd also look good if each bar started consistently.
- If the pattern width is _large_ relative to the vertical bar thickness, then starting at the same pixel column will yield an unbalanced pattern application, eg. only the left side of the pattern is ever used 
- Histograms are more like area charts than general bar charts, because (I think) the discontinuity caused by starting each shape consistently is worse than each bar pattern starting at a different place

Alternatives: 
1. include the referenced issue in the scope of this PR so it solves it, at the expense of leaving this PR open for longer
2. do it as a follow-up PR, as this PR should already improve on the situation

I'm somewhat leaning toward 2 (let's do it as separate PR), as the referenced issue would need a bit more clarification as to what's the best option, design and configuration wise. It feels like we should allow a toggle for per-chart or per-element flooding (the current PR is per-chart flooding). And other options would customize the anchor point (origin) for the pattern pasting, with defaults, eg. for a vertical bar, x: either left edge or bar center; y: either zero base or middle of bar or top of bar

packages/charts/api/charts.api.md
monfera: Yes, can do, and iirc it had some benefit besides consistency, can you remind me please? No worries if there's no technical benefit though, even if fewer lines and less repetition have some appeal

packages/charts/src/chart_types/xy_chart/utils/texture.ts
monfera: It won't throw, as the spec says `ctx.createPattern` invoked with `HTMLCanvasElement` (which `patternCanvas` is guaranteed to be) has no `null` case. 

Details:

It'd throw if `pattern` ended up being `null` (there are other throw conditioins but TS guards against most(*) of those already; example: the `"repeat"` literal is OK here).

The [spec](https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-createpattern-dev) for `createPattern` says:

The only potential for getting a `null` is point 2:
> 2. If _usability_ is bad, then return null

It's the preceding point that references _usability_:
> 1. Let _usability_ be the result of [checking the usability of image](https://html.spec.whatwg.org/multipage/canvas.html#check-the-usability-of-the-image-argument)

The "switch" at the link has no "bad" usability case for `HTMLCanvasElement`, so there can't be a `null`, so the `!` is safe and `pattern` will never be `null`.

(*) If my 2nd reading prompted by your question is correct, then a zero dimension canvas would lead to a DOM exception. TS doesn't complain about this opportunity, nor does current `master` wrap it in `throw`/`catch` so this aspect is not made worse by the PR. Still, I'll check if we defend against a zero-dimension canvas upstream, eg. by not creating such a canvas, or by applying a minimum 1px by 1px dimension. The best combo is,avoiding a `throw` condition so that a `throw/catch` wrapper is not needed

packages/charts/api/charts.api.md
markov00: - we can change the value of one enum item without breaking the APIs
- we don't have to change every line that uses directly the value if we want to change that value

In the current case, I don't see the reason to rename the values as they are standardized so the enum here can be used just for consistency

packages/charts/src/chart_types/xy_chart/renderer/canvas/panels/panel_title.ts
markov00: great idea

packages/charts/src/chart_types/xy_chart/renderer/canvas/primitives/rect.ts
markov00: I'm totally up to improve this in a follow-up PR.
It definitely needs a bit more research to understand the various cases and design depending on the configuration.

markov00: We could also add an `autofill` option: it will fill the entire width of the geometry with the pattern without truncation.

packages/charts/api/charts.api.md
monfera: Let's discuss and consider it for a subsequent PR, as all these are not object "enum" and they come from DOM standards, so the likelihood of name changing is next to zero, and the value in brevity may or may not outweigh a natural or enforced convention. There are already quite a few noise lines with all the `/** @public */` 
```
const FONT_WEIGHTS_NUMERIC = [100, 200, 300, 400, 500, 600, 700, 800, 900];
const FONT_WEIGHTS_ALPHA = ['normal', 'bold', 'lighter', 'bolder', 'inherit', 'initial', 'unset'];

/** @public */
export const FONT_WEIGHTS = Object.freeze([...FONT_WEIGHTS_NUMERIC, ...FONT_WEIGHTS_ALPHA] as const);
/** @public */
export const FONT_VARIANTS = Object.freeze(['normal', 'small-caps'] as const);
/** @public */
export type FontVariant = typeof FONT_VARIANTS[number];
/** @public */
export type FontWeight = typeof FONT_WEIGHTS[number];
/** @public */
export const FONT_STYLES = Object.freeze(['normal', 'italic', 'oblique', 'inherit', 'initial', 'unset'] as const);
/** @public */
export type FontStyle = typeof FONT_STYLES[number];
/** @public */
export type PartialFont = Partial<Font>;
/** @public */
export const TEXT_ALIGNS = Object.freeze(['start', 'end', 'left', 'right', 'center'] as const);
/** @public */
export type TextAlign = typeof TEXT_ALIGNS[number];
/** @public */
export type TextBaseline = typeof TEXT_BASELINE[number];
```

monfera:
jenkins test this

monfera:
jenkins test this

packages/charts/api/charts.api.md
nickofthyme: Postpone to later date when able to migrate to `const enums`

packages/charts/src/chart_types/xy_chart/utils/texture.ts
nickofthyme: > zero dimension canvas would lead to a DOM exception

Good to know!

nickofthyme (approved):
Latest changes LGTM.

packages/charts/src/chart_types/xy_chart/utils/texture.ts
monfera: Added this and a bunch of other things here into https://github.com/elastic/elastic-charts/issues/1303

PR #1287, opened 2021-08-06
build: fix storybook build

nickofthyme:
## Summary

Fix storybook build

## Details

Replaces [`storybook-addon-themes`](https://github.com/tonai/storybook-addon-themes) with custom [`storybook-addon-theme-toggle`](https://github.com/nickofthyme/storybook-addon-theme-toggle). Also replaces [`@storybook/addons-backgrounds`](https://storybook.js.org/docs/react/essentials/backgrounds) with custom [`storybook-addon-background-toggle`](https://github.com/nickofthyme/storybook-addon-background-toggle).

The reasoning is that changes make to forks of each were problematic. Also both were not designed very great especially for our use case. The theme addon also has a bug only for static builds where it tries to access the iframe to change styles, which was throwing an error that broke everything. This is fixed by using a decorator to change the dom that is already within the `iframe` context.

Storysource addon is disabled for now until it allows bucketed `___.stories.tsx` files to be used.


nickofthyme:
Jenkins test this please

PR #1289, opened 2021-08-06
ci: fix package versioning script

nickofthyme:
## Summary

Fix package versioning script

## Details
The path to the `pkgRoot` in the `@semantic-release/npm` config should be a full resolved path as shown below.

https://github.com/semantic-release/npm/blob/d5cab62ee88e38f2b866feda5c7af8fce95c0173/lib/prepare.js#L5-L6


nickofthyme:
didn't work 😢

PR #1293, opened 2021-08-10
feat(annotations): add onClickHandler for annotations

### LINKED ISSUE #1211: [Charts] Annotations should have click handler
When using the chart annotations, I am unable to hook into interactions with the annotation. 

I would like to have an event handler (preferably click) that I can pass in so that I can easily trigger some other action from a click on the annotation. The handler I pass in would probably expect the annotation data as a parameter.

No alternatives I could find for the use case I'd like. 

**Kibana Cross Issues**

**Target Release**
Would be great to get this in soon, if possible. 7.14 would be great but mindful it's a bit of a squeeze so 7.15 would be great, also. Thank you!

**Checklist**

rshen91:
## Summary

The `onAnnotationClick` prop is now available in the `Settings` component to listen to click events on `LineAnnotation` (marker only) and `RectAnnnotation`.

## Details

Example code snippet from the new storybook example in annotations > rects  > click-handle-rect-annotation:

``` tsx
  const onAnnotationClick = action('onAnnotationClick');

  return (
    <Chart>
      <Settings baseTheme={useBaseTheme()} onAnnotationClick={onAnnotationClick} />
      <RectAnnotation
        id="rect1"
        dataValues={[
          {
            coordinates: {
              x0: 0,
              x1: 1,
              y0: 0,
              y1: 4,
            },
            details: 'details about this annotation',
          },
        ]}
        style={{ fill: 'red' }}
      />
      <RectAnnotation
        id="rect2"
        dataValues={[
          {
            coordinates: {
              x0: 1,
              x1: 2,
              y0: 1,
              y1: 5,
            },
            details: 'details about this other annotation',
          },
        ]}
        style={{ fill: 'blue' }}
      />
      <LineAnnotation
        id="line_annotation"
        domainType={AnnotationDomainType.XDomain}
        dataValues={[{ dataValue: 2, details: `detail-${0}` }]}
        marker={<Icon type="alert" />}
        markerPosition={Position.Top}
      />
...
```

## Issues
Closes #1211 request from the ML team

### Checklist


monfera:
What's the approach for the mouse capture area padding? For example, the actual line of a line annotation is very thin, and Fitts law needs significant meat for being able to reliably select. Then again, there may be some tightly spaced lines, in which case the generous padding is constrained so that the midline between two nearby parallel lines determines if this or that line gets selected (otherwise the capture zone of one line can occlude the other line). So, a one-dimensional Voronoi version of [this 2D example](https://bl.ocks.org/Kcnarf/6d5ace3aa9cc1a313d72b810388d1003)—or the Voronoi picker @nickofthyme added to charts—at least conceptually, and the 1D Voronoi logic can be extracted into a `common` directory for later reuse.

Capture zone padding gets more complicated with mixed or overlapping annotations: both horizontal and vertical lines; lines over rectangles; overlapping rectangles; annotation over the text of some other annotation; annotation over bars, data point markers or other interactive elements. 

So, just curious about what we have as current goals for this, maybe on the original issue, or on a separate issue that goes into this kind of detail (or discussed as part of the PR)

rshen91:
> What's the approach for the mouse capture area padding? For example, the actual line of a line annotation is very thin, and Fitts law needs significant meat for being able to reliably select. Then again, there may be some tightly spaced lines, in which case the generous padding is constrained so that the midline between two nearby parallel lines determines if this or that line gets selected (otherwise the capture zone of one line can occlude the other line). So, a one-dimensional Voronoi version of [this 2D example](https://bl.ocks.org/Kcnarf/6d5ace3aa9cc1a313d72b810388d1003)—or the Voronoi picker @nickofthyme added to charts—at least conceptually, and the 1D Voronoi logic can be extracted into a `common` directory for later reuse.
> 
> Capture zone padding gets more complicated with mixed or overlapping annotations: both horizontal and vertical lines; lines over rectangles; overlapping rectangles; annotation over the text of some other annotation; annotation over bars, data point markers or other interactive elements.
> 
> So, just curious about what we have as current goals for this, maybe on the original issue, or on a separate issue that goes into this kind of detail (or discussed as part of the PR)

Hey! I might be misinterpreting, but this only should make the line annotation marker clickable if passed an onAnnotationClick listener not the whole line.

monfera:
> this only should make the line annotation marker clickable if passed an onAnnotationClick listener not the whole line

Thanks Rachel! It wasn't clear to me, as the issue (to which I need to compare the PR) mentions annotations, and specifically the line and rectangle annotations. Could the scope be added to the gh issue, ie. it's just the annotation markerand not the entire rectangle or line, so I only review things in the PR that are agreed to change.

I guess that for the annotation marker, it'll stay as it is, basically no padding. @markov00 do we want to add padding to the markers, and are there signs of needs for capture events on the lines/rectangles?

packages/charts/src/chart_types/xy_chart/specs/rect_annotation.tsx
```diff
@@ -34,5 +34,7 @@ export const RectAnnotation: React.FunctionComponent<
       >
     >
 > = getConnect()(
-  specComponentFactory<RectAnnotationSpec, 'groupId' | 'annotationType' | 'zIndex' | 'style'>(defaultProps),
+  specComponentFactory<RectAnnotationSpec, 'groupId' | 'annotationType' | 'zIndex' | 'style' | 'onClickHandler'>(
```
nickofthyme, line 37:
```suggestion
  specComponentFactory<RectAnnotationSpec, 'groupId' | 'annotationType' | 'zIndex' | 'style'>(
```

packages/charts/src/chart_types/xy_chart/state/chart_state.interactions.test.ts
```diff
@@ -1170,3 +1170,66 @@ describe('Negative bars click and hover', () => {
...
+              details: 'details about this annotation',
+            },
+          ],
+          style: { fill: 'red' },
+        }),
+        MockAnnotationSpec.line({
+          id: 'line_annotation',
+          domainType: AnnotationDomainType.XDomain,
+          dataValues: [{ dataValue: 2, details: 'foo' }],
+          // marker: <div /> ,
```
nickofthyme, line 1219:
```suggestion
```

packages/charts/src/chart_types/xy_chart/state/chart_state.interactions.test.ts
```diff
@@ -1170,3 +1170,66 @@ describe('Negative bars click and hover', () => {
...
+  });
+  test('click rect1 annotation', () => {
+    store.dispatch(onPointerMove({ x: 130, y: 217 }, 0));
+    store.dispatch(onMouseDown({ x: 130, y: 217 }, 100));
+    store.dispatch(onMouseUp({ x: 130, y: 217 }, 200));
+
+    expect(onAnnotationClick).toBeCalled();
+    const callArgs = onAnnotationClick.mock.calls[0][0];
+    expect(callArgs.rects[0].id).toEqual('rect1');
+  });
```
nickofthyme, line 1234: Could you also add a test for line marker clicks and clicking two rect annotations (i.e. where `rects: [{ ... }, { ... }]`)

packages/charts/src/chart_types/xy_chart/state/selectors/get_cursor_pointer.ts
```diff
@@ -47,6 +47,14 @@ export const getPointerCursorSelector = createCustomCachedSelector(
     if (highlightedGeometries.length > 0 && (settingsSpec.onElementClick || settingsSpec.onElementOver)) {
       return 'pointer';
     }
+
+    // no pointer if there is a onAnnotionClick but there is a highlighted geometry
```
nickofthyme, line 51:
```suggestion
    // no pointer if there is a onAnnotationClick but there is a highlighted geometry
```

packages/charts/src/chart_types/xy_chart/state/selectors/get_cursor_pointer.ts
```diff
@@ -47,6 +47,14 @@ export const getPointerCursorSelector = createCustomCachedSelector(
...
+    if (highlightedGeometries.length > 0 && (!settingsSpec.onElementClick || !settingsSpec.onElementOver)) {
+      return DEFAULT_CSS_CURSOR;
+    }
+    if (settingsSpec.onAnnotationClick) {
+      return 'pointer';
+    }
```
nickofthyme, line 57: So this is applied even if there are no annotations actively hovered?

packages/charts/src/chart_types/xy_chart/state/selectors/on_click_caller.ts
```diff
@@ -38,18 +40,27 @@ export function createOnClickCaller(): (state: GlobalChartState) => void {
...
         if (!isClicking(prevClick, lastClick)) {
           return;
         }
         const elementClickFired = tryFiringOnElementClick(indexedGeometries, onElementClick);
-        if (!elementClickFired) {
+        if (onAnnotationClick && tooltipState) {
+          tryFiringOnAnnotationClick(tooltipState, onAnnotationClick);
+        } else if (!elementClickFired) {
           tryFiringOnProjectionClick(values, onProjectionClick);
         }
```
nickofthyme, line 65: I don't think we should fire the annotation click **AND** the element click handler, but instead just the element or just the annotation, where element is higher priority.

packages/charts/src/chart_types/xy_chart/state/selectors/on_click_caller.ts
```diff
@@ -84,3 +94,32 @@ function tryFiringOnProjectionClick(
...
+    const lineAnnotations = [];
+    if (annotationState.annotationType === 'rectangle') {
+      rectangularAnnotations.push({
+        id: annotationState.id,
+        datum: {
+          // @ts-ignore
+          coordinates: annotationState.datum.coordinates,
+          details: annotationState.datum.details,
+        },
+      });
```
nickofthyme, line 113: There is a type guard for this that will likely fix the `@ts-ignore` below.

packages/charts/src/chart_types/xy_chart/state/selectors/on_click_caller.ts
```diff
@@ -84,3 +94,32 @@ function tryFiringOnProjectionClick(
...
   return true;
 }
+
+function tryFiringOnAnnotationClick(
+  annotationState: AnnotationTooltipState,
+  onAnnotationClick: SettingsSpec['onAnnotationClick'],
+): boolean {
+  if (annotationState && onAnnotationClick) {
+    const rectangularAnnotations = [];
+    const lineAnnotations = [];
```
nickofthyme, line 104: nit

```suggestion
    const rects = [];
    const lines = [];
```

Making this the eventual object keys simplifies line 121. Unless there is a good case for this I tend to match the keys. Since the context of this function is already annotation (i.e. `tryFiringOnAnnotationClick`) I think this is just easier.

packages/charts/src/chart_types/xy_chart/state/selectors/on_click_caller.ts
```diff
@@ -84,3 +94,32 @@ function tryFiringOnProjectionClick(
...
+          details: annotationState.datum.details,
+        },
+      });
+    } else if (annotationState.annotationType === 'line') {
+      lineAnnotations.push({
+        id: annotationState.id,
+        // @ts-ignore
+        datum: { dataValue: annotationState.datum.dataValue, details: annotationState.datum.details },
+      });
+    }
```
nickofthyme, line 120: There is a type guard for this that will likely fix the @ts-ignore below.

packages/charts/src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -921,6 +921,10 @@ export type RectAnnotationSpec = BaseAnnotationSpec<
    * Dimension, either height or width, of outside annotation
    */
   outsideDimension?: number;
+  /**
+   * Allows the consumer to pass an on click event for the annotation
+   */
+  onClickHandler?: MouseEventHandler<Element>;
```
nickofthyme, line 927:
```suggestion
```

Stale remnant

packages/charts/src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -948,6 +952,10 @@ export interface BaseAnnotationSpec<
   S extends RectAnnotationStyle | LineAnnotationStyle
 > extends Spec,
     AnnotationPortalSettings {
+  /**
+   * Id provided by the user for the annotation
+   */
+  id: string;
```
nickofthyme, line 958:
```suggestion
```

This is redundant as it extends `Spec` which has `{ id: string; }`

packages/charts/src/components/portal/tooltip_portal.tsx
```diff
@@ -44,6 +44,10 @@ type PortalTooltipProps = {
    * Chart Id to add new anchor for each chart on the page
    */
   chartId: string;
+  /**
+   * onClickHandler attached to the portal
+   */
+  onClickHandler?: MouseEventHandler<Element>;
```
nickofthyme, line 50:
```suggestion
```

Stale remnant

storybook/main.js
```diff
@@ -9,6 +9,7 @@
 module.exports = {
   stories: ['./stories/**/*.stories.ts', './stories/**/*.stories.tsx'],
   addons: [
+    '@storybook/addon-actions',
```
nickofthyme, line 12: ❤️

nickofthyme (changes requested):
Looking good! I made a few suggestions along with some questions. Ping me if something doesn't make sense.

packages/charts/src/chart_types/xy_chart/state/selectors/get_cursor_pointer.ts
rshen91: As far as Im understanding, the highlighted geometries does not include the annotations. That pointer logic is tied to the annotation tooltip. I'm noticing though that the current implementation isn't working --- the pointer is being applied on all parts of the chart that aren't the elements if there's an annotation. Thank you for the catch, I'll make sure to update with a fix

packages/charts/src/chart_types/xy_chart/specs/rect_annotation.tsx
rshen91: Thank you! afa3069

packages/charts/src/chart_types/xy_chart/state/chart_state.interactions.test.ts
rshen91: Thank you! afa3069 for this change

packages/charts/src/chart_types/xy_chart/state/selectors/get_cursor_pointer.ts
rshen91: Lols my bad, thanks afa3069

packages/charts/src/chart_types/xy_chart/utils/specs.ts
rshen91: Thank you! afa3069

packages/charts/src/components/portal/tooltip_portal.tsx
rshen91: Thank you! afa3069

packages/charts/src/chart_types/xy_chart/utils/specs.ts
rshen91: Missed this! 8cc5e7f for changes

packages/charts/src/chart_types/xy_chart/state/selectors/get_cursor_pointer.ts
rshen91: 8cc5e7f has updates thank you for the catch!

packages/charts/src/chart_types/xy_chart/state/selectors/on_click_caller.ts
rshen91: Thank you for this catch! I agree, only elementclick or annotationclick should happen in one click action. 8cc5e7f for changes thank you!

rshen91: Sorry, wrong commit 7d7700d for changes thank you

packages/charts/src/chart_types/xy_chart/state/selectors/on_click_caller.ts
rshen91: Thank you for the zoom to clarify 7d7700d for changes

packages/charts/src/chart_types/xy_chart/state/selectors/on_click_caller.ts
rshen91: 7d7700d for changes thanks!

packages/charts/src/chart_types/xy_chart/state/selectors/on_click_caller.ts
rshen91: 7d7700 for changes thank you!

packages/charts/src/chart_types/xy_chart/state/chart_state.interactions.test.tsx
```diff
@@ -1232,4 +1225,112 @@ describe('Clickable annotations', () => {
...
+      store,
+    );
+    // the overlap of the blue and red rect
+    store.dispatch(onPointerMove({ x: 200, y: 195 }, 0));
+    store.dispatch(onMouseDown({ x: 200, y: 195 }, 100));
+    store.dispatch(onMouseUp({ x: 200, y: 195 }, 200));
+
+    expect(onAnnotationClick).toBeCalled();
+    const callArgs = onAnnotationClick.mock.calls[0][0];
+    expect(callArgs.rects[0].id).toEqual('rect2');
```
nickofthyme, line 1291: wouldn't you expect both to be in `rects`?

```suggestion
    expect(callArgs.rects[0].id).toEqual('rect1');
    expect(callArgs.rects[1].id).toEqual('rect2');
```

packages/charts/src/chart_types/xy_chart/state/selectors/get_cursor_pointer.ts
```diff
@@ -48,11 +51,11 @@ export const getPointerCursorSelector = createCustomCachedSelector(
...
 
-    // no pointer if there is a onAnnotionClick but there is a highlighted geometry
+    // no pointer if there is a onAnnotationClick but there is a highlighted geometry
     if (highlightedGeometries.length > 0 && (!settingsSpec.onElementClick || !settingsSpec.onElementOver)) {
       return DEFAULT_CSS_CURSOR;
     }
-    if (settingsSpec.onAnnotationClick) {
+    if (settingsSpec.onAnnotationClick && tooltipState) {
       return 'pointer';
     }
```
nickofthyme: I see why this is here now. I think you can simplify this logic to just this...

```suggestion
    if (highlightedGeometries.length === 0 && settingsSpec.onAnnotationClick && tooltipState) {
      return 'pointer';
    }
```

Are you sure you don't need to check the `tooltipState` for anything? Just if it exists that means it has something?

packages/charts/src/chart_types/xy_chart/state/selectors/get_cursor_pointer.ts
```diff
@@ -34,6 +36,7 @@ export const getPointerCursorSelector = createCustomCachedSelector(
     projectedValues,
     { chartDimensions },
     isBrushAvailable,
+    tooltipState,
```
nickofthyme, line 39: This context is not obvious I think we rename this to be clear.

```suggestion
    annotationTooltipState,
```

packages/charts/src/chart_types/xy_chart/state/selectors/get_cursor_pointer.ts
```diff
@@ -48,11 +51,11 @@ export const getPointerCursorSelector = createCustomCachedSelector(
       return 'pointer';
     }
 
-    // no pointer if there is a onAnnotionClick but there is a highlighted geometry
+    // no pointer if there is a onAnnotationClick but there is a highlighted geometry
     if (highlightedGeometries.length > 0 && (!settingsSpec.onElementClick || !settingsSpec.onElementOver)) {
       return DEFAULT_CSS_CURSOR;
     }
```
nickofthyme: Not needed with change in logic

```suggestion
```

packages/charts/src/chart_types/xy_chart/state/selectors/on_click_caller.ts
```diff
@@ -58,7 +64,7 @@ export function createOnClickCaller(): (state: GlobalChartState) => void {
           return;
         }
         const elementClickFired = tryFiringOnElementClick(indexedGeometries, onElementClick);
-        if (onAnnotationClick && tooltipState) {
+        if (!elementClickFired && onAnnotationClick && tooltipState) {
```
nickofthyme, line 67: 👍🏼

nickofthyme (changes requested):
Most changes were addressed, just a few more comments.

packages/charts/src/chart_types/xy_chart/state/selectors/get_cursor_pointer.ts
nickofthyme: My first spelling correction that's not my own 🏆

packages/charts/src/chart_types/xy_chart/state/selectors/get_cursor_pointer.ts
rshen91: Removed in cb44d3e thank you!

packages/charts/src/chart_types/xy_chart/state/selectors/get_cursor_pointer.ts
rshen91: Smart suggestion cb44d3e

packages/charts/src/chart_types/xy_chart/state/selectors/get_cursor_pointer.ts
rshen91: Yep! Good call cb44d3e

packages/charts/src/chart_types/xy_chart/state/chart_state.interactions.test.tsx
rshen91: I attempted a large scale refactoring but then backtracked in 495fc1b. I think it might introduce a breaking change in what is returned for tooltips so it might be better for a new PR.

.github/workflows/pull_request.yaml
```diff
@@ -0,0 +1,182 @@
+name: '@elastic/charts - CI'
```
nickofthyme, line 1: This was an old file picked up from merging. You can delete this.

packages/charts/src/chart_types/xy_chart/annotations/tooltip.ts
```diff
@@ -60,6 +61,49 @@ export function computeRectAnnotationTooltipState(
...
+      chartRotation,
+      chartDimensions,
+      spec.id,
+    );
+    if (rectAnnotationTooltipState) {
+      allRectAnnotations.push({
+        ...rectAnnotationTooltipState,
+        tooltipSettings,
+        customTooltip,
+        customTooltipDetails: customTooltipDetails ?? spec.renderTooltip,
```
nickofthyme, line 100: So this is in lieu of the refactoring that backfired?

rshen91: Cleaned up here: a4c1713

.github/workflows/pull_request.yaml
rshen91: Removed in a4c1713 thanks!

packages/charts/src/chart_types/xy_chart/renderer/dom/annotations/annotations.tsx
```diff
@@ -77,6 +80,7 @@ function renderAnnotationLineMarkers(
         chartDimensions={chartDimensions}
         onDOMElementEnter={onDOMElementEnter}
         onDOMElementLeave={onDOMElementLeave}
+        onLineMarkerClickHandler={onLineMarkerClickHandler}
       />,
```
nickofthyme, line 84: Probably best to use a consistent naming `onDOMElementClick`

packages/charts/src/chart_types/xy_chart/state/chart_state.interactions.test.tsx
```diff
@@ -1319,18 +1318,16 @@ describe('Clickable annotations', () => {
...
     );
     // the line marker
-    store.dispatch(onPointerMove({ x: 10, y: 10 }, 0));
-    store.dispatch(onMouseDown({ x: 10, y: 10 }, 100));
-    store.dispatch(onMouseUp({ x: 10, y: 10 }, 200));
+    store.dispatch(onPointerMove({ x: 8, y: 10 }, 0));
+    store.dispatch(onMouseDown({ x: 8, y: 10 }, 100));
+    store.dispatch(onMouseUp({ x: 8, y: 10 }, 200));
 
     expect(onAnnotationClick).toBeCalled();
```
nickofthyme, line 1331: Can you say something like `expect(onAnnotationClick).toBeCalledWith({ lines: [...] });`?

rshen91: I changed the test a little in 726d952 let me know if this seems valid

packages/charts/src/chart_types/xy_chart/state/chart_state.interactions.test.tsx
```diff
@@ -1290,44 +1292,35 @@ describe('Clickable annotations', () => {
...
     // the line marker
-    store.dispatch(onPointerMove({ x: 8, y: 10 }, 0));
-    store.dispatch(onMouseDown({ x: 8, y: 10 }, 100));
-    store.dispatch(onMouseUp({ x: 8, y: 10 }, 200));
-
-    expect(onAnnotationClick).toBeCalled();
+    const annotation = wrapper.find('.echAnnotation');
+    expect(annotation).toHaveLength(1);
+    annotation.simulate('click');
+    expect(annotation.debug()).toContain('onDOMElementClick');
```
nickofthyme, line 1324: This test doesn't assert what the interaction does just that a function was passed

rshen91: I also missed updating the two rectangle annotation test - f37ec94 for changes. Lmk what you think thank you

rshen91:
In aa8c050 I updated the failing vrts - these were exclusively screenshots that had line annotations and had slight pixel changes

nickofthyme (changes requested):
The cursor pointer logic does not match the actual logic. The `onAnnotationClick` event should **NOT** fire over element geometry. Also check how this acts with line charts when clicking points above the rect annotations.


https://user-images.githubusercontent.com/19007109/131392512-c6e45a8e-81ee-4ea7-96b0-095894430264.mp4

rshen91:
> The cursor pointer logic does not match the actual logic. The `onAnnotationClick` event should **NOT** fire over element geometry. Also check how this acts with line charts when clicking points above the rect annotations.
> 
>  Screen.Recording.2021-08-30.at.02.13.54.PM.mp4

Please see 4c07301 for the fix thank you for catching this!

nickofthyme (approved):
Looks great @rshen91! Great work.

PR #1294, opened 2021-08-10
feat: auto interval approximation

### LINKED ISSUE #840: Default min interval for single values
Yes, when using a histogram chart with a single bar and custom `xDomain` (setting `min` and/or `max`). The single bar is not displayed because we cannot compute a sensible interval.

Related to https://github.com/elastic/elastic-charts/issues/836

See https://codesandbox.io/s/flamboyant-tdd-b0hzt?file=/src/App.tsx

A fallback `minInterval` of say `1/16` or `1/20` of the domain should be used to at least render the single value.

There should also be a development warning to indicate that we are shooting in the dark trying to set a practical interval based on the data and domain provided, and that the consumer should provide a value.

Set the `minInterval` whenever there is a single value.

### LINKED ISSUE #1184: [Heatmap] Time scale type with a single row of data breaks the chart
**Describe the bug**
Passing a single row of data with a time scale on the X-axis breaks the chart. 

**To Reproduce**
Go to https://codesandbox.io/s/cool-sound-8j9m7?file=/src/App.tsx

> Notice by not setting the `xDomain` with a single values, causes the error below.

**Expected behaviour**
The chart is rendered with a single row of data 

 - OS: [e.g. iOS] macOS
 - Browser: [e.g. chrome, safari] Chrome
 - Elastic Charts: [e.g. 1.0.1] latest

**Errors in browser console**
```
quantize.js:18 Uncaught RangeError: Invalid array length
    at rescale (quantize.js:18)
    at Function.scale.range (quantize.js:28)
    at Object.shapeViewModel (viewmodel.js:81)
    at Object.render (scenegraph.js:30)
    at geometries.js:42
    at index.js:70
    at index.js:30
    at index.js:84
    at index.js:30
```

nickofthyme:
## Summary

The PR adds auto approximation of the `minInterval` when not directly inferable. This includes both time and non-time based data.

image

## Details

This PR adds a simplified approximation of the `minInterval` for single point datasets when a `min` and `max` `xDomain` is provided without a `minInterval`. Currently, we default the interval in such cases to `1`. This is problematic as this often prohibits rendering the bar given its infinitesimal width.

The solution in this case is to determine the best unit interval that would _reasonably_ display the single data point. This is done by looping through different time units of decreasing scale to determine the best time unit for the provided domain extents. Some examples include:

| domain extents | approx. minInterval |
|:-:|:-:|
| 40 years | 1 year |
| 10 years | 1 year |
| 52 weeks | 1 week |
| 4 weeks | 1 week |
| 1 week | 1 day |
| 8 days | 1 day |
| 7 days | 1 day |
| 1 days | 1 hour |
| 24 hours | 1 hour |
| 1 hour | 1 minute |
| 60 minutes | 1 minute |

> Notice all approximate `minInterval`s are a single unit of time. It has been noted as a future enhancement to this logic to account for multi-factor units such as `12 hours`, to best fit the given data.

A development warning is surfaced in the console to promote adding the actual `minInterval` based on the data.

image

## Issues

fixes #840, fixes #1184

### Checklist


packages/charts/src/chart_types/xy_chart/domains/x_domain.ts
```diff
@@ -127,18 +133,77 @@ function getMinInterval(computedMinInterval: number, size: number, customMinInte
...
+ * iteratively starting from the largest unit of time going to the smallest
+ *
+ * TODO: build in rules to apply multiplier of the unit interval (e.g. 12h)
+ */
+function approximateNearestDuration(value: number): number {
+  if (!value) return value;
+
+  for (let i = 0; i < approximatedOptions.length; i++) {
+    const { unit, min = 1, max = 10 } = approximatedOptions[i];
+    const unitInterval = moment.duration(1, unit).asMilliseconds();
```
markov00, line 162: can we just add the millisecond resolution in the `approximatedOptions`?

packages/charts/src/chart_types/xy_chart/domains/x_domain.ts
```diff
@@ -127,18 +133,77 @@ function getMinInterval(computedMinInterval: number, size: number, customMinInte
...
  */
-export function findMinInterval(xValues: number[]): number {
+export function findMinInterval(xValues: number[]): number | null {
   const valuesLength = xValues.length;
   if (valuesLength <= 0) {
     return 0;
   }
   if (valuesLength === 1) {
-    return 1;
+    return null;
```
markov00, line 206: Instead of returning null, can we just call `guessMinInterval(seriesXComputedDomains, fallbackScale ?? type);` here?
```suggestion
    return guessMinInterval(seriesXComputedDomains, fallbackScale ?? type);
```

packages/charts/src/chart_types/xy_chart/domains/x_domain.ts
nickofthyme: Not sure what you are suggesting here. Do you want me to remove the `milliseconds` option altogether?

packages/charts/src/chart_types/xy_chart/domains/x_domain.ts
nickofthyme: I thought about that but then I would need to add two more arguments to this function. Which would you prefer?

packages/charts/src/chart_types/xy_chart/domains/x_domain.ts
```diff
@@ -127,18 +133,77 @@ function getMinInterval(computedMinInterval: number, size: number, customMinInte
   return customMinInterval;
 }
 
+const approximatedOptions: {
```
monfera, line 136: Would there be a more telling name for anything in the code change with approx* in it, as what is going on here doesn't look like approximation, more like, inferred min interval? Yeah any min interval is inferred that's not expressly given, so if it's ambiguous, it could be more explicit, like singleton min interval or some such, easy to grep. 

The preexisting `minInterval` itself is already a bit of a misnomer, esp. for the case of a singleton bar, when there's no interval by definition. I see such alternatives in dataviz land as
- pitch, temporal pitch
- cadence
- bin width, temporal bin width

So it might be `binWidthFromSinglePoint` even

packages/charts/src/chart_types/xy_chart/domains/x_domain.ts
```diff
@@ -127,18 +133,77 @@ function getMinInterval(computedMinInterval: number, size: number, customMinInte
...
+  max?: number;
+}[] = [
+  { unit: 'year' },
+  { unit: 'day', min: 6, max: 8 },
+  { unit: 'week', max: 52 },
+  { unit: 'day', max: 32 },
+  { unit: 'hour', max: 24 },
+  { unit: 'minute', max: 60 },
+  { unit: 'second', max: 60 },
+  { unit: 'millisecond', max: 1000 },
```
monfera, line 148: Does it mean that if the specified time domain is such that there'd be 61 seconds in the domain, then it'll reject `second`, and will consider the next one which is `minute`? 

What drives the numbers? 

Asking it as we already have problems with enormous bar widths that no dataviz designer would manually consider. I realize it's an edge case, but a minute-bar shown on a 61-second domain will be almost full Cartesian area width, or if there's some nicing of the time axis(?) 1/3rd to half of the Cartesian width.

Also, shouldn't the whole thing depend on the screenspace Cartesian width ie. pixels, or estimated bar width pixels or bar pitch pixels? It feels like showing a minute-bar in a 61-second domain is great if the chart is very narrow, but not so great if it spans a dashboard horizontally

packages/charts/src/chart_types/xy_chart/domains/x_domain.ts
monfera: fwiw I tend to trade argument list brevity for return type simplicity so I wouldn't mind having two extra args, although there may be some better solution (`NaN` would make it `number` consistently, but if you need an `isNaN` downstream due to this, then it'd be a bit superficial)

packages/charts/src/chart_types/xy_chart/domains/x_domain.ts
```diff
@@ -127,18 +133,77 @@ function getMinInterval(computedMinInterval: number, size: number, customMinInte
...
+    scaleType === ScaleType.Time
+      ? approximateNearestDuration(domainGap)
+      : domainGap === 0
+      ? 1
+      : Math.round(domainGap / DEFAULT_INTERVAL_COUNT);
+
+  Logger.warn(
+    `The minInterval, given only a single data point, has been approximated as ${
+      scaleType === ScaleType.Time ? moment.duration(value).humanize() : value
+    }. This is not ideal as this data point likely does not coincide with the proper interval. Please set the minInterval in the xDomain.`,
```
monfera, line 189: Nit: the linter complains of line length but it's a template string, what do we want to generally do in cases like this?

packages/charts/src/chart_types/xy_chart/domains/x_domain.ts
```diff
@@ -127,18 +133,77 @@ function getMinInterval(computedMinInterval: number, size: number, customMinInte
...
   const valuesLength = xValues.length;
   if (valuesLength <= 0) {
     return 0;
   }
   if (valuesLength === 1) {
-    return 1;
+    return null;
   }
   const sortedValues = xValues.slice().sort(compareByValueAsc);
   let i;
```
monfera, line 209: Might as well put the `let` inside the `for`

packages/charts/src/chart_types/xy_chart/domains/x_domain.ts
```diff
@@ -127,18 +133,77 @@ function getMinInterval(computedMinInterval: number, size: number, customMinInte
...
 /**
  * Find the minimum interval between xValues.
- * Default to 0 if an empty array, 1 if one item array
+ * Default to 0 if an empty array, null if one item array
  * @internal
  */
-export function findMinInterval(xValues: number[]): number {
+export function findMinInterval(xValues: number[]): number | null {
   const valuesLength = xValues.length;
   if (valuesLength <= 0) {
```
monfera, line 202: Preexisting: [].length can't be less than 0 (I'm OK with a `<=` convention though if we want to save on bytes)

packages/charts/src/chart_types/xy_chart/domains/x_domain.ts
monfera: Putting aside the question around `null` the function is basically a reduction:
```
export function findMinInterval(xValues: number[]): number | null {
  const sortedValues = xValues.slice().sort(compareByValueAsc);
  return xValues.length <= 0
    ? 0
    : xValues.length === 1
    ? null
    : sortedValues.reduce(
        (p, n, i, a) => Math.min(p, (i < a.length - 1 ? a[i + 1] : Infinity) - n),
        sortedValues[1] - sortedValues[0],
      );
}
```
Somehow it feels asymmetrical that for a zero length array, we return zero, and for a 1-length array, it's null, then from 2, it's a number again

monfera: Oh my point about this being a reduction: probably it belongs in some basic array processing util, not `x_domain.ts`, so 
```
export function findMinInterval(xValues: number[]): number | null {
  const sortedValues = xValues.slice().sort(compareByValueAsc);
  return xValues.length <= 0
    ? 0
    : xValues.length === 1
    ? null
    : minDelta(sortedValues)
}

```

packages/charts/src/chart_types/xy_chart/domains/x_domain.ts
```diff
@@ -127,18 +133,77 @@ function getMinInterval(computedMinInterval: number, size: number, customMinInte
...
     return 0;
   }
   if (valuesLength === 1) {
-    return 1;
+    return null;
   }
   const sortedValues = xValues.slice().sort(compareByValueAsc);
```
monfera, line 208: Since we sort, there's no need for the `Math.abs` in `const interval = Math.abs(next - current);`

packages/charts/src/chart_types/xy_chart/domains/x_domain.ts
```diff
@@ -127,18 +133,77 @@ function getMinInterval(computedMinInterval: number, size: number, customMinInte
   return customMinInterval;
 }
 
+const approximatedOptions: {
+  unit: moment.unitOfTime.Base;
+  min?: number;
+  max?: number;
+}[] = [
+  { unit: 'year' },
```
monfera, line 141: `{ unit: 'year' , max: Infinity}`

packages/charts/src/chart_types/xy_chart/domains/x_domain.ts
```diff
@@ -127,18 +133,77 @@ function getMinInterval(computedMinInterval: number, size: number, customMinInte
...
+ */
+function approximateNearestDuration(value: number): number {
+  if (!value) return value;
+
+  for (let i = 0; i < approximatedOptions.length; i++) {
+    const { unit, min = 1, max = 10 } = approximatedOptions[i];
+    const unitInterval = moment.duration(1, unit).asMilliseconds();
+    const unitDiff = Math.floor(value / unitInterval);
+
+    if ((unitDiff > min && unitDiff <= max) || (unit === 'year' && unitDiff > max)) {
```
monfera, line 165: ... then no need for the extra `|| (unit === 'year' && unitDiff > max)` check

monfera: Minor: it's also fine, and maybe more intuitive here to permit the min value too, and using a math-like ordering `min <= unitDiff <= max`

```
min <= unitDiff && unitDiff <= max
```

nickofthyme:
Closing in favor of different approach, that being to require `min-interval` on `BarSeriesSpec` and remove from `xDomain`. Keep comments for future consideration.

PR #1300, opened 2021-08-15
ci: cleanup storybook vrt styles and colors

nickofthyme:
## Details

Cleanup storybook styles to match given theme. Update colors to be consistent and accurate across theme and background color.


markov00 (commented):
Cleanup looks good to me

PR #1307, opened 2021-08-17
chore(xy): avoid large dimensions in ctx.clearRect

monfera:
## Summary

Removes the `ctx.clearRect(-200000, -200000, 400000, 400000)` present in a couple of chart types. Also, more places use the common, `clearCanvas` method that got updated for transform independence 

Also, removed one or two unnecessary `withContext` and lightly restructured some surrounding code 

## Details

The global canvas erasure (in `master` and in this branch too) is currently possible, because the canvas isn't shared: each chart type possesses it, so it's safe to clear everything. In the future, we do need to refine this, for example, if
- we want heterogeneous small multiples or tiling layouts
- we want to update just one panel, while leaving the rest untouched (eg. for perf reasons, no need to recalc/rerender)

## Issues

Follow-up on #1286 which also eliminated some large canvas dimension operations

### Checklist

### DIFF
--- a/packages/charts/src/chart_types/goal_chart/renderer/canvas/canvas_renderers.ts
+++ b/packages/charts/src/chart_types/goal_chart/renderer/canvas/canvas_renderers.ts
@@ -30,10 +30,9 @@ export function renderCanvas2d(ctx: CanvasRenderingContext2D, dpr: number, geomO
 
     renderLayers(ctx, [
       // clear the canvas
-      (context: CanvasRenderingContext2D) => clearCanvas(context, 200000, 200000),
-
+      clearCanvas,
       (context: CanvasRenderingContext2D) =>
-        withContext(context, (ctx) => geomObjects.forEach((obj) => withContext(ctx, (ctx) => obj.render(ctx)))),
+        geomObjects.forEach((obj) => withContext(context, (ctxt) => obj.render(ctxt))),
     ]);
   });
 }
--- a/packages/charts/src/chart_types/partition_chart/renderer/canvas/partition.tsx
+++ b/packages/charts/src/chart_types/partition_chart/renderer/canvas/partition.tsx
@@ -148,10 +148,7 @@ class PartitionComponent extends React.Component<PartitionProps> {
       a11ySettings,
       debug,
     } = this.props;
-    if (!initialized || width === 0 || height === 0) {
-      return null;
-    }
-    return (
+    return width === 0 || height === 0 || !initialized ? null : (
       <figure aria-labelledby={a11ySettings.labelId} aria-describedby={a11ySettings.descriptionId}>
         <canvas
           ref={forwardStageRef}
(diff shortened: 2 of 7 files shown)


packages/charts/src/chart_types/partition_chart/renderer/canvas/partition.tsx
```diff
@@ -148,10 +148,7 @@ class PartitionComponent extends React.Component<PartitionProps> {
       a11ySettings,
       debug,
     } = this.props;
-    if (!initialized || width === 0 || height === 0) {
-      return null;
-    }
-    return (
+    return width === 0 || height === 0 || !initialized ? null : (
```
markov00, line 151: I love how the end of this line looks like :(

packages/charts/src/chart_types/goal_chart/renderer/canvas/canvas_renderers.ts
```diff
@@ -30,10 +30,9 @@ export function renderCanvas2d(ctx: CanvasRenderingContext2D, dpr: number, geomO
 
     renderLayers(ctx, [
       // clear the canvas
-      (context: CanvasRenderingContext2D) => clearCanvas(context, 200000, 200000),
-
+      clearCanvas,
       (context: CanvasRenderingContext2D) =>
-        withContext(context, (ctx) => geomObjects.forEach((obj) => withContext(ctx, (ctx) => obj.render(ctx)))),
+        geomObjects.forEach((obj) => withContext(context, (ctxt) => obj.render(ctxt))),
```
markov00, line 35: we should find a better solution to this: `context -> ctx -> ctxt` what would be the next?

markov00 (approved):
LGTM

packages/charts/src/chart_types/goal_chart/renderer/canvas/canvas_renderers.ts
monfera: I see the motivation for avoiding shadowing, yet there are harmless cases. I don't think this rule purchases a lot, so we could deactivate it in some future PR. There are other options like adding noise, code bulk and cognitive load by factoring out all inner parts into top level functions. 

Ultimately it'd be great to more declaratively generate the canvas commands from the intents (as with some of our research projects) and that means, there's no need for the `withContext`. It's a safety harness now as it reliably pairs a ctx.save / ctx.restore

packages/charts/src/chart_types/partition_chart/renderer/canvas/partition.tsx
monfera: Yes it's interesting formatting, I tried to format it sensibly but it reverted. Would be in favor of giving more code formatting freedom to the developer, and we've been discussing about moving away from prettier for other slightly undesirable cases too. Here it's not even that bad, as the 1st line quickly gets the bail condition out of the way (lets face it, it's noise so the briefer the better) so at least the focus is on the actual non-null return value on 1st blink

packages/charts/src/chart_types/goal_chart/renderer/canvas/canvas_renderers.ts
monfera: ... here I should've written 
```
obj.render
```
instead of
```
(ctxt) => obj.render(ctxt)
```
I had the fleeting assumption that a `.bind` would be needed, but on a 2nd look, nope, so your comment became acted upon (next PR)

monfera: became
```
(context) => geomObjects.forEach(({ render }) => withContext(context, render))
```

monfera: `withContext` could be repurposed, if we want: the callback function would lose the `ctx` arg, so any mention of `ctx` in the callback function would become a closure variable. `withContext` could be renamed to `withRestore` or `saveRestore`. @markov00 @nickofthyme lmk what you think. Probably we never exploit the flexibility of passing through the context to the callback fun

monfera: Solved with https://github.com/elastic/elastic-charts/pull/1313 - also, the binding of the mark was needed afterall, as it's a pseudoclass object

PR #1309, opened 2021-08-18
test: canvas methods in jest and ci

monfera:
This is just a test PR for discussing the lack of availability of  `ctx.getTransform` and `DOMMatrix.invertSelf()`

Apparently, `jsdom` has support for both; this passes: https://codesandbox.io/s/practical-lederberg-wfkgg?file=/src/index.js

`jest-canvas-mock` also supports `ctx.getTransform` (but has no support for `DOMMatrix.invertSelf()` because it's not a method on the canvas context object, but on a `DOMMatrix` instance) 

As seen in the last commit, I updated purportedly all of the packages that had `jsdom` in it (recursed the yarn.lock manually...). Now something became inconsistent, all the test fail to run, reporting 
>  The error below may be caused by using the wrong test environment, see https://jestjs.io/docs/configuration#testenvironment-string.
    Consider using the "jsdom" test environment

I also ran `yarn cache clean`

Maybe there's a bunch of other packages to update


markov00:
@monfera I think with the latest changes in the `canvas` dependency and the removed `jest-canvas-mock` we now have the required method in the jest tests. If you want to make a try

PR #1313, opened 2021-08-19
refactor: reduction of canvas context threading

monfera:
## Summary

Callbacks for `withContext` and, occasionally, other such wrappers to use the outer ctx, as we don't juggle them anyway. The main motivation is to avoid the awkwardness of avoidance of shadowing, while keeping the somewhat sensible rule against shadowing. Stemming from this [ask](https://github.com/elastic/elastic-charts/pull/1307#discussion_r691952011) by @markov00 

### Checklist

### DIFF
--- a/packages/charts/src/chart_types/goal_chart/renderer/canvas/canvas_renderers.ts
+++ b/packages/charts/src/chart_types/goal_chart/renderer/canvas/canvas_renderers.ts
@@ -31,7 +31,7 @@ export function renderCanvas2d(ctx: CanvasRenderingContext2D, dpr: number, geomO
     renderLayers(ctx, [
       // clear the canvas
       clearCanvas,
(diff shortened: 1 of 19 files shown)


nickofthyme (approved):
LGTM, what was in original use case for using the callback `ctx`?

monfera:
@nickofthyme it's a bit of an FP approach that instead of relying on capturing some variable at the place where the callback is defined (here, the canvas context) it gets passed on explicitly. Which is still doable btw. But we can't enforce variable capture even if it'd be great for a more FP style, at least I'm not aware of a linter that shouts if you capture variables. So the only serious way to defend against that if we made it a policy to define every and all callbacks at the top level of every file. 

In practical terms, you can do some higher order function manipulation if the context var isn't captured. For example, you can play, reorder, optimize etc. callback functions on their own, without them needing to carry a specific context. But we never do that today, and again, this option is still there if/when we need it

PR #1315, opened 2021-08-19
chore: removed more context variable shadowing and withContext

monfera:
## Summary

Removed 
- a bunch more context variable shadowing, and renamed some `context` back to `ctx` (fix)
- unsafe `[].push(...possiblyLongArray)` calls (fix)
- some more redundant `withContext` calls (perf)
- some unneeded curlies and CRLF where I touched code (chore)
- unused code I ran into (see per commit) (fix)
- single-use function (see per commit) (refactor)
- one last 0.001 in favor of `MIN_STROKE_WIDTH` (chore)
- some unnecessary bail conditions
- 340  lines (refactor)

There's also a small type fix, which needed the `isCanvasRenderer` type guard function, and a type `CanvasRenderer` was added, as the `(ctx) => void` got a bit non-DRY

A couple of `[].map().sort().forEach()` got replaced with `[].slice().sort().forEach()` mostly for data flow simplicity and avoidance of unnecessary object allocation (`map` allocated new tuples as array elements, while`slice` doesn't)

### Checklist


packages/charts/src/chart_types/xy_chart/renderer/canvas/bubbles.ts
```diff
@@ -32,36 +32,19 @@ export function renderBubbles(ctx: CanvasRenderingContext2D, props: BubbleGeomet
...
-    const pointStyles: Record<SeriesKey, PointStyle> = {};
 
-    const allPoints = bubbles.reduce<PointGeometry[]>(
-      (acc, { value: { seriesIdentifier, seriesPointStyle, points } }) => {
-        geometryStyles[seriesIdentifier.key] = getGeometryStateStyle(
-          seriesIdentifier,
-          sharedStyle,
-          highlightedLegendItem,
-        );
-        pointStyles[seriesIdentifier.key] = seriesPointStyle;
```
nickofthyme, line 44: 👍🏼

nickofthyme (approved):
LGTM

monfera:
Regarding the `[].push()` issue, there's nothing in the EcmaScript spec that'd limit a function argument count, though some earlier tests on the internet (2017) showed limits like 64k function arguments for Chrome. 

So it might be perfectly safe to do the `[].push(...longArray)`

Glad to revert [this commit](https://github.com/elastic/elastic-charts/pull/1315/commits/645d2539c234f7afffcca374e4399b7e260cda50) with the three replacements if you say so

It would be useful to not revert [this commit](https://github.com/elastic/elastic-charts/pull/1315/commits/36ceb040f4d1568ee87c826720e2930c1169438e) though as it's a shorter / clearer solution (suggests what happens right on the top with `flatMap`)

packages/charts/src/chart_types/xy_chart/renderer/dom/annotations/annotations.tsx
```diff
@@ -115,7 +115,7 @@ const AnnotationsComponent = ({
           onDOMElementEnter,
           onDOMElementLeave,
         );
-        markers.push(...lineMarkers);
+        lineMarkers.forEach((m) => markers.push(m));
```
nickofthyme, line 118: reasoning for this?

nickofthyme (approved):
LGTM, what's your aversion to `push(...things)`?

packages/charts/src/chart_types/xy_chart/renderer/dom/annotations/annotations.tsx
monfera: a = [].push(...[...new Array(200000)].map(() => Math.random()))

errors out—there must be some high but arbitrary limit

monfera: (see description & commit title - concern about arg count limit)

monfera: errors out = crashes the tab (chrome)

monfera: see this comment too https://github.com/elastic/elastic-charts/pull/1315#issuecomment-902146581

packages/charts/src/chart_types/xy_chart/renderer/canvas/annotations/index.ts
```diff
@@ -34,20 +35,22 @@ export function renderAnnotations(
...
-      if (isLineAnnotation(spec)) {
-        const lineStyle = mergeWithDefaultAnnotationLine(spec.style);
-        renderLineAnnotations(ctx, annotation as AnnotationLineProps[], lineStyle, rotation, renderingArea);
-      } else if (isRectAnnotation(spec)) {
-        const rectStyle = mergeWithDefaultAnnotationRect(spec.style);
-        renderRectAnnotations(ctx, annotation as AnnotationRectProps[], rectStyle, rotation, renderingArea);
+  withContext(ctx, () =>
+    annotationDimensions.forEach((annotation, id) => {
+      const spec = getSpecsById<AnnotationSpec>(annotationSpecs, id);
+      if (!spec) {
```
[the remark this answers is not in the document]
monfera: I added this b/c I also removed two others (where `renderAnnotations` was called). I didn't want to push my luck as the next safety barrier is kinda deep (within forEach/withPanelTransform) so let's consider it in a next round (say, we refactor the forEach bit and it becomes unsafe, so it's slightly riskier due to the higher level of indirection)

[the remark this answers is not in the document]
monfera: After further thinking I heeded your advice, it's the kind of thing that should easily show up on VRT so why not https://github.com/elastic/elastic-charts/pull/1315/commits/1d1437c9d2760604910f7cb4a07ed623fdc21a84

PR #1316, opened 2021-08-19
fix(legend): no truncation with single value

### LINKED ISSUE #1312: [legend] top and bottom legend position doesn't work well with new truncate setting
Recently we added an option to truncate the legend in all Visualize, TSVB and Lens editors. It means that the user can 
- choose not to truncate their legend at all
- truncate their legend items between 1 to 5 lines

When user chooses a legend position `top` or `bottom`, the space taken by the legend is around 1.5 lines. Which was fine before, because the shade effect indicated that space is scrollable. I think it's fine for 1 line or not-truncated line, but as we give an option to truncate from 1 to 5 lines, maybe we could in these cases expand this space and show full truncated strings.  

Screenshot 2021-08-19 at 11 12 02

https://user-images.githubusercontent.com/4283304/130046376-7e0b1614-5745-4d38-b5e7-ed0c39710b4a.mp4

nickofthyme:
## Summary

Fix issue with single item legends limiting width of legend.

Screen Recording 2021-08-19 at 10 04 33 AM

## Details

The legend grid styles logic attempts to create the best fit for all cases. This is not always the case as the limiting size may not be ideal for **(a)** the number of series and/or **(b)** the width of the chart, leading to unnecessary white space. This option is configurable via the `verticalWidth` theme parameter however in the case of a single value the legend can always be set to the full width.

## Issues

fixes #1312

### Checklist


markov00 (approved):
Code looks good to me.
The legend needs to be anyway redesigned to accommodate more edge cases, but this solves the linked issue

nickofthyme:
> Code looks good to me.
> The legend needs to be anyway redesigned to accommodate more edge cases, but this solves the linked issue

Yup I added the issue to #580

PR #1319, opened 2021-08-20
test: vite

markov00:
### DO NOT MERGE, JUST FOR TESTING

I've taken the chance to play quickly with `vite` today and I'd like to share here what was the process/issues I've found.

Reasons for this testing:
- current webpack configuration for screenshot testing takes a long time to compile, 30 seconds at least. On every change, another 30 seconds are spent on the compilation
- The server startup on the CI for VRTs tooks nearly 2 minutes
- Storybook took a lot to start, we can convert in the future to our own simplified version of Storybook, just using EUI + our charts. We can benefit from `vite` on speed and from EUI to have increased control over the chart configuration

Followed process:

- Created a simple example folder in `packages/docs`
- used `yarn create vite docs --template react-ts` to build the scafolding project
- Added the `@elastic/charts` in the `package.json` with the exact same version as the one available in the workspace package `charts`:
    - Drawback: we need to keep that in sync to be sure that the folder is symlinked correctly, if not an npm version will be downloaded.
    - Workaround: we can change the version to `0.0.0-development` as recommended by `semantic-release`. This should avoid sync the version every time and semantically is more correct
- To run the vite project `yarn dev` from the `packages/docs` folder. It starts in milliseconds and is immediately available
- I started adding a chart to play with. 
    - Problems: the`@elastic/charts` package is configured to be used from the `dist` folder. This needs a separate compilation always running in watch mode. 
    - Workaround: Instead of importing from `@elastic/charts` I've imported from the `src` directory `import {Charts} from '@elastic/charts/src'`. How can we fix that? 
- During the initial startup of Vite throws some errors, mostly related to:
    - SCSS: the current imported EUI scss files were not found. I've brutally solved removing the tilde `~` at the beginning, but not sure what was the exact problem
    - Mixed exported types and actual functions/const. Solved moving all type only exports to use `export type` syntax and everything else the usual `export`
- I removed the `resize-observer-polyfill` because it throws errors (probably due to a different react version
- Aligned the react version between vite project and charts one. I've seen that we don't have `@types/react` as dev dep in charts, should we use one?

Observations:
- the server starts up in milliseconds, but it will takes 5/10 seconds for the initial compilation (not sure what is the bottleneck here). It is a very simple app with one chart but the configuration is the one straight from the vite scaffolding, so probably we can improve it
- the changes, both on the charts code or the app code are picked up automatically, the page is refreshed in milliseconds 

https://user-images.githubusercontent.com/1421091/130253695-64366408-a1b7-4ec7-afd6-4965bc94c793.mp4

 - I've also created a production build from the simple app, and it tooks 2 minutes to complete (it doesn't use esbuild here but tsc)


packages/charts/src/_eui_imports.scss
```diff
@@ -1,3 +1,3 @@
-@import '~@elastic/eui/src/global_styling/functions/index';
-@import '~@elastic/eui/src/global_styling/variables/index';
-@import '~@elastic/eui/src/global_styling/mixins/index';
```
nickofthyme, line 3: 👍🏼  I found out this syntax is a `sass-loader` thing not a `sass` thing and is not the best practice. 😞

packages/charts/src/index.ts
```diff
@@ -8,50 +8,50 @@
...
-export { SimplePadding, Padding } from './utils/dimensions';
+export type { ContinuousDomain, OrdinalDomain } from './utils/domain';
+export type { SimplePadding, Padding } from './utils/dimensions';
 export { timeFormatter, niceTimeFormatter, niceTimeFormatByDay } from './utils/data/formatters';
-export { SeriesIdentifier, SeriesKey } from './common/series_id';
-export { XYChartSeriesIdentifier, DataSeriesDatum, FilledValues } from './chart_types/xy_chart/utils/series';
-export {
+export type { SeriesIdentifier, SeriesKey } from './common/series_id';
+export type { XYChartSeriesIdentifier, DataSeriesDatum, FilledValues } from './chart_types/xy_chart/utils/series';
+export type {
```
nickofthyme, line 25: Can we do this at a file level? All these types are unknown here.

Also are these treated a different way with this syntax even if they are only `type` exports from the original file?

nickofthyme:
This is amazing! 🤩  I wonder how fast are the changes to `.scss` files? I'm sure we could resolve the bottleneck somehow 😉 .

---

> Workaround: we can change the version to 0.0.0-development as recommended by semantic-release. This should avoid sync the version every time and semantically is more correct

Good idea 👍 

> It starts in milliseconds and is immediately available
💨 

> Instead of importing from @elastic/charts I've imported from the src directory import {Charts} from '@elastic/charts/src'. How can we fix that?

Yeah this is something I had issues with too. I think we could just add a module resolution to point to source so it's basically the same thing but easier. We can also start to limit importing deeper than `src/`.

> Mixed exported types and actual functions/const. Solved moving all type only exports to use export type syntax and everything else the usual export

Are you saying this was due to naming collisions?

packages/charts/src/index.ts
markov00: this is already done at file level (we already use `export type` `export interface` export const` `export function`)
but this is the final export. It just complains about missing object when not exporting the types as types

markov00:
> Mixed exported types and actual functions/const. Solved moving all type only exports to use export type syntax and everything else the usual export
> 
> Are you saying this was due to naming collisions?

From esbuild docs:
>  This is parsed and ignored: Type-only exports like: export type {Type} from 'foo'
> Handled by ignoring missing exports in TypeScript files: export {Type} from 'foo'

bit it doesn't seem to fully ignore missing exports in typescript files when types exports are mixed with object ones

nickofthyme:
Just saw https://vitest.dev/guide/features.html and think when/if we switch to vite could be good to switch `jest` to `vitext` too.

PR #1320, opened 2021-08-20
refactor: cleanup colors

markov00:
## Summary

During the heatmap legend fixes I've noticed that we have multiple places where some standard colors are allocated.
I've moved the `Color` type + some static colors (red, white, black, transparent) into the `common/color.ts` file.
This is not a complete refactoring but the first part of the initiative to cleanup the folders the typings.

### Checklist


packages/charts/src/common/color.ts
```diff
@@ -0,0 +1,26 @@
...
+export const WHITE_COLOR: Color = 'rgba(255, 255, 255, 1)';
+/** @internal */
+export const BLACK_COLOR: Color = 'rgba(0, 0, 0, 1)';
+/** @internal */
+export const RED_COLOR: Color = 'rgba(255, 0, 0, 1)';
+
+/** @internal */
+export const RED_RGBOBJ_COLOR: RgbObject = { r: 255, g: 0, b: 0, opacity: 1 };
+/** @internal */
+export const TRANSPARENT_RBGOBJ_COLOR: RgbObject = { r: 0, g: 0, b: 0, opacity: 0 };
```
nickofthyme, line 26: What would you think about putting these in an `object` or `class` a la Flutter (https://api.flutter.dev/flutter/material/Colors-class.html). That way we lessen the import mess and the less redundant naming, `RED_COLOR` would just be `Colors.Red`.

That way we still preserve the `Color` as a type.

storybook/stories/goal/10_band_in_band.tsx
```diff
@@ -12,8 +12,8 @@ import { Chart, Goal, Settings } from '@elastic/charts';
 import { config } from '@elastic/charts/src/chart_types/goal_chart/layout/config/config';
 import { BandFillColorAccessorInput } from '@elastic/charts/src/chart_types/goal_chart/specs';
 import { GoalSubtype } from '@elastic/charts/src/chart_types/goal_chart/specs/constants';
-import { Color } from '@elastic/charts/src/utils/common';
 
+import { Color } from '../../../packages/charts/src/common/color';
 import { useBaseTheme } from '../../use_base_theme';
 
 const subtype = GoalSubtype.Goal;
```
nickofthyme, line 19: ??? Why is this changed? I think your IDE is playing tricks on you.

nickofthyme (commented):
LGTM, I would suggest a slightest simplified API. See comments.

packages/charts/src/common/color.ts
nickofthyme: You could even do `Colors.rgbo.Red` or something.

monfera (approved):
LGTM, good to merge once it passes and conflicts are resolved

Switching to some color object (or fake enum) would be sensible if we expect significant growth to the count of these constants, or a `COLOR_` prefix at least. With the current handful of colors, I don't feel strongly either way.  Worth doing a last minute grep for all string occurrences `'black'` or even without quotes to ferret out template string occurrences or similar

packages/charts/src/common/color.ts
```diff
@@ -12,15 +12,21 @@ import { RgbObject } from './color_library_wrappers';
...
+  },
+  Black: {
+    rgba: 'rgba(0, 0, 0, 1)',
+    rgbaChannels: { r: 0, g: 0, b: 0, opacity: 1 },
+  },
+  Transparent: {
+    rgba: 'rgba(0, 0, 0, 0)',
+    rgbaChannels: { r: 0, g: 0, b: 0, opacity: 0 },
+  },
+};
```
nickofthyme, line 32: ❤️ Nice! The merging of like colors was an added perk! 👍🏼

nickofthyme (approved):

markov00:
just an update: I've found that `transparent` and `rgba(0,0,0,0)` are not treated the same way in the partition chart. I'm currently investigating to fix this

markov00:
close in favour of #1397

Issue #123, opened 2019-03-27
Crosshair works incorrectly after page scroll


sulemanof:
**Describe the bug**
Crosshair works incorrectly after page scroll. User unable to see a tooltip while hover on a chart point

**To Reproduce**
1. Create any chart with data
2. Hover on any point -> user is able to see a tooltip
3. Scroll down a bit
4. Hover on any point -> user can't see any tooltip
5. Move the mouse down on the same tick -> in one moment the tooltip will appear as on point hover

**Expected behavior**
A page scroll shouldn't break the chart positioning

**Screenshots**
![chart_scroll_issue](https://user-images.githubusercontent.com/31325372/55064105-77f89700-508a-11e9-85ee-8635c39838b3.gif)

**Version (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome 73
 - Elastic Charts: 3.4.0

**Additional context**
n/a

**Errors in browser console**
n/a

**Kibana Cross Issues**
n/a

**Checklist**
- [x] every related Kibana issue is listed under `Kibana Cross Issues` list
- [x] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

Issue #1310, opened 2021-08-18
Timeslip integration in progress


markov00:

ghudgins:
linking up kibana issue - https://github.com/elastic/kibana/issues/51227

monfera:
Adding a `charts` generated snapshot of interim steps toward the listed goals:

![image](https://user-images.githubusercontent.com/1548516/131577155-d1895e31-2d03-4dac-964e-2a27d0319f8e.png)

Issue #758, opened 2020-07-21
Charts appear shrunk when using long tooltip header


wylieconlon:
I couldn't find an identical bug report, but it's possible that there is one. In Lens we are setting the tooltip headerFormatter to show the X axis value, with a code snippet like this:

```
<Settings
  tooltip={{ headerFormatter: () => 'Winnipeg / James Armstrong Richardson International Airport' }}
/>
```

The rendered result looks like this:

<img width="686" alt="Screen Shot 2020-07-21 at 5 18 06 PM" src="https://user-images.githubusercontent.com/666475/88108120-4efcef00-cb76-11ea-9e24-f1ca3ea11ee0.png">

This can be reproduced in the storybooks, so I think this is a real issue. It's not super high severity, but it's definitely noticeable.

nickofthyme:
Thanks @wylieconlon, we'll look into it!

markov00:
This is not because of the tooltip header has a long name, but happens when the data category contains such long names.
In fact, you see that all the tick labels are hidden because all of them overlap with each other. The chart also shrinks to accommodate "correctly" display the last label without truncations but fails in that.
It's definitely a bug and we should fix this in 3 steps:
- provide a way to truncate label text
- if hiding the overlapping labels, these labels should not be used to compute the chart size
- we should add the ability to wrap labels into multiple lines

Issue #1303, opened 2021-08-16
Code improvements meta issue


monfera:
- [ ] Check for zero dimension canvas mentioned at the bottom of [this](https://github.com/elastic/elastic-charts/pull/1286#discussion_r688151382) as otherwise the current code (pre or post merge) may throw
- [ ] Reconsider the [array enums](https://github.com/elastic/elastic-charts/pull/1286#discussion_r685358962) and object enums as the time for switching to proper TS enums may have come, and/or it's sometimes possible, and better, to piggyback on a preexisting HTML DOM type library instead of explicitly listing values (though the latter would help generate configuration UIs automatically, cf. `Config` object in partition charts) 
- [ ] (maybe for a separate, functiona meta issue?) More pattern filling options, eg. round number of tiling, or constrained, approximate tile sizes (in pixels) as well as scaling options upon resize (initial [discussion](https://github.com/elastic/elastic-charts/pull/1286#discussion_r688138312))
- [ ] Consider renaming`ON_TOGGLE_DESELECT_SERIES` ([comment](https://github.com/elastic/elastic-charts/pull/1273#discussion_r689314258))
- [ ] Consider moving away from the use of `specId` in heatmap to encode information (came up eg. [here](https://github.com/elastic/elastic-charts/pull/1273#commitcomment-54945011))

PR #1324, opened 2021-08-23
refactor(xy): axis title and related simplifications

monfera:
## Summary

- safer text measurement, no need for manual `destroy`, no need for class
- switched away from the svg measurer in tests and removed both DOM based measurers
- [].sort predicate `compareByValueAsc` equipped to deal with duplicates
- simplified and split `getSimplePadding`
- simpler `computeAxesSizes`, `renderBars` etc.
- removal of remaining shadowing `(ctx) =>` cases
- unified horizontal and vertical titles, due to much shared code
- unified global and panel titles, due to much shared code
- removed parallel redundant switch/case structures etc. from `values.bar.ts`
- replaced two other versions of `clamp` with `clamp`
- removed the git commit and husky
- minor: `forEach` where a `for` is essentially that, etc.

Prepares the ground for https://github.com/elastic/elastic-charts/issues/1310 and https://github.com/elastic/elastic-charts/issues/758 

### Checklist


packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -38,163 +38,161 @@ export function renderBars(
...
-    }
-
-    if (absMinHeight !== undefined && height !== 0 && Math.abs(height) < absMinHeight) {
-      const heightDelta = absMinHeight - Math.abs(height);
-      if (height < 0) {
-        height = -absMinHeight;
-        y += heightDelta;
+
+      let y: number | null;
+      let y0Scaled;
```
monfera, line 54: Just for the record, I didn't write these `let`s, git lost track of what's happening, still I plan to fix them at some point

.huskyrc.js
```diff
@@ -1,5 +1,3 @@
 module.exports = {
-  hooks: {
-    'commit-msg': 'commitlint -E HUSKY_GIT_PARAMS && yarn typecheck:all && yarn pq && lint-staged',
-  },
+  hooks: {},
```
markov00, line 2: We can completely get rid of the file and remove also the `husky` dependency

packages/charts/src/state/selectors/get_legend_size.ts
```diff
@@ -25,6 +24,8 @@ const MARKER_WIDTH = 16;
 const SHARED_MARGIN = 4;
 const VERTICAL_PADDING = 4;
 const TOP_MARGIN = 2;
+const MAGIC_FONT_FAMILY =
+  '"Inter UI", -apple-system, BlinkMacSystemFont, "Segoe UI", Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol"';
```
markov00, line 28: 🤦 We should find a solution here.
We probably have to wait until EUI has completely switched to CSS-in-JS so we can easily pick the right font from them

packages/charts/src/utils/bbox/canvas_text_bbox_calculator.ts
```diff
@@ -6,44 +6,37 @@
...
+/** @internal */
+export const withTextMeasure = <T>(fun: (textMeasure: TextMeasure) => T) => {
+  const canvas = document.createElement('canvas');
+  canvas.style.display = 'none';
+  const ctx = canvas.getContext('2d');
+  const root = document.documentElement;
+  root.appendChild(canvas);
+  const textMeasure: TextMeasure = ctx
+    ? (text: string, padding: number, fontSize = 16, fontFamily = 'Arial', lineHeight = 1, fontWeight = 400) => {
+        ctx.font = `${fontWeight} ${fontSize}px ${fontFamily}`;
```
markov00, line 34: you can use the function `cssFontShorthand` here

monfera: good point! though it constructs than destructs an object for very little gain here, and they need the full Font spec (though it's a pro argument as well)

packages/charts/src/state/selectors/get_legend_size.ts
monfera: I'll put this and a couple of other feedback elements on our code improvement metaissue, and/or fix a couple of them in this PR

packages/charts/src/utils/bbox/canvas_text_bbox_calculator.ts
monfera: referenced in #1303

packages/charts/src/state/selectors/get_legend_size.ts
monfera: referenced in #1303

packages/charts/src/utils/bbox/canvas_text_bbox_calculator.ts
monfera: added a point to #1303

markov00: yep, I just prefer use the function so I don't have to remember the order of arguments

package.json
```diff
@@ -130,7 +130,6 @@
     "eslint-plugin-react-hooks": "^4.2.0",
     "eslint-plugin-unicorn": "^25.0.1",
     "html-webpack-plugin": "^4.5.2",
-    "husky": "^4.3.6",
```
nickofthyme, line 133: 🎉 🥂 🍾

nickofthyme (approved):
A cursory review, code changes LGTM. Thanks for always pushing code quality!

package.json
monfera: I've second thoughts, maybe we should retain `yarn pq`?

monfera: Though in my case at least, repo reformatting is already triggered automatically when saving a file (any file)

PR #1326, opened 2021-08-24
feat(xy): mutilayer time axis step 1

monfera:
## Summary

Adds a preview of the hierarchical time axis via a new multiaxis story, including the minor code update to realize the preexisting left-align specification for the horizontal axis ticks. Refactors and improves on a lot of the Cartesian axis related code

image

 ### BREAKING CHANGE

- feat: removes the axis deduplication feature
- fix: `showDuplicatedTicks` now causes a duplication check on the actual axis tick label (possibly yielded by `tickFormat` rather than the more general `labelFormat`)

## Details

- adds a hierarchical time axis story
- fix/feat: the small update to actually left-align the horizontal axis ticks
- BREAKING feat: removes the axis deduplication feature
- BREAKING fix: `showDuplicatedTicks` causes a duplication check on the actual axis tick label (possibly yielded by `Axis.tickLabel` rather than the more general `tickFormat`)
- fix: turned all `[].sort()` calls safe (some API inputs used to be modified in the past; also, brittle to future code changes)
- fix(test): makes certain array comparison tests use `toIncludeSameMembers` instead of `toEqual` so the test remain neutral about the order of array elements (not all arrays need to have a specific order)
- fix: `axisTickLabel` (populated by `Axis.labelFormat`) is now precomputed and used for bounding box calculation and wherever it's needed for the actual axis tick rendering, as the preexisting `label` (populated by `Axis.tickFormat`) was generic, used eg. for tooltip formatting
- test: custom font that's fairly usable in live storybook mode (font for a11y: Atkinson Hyperlegible)
- refactor: swathes of axis related code eg. 350 lines removed from `axis_utils.ts` alone, not counting the deletion of the axis dedupe fun (refactoring is guided by certain objectives such as clarity of data flow, performance, simple types, avoidance of special cases and bails, expressions instead of control structures, succinctness)

## Issues

Step 1 toward https://github.com/elastic/elastic-charts/issues/1310 and https://github.com/elastic/kibana/issues/51227

### Checklist


monfera:
jenkins test this

packages/charts/src/common/color_calcs.ts
```diff
@@ -162,14 +162,13 @@ export function getTextColorIfTextInvertible(
         : `rgba(${255 - tr}, ${255 - tg}, ${255 - tb}, ${to})`
       : textColor;
   }
-  if (textContrast === true && typeof textContrast !== 'number') {
+  if (textContrast === true) {
```
monfera, line 165: If it equals `true`, then it's a boolean, can't be a number

packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts
```diff
@@ -344,7 +344,10 @@ export function getTickLabelProps(
   const verticalAlign = getVerticalAlign(position, rotation, textAlignment?.vertical);
 
   const userOffsets = getUserTextOffsets(tickDimensions, textOffset);
-  const textOffsetX = getHorizontalTextOffset(maxLabelTextWidth, horizontalAlign) + userOffsets.local.x;
+  // getHorizontalTextOffset needs to be used for vertical axis labels, and vertical labels of horizontal axes
+  const textOffsetX =
+    (isHorizontalAxis(position) && rotation === 0 ? 0 : getHorizontalTextOffset(maxLabelTextWidth, horizontalAlign)) +
+    userOffsets.local.x;
```
monfera, line 350: This was the only change needed for the multilayer axis prototype (it's needed for the left aligned tick labels)

monfera:
jenkins test this

packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts
```diff
@@ -141,8 +140,7 @@ export function axisViewModel(
     maxLabelBboxHeight,
     maxLabelTextWidth,
     maxLabelTextHeight,
-    tickValues,
-    tickLabels,
+    tickLabels: tickLabels.join('|'),
```
monfera, line 143: This ensures that axes with identical tick sets get merged. The algorithmic complexity is roughly the same:
the `[].join` is O(n), but carrying around a single string, and comparing string equality is more efficient than retaining an array with poor memory locality and doing up to 3 checks in the O(n^2) comparison phase. The number of tick labels is bounded in practice to a few, or rarely, a dozen or two

packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts
```diff
@@ -344,123 +249,41 @@ export function getTickLabelProps(
...
+        ...rest,
+      }
+    : {
+        x: tickPosition,
+        y: isAxisTop ? axisSize.height - tickDimension - labelInnerPadding : tickDimension + labelInnerPadding,
+        offsetX: userOffsets.global.x,
+        offsetY: (isAxisTop ? -maxLabelBboxHeight / 2 : maxLabelBboxHeight / 2) + userOffsets.global.y,
+        ...rest,
+      };
 }
```
monfera, line 273: Uused functions

packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts
```diff
@@ -344,123 +249,41 @@ export function getTickLabelProps(
...
-      return { minRange: height, maxRange: 0 };
-  }
+): { minRange: number; maxRange: number } {
+  const horizontal = isHorizontalAxis(axisPosition);
+  const flipped = horizontal
+    ? chartRotation === -90 || chartRotation === 180
+    : chartRotation === 90 || chartRotation === 180;
+  return horizontal
+    ? { minRange: flipped ? width : 0, maxRange: flipped ? 0 : width }
+    : { minRange: flipped ? 0 : height, maxRange: flipped ? height : 0 };
```
monfera, line 286: 85% smaller by weight 😅

packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts
```diff
@@ -526,53 +319,35 @@ export function enableDuplicatedTicks(
   fallBackTickFormatter: TickFormatter,
   tickFormatOptions?: TickFormatterOptions,
 ): AxisTick[] {
-  const ticks = scale.ticks();
-  const allTicks: AxisTick[] = ticks.map((tick) => ({
+  const allTicks: AxisTick[] = scale.ticks().map((tick) => ({
     value: tick,
     // TODO handle empty string tick formatting
```
monfera, line 324: @nickofthyme  is this now handled below here? Or you meant something else?

monfera:
jenkins test this

monfera:
yarn test this

packages/charts/src/chart_types/xy_chart/utils/axis_utils.test.ts
```diff
@@ -991,7 +815,7 @@ describe('Axis computational utils', () => {
       x: 10,
     });
     expect(verticalAxisSpecWTitleGeoms?.size).toEqual({
-      width: 30,
+      width: 50,
```
markov00, line 818: what has been changed here that the width is actually 20px bigger?

monfera: I don't know, wanted to get input, don't immediately see

monfera: Just for the record, @markov00 has run the question in isolation and it yielded 50, so most likely, the 30 was some unintended state capture in the preexisting version of the test

packages/charts/src/chart_types/xy_chart/utils/series.ts
```diff
@@ -299,7 +299,7 @@ const getSortedDataSeries = (
   xValues: Set<string | number>,
   xScaleType: ScaleType,
 ): DataSeries[] =>
-  dataSeries.map(({ data, ...rest }) => ({
+  [...dataSeries].map(({ data, ...rest }) => ({
     ...rest,
     data: data.sort(datumXSortPredicate(xScaleType, [...xValues.values()])),
   }));
```
markov00, line 305:
```suggestion
   dataSeries.map(({ data, ...rest }) => ({
    ...rest,
    data: [...data].sort(datumXSortPredicate(xScaleType, [...xValues.values()])),
  }));
```
It looks like we need to shallow copy the `data` object that you are sorting, not the `dataSeries` one that is not changed

packages/charts/src/specs/settings.tsx
```diff
@@ -49,7 +49,7 @@ import {
 /** @public */
 export interface LayerValue {
   /**
-   * The category value as retrieved by the `groupByRollup` callback
+   * The category value as retrieved by the `groupByRollup` callback§
```
markov00, line 52: typo or wanted symbol?

packages/charts/src/chart_types/xy_chart/state/selectors/compute_axis_ticks_dimensions.ts
```diff
@@ -20,56 +27,61 @@ import { getAxisSpecsSelector, getSeriesSpecsSelector } from './get_specs';
...
+                  return sizes;
+                },
+                { maxLabelBboxWidth: 0, maxLabelBboxHeight: 0, maxLabelTextWidth: 0, maxLabelTextHeight: 0 },
+              );
+              axesTicksDimensions.set(id, { ...maxLabelSizes, isHidden: axisSpec.hide && gridLineVisible });
+            } else {
+              Logger.warn(`Cannot compute scale for axis spec ${axisSpec.id}. Axis will not be displayed.`);
+            }
+          }
+          return axesTicksDimensions;
```
markov00, line 83: In cases like that one I personally prefer to return early than create a very deep nested function.
In this case we can break and return early with `if(!gridlineVisible && hide) return axesTicksDimensions`
and also return early if the scale is not set showing a warning.
This to me simplify a bit the readiness of the code and avoid the 9x2 indentation spaces

packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts
```diff
@@ -684,164 +432,71 @@ export function getAxesGeometries(
...
+    barsPadding,
+    enableHistogramMode,
+  );
+  return [...axisDimensions].reduce(
+    (acc: PerSideDistance & { geoms: AxisGeometry[] }, [axisId, axisDim]: [string, TickLabelBounds]) => {
       const axisSpec = getSpecsById<AxisSpec>(axisSpecs, axisId);
-
-      if (!axisSpec) {
-        return acc;
+      if (axisSpec) {
```
markov00, line 447: similar to a previous comment: what is the benefit of keep open a long if instead of returning earlier?
https://github.com/elastic/kibana/blob/master/STYLEGUIDE.mdx#returnthrow-early-from-functions

packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts
```diff
@@ -344,123 +215,35 @@ export function getTickLabelProps(
...
-    offsetY: offsetY + userOffsets.global.y,
-    textOffsetX,
-    textOffsetY,
-    horizontalAlign,
-    verticalAlign,
-  };
-}
-
-/** @internal */
-export function getVerticalAxisTickLineProps(
```
markov00, line 381: these functions are not used anywhere except in tests, so it's fine to remove them

packages/charts/src/chart_types/xy_chart/utils/axis_utils.test.ts
```diff
@@ -209,107 +190,6 @@ describe('Axis computational utils', () => {
...
-        1,
-        textMeasure,
-        NO_ROTATION,
-        axes,
-        (v) => `${v}`,
-      );
-      expect(axisDimensions).toBe(null);
-    }));
-
-  test('should compute axis dimensions with timeZone', () => {
```
markov00, line 261: I think we should introduce some VRTs to replace those tests with different timeZone.
We already have a story to change the timezone, but we just test for UTC, we should add screenshots for the other timezones: https://elastic.github.io/elastic-charts/?path=/story/scales--timezone-configuration&globals=theme:light
(can be done in a different PR)

packages/charts/src/chart_types/xy_chart/utils/axis_utils.test.ts
```diff
@@ -7,45 +7,44 @@
  */
 
 import { DateTime } from 'luxon';
+// @ts-ignore
 import moment from 'moment-timezone';
 
+import 'jest-extended';
```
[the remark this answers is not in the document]
monfera: Indeed it's not needed for the runtime, it's for the types only, otherwise it'll put red squigglies under the extended method. I read the route of adding the import expressly on iirc the gh site of the `jest-extended` repo

packages/charts/src/chart_types/xy_chart/utils/series.ts
monfera: Thanks, what was I thinking with this? 🤣

packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts
monfera: Thanks! I'm totally open to changing these couple of places I added; glad to chat about the pros/cons eg. me highlighting what it buys, and if still wanted, do this minor change in this PR or the next one

packages/charts/src/chart_types/xy_chart/utils/axis_utils.test.ts
monfera: Thanks, I'll probably need help with covering the logic sufficiently with VRTs

packages/charts/src/chart_types/xy_chart/state/selectors/compute_axis_ticks_dimensions.ts
monfera: I'd like to make it so that there's no need for either the `if` wrapping or an early return, though probably not in this PR. There's some level of pain both with early bail and `if` wrapping, so the best thing is to fix the root cause 😃

monfera: ... just to show progress along these lines, a lot of bail/wrapping needs had already been removed in this PR, simply by using diferent function boundaries, so there's existence proof that these can be reduced or eliminated with reasonable code changes

monfera: This is one example, one reduction instead of two ones 46d02215816f3ebf5e1766c1a2ebc1dda2e169a6

packages/charts/src/specs/settings.tsx
monfera: b22c647a1fce16c8e5f47ecaf003e64f6962a4f1

packages/charts/src/chart_types/xy_chart/utils/series.ts
monfera: b22c647a1fce16c8e5f47ecaf003e64f6962a4f1

packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts
monfera: Resolving it here, keeping the other equivalent note

packages/charts/src/chart_types/xy_chart/utils/axis_utils.test.ts
```diff
@@ -1003,33 +827,25 @@ describe('Axis computational utils', () => {
...
-      0,
-      true,
+      { top: cumTopSum, bottom: cumBottomSum, left: cumLeftSum, right: cumRightSum },
     );
-
     const expectedLeftAxisPosition = {
       dimensions: {
         height: 100,
-        width: 40,
+        width: 50,
```
monfera, line 842: There was a bunch of +10 changes when switching from `getAxisPosition` (whose tests got some direct input eg. the last args `10, 0, true`) to `getPosition` which feeds from the slightly more composite input arguments—I'd appreciate eyeballs on it, and maybe a suggestion as to how to best alter the input so the output value expectations need not change (can be reverted)

packages/charts/src/chart_types/xy_chart/utils/axis_utils.test.ts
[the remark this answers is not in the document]
nickofthyme: Ugh this is so dumb Idk, what is the issue but I'll have to dig deeper another time.

packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts
nickofthyme: Not sure. I think that was just to handle someone passing `() => ''` to the `tickFormat`.

packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts
nickofthyme: Yeah I'm not sold early return is best either. I don't see any performance pros and cons. But some times the fall-through case is usefully to where you don't have to do something like...

```js
const myFunc(value: any) {
  if (!value) return false;
  if (value % 2) return true;
  return false;
};
```

I know it is in the kibana style guide but I don't love it.

packages/charts/src/chart_types/xy_chart/utils/series.ts
```diff
@@ -490,8 +490,7 @@ function getSeriesNameFromOptions(
   }
 
   return (
-    options.names
-      .slice()
+    [...options.names]
```
nickofthyme, line 493: I'm curious what your reasoning behind this?

image

packages/charts/src/common/color_calcs.ts
nickofthyme: Nice catch. Wonder why TS doesn't say anything 🤔

storybook/stories/axes/12_duplicate_ticks.story.tsx
```diff
@@ -8,7 +8,7 @@
 
 import { boolean, select } from '@storybook/addon-knobs';
 import { DateTime } from 'luxon';
-import moment from 'moment-timezone';
+import * as moment from 'moment-timezone';
```
nickofthyme, line 11: You said this throws a runtime error?

storybook/style.scss
```diff
@@ -1,5 +1,25 @@
...
+/* atkinson-hyperlegible-700 - latin */
+@font-face {
+  font-family: 'Atkinson Hyperlegible';
+  font-style: normal;
+  font-weight: 700;
+  src: local(''), url('../public/fonts/atkinson-hyperlegible-v1-latin-700.woff2') format('woff2'),
+    /* Chrome 26+, Opera 23+, Firefox 39+ */ url('../public/fonts/atkinson-hyperlegible-v1-latin-700.woff')
+      format('woff'); /* Chrome 6+, Firefox 3.6+, IE 9+, Safari 5.1+ */
+}
+
```
nickofthyme, line 22: What's so special about this font?

nickofthyme (commented):
Code changes LGTM. Didn't have time yet to go through the axis diff in great detail.

I think we should have a replacement VRT for every test that was deleted before merging this PR.

storybook/style.scss
monfera: It needs not be that special, though I think it looks rad as well as useful, added https://github.com/elastic/elastic-charts/issues/1365 as a brief response 🤣

packages/charts/src/chart_types/xy_chart/state/selectors/compute_axis_ticks_dimensions.ts
markov00: totally agree, let's see in a different PR if we can get to the root cause

packages/charts/src/chart_types/xy_chart/utils/series.ts
markov00: ⬆️

storybook/style.scss
markov00: I think we can probably use `@font-face` with the current Inter font, applying the font settings for disambiguation, and reuse that in the canvas. I will try when this PR is merged

packages/charts/src/chart_types/xy_chart/utils/axis_utils.test.ts
markov00: The changes are valid: in the previous code, the `getAxisPosition()` function receive the tickDimension as parameter. Instead on the current one the tick dimension is computed within the function.
The previous test used a tickDimension of `10` where in your code, the. tick dimension is `20` the right value that comes from the AxisStyle, so good to go with this change

markov00 (approved):
I've tested it locally and works great (with the various fixed to the tickFormatter/labelFormatter now the tooltip header works great)

I think is good to merge after solving the legit Nick's comments

packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts
monfera: Yes the single return is nice, esp. with our pattern with `[]`reduce` that efficiently updates a reduction object in place, where you always return the object. I seem to prize data flow simplicity (one path instead of more; also, if we follow this coding style, the code reader can assume single-return rather than jumping in, "OK let's first discover the combinatorics of control and return structures". There's cost too, losing a couple of columns from the source text.

Having said this I'd like to strive for coding conventions where early bails are prevented through assurances that the thing being processed will always enjoys the ride in STP ([straight through processing](https://en.wikipedia.org/wiki/Straight-through_processing))

packages/charts/src/chart_types/goal_chart/layout/viewmodel/geoms.ts
```diff
@@ -405,8 +405,7 @@ export function geoms(
   const angleScale = (x: number) => angleStart + (angleRange * (x - domain[0])) / domainExtent;
   const clockwise = angleStart > angleEnd; // todo refine this crude approach
 
-  const geomObjects = abstractGeoms
-    .slice()
+  const [...geomObjects] = abstractGeoms
     .sort((a, b) => a.order - b.order)
```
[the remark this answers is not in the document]
monfera: Thanks for the catch, I wanted to re-spread the `abstractGeoms` but ended up committing this weird typo

packages/charts/src/chart_types/xy_chart/utils/series.ts
monfera: It's hard to effectively micro-benchmark this one, it's bound to be both negligible and stochastic w.r.t. which browsers, and which of their versions do what kind of flow analytics, as they have the same meaning for this kind of use. So I simply used the less verbose, less noisy, somewhat visually more explicit version, which also avoids the need to add yet another line or element into a possibly cascading flow of fluent-style methods. For me at least, I can see 
```
[...someArray].sort()  // neat, a fresh [...].sort MUST be safe
```
adjacent in a single line is more primally safe and "re-wrapped" than having to follow a cascade
```
  const result = someArray  // hmm is it protected? got to see further
                  .slice()
                  .sort()  // hmm are we sorting a buffered array? check stuff directly upstream
```

Do y'all prefer `[].slice` over spread? Glad to switch back, it's a super minor point that fell one way while coding but can fall the other

storybook/stories/axes/12_duplicate_ticks.story.tsx
monfera: `import * as moment from 'moment-timezone';` caused a runtime error for me when running `yarn test`

monfera:
@nickofthyme @markov00 
> I think we should have a replacement VRT for every test that was deleted 

My thought as well, to lock in the outcomes of the logic incl. corner case handling whose unit tests were removed. It's critical to aim for an as high ratio of code coverage with VRT as possible, hopefully 100% over time, at least everything that ultimately influences what gets rendered statically or upon interaction 

> before merging this PR

Due to the high volume of code that's changed laboriously and non-automatically, merging this soon would avoid nasty merge conflicts, and the much more limited (merge conflict free) test updates could be an immediate follow-up. Besides the merge conflict angst, the PR contains a few fixes (and VRT's them) and added safety so the balance may be positive. An earlier merge brings up potential regressions sooner. My recent stint with refactoring PRs removed around a couple thousand runtime lines and changed another few thousands, and where there's change there's risk of regression, so being exposed to usage feedback earlier has some merit.

Whether you prefer merging sooner or you'd rather await VRT coverage, I'd like to get help from the two of you, as sometimes it's not immediately clear what a unit test meant to lock in, in terms of user input to final results. As you have more `elastic-charts` experience with the Cartesians, and wrote the tests, it's also more efficient to go parallel on this, or at least walk me through this on a call so we gather what images can cover the unit tests that got disused for one reason or another

Also linking Marco's "one for the team" [PR](https://github.com/elastic/elastic-charts/pull/1363) as it covers some of what went away in unit tests here

packages/charts/src/chart_types/goal_chart/layout/viewmodel/geoms.ts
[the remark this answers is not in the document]
monfera: Done in 44f76edb658ae9ecc3ca1f057dc20893144b8008

nickofthyme:
@monfera sure that's understandable. I think we just need to add some some storybook tests to replicate some of the unit test setup.

storybook/stories/interactions/11_brush_time.story.tsx
```diff
@@ -9,7 +9,7 @@
 import { action } from '@storybook/addon-actions';
 import { boolean } from '@storybook/addon-knobs';
 import { DateTime } from 'luxon';
-import moment from 'moment-timezone';
+import * as moment from 'moment-timezone';
```
monfera, line 12: @nickofthyme this occurrence passed in runtime; the ones that didn't pass were like usages in the unit test files that are run via `yarn test`

storybook/stories/axes/12_duplicate_ticks.story.tsx
nickofthyme: Ok I'll take a look.

packages/charts/src/chart_types/xy_chart/utils/series.ts
nickofthyme: No preference really just curious 🤨. I think we need to just be vigilant about people removing the spread before sort thinking it's not needed.

nickofthyme (approved):

packages/charts/src/chart_types/xy_chart/utils/axis_utils.test.ts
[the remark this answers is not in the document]
nickofthyme: To be clear, this is for typing only in the IDE not in the `ts-jest` process.

PR #1329, opened 2021-08-24
fix(partition): small multiples panel title color

markov00:
## Summary

When rendering a pie chart in small multiples, the panel title now correctly use the theme `axisPanelTitle` configuration.

image

## Details

The previous implementation made use of a static color and font for the panel title.
The code was refactored to use the `theme.axes.axisPanelTitle` fontface and size.

A small fix was also pushed to the VRT background/theme hooks to correctly pass the backgroundColor to the BackgroundProvider.

## Issues

fix #1327

### Checklist

- [x] The proper **chart type** label has been added (e.g. `:xy`, `:partition`)
- [x] The proper **feature** labels have been added (e.g. `:interactions`, `:axis`)
- [x] All related issues have been linked (i.e. `closes #123`, `fixes #123`)
- [x] Unit tests have been added or updated to match the most common scenarios
- [x] The proper documentation and/or storybook story has been added or updated
- [x] Visual changes have been tested with all available themes including `dark`, `light`, `eui-dark` & `eui-light`

PR #1336, opened 2021-08-25
fix(partition): small multiples panel title color (#1329) [34.1.x]

nickofthyme:
Backports the following commits to 34.1.x:
 - fix(partition): small multiples panel title color (#1329)


markov00 (commented):
LGTM

PR #1342, opened 2021-08-30
feat(heatmap): add text color contrast to heatmap cells

### LINKED ISSUE #1296: [Heatmap] Cell values should automatically invert color based on contrast
**Describe the bug**

Cell values are displayed with the selected color without taking into account the background contrast.
Here's an example:

**Expected behaviour**

The background color should be considered when assigning the font color to the cell value accordingly with the contrast measurement.
Extra feature: add a way to pass the page background color as well to handle constrast measurement with `rgba` values (we hit this bug already in table Lens dynamic coloring).

**Kibana Cross Issues**
[Kibana cross issue](https://github.com/elastic/kibana/issues/105410)

**Checklist**

rshen91:
## Summary
Text will now take into account the background color of the cell and adjust its contrast to 4.5. 

## Details
Still to do:
- [x] Confirm that true textContrast will default to 4.5
- [x] Fix the makeHighContrastColor function for cases where contrast is hard/impossible (commit 2eaa8e6)
- [x] Take into account the background color of the chart in addition to the background color (commit 5c6b739)

## Issues
Closes #1296 

### Checklist

-  [x] New public API exports have been added to `packages/charts/src/index.ts`


markov00:
@rshen91 I've few doubts on your todo list:
> Fix bullet chart stories without having contrast defined

we should fix the bullet chart in a different PR not this one. I'm also not sure what you mean by fix stories without having contrast defined.

> Take into account the background color of the chart in addition to the background color

Can you please explain to me why? in heatmap we are rendering the text on top of a cell, that is colored. Why we need to consider the background color of the chart? for semi-transparent colors?

rshen91:
Hi @markov00 good questions. Sorry, these points might have been misleading:

> Fix bullet chart stories without having contrast defined
> 
> we should fix the bullet chart in a different PR not this one. I'm also not sure what you mean by fix stories without having contrast defined.
I have some stories locally that look like they're failing these changes but they seem to be fine in vrt testing in master. I'm thinking because they're not failing in master then it's fine? 


In https://github.com/elastic/elastic-charts/issues/1296, they mention

> Extra feature: add a way to pass the page background color as well to handle constrast measurement with rgba values (we hit this bug already in table Lens dynamic coloring).

Should I do this in a future PR? I might be misunderstanding thank you!

packages/charts/src/chart_types/xy_chart/renderer/canvas/primitives/text.ts
```diff
@@ -31,13 +34,20 @@ export function renderText(
...
   withContext(ctx, () => {
     ctx.translate(origin.x, origin.y);
     ctx.rotate(degToRad(angle));
     ctx.translate(translateX, translateY);
     ctx.scale(scale, scale);
-    ctx.fillStyle = font.textColor;
+    ctx.fillStyle =
+      textContrast && cellColor
+        ? fillTextColor(font.textColor, true, textContrast, RGBtoString(cellColor), convertPageBackgroundColor)
+        : font.textColor;
```
markov00, line 50: I think we should not patch the code at the renderer level.
This rendering method should be agnostic and should not know If we are rendering a heatmap cell text or something else.
It renders a text with a color specified by the `font` property.
The optimal solution is to move the calculation of the correct `textColor` within the cell calculation code, you can probably add to the `Cell` type a new property to describe the textColor and change the way the `renderText` method is called using that `textColor` instead of the default one coming from the `config` like:
```
config.cell.label.visible &&
        withContext(ctx, () => {
          // render text on cells
          const { x, y } = heatmapViewModel.gridOrigin;
          ctx.translate(x, y);
          filteredCells.forEach((cell) => {
            if (cell.visible && !Number.isNaN(cell.fontSize))
              renderText(ctx, { x: cell.x + cell.width / 2, y: cell.y + cell.height / 2 }, cell.formatted, {
                ...config.cell.label,
                textColor: cell.textColor,
              });
          });
        }),
```

packages/charts/src/common/color_calcs.ts
```diff
@@ -117,9 +118,14 @@ export function makeHighContrastColor(foreground: Color, background: Color, rati
...
-    // catch if the ideal contrast may not be possible
+    // catch if the ideal contrast may not be possible, switch to the other extreme color contrast
     if (scaledOldContrast === scaledContrast) {
-      break;
+      const contrastColor =
+        originalhighContrastTextColor === 'rgba(255, 255, 255, 1)' || originalhighContrastTextColor === '#fff'
+          ? '#000'
+          : '#fff';
+      // make sure the new text color hits the ratio, if not, then return the scaledContrast since we tried earlier
+      return getContrast(contrastColor, background) > ratio ? contrastColor : scaledContrast.toString();
```
markov00, line 128: two comments here:
1- can you explain why you have made these changes? 
2- instead of use both hex and rgba values, can you please always use `rgba` strings instead where possible?

markov00 (changes requested):

packages/charts/src/chart_types/xy_chart/renderer/canvas/primitives/text.ts
```diff
@@ -31,13 +34,20 @@ export function renderText(
   translateX: number = 0,
   translateY: number = 0,
   scale: number = 1,
+  pageBackgroundColor: Color = '#000000',
+  cellColor?: RgbObject,
 ) {
+  const { textContrast } = font;
+  const convertPageBackgroundColor = getHexValue(pageBackgroundColor);
```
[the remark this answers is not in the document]
rshen91: For sure! So the issue I was facing was using rgba(0,0,0,0) instead of #00000 both pointing to a transparent color. The fillTextColor() has a warning for rgba values with an alpha value of 0. But yes, overall I do think it makes the most sense to have the fillTextColor() just using rgba values.

packages/charts/src/common/color_calcs.ts
rshen91: 1. I was noticing one of the cell background colors (rgba(44, 127, 184, 1),  #2c7fb8) was going to white text with worse contrast than if it was black text. I hunted the issue down to this edge case and added this to catch this issue in the future.

2. I agree, I will make sure to refactor to use rgba values where possible instead of hex. I got hung up on the transparent hex value being passed by default but I agree that the rgba value makes the most sense. Will include this in my refactoring, thanks!

packages/charts/src/chart_types/xy_chart/renderer/canvas/primitives/text.ts
rshen91: Thank you! I think I am following - f32b32d for changes

packages/charts/src/chart_types/heatmap/layout/types/config_types.ts
```diff
@@ -87,6 +87,7 @@ export interface Config {
       align: TextAlign;
       baseline: TextBaseline;
       visible: boolean;
+      textContrast: boolean | number;
```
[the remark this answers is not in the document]
rshen91: f32b32d for changes thanks!

packages/charts/src/chart_types/heatmap/layout/types/config_types.ts
```diff
@@ -20,9 +20,9 @@ export interface Config {
   margin: { left: SizeRatio; right: SizeRatio; top: SizeRatio; bottom: SizeRatio };
   maxRowHeight: Pixels;
   maxColumnWidth: Pixels;
+  pageBackgroundColor: Color;
```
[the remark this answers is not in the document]
rshen91: Can you let me know if I did this right in f32b32d? thanks!

packages/charts/src/common/color_library_wrappers.ts
```diff
@@ -31,6 +32,11 @@ export const defaultD3Color: D3RGBColor = d3Rgb(defaultColor.r, defaultColor.g,
 /** @internal */
 export type OpacityFn = (opacity: number, seriesOpacity?: number) => number;
 
+/** @internal */
+export function getHexValue(color: Color) {
+  return RGBAToHex(RGBtoString(getColor(color)));
+}
+
```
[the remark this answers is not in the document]
rshen91: Refactored in f32b32d thank you!

rshen91:
jenkins, test this

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
```diff
@@ -205,6 +209,13 @@ export function shapeViewModel(
       value: d.value,
       visible: !isValueHidden(d.value, bandsToHide),
       formatted: spec.valueFormatter(d.value),
+      textColor: fillTextColor(
+        config.cell.label.textColor,
+        true,
+        config.cell.label.textContrast ? 4.5 : false,
+        color,
+        theme.background.color === 'transparent' ? '#000000' : theme.background.color,
```
markov00, line 217: if with this code you mean: if the theme background color is `transparent` keyword use `black` as the fallback, I think we need to change that to use `white` instead, as the default theme is the LIGHT one
Use white rgba if possible `rgba(255,255,255,1)`

packages/charts/src/common/color_calcs.ts
```diff
@@ -117,9 +118,14 @@ export function makeHighContrastColor(foreground: Color, background: Color, rati
...
     contrast = getContrast(highContrastTextColor, background);
     const scaledContrast = Math.round(contrast * precision) / precision;
-    // catch if the ideal contrast may not be possible
+    // catch if the ideal contrast may not be possible, switch to the other extreme color contrast
     if (scaledOldContrast === scaledContrast) {
-      break;
+      const contrastColor =
+        originalhighContrastTextColor === 'rgba(255, 255, 255, 1)' || originalhighContrastTextColor === '#fff'
+          ? '#000'
+          : '#fff';
```
markov00, line 126:
```suggestion
          ? 'rgba(0,0,0,1)'
          : 'rgba(255,255,255,1)';
```

packages/charts/src/common/color_library_wrappers.ts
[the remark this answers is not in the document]
markov00: it looks like the function still there

packages/charts/src/common/fill_text_color.ts
```diff
@@ -55,7 +55,10 @@ export function fillTextColor(
   }
 
   const adjustedTextColor: string | undefined = textColor;
-  const containerBackground = combineColors(shapeFillColor, backgroundColor);
+  const containerBackground = combineColors(
+    shapeFillColor,
+    backgroundColor === 'transparent' ? '#000000' : backgroundColor,
```
markov00, line 60: similar to a previous comment:
the background fallback for transparent color should be white (because the default theme is the light one)
So if you need to replace the `transparent` with a valid fallback please use `rgba(255,255,255,1)`

storybook/stories/heatmap/heatmap.stories.tsx
```diff
@@ -13,3 +13,4 @@ export default {
 export { Example as basic } from './1_basic.story';
 export { Example as time } from './3_time.story';
 export { Example as categorical } from './2_categorical.story';
+export { Example as textColorContrast } from './4_color_text.story';
```
markov00, line 16: instead of creating a new vrt for that, just show by default the value labels on the categorical example,we can save two new tests (please also remove the added test as this replace the need of the vrt added)

packages/charts/src/chart_types/xy_chart/renderer/canvas/primitives/text.ts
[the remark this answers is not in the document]
markov00: `rgba(0,0,0,0)` is `transparent`
If you want to use black is `rgba(0,0,0,1)` with full opacity

storybook/stories/heatmap/heatmap.stories.tsx
rshen91: I think I did this in 0a9a914 thank you

packages/charts/src/common/fill_text_color.ts
rshen91: Good catch thank you 0a9a914

packages/charts/src/common/color_calcs.ts
rshen91: Yikes sorry! 0a9a914 thanks

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
rshen91: 0a9a914 thank you!

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
```diff
@@ -205,6 +209,13 @@ export function shapeViewModel(
       value: d.value,
       visible: !isValueHidden(d.value, bandsToHide),
       formatted: spec.valueFormatter(d.value),
+      textColor: fillTextColor(
+        config.cell.label.textColor,
+        true,
+        config.cell.label.textContrast ? 4.5 : false,
```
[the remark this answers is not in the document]
rshen91: Ok I think Im following 0a9a914 thanks!

packages/charts/src/chart_types/heatmap/layout/types/config_types.ts
```diff
@@ -87,6 +86,7 @@ export interface Config {
       align: TextAlign;
       baseline: TextBaseline;
       visible: boolean;
+      textContrast: boolean;
```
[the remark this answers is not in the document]
rshen91: Ok cool sounds good to me 0a9a914 for changes thanks!

packages/charts/src/common/color_library_wrappers.ts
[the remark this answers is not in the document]
rshen91: removed in 0a9a914 sorry!

markov00 (approved):
I'm ok merging this PR for now.
I will take a look and clean up the color contrast functions in a next PR because it doesn't look very clean and readable.

PR #1350, opened 2021-09-01
refactor(xy): drop triple equals true on booleans

wayneseymour:
## Summary

Refactor `textContrast` check to use `truthy`,
instead of using `=== true`.

## Details

Remove a little duplication, 
while in the "neighbourhood".

## Issues

Helps with: https://github.com/elastic/elastic-charts/issues/1303
 - Addresses `check for === true, === false, return true etc. redundancy candidates and eliminate`


packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts
```diff
@@ -534,9 +534,8 @@ export function enableDuplicatedTicks(
     position: (scale.scale(tick) ?? 0) + offset,
   }));
 
-  if (axisSpec.showDuplicatedTicks === true) {
-    return allTicks;
-  }
+  if (axisSpec.showDuplicatedTicks) return allTicks;
+
   return getUniqueValues(allTicks, 'label', true);
```
monfera, line 539: Would prefer `return axisSpec.showDuplicatedTicks ? allTicks : getUniqueValues(allTicks, 'label', true);`

packages/charts/src/common/color_calcs.ts
```diff
@@ -142,6 +142,12 @@ export function colorIsDark(color: Color): boolean {
   return luminance < 0.2;
 }
 
+const colors = (tr: any) => (tg: any) => (tb: any) => ({
```
monfera, line 145: Good idea to abstract out this util indeed! In this case there's no current or foreseeable partial application per color channel, so it could be a single parameter list, and the symbols can simplify to r, g, b as the t prefix carries no meaning in the scope of this function

monfera:
Hi @wayneseymour thanks for the contrib! 

There's an eslint issue: we recently got rid of husky (blame me), so the code autoformatting on commit went away, we pretty much lean on our editors linking to the ever current eslint rules, not just for the red squigglies but also for reformatting on autosave:
image

The "Semantic Pull Request" check was red as it enforces commitizen rules, in this case, 
- one of refactor, test, chore, feat, 
- plus, if applicable, the area in braces, such as xy, partition, heatmap etc.

It's late and I haven't checked the actual logic, but it looks like at least the 3rd visual regression test may have picked up a logic change, worth checking, as here the outcomes are expected to remain the same. Glad to get you started on ways to run the VRT locally for a sometimes faster turnaround

packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts
wayneseymour: Oh totally!  Much cleaner!

packages/charts/src/common/color_calcs.ts
wayneseymour: Ah, that makes sense.

markov00: This function is basically a way to invert color channels.
There was also another, definitely simplifiable, piece of code:
```
? to === undefined
        ? `rgb(${255 - tr}, ${255 - tg}, ${255 - tb})`
        : `rgba(${255 - tr}, ${255 - tg}, ${255 - tb}, ${to})`
```
that just create an `rgb` or an `rgba` color string depending if the opacity is defined or not.

I think we can easily replace this by always rendering `rgba` strings (non `rgbA` 😉) where the opacity is always `1` if the param is `undefined`, something along those lines:

```
function invertColor(r:RGB, g:RGB, b:RGB, o = 1) {
   return `rgba(${255 - r}, ${255 -g},${255 - b},${o})`;
}
```

packages/charts/src/common/color_calcs.ts
```diff
@@ -155,27 +161,29 @@ export function getTextColorIfTextInvertible(
 ): Color {
   const inverseForContrast = specifiedTextColorIsDark === backgroundIsDark;
   const { r: tr, g: tg, b: tb, opacity: to } = stringToRGB(textColor);
-  if (!textContrast) {
```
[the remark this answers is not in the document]
wayneseymour: What I did is just as "unattractive": https://github.com/elastic/kibana/blob/master/src/dev/code_coverage/ingest_coverage/either.js#L49

packages/charts/src/common/color_calcs.ts
wayneseymour: Wow, that's waaaaaay more simple!

wayneseymour: Full disclosure, I had no idea of the context of all this **rgb** "stuff" lol

wayneseymour: Ok, I'm going to push what I have so far, but I'm also going to go back through and change it to have this `invertColor` fn.  I really like the simplicity.

monfera:
jenkins test this

packages/charts/src/common/color_calcs.ts
```diff
@@ -142,6 +142,10 @@ export function colorIsDark(color: Color): boolean {
   return luminance < 0.2;
 }
 
+const reducer = (acc: any, curr: any, idx: any) => (idx !== 0 ? `${acc}, ${curr}` : `${curr}`);
+
+const invert = (r: number, g: number, b: number) => [r, g, b].map((x) => 255 - x).reduce(reducer, '');
```
monfera, line 147: `[].join(', ')` removes the need for the `[].reduce` and the overly generically named `reducer`, though even with that, the mild repetition 
```
const invert = (r: number, g: number, b: number) => `${255 - r}, ${255 - g}, ${255 - b}`
```
would likely lead to fewer cycles, fewer bytes allocated and a bit shorter code, plus the code reader sees what's going on on a 1st blink. I still wouldn't name it `invert` as it returns a partial CSS color string, besides the inversion.

Combined with Marko's pitch for using `rgba` even if `rgb` would suffice, it could be
```
const invertedColor = (r: number, g: number, b: number, a: number) => `rgba(${255 - r}, ${255 - g}, ${255 - b}, a)`
```

wayneseymour:
jenkins test this

packages/charts/src/common/color_calcs.ts
wayneseymour: Ahhh, ok.  That makes a lot of sense.  I'll update.

wayneseymour: Does that mean `rgb` and `rgbA` are nearly interchangeable?

wayneseymour: Ahh, I see, it's about opacity. I get it now.

monfera:
@wayneseymour the resolution for the `axis_utils.ts` conflict is this: 
```
  return axisSpec.showDuplicatedTicks ? allTicks : getUniqueValues(allTicks, 'axisTickLabel', true);
```
I did this conversion to ternary when we looked at axis code together, but then I also introduced `axisTickLabel` so you can just accept the current one on merge

wayneseymour:
Running the vrt now on my latest changes, if it passes, I'll push

packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts
```diff
@@ -320,6 +320,7 @@ export function enableDuplicatedTicks(
     axisTickLabel: (axisSpec.labelFormat ?? axisSpec.tickFormat ?? fallBackTickFormatter)(tick, tickFormatOptions),
     position: (scale.scale(tick) ?? 0) + offset,
   }));
+
```
wayneseymour, line 323: This could be dropped.

packages/charts/src/common/color_calcs.ts
```diff
@@ -159,27 +161,24 @@ export function getTextColorIfTextInvertible(
...
+
+  function contrast(ratio?: any) {
     return inverseForContrast
       ? to === undefined
-        ? makeHighContrastColor(`rgb(${255 - tr}, ${255 - tg}, ${255 - tb})`, backgroundColor, textContrast)
-        : makeHighContrastColor(`rgba(${255 - tr}, ${255 - tg}, ${255 - tb}, ${to})`, backgroundColor, textContrast)
-      : makeHighContrastColor(textColor, backgroundColor, textContrast);
+        ? makeHighContrastColor(`rgb(${inverted})`, backgroundColor, ratio || undefined)
+        : makeHighContrastColor(`rgba(${inverted}, ${to})`, backgroundColor, ratio || undefined)
+      : makeHighContrastColor(textColor, backgroundColor, ratio || undefined);
```
wayneseymour, line 180: Now that I think about it, this could be simply `, ratio`...should be undefined if it's not passed.

wayneseymour: @monfera thoughts?

wayneseymour:
jenkins test this

wayneseymour:
jenkins test this

packages/charts/src/common/color_calcs.ts
markov00: I can explain my way why we should refactor this function in a different way:

1. calling `makeHighContrastColor` with an rgba or rgb color is exactly the same. The internal code take care of the values.
2. we can always write an `rgba` color starting from an `rgb` color adding opacity 1
3. the `ratio` or is defined and a number if not it should be left undefined living the `makeHighContrastColor` use the default value


```ts
export function getTextColorIfTextInvertible(
  textColorIsDark: boolean,
  backgroundIsDark: boolean,
  textColor: Color,
  textContrast: TextContrast,
  backgroundColor: Color,
): Color {
  const finalTextColor = textColorIsDark === backgroundIsDark ? invertColor(stringToRGB(textColor)) : textColor;
  const contrastRatio = typeof textContrast === 'number' ? textContrast : undefined;
  return textContrast === false
    ? finalTextColor
    : makeHighContrastColor(finalTextColor, backgroundColor, contrastRatio);
}

function invertColor({ r, g, b, opacity }: RgbObject): Color {
  return `rgba(${255 - r}, ${255 - g}, ${255 - b}, opacity: ${opacity})`;
}
```

I have some doubts on the real need of this function anyway, but if we extract it out of the context this is my solution

PR #1351, opened 2021-09-01
fix(heatmap): filter out tooltip picked shapes in x-axis area

### LINKED ISSUE #1215: Paginated heatmap shows tooltip outside the projection area
**Describe the bug**
A heatmap can be paginated: depending on the cell/grid configuration is possible to reduce the number of rows shown to fit into the screen. 
We track the mouse position outside the projection area. The geometry picking logic doesn't exclude the non-rendered cells causing the tooltip to appear when hovering near the axis but outside the projection area.

 

**To Reproduce**
1. Go to https://elastic.github.io/elastic-charts/?path=/story/heatmap-alpha--categorical
2. move the mouse near the bottom part of the chart, over the x axis
3. the tooltip, related to non-visible cells will appear 

**Expected behaviour**
The tooltip should only be active for the visible cells

**Screenshots**
Jun-22-2021 19-54-38

 - Elastic Charts: 30.x

**Additional context**
n/a

**Errors in browser console**
n/a

rshen91:
## Summary

This bug fix removes the tooltip when hovering over the x-axis. 

### Before

https://user-images.githubusercontent.com/20343860/131744782-b8ec9192-d36e-4d36-bd3c-652741ec77e6.mp4

### After

https://user-images.githubusercontent.com/20343860/131744852-35e31147-40e2-4d2e-952f-a8a8cae9cfd2.mp4

## Details
- [x] Adjust cursor logic 

## Issues
Closes #1215 

### Checklist

### DIFF
--- a/packages/charts/src/chart_types/heatmap/state/selectors/picked_shapes.ts
+++ b/packages/charts/src/chart_types/heatmap/state/selectors/picked_shapes.ts
@@ -10,6 +10,7 @@ import { GlobalChartState } from '../../../../state/chart_state';
 import { createCustomCachedSelector } from '../../../../state/create_selector';
 import { Cell, TextBox } from '../../layout/types/viewmodel_types';
 import { geometries } from './geometries';
+import { getGridHeightParamsSelector } from './get_grid_full_height';
 
 function getCurrentPointerPosition(state: GlobalChartState) {
   return state.interactions.pointer.current.position;
@@ -17,10 +18,13 @@ function getCurrentPointerPosition(state: GlobalChartState) {
 
 /** @internal */
 export const getPickedShapes = createCustomCachedSelector(
-  [geometries, getCurrentPointerPosition],
-  (geoms, pointerPosition): Cell[] | TextBox => {
+  [geometries, getCurrentPointerPosition, getGridHeightParamsSelector],
+  (geoms, pointerPosition, gridParams): Cell[] | TextBox => {
     const picker = geoms.pickQuads;
     const { x, y } = pointerPosition;
-    return picker(x, y);
+    const arrayPicker = picker(x, y) as Cell[];
+    return Array.isArray(picker(x, y))
+      ? arrayPicker.filter(({ y }) => y < gridParams.gridCellHeight * gridParams.pageSize)
+      : picker(x, y);
   },
 );
--- a/packages/charts/src/chart_types/heatmap/state/selectors/picked_shapes.ts
+++ b/packages/charts/src/chart_types/heatmap/state/selectors/picked_shapes.ts
@@ -22,9 +22,9 @@ export const getPickedShapes = createCustomCachedSelector(
   (geoms, pointerPosition, gridParams): Cell[] | TextBox => {
     const picker = geoms.pickQuads;
     const { x, y } = pointerPosition;
-    const arrayPicker = picker(x, y) as Cell[];
-    return Array.isArray(picker(x, y))
-      ? arrayPicker.filter(({ y }) => y < gridParams.gridCellHeight * gridParams.pageSize)
-      : picker(x, y);
+    const pickedData = picker(x, y);
+    return Array.isArray(pickedData)
+      ? pickedData.filter(({ y }) => y < gridParams.gridCellHeight * gridParams.pageSize)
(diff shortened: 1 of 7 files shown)


markov00 (commented):
Logically looks correct, do you think we can apply this `pickedShapes` filter directly into the `getPickedShapes` selector to control and fix that in a single place?

rshen91:
> Logically looks correct, do you think we can apply this `pickedShapes` filter directly into the `getPickedShapes` selector to control and fix that in a single place?

I'm definitely going to try -- the issue I'm facing is the return type of pickedShapes can be an array of Cells or TextBox so it's more complicated than I had hoped :)

rshen91:
@markov00 in 8358dd8 I cleaned it up in the picked_shapes selector but I was struggling to find a cleaner way to deal with the cursor logic in get_cursor_pointer.ts (like a way to avoid the length check in get_cursor_pointer). I was trying to move that length check into picked_shapes but kept facing issues with typescript and returning boolean values. Is there a cleaner way I'm overlooking or is the refactoring in 8358dd8 ok? Thanks

rshen91:
jenkins test this

packages/charts/src/chart_types/heatmap/state/selectors/picked_shapes.ts
```diff
@@ -10,17 +10,21 @@ import { GlobalChartState } from '../../../../state/chart_state';
...
-  (geoms, pointerPosition): Cell[] | TextBox => {
+  [geometries, getCurrentPointerPosition, getGridHeightParamsSelector],
+  (geoms, pointerPosition, gridParams): Cell[] | TextBox => {
     const picker = geoms.pickQuads;
     const { x, y } = pointerPosition;
-    return picker(x, y);
+    const arrayPicker = picker(x, y) as Cell[];
+    return Array.isArray(picker(x, y))
+      ? arrayPicker.filter(({ y }) => y < gridParams.gridCellHeight * gridParams.pageSize)
+      : picker(x, y);
```
markov00, line 28: My suggestion is: 
- first of all call the `picker` function only one time
- don't cast it to `Cell[]` because it is not required in the code
- the ternary check `isArray` already allows TS and also JS to define if the pickedData is an array of Cell or a TextBox 
 

```suggestion
    const pickedData = picker(x, y);
    return Array.isArray(pickedData)
      ? pickedData.filter(({ y }) => y < gridParams.gridCellHeight * gridParams.pageSize)
      : pickedData;
```

markov00:
Can you also please update the story with a knob that remove the filter on the BABY_NAME data so reviewers can test this easily without touching the code? thanks

packages/charts/src/chart_types/heatmap/state/selectors/picked_shapes.ts
rshen91: Thank you! 6785db6 for changes and added the story changes into this commit as well

storybook/stories/heatmap/2_categorical.story.tsx
```diff
@@ -15,7 +16,8 @@ import { BABYNAME_DATA } from '@elastic/charts/src/utils/data_samples/babynames'
 import { useBaseTheme } from '../../use_base_theme';
 
 export const Example = () => {
-  const data = BABYNAME_DATA.filter(([year]) => year > 1950 && year < 1960);
+  const filterData = boolean('filter BABYNAME_DATA', false);
+  const data = filterData ? BABYNAME_DATA : BABYNAME_DATA.filter(([year]) => year > 1950 && year < 1960);
```
markov00, line 20: Isn't that the inverse?
If I check the `filter` knob to enable the filter I should see only a partial set.
If the filter check is off, then I should see all the data available.
Two code change:
- reverse the ternary operation and change the default value of the checkbox to true
- rename the filter to: `filter dataset`. The BABYNAME_DATA is something we don't want to show to the user

markov00 (approved):
LGTM, good to merge after fixing the highlighted code change

storybook/stories/heatmap/2_categorical.story.tsx
rshen91: Yikes good catch thanks 38147b4

PR #1352, opened 2021-09-02
feat(heatmap): reduce font size to fit label within cells

### LINKED ISSUE #1295: [Heatmap] Values overflow cells overlapping other labels
**Describe the bug**

When cell values are shown in a heatmap, there's no overflow check for the value when displayed, leading to some unreadable scenarios like the following:

As shown in the image the overflow is both vertical and horizontal.

**Expected behaviour**

No clear "good" solution here, but some thoughts on:
* hide a label if another one is overflowing its space (leads to a partial display, but at least readable)
* scale the value font size based on the cell dimension (probably with an upper and lower bound size: stop increase after the upper bounds, hide below the lower bound?)

**Kibana Cross Issues**
[Kibana cross issue](https://github.com/elastic/kibana/issues/105410)

**Checklist**

markov00:
## Summary

We introduced the ability to scale the font size of numeric values within the heatmap cells if the label doesn't fit within the label using the default font size. The resulting font size is constrained within a user-configured range. If the label doesn't fit using the min font size, the label is not rendered.

### BREAKING CHANGE
The `config.label.fontSize` prop is replaced by `config.label.minFontSize` and `config.label.maxFontSize`. You can specify the same value for both properties to have a fixed font size.
The `config.label.align` and `config.label.baseline` props are removed from the `HeatmapConfig` object. These props handle the alignment of the text within the text and they are not currently compatible with the font size scaling feature introduced. Is not very common to change the alignment of text within rectangular boxes where the centered option is the most used one.

Screenshot 2021-09-02 at 11 25 28

## Details

The changes are actually suboptimal: 
- from the API point of view we can have a single union type for the `fontSize` like `Pixels | [minFontSize: Pixels, maxFontSize:Pixels]` or something along those lines. Is a union type but it should be converted to the extended form after the API layer.
- The value volatility can cause noticeable differences between the font sizes, if min and max are exaggerated. We can introduce an option to use the same min font size on every cell
- we can also introduce the ability to add padding around the text to avoid having text touching each other along the edges.  

## Issues

fix #1295 

### Checklist


packages/charts/api/charts.api.md
```diff
@@ -356,6 +356,8 @@ export interface Cell {
     // (undocumented)
     fill: Fill;
     // (undocumented)
+    fontSize: Pixels;
+    // (undocumented)
```
monfera, line 360: Can we add documentation to at least the new API elements, and maybe whatever is in the direct surroundings? If it's a goal that we increase doc coverage

markov00: I'm wondering if documenting self-explanatory properties like `fontSize` is necessary or not. What do you think?

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
```diff
@@ -177,6 +177,9 @@ export function shapeViewModel(
     };
   });
 
+  const cellWidthNoBorder = cellWidth - gridStrokeWidth * 2;
+  const cellHeightNoBorder = cellHeight - gridStrokeWidth * 2;
```
monfera, line 181: Nit: name could be `cellWidthWithinBorders`, or we can lean on standardized words from the [CSS box model](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Introduction_to_the_CSS_box_model#padding_area), it could be `cellPaddedWidth`; or for simplicity, it could be `cellWidthInner` or some such. Either of these three alternatives look approx equally good as it's easy to look up the def in the module context

packages/charts/src/chart_types/heatmap/renderer/canvas/canvas_renderers.ts
```diff
@@ -72,13 +72,13 @@ export function renderCanvas2d(
...
           ctx.translate(x, y);
           filteredCells.forEach((cell) => {
-            if (cell.visible)
-              renderText(
-                ctx,
-                { x: cell.x + cell.width / 2, y: cell.y + cell.height / 2 },
-                cell.formatted,
-                config.cell.label,
-              );
+            if (cell.visible && !Number.isNaN(cell.fontSize))
```
monfera, line 75: Minor: `Number.isFinite` lets you state it positively, though it also rules out the infinities, which I think is OK though likely can't happen

packages/charts/src/common/text_utils.ts
```diff
@@ -134,3 +134,22 @@ export function fitText(
   const { width } = measure(fontSize, [{ ...font, text }])[0];
   return { width, text };
 }
+
+/** @internal */
+export function maximiseFontSize(
```
monfera, line 139: We can avoid a single-use function that has the negative of a 8-line parameter list by directly setting up the `response` function in the context where `maximiseFontSize` is used now, and then it's a single-line call to `monoliticHillClimb` there. It also avoids an export and reduces code noise. 

If the motivation is to move the apparent complexity of the negations etc. in the `response` callback, another option could be something like
```
const boxResponseFun = (measure, args, boxWidth, boxHeight) => (fontSize) => {
    const [{ width }] = measure(fontSize, [args]);
    const widthDiff = boxWidth - width;
    const heightDiff = boxHeight - fontSize;
    return -Math.min(widthDiff, heightDiff);
  }
```
So you can do the one-liner 
```
const fontSize = monotonicHillClimb(
    boxResponseFun(measure, args, boxWidth, boxHeight),
    maxFontSize, 
    0, 
    integerSnap, 
    minFontSize
)
```
This also avoids the relatively long param list of `maximiseFontSize` and easier to see in situ what's going on, while the negations, `Math.min` impl. detail are put out of sight.

Maybe the theme for such a change is, reducing argument-passing noise at the slight expense(?) of somewhat lower abstraction level

monfera: ... or if you decide on retaining the abstraction, maybe it could be reused elsewhere if feasible

monfera (approved):
LGTM 🎉  The three limitations mentioned in the description would be useful to solve in this or an eventual PR, depending on the priority landscape. Two of those items have visual impact and it may be better to offer the more defensible options first, and by default (allowing a slight padding, and allowing a capping of the font size a bit like what you asked for for the treemap back then). It's hard to wean users off of suboptimal defaults as they desensitize users for their weaknesses while accustom them for their (maybe lesser) benefits. But there may be reasons for expediency, or strong demand for the volatile font size

While the slight disentanglement of the union type is implementation detail, it's probably easier while the code is in working memory, and gain experience with a shift to a disambiguation layer. Maybe at some point it could be a listed PR eval criterion (no propagation of user facing union types into internals if it's reasonably solvable upstream)

storybook/stories/heatmap/2_categorical.story.tsx
```diff
@@ -49,12 +57,17 @@ export const Example = () => {
...
+            },
           },
           cell: {
             maxWidth: 'fill',
-            maxHeight: 20,
             label: {
-              visible: true,
+              minFontSize,
+              maxFontSize,
+              visible: showLabels,
```
monfera, line 70: Would it be possible to move away from the problematic yellow-on-white color use in this story while changing it?

nickofthyme (approved):
LGTM

packages/charts/api/charts.api.md
monfera: It's self-explanatory also resulting from the documentype `Pixels` etc. I'm not privy of the needs, just remember that CI logs mention if missing documentation increases. Adding noise is pointless indeed

packages/charts/src/common/text_utils.ts
markov00: You are right about the single-use, but I was thinking it could be used more in the future where needed. I'm pretty sure there will be other uses for that function, not just on heatmap, but for example on the bar chart value labels.

monfera: Yes, if there are short to mid term plans for reusing it, it makes a ton of sense

packages/charts/src/chart_types/heatmap/renderer/canvas/canvas_renderers.ts
markov00: done https://github.com/elastic/elastic-charts/pull/1352/commits/3deb1c76325fc510bb16736e1938dcca27f75742

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
markov00: done https://github.com/elastic/elastic-charts/pull/1352/commits/3deb1c76325fc510bb16736e1938dcca27f75742

darnautov (approved):
LGTM

markov00:
> LGTM 🎉 The three limitations mentioned in the description would be useful to solve in this or an eventual PR, depending on the priority landscape. Two of those items have visual impact and it may be better to offer the more defensible options first, and by default (allowing a slight padding, and allowing a capping of the font size a bit like what you asked for for the treemap back then). It's hard to wean users off of suboptimal defaults as they desensitize users for their weaknesses while accustom them for their (maybe lesser) benefits. But there may be reasons for expediency, or strong demand for the volatile font size

@monfera please take a look at 9017878 and 3c2dc77 where I've added the text padding (it looks like a sensible default for now but happy to play with the value a bit more) and the default configuration that use the global minimum fontSize instead of a fontSize relative to each cell text value.

> While the slight disentanglement of the union type is implementation detail, it's probably easier while the code is in working memory, and gain experience with a shift to a disambiguation layer. Maybe at some point it could be a listed PR eval criterion (no propagation of user facing union types into internals if it's reasonably solvable upstream)

markov00:
@monfera I've seen @nickofthyme working on cleaning up the various interfaces/types/api and I prefer to wait him before adding a disambiguation layer for the `fontSize` at least

PR #1353, opened 2021-09-02
build: update webpack server to v4

markov00:
I've updated webpack dev server to v4
(it just clears up a bit the yarn install warnings about wrong dependencies)

update:
Updated few other SASS libraries (removed `sass` in favor of `node-sass`) we should move to `dart-sass` if we want to continue using SASS as `node-sass` is deprecated


storybook/package.json
```diff
@@ -33,12 +33,7 @@
     "@storybook/source-loader": "^6.3.7",
     "@storybook/theming": "^6.3.7",
     "circular-dependency-plugin": "^5.2.2",
-    "sass": "~1.32.0",
-    "sass-loader": "^10.1.1",
     "storybook-addon-theme-toggle": "^0.0.1",
-    "storybook-addon-background-toggle": "^0.0.1",
-    "terser-webpack-plugin": "^4.2.3",
-    "ts-loader": "^7.0.5",
-    "typescript": "^4.1.3"
```
nickofthyme, line 42: Why are we doing this? The idea behind moving these deps deeper was to isolate package ownership between different workspaces. If you want these for the integration server you should add them to the top-level `package.json` until we make integration it's own workspace.

storybook/package.json
```diff
@@ -33,12 +33,7 @@
     "@storybook/source-loader": "^6.3.7",
     "@storybook/theming": "^6.3.7",
     "circular-dependency-plugin": "^5.2.2",
-    "sass": "~1.32.0",
```
nickofthyme, line 36: What's the reasoning for switching back to `node-sass`, is is faster than the dart compiled version? I also agree with moving to `dart-sass`.

nickofthyme (changes requested):
Requesting the dependencies stay in their relative workspace, see comment below.

storybook/package.json
markov00: chatted online, reverted change

storybook/package.json
markov00: chatted online, reverted change

nickofthyme (approved):

PR #1356, opened 2021-09-02
feat(heatmap): dark mode with theme controls

### LINKED ISSUE #1335: [Heatmap] visualizations don't adjust color values for dark mode
**Describe the bug**
Heatmap doesn't adapt to the currently running Kibana theme

**To Reproduce**
 - Enable dark mode in Kibana
 - Visit ML anomaly viewer

Expected: Heatmap adapts to theme
Actual: Heatmap does not adapt to theme

Note: Lens works around this by explicitly setting `background` color

image

 - OS: OSX
 - Browser: chromium edge 92
 - Elastic Charts: version @ Kibana 7.15 BC1

**Kibana Cross Issues**
https://github.com/elastic/kibana/issues/97636

**Checklist**

- [X] every related Kibana issue is listed under `Kibana Cross Issues` list
- [X] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

nickofthyme:
## Summary

Screen Recording 2021-09-02 at 02 31 07 PM

### BREAKING CHANGES

The `HeatmapSpec.config` prop is removed. All properties have been moved/renamed under new `Theme.heatmap` options with the following exceptions:

- `HeatmapSpec.margin` was not used, `Theme.chartMargins` should be used in the future.
- `HeatmapSpec.fontFamily` `HeatmapSpec.width` and `HeatmapSpec.height` all removed as they were never used.

## Details

The `Heatmap`, `Partition` and `WordCloud` specs all have no connection to the chart theme and thus have no canonical inheritance to dark mode. This PR fixes this for the `Heatmap` chart and adds some added styles to the text labels as exposed in xy charts. The `HeatmapConfig` was broken down and moved into `Theme` or the `HeatmapSpec` accordingly. We should avoid having a `Config` option to control theme as this not only creates a disconnect but also allows for multiple sources of truth for similar options and another property requirement to toggle for theming between dark and light modes.

## Issues

closes #1335

### Checklist


packages/charts/src/utils/themes/dark_theme.ts
```diff
@@ -230,4 +230,85 @@ export const DARK_THEME: Theme = {
       stroke: 'white',
     },
   },
+  heatmap: {
+    maxRowHeight: 30,
+    maxColumnWidth: 30,
+    brushArea: {
+      visible: true,
+      fill: 'white', // white === transparent
```
markov00, line 238: what do you mean by `white === transparent`

packages/charts/src/utils/themes/theme.ts
```diff
@@ -186,6 +186,76 @@ export interface GoalStyles {
   maxFontSize: number;
 }
 
+/**
+ * @public
+ */
+export interface HeatmapStyles {
```
markov00, line 192: What if this is renamed as singular? because it covers the style of the heatmap, not multiple styles.

markov00: I know that this goes in tandem with the GoalStyles (that IMHO we should rename to singular)
Happy to discuss this

markov00 (changes requested):
I can see a visible difference on the `echHighlighterMask` in this PR in dark mode:
the Y axis and the selected areas are not transparent anymore as in the old implementation:
This PR:
Screenshot 2021-09-06 at 12 21 09
Current master:
Screenshot 2021-09-06 at 12 30 14

Despite the black text, the behaviour on master is the correct one. clearly visible also in the light theme
Screenshot 2021-09-06 at 12 21 02

markov00:
just remember that the `heatmap` is in `@alpha` and we can also avoid having BREAKING CHANGES there

packages/charts/src/utils/themes/theme.ts
nickofthyme: Yeah probably right about being consistent with others like `BarSeriesStyle`, and like `<div style={{}} />`. I just think of it as there are multiple styles to define, not a singular style.

packages/charts/src/utils/themes/dark_theme.ts
nickofthyme: this was an old comment from the config that I kept.

nickofthyme:
> I'd also like to chat about when a style needs to be pushed into the theme and when it needs to be local on the spec.

Aside from what we discussed in todays meeting, I tend to through all function thing like formatters in the spec rather than the theme.

packages/charts/src/utils/themes/dark_theme.ts
nickofthyme: This `fill` was applied to a `<mask>` for some reason. Made it optional to be transparent by default unless someone wants to color the mask `#ff0000` or something. See c8b5ea7.

Screen Recording 2021-09-07 at 02 57 53 PM

packages/charts/src/utils/themes/theme.ts
nickofthyme: 4f91559

markov00 (approved):
Looks good to me
tested locally and all the issues are now gone!

PR #1360, opened 2021-09-06
ci: disable local puppeteer download 

markov00:
Disabling the Puppeteer Download when installing the npm dependencies, following the advice here: https://github.com/gidztech/jest-puppeteer-docker

Could save few seconds on CI :)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,6 +48,8 @@ jobs:
         uses: bahmutov/npm-install@v1
         with:
           useRollingCache: true
+        env:
+          PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: true
 
   eslint:
     name: Eslint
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yaml
```diff
@@ -48,6 +48,8 @@ jobs:
         uses: bahmutov/npm-install@v1
         with:
           useRollingCache: true
+        env:
+          PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: true
```
nickofthyme, line 52: We could assign `env` at the top level too

markov00: oh nice, didn't know that :) 
I can fix this in the next CI change PR if you are fine

nickofthyme: No rush just saying.

PR #1361, opened 2021-09-06
test: fix flaky test with multi-chart interaction tooltip

markov00:
This hopes to fix the flaky test that appears when testing the pointer position across multiple charts.
I've added a delay of 1s for both tests. There was already another test with the same delay so I've just aligned the others.

image

We could reduce this delay if we want, but we just have 3 screenshot with that delay so it doesn't influence much the testing


markov00:
I need also to unskip one of the test

monfera (approved):
LGTM, passed on first run 🎉

markov00:
jenkins test this

nickofthyme (approved):
LGTM, We should revisit this delay when we switch to playwright.

Side note -- I thought this was already added but must have been the one below. Let's see if it behaves.

markov00:
jenkins test this

PR #1362, opened 2021-09-07
fix(build): fix license in package.json

markov00:
After migrating to `Elastic License 2.0` we forgot to upgrade the package.json license header to the right one.

cc @ghudgins 

ref: https://github.com/elastic/kibana/issues/105614


nickofthyme (approved):
Good catch!

PR #1364, opened 2021-09-07
fix(heatmap): remove values when brushing only over axes

### LINKED ISSUE #1338: [Heatmap] Brushing action over the Y-axis labels
**Describe the bug**
Brushing action over the Y-axis labels triggers highlightes cells to the right and invoke `onBrushEnd` event. 

**Expected behaviour**
Cells should not be highlighted because the cursor never appears on top. `onBrushEnd` event probably still can be triggered but it shouldn't contain any values from the X-axis labels. 

**Screenshots**
https://user-images.githubusercontent.com/5236598/130745236-55acdf4e-5efe-465b-a6e7-0d211de265b5.mov

rshen91:
## Summary
If a user brushes only over the y axis, x values should not be calculated in the pickedCells. If a user brushes only over the x axis, the y values should not be populated.

## Details
This PR removes x values in picked cells when brushing only over y-axis and remove y values in picked cells when brushing over the x axis.

## Issues
Closes #1338 

### Checklist


markov00:
@rshen91 can you please add a link to an issue for this, or describe it here?
I'm not sure what are you talking about, but if the reason is to disable brushing over the Y axis I think this is not desirable, or at least not from the ML team

rshen91:
Hi @markov00 sorry I'm still trying to figure out implementation for this. I think I'm following that brushing should be disabled when the pointer starts at a position over the y axis. The issue I'm trying to fix is #1338, thanks

packages/charts/src/chart_types/heatmap/renderer/dom/highlighter.tsx
```diff
@@ -36,11 +38,14 @@ export const HighlighterCellsComponent: FC<HighlighterCellsProps> = ({
...
   brushMask,
+  pointer,
 }) => {
   if (!initialized || dragShape === null) return null;
 
   const maskId = `echHighlighterMask__${chartId}`;
-  return (
+  const pointerOnLeftAxis =
+    pointer && pointer.lastDrag ? pointer.lastDrag?.start.position.x < canvasDimension.left : false;
+  return !pointerOnLeftAxis ? (
```
monfera, line 48: Nice ternary here! I try not to negate a singular predicate in such cases, no real issue with `null` being in the 1st outcome of the ternary, in fact a bit nicer as it gets the non-result out of the way and the code reader can focus on the DOM fragment

packages/charts/src/chart_types/heatmap/renderer/dom/highlighter.tsx
```diff
@@ -36,11 +38,14 @@ export const HighlighterCellsComponent: FC<HighlighterCellsProps> = ({
...
   brushArea,
   brushMask,
+  pointer,
 }) => {
   if (!initialized || dragShape === null) return null;
 
   const maskId = `echHighlighterMask__${chartId}`;
-  return (
+  const pointerOnLeftAxis =
+    pointer && pointer.lastDrag ? pointer.lastDrag?.start.position.x < canvasDimension.left : false;
```
monfera, line 47: When there's a `true` or `false` as a value expressly yielded eg. via a ternary, it almost always means, it can be simplified: 
```
  const pointerOnLeftAxis =
    pointer && pointer.lastDrag && pointer.lastDrag?.start.position.x < canvasDimension.left;
```
Explanation: if `pointer` or `pointer.lastDrag` are falsey, the expression evaluates to falsey anyway.

The question mark after the second use of `lastDrag` isn't needed, because TS will type guard it anyway, since you never land on that if `pointer.lastDrag` is nullish:
```
  const pointerOnLeftAxis =
    pointer && pointer.lastDrag && pointer.lastDrag.start.position.x < canvasDimension.left;
```

Also, `pointer && pointer.lastDrag` can usually be collapsed like this:
```
  const pointerOnLeftAxis =
    pointer?.lastDrag && pointer.lastDrag.start.position.x < canvasDimension.left;
```
It still has some repetition: the `pointer.lastDrag` chain. We can NOT simply say, 
```
  const pointerOnLeftAxis = pointer?.lastDrag?.start.position.x < canvasDimension.left;
```
in part because TypeScript prevents us from doing automatic type coercion, so only a number can be `<` compared with another number (or eg. string with string); in part because a nullish comparison in JS _may_ yield unexpected results, which is why TS prevents coercion to begin with (eg. `undefined < 3 === false` but `null < 3` is true, as `null` gets coerced to zero (`null == 0` is false though...)

However we can use a value for nullish coalescing, which will be always greater than `canvasDimension.left`, avoiding the above repetition as well as the ternary:
```suggestion
  const pointerOnLeftAxis = (pointer?.lastDrag?.start.position.x ?? Infinity) < canvasDimension.left;
```
(Caveats, I haven't tried these TS-wise or runtime-wise, pls. mention if it doesn't work out; also, I'm not 100% sure if it's a `<` or `<=` predicate that's better here)

packages/charts/src/chart_types/heatmap/renderer/dom/highlighter.tsx
```diff
@@ -36,11 +38,14 @@ export const HighlighterCellsComponent: FC<HighlighterCellsProps> = ({
   canvasDimension,
   brushArea,
   brushMask,
+  pointer,
 }) => {
   if (!initialized || dragShape === null) return null;
 
   const maskId = `echHighlighterMask__${chartId}`;
```
monfera, line 45: While at it, I'd inline `maskId`: it has a single use; its meaning is clear already at that place of use (`<mask id={...}>` is clearly a mask id; the variable name doesn't add info); and no need to precompute it as we may still yield `null`. So, the suggestion is, `<mask id={`echHighlighterMask__${chartId}`}>`

packages/charts/src/chart_types/heatmap/renderer/dom/highlighter.tsx
```diff
@@ -36,11 +38,14 @@ export const HighlighterCellsComponent: FC<HighlighterCellsProps> = ({
   canvasDimension,
   brushArea,
   brushMask,
+  pointer,
 }) => {
   if (!initialized || dragShape === null) return null;
```
monfera, line 43: Also a preexisting code improvement opportunity: Here, we yield `null` if these predicates are true. But the continuation branch also yields `null` if `pointerOnLeftAxis` is true. So the bail can be deleted like this:
```
}) => {
  const pointerOnLeftAxis = (pointer?.lastDrag?.start.position.x ?? Infinity) < canvasDimension.left;
  return dragShape === null || !initialized || pointerOnLeftAxis ? null : (
```
A possible objection is that maybe it's stupid to calculate `pointerOnLeftAxis` if eg. `dragShape === null`, its value won't be used. However, the way current JS engines work, it likely won't be computed, as JS engines are free to reorder operations, so most likely, it won't evaluate the `pointerOnLeftAxis` expression unless needed so. It could be a different situation if the `pointerOnLeftAxis` expression called some expensive function.

Also, it's such a cheap calculation even if it were done, performance is not a concern, any difference is unmeasurable.

The gain is not just fewer lines to read (us) and send to the browser (payload) but the simpler, linear start to finish flow. Unless there's a strong reason, it's nice to work with expressions rather than statements, and it's nice to have a singular return point

packages/charts/src/chart_types/heatmap/renderer/dom/highlighter.tsx
```diff
@@ -22,6 +23,7 @@ export interface HighlighterCellsProps {
   dragShape: DragShape | null;
   brushMask: Config['brushMask'];
   brushArea: Config['brushArea'];
+  pointer: PointerStates | null;
```
monfera, line 26: Let's avoid adding a union type: an initial value, currently defined as `null`, can be replaced with something like 
```
pointer = { dragging: false, current: { position: { x: NaN, y: NaN}, time: -Infinity }, down: null, up: null, lastDrag: null, lastClick: null }
```
Yes it's a bit verbose, but then the `const pointerOnLeftAxis = ...` predicate can further simplify (there's always a pointer object, etc.)

Having said that, the entirety of `PointerStates`, or its props except the first 1 or 2 props could be profitably replaced with a `Map`, as it is already a Map in spirit. No need for all these ternaries. Or `PointerStates` props could be retained as they are, except removing the `| null` ternary (initial values are trivial) or at least, not needing a `null` (can be an optional value, left `undefined` if unused) 

@markov00 do you agree with such eventual refactoring?

rshen91:
@markov00 and @monfera thanks for the initial review I think I'm understanding the solution better 😅 in commit 7c131ed
@monfera I removed the older code you had commented on since I was misunderstanding the problem sorry for the hassle in reviewing!

packages/charts/src/chart_types/heatmap/renderer/dom/highlighter.tsx
rshen91: Sounds good to me - small fix in e5ca9d9 thanks

monfera (changes requested):
As we discussed while screensharing:

Brushing on the Y axis (in terms of gesture, click-dragging over the axis tick labels of the Y axis) in the Heatmap/Categorical story had a couple of undesirable results:

1. The highlighting itself works, but it only highlighs the first column, rather than the entire width of the highlighted rows, which looks like a preexisting issue in `master`
2. The `onBrushEnd` event also had data for the X axis, which, as I understand, is what the original issue was
3. This x axis info specifies only the first column (first value on the X axis)
4. (this may or may not be a Storybook artifact) After I click on the Actions data on the right, I'm sometimes unable to highlight the approx. same area again, unless I highlight something else. It may be a bug, as it seems to happen if I highlight within the area of a previous highlight

All these are shown in this video (starting with item 4):

Sep-10-2021 18-17-41

It's not clear to me if solving problem no. 1 is in scope for this PR or not, @markov00 pls. weigh in on that

monfera (approved):
The code looks good and the PR addresses the second point of the issue: the raised event is not specific to X values.
As it's a clear impovement, and helps the requestor test the changes, I think it's a positive step, so I'm approving it.

The first half of the issue is not addressed by this PR right now: the highlighting only highlighs the first column, rather than the entire width of the highlighted rows. So please remove the "closes [issue number]" to avoid closing the issue if it gets merged without this half

Also, while the issue was filed with reference to the Y axis, it should be the same behavior for the X axis too, otherwise users will be confused.

Another observation, may or may not be a Storybook artifact: after I click on the Actions data on the right, I'm sometimes unable to highlight the approx. same area again, unless I highlight something else. It may be a bug, as it seems to happen if I highlight within the area of a previous highlight

So it means there'd be two more tasks for full closure of the issue, as well as the filing of a bug report for this selection glitch, if needed

rshen91:
@monfera good points. I added 11a84ec to add the logic to the x-axis dragging. Thanks!

markov00:
nit: please reduce and simplify the title length, and move it to the description

PR #1367, opened 2021-09-09
build(deps): update canvas to 2.8.0

markov00:
I've updated `canvas` dependency to 2.8.0 to fix a missing function https://github.com/Automattic/node-canvas/pull/1769 and I've removed `jest-canvas-mock` because `canvas` already mocks everything we need.

@monfera I've also tried the failing changes in https://github.com/elastic/elastic-charts/pull/1309/files replacing the `clearCanvas` method with:
```ts
export function clearCanvas(ctx: CanvasRenderingContext2D) {
  const { a, d } = ctx.getTransform().invertSelf().scale(ctx.canvas.width, ctx.canvas.height);
  ctx.clearRect(0, 0, a, d);
}
```
and it works correctly now


nickofthyme:
Are there any cases where we would not want the full implementation and instead just want to mock out any canvas elements in unit tests?

markov00:
> Are there any cases where we would not want the full implementation and instead just want to mock out any canvas elements in unit tests?

Probably in most cases, but the think is that the `jest-canvas-mock` doesn't mocks everything correctly see https://github.com/hustcc/jest-canvas-mock/issues/86 and https://github.com/hustcc/jest-canvas-mock/issues/85 so we can switch it off completely.
The other benefit from the `canvas` only library is that we can "rely" on a real implementation for the textMeasure API in canvas, where the mock just return the string length.

markov00:
There are two possible solution to the above mentioned issue in Jenkins:
- ask Infra if these libraries are available in the Jenkins installation and in case add the correct PATH to LD_LIBRARY_PATH as described in https://stackoverflow.com/questions/20357033/usr-lib-x86-64-linux-gnu-libstdc-so-6-version-cxxabi-1-3-8-not-found
- we don't need the `canvas` package when running VRTs. It's picked up automatically by the jest environment, but we don't really need it, @nickofthyme do you know if we can run `yarn install` ignoring the `canvas` package? should we just run `yarn remove canvas` after the install?

nickofthyme:
@markov00 It seems to be using the `canvas` on the VRT which it should not

```
04:46:47  FAIL  integration/tests/interactions.test.ts
04:47:39   ● Test suite failed to run
04:47:39 
04:47:39     /lib64/libstdc++.so.6: version `CXXABI_1.3.9' not found (required by /var/lib/jenkins/workspace/elastic+elastic-charts+pr-vrts/node_modules/canvas/build/Release/canvas.node)
04:47:39 
04:47:39       at Runtime._loadModule (../node_modules/jest-runtime/build/index.js:893:29)
04:47:39       at Object.<anonymous> (../node_modules/canvas/lib/bindings.js:3:18)
```

nickofthyme (approved):

PR #1369, opened 2021-09-09
feat(heatmap): move onBrushEnd from config to Settings

### LINKED ISSUE #1214: Move heatmap onBrushEnd callback to Settings
The `onBrushEnd` action usually is configured in the `Settings` component.
For the `Heatmap` instead, the callback is wrongly configured in the `config`.

We should replace the `config.onBrushEnd` with the default `Settings.onBrushend` one.

rshen91:
### BREAKING CHANGE

The `onBrushEnd` prop is removed from the heatmap config. It can now be pulled off the Setting to match the other chart types. 

## Details
In `settings.tsx` I created a new type `BrushArea`:

```js
/** @public */
export type BrushArea = XYBrushArea | HeatmapBrushEvent;

/** @public */
export type BrushEndListener = (brushArea: BrushArea) => void;
```

This satisfies `chart_state.interactions.test.tsx` and `on_brushed_highlighted_shapes.test.ts`

## Issues
Closes #1214 

### Checklist


packages/charts/src/specs/settings.tsx
```diff
@@ -140,8 +140,20 @@ export type ElementClickListener = (
 export type ElementOverListener = (
   elements: Array<XYChartElementEvent | PartitionElementEvent | HeatmapElementEvent | WordCloudElementEvent>,
 ) => void;
+
+/** @public */
+export type BrushArea = XYBrushArea | HeatmapBrushEvent;
+
 /** @public */
-export type BrushEndListener = (brushArea: XYBrushArea) => void;
+export type BrushEndListener = (brushArea: BrushArea) => void;
```
markov00, line 148: @nickofthyme @monfera  @rshen91 
I'd like to have your opinion here: this is not the only place where we have an union type as an argument of a chart callback.
It happens on the onClick event, but also in other situations.
Most chart consumers automatically cast to the right type depending on their needs: e.g. I'm using a bar chart so I'm cast with `as XYBrushArea` the brushArea. 
Right now we don't offer a Generic type for the chart and I also don't fully know if and how this can work.
I'm wondering if we can, in the meantime, find a simpler and cleaner solution to the problem.
Tagging the callback argument type with the chart type could be a solution:

```ts
type HeatmapBrushEvent = {
  type: 'heatmap';
  cells: Cell[];
  x: (string | number)[];
  y: (string | number)[];
};

interface XYBrushArea {
  type: 'cartesian';
  x?: [number, number];
  y?: Array<GroupBrushExtent>;
}
```

what do you think?

rshen91: This seems cool! Let me try and refactor to implement this, thanks 👍🏻

packages/charts/src/specs/settings.tsx
```diff
@@ -140,8 +140,20 @@ export type ElementClickListener = (
 export type ElementOverListener = (
   elements: Array<XYChartElementEvent | PartitionElementEvent | HeatmapElementEvent | WordCloudElementEvent>,
 ) => void;
+
+/** @public */
+export type BrushArea = XYBrushArea | HeatmapBrushEvent;
```
nickofthyme, line 145: nit: I would keep this consistent and `___event` is most fitting. Also since this is breaking we should change `XYBrushArea` to `XYBrushEvent`.

```suggestion
export type BrushEvent = XYBrushEvent | HeatmapBrushEvent;
```

packages/charts/src/specs/settings.tsx
nickofthyme: So I was thinking one way to do this would be in the function somehow like this...

```tsx
<Settings
  onBrushEnd={('heatmap', event: HeatmapBrushEvent) => null}
  // or with multiple
  onBrushEnd={(['line', 'area'], event: LineBrushEvent | AreaBrushEvent) => null}
/>
```

The issue with this is that they could pass `'heatmap` when there is no `HeatmapSpec` defined.

```tsx
<Settings
  onBrushEnd={{
    types: ['heatmap'],
    callback: (event: HeatmapBrushEvent) => null
  }}
/>
```

The idea behind the types is something like (see [TS Playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgBITmAtnADgUQDcJxkBvAKGWuTAE9cIAuZAIgAsNs9WqaBzCGHSYcuFgAoAlMgC8APnIBfCioqhIsRCgAaATSIkw5PtXqMWrAB51eNZILD7JMhctUUKAei-IAKpxQKDh0yABGKHDIADYQAM5xyMRQYQD2cSgA7nChqTDIACapoPy07MBxAHTVFOYoAMJw0dFhiADWADx+DBCKsv49yBBWkCAFiQDknKJ4E8gA-KZoXGKGpAA+yI3NrQid+FYI0QCuBRBdPQA0Uyuz8opMS92MQyMk48gTNnOL9vb6a2Mm22LXaHQOR1O52eEGu33uyEefxoIAgyU8dWWM1w9TCci2TVBew6N2xE3kAG5aoN9Lj8SDdp0vnRyVTMQBBEB0On9Blg5kTdak7i4VlAA)...


https://user-images.githubusercontent.com/19007109/133267907-45f122e1-c576-408d-aebe-5d7bfcecd1ec.mp4

packages/charts/src/specs/settings.tsx
rshen91: Good naming consistency! Thanks!

rshen91: 6b4a4b9

packages/charts/src/specs/settings.tsx
nickofthyme: I found a slightly simpler way to avoid the type recursion. See [demo](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgBITmAtnADgUQDcJxkBvAKGWuTAE9cIAuZAIgAsNs9WqaBzCGHSYcuFgAoAlMgC8APnIBfCioqhIsRCgAaATSIkw5PtXqMWrAB51eNZILD7JMhctXrw0eEmSHwALJ4JvYA5JyieKEsItwExOAA3KbIoTbRyPr+YMlqAPR5yAAqnFAoOHTIAEYocMgANhAAzk3IxFBVAPZNKADucJWdMMgAJp2g-LTswE0AdPMU5igAwnD19VWIANYAPEUMEIqyxQfIEFaQICOtWxB0Q34JYEG4yAD8KdkvANr7jAC6yBYIAg7Qoi1OsTEyyqcmQq3WmwQu3CXDEoXkiWQBTQaLw2QhjEyehhcIRG22OzSdAxWJxWSehJQAEEQHRScdyUiUekAD6oyK4WnYwpQ-FPZC84nZIA)

```ts
interface HeatmapEvent {
    type: "heatmap"
    getHeatmap: () => {}
}

interface XYEvent {
    type: "xy"
    getXY: () => {}
}
interface EventMap {
    'heatmap': HeatmapEvent;
    'xy': XYEvent;
}

// There may be a less verbose way of doing this...
type Callback<Type> = Type extends keyof EventMap ?
    EventMap[Type] : never

type HeatmapCb = Callback<'heatmap'>; // HeatmapEvent
type XYCb = Callback<'xy'>; // XYEvent
type AnyCb = Callback<'xy'|'heatmap'>; // HeatmapEvent | XYEvent
```

nickofthyme: After looking at how we could implement this in the brush end listener, it is not currently easy to use a generic via `Settings` props. @rshen91 is opening an issue to discuss a future enhancement.

CHANGELOG.md
```diff
@@ -1,31 +1,3 @@
-# [35.0.0](https://github.com/elastic/elastic-charts/compare/v34.2.1...v35.0.0) (2021-09-13)
```
nickofthyme: Ugh sorry for got about `git checkout` would also wipe any auto changes in between. I just `git reset` to the good pr with the unit test fix on top 🍒 .

markov00:
jenkins test this please

nickofthyme (approved):
LGTM! 🎉

PR #1377, opened 2021-09-13
refactor(partition): remove width height from partition config

### LINKED ISSUE #1185: Partition chart should pull from theme and spec not config
**Describe the bug**
The `config.width` and `config.height` in the partition chart, at least in the pie chart layout, are overwritten by the parentDimensions.

The first part where the `width` and `height` are overwritten is visible in the `getShapeViewModel` before merging the partial config.

The second place is in the `partitionMultiGeometries` where we compute the `chartWidth` and `chartHeight` as derived from the `marginedWidth/Height` (always coming from the `parentDimensions`). On the same function we should also check the `innerWidth/Height` that also comes from the `parentDimensions`

**To Reproduce**
1. Go to https://codesandbox.io/s/youthful-curran-egqkv?file=/src/App.tsx
2. Try to use `width`, `height` in the config, instead of the currently proposed workaround with margin/outerSizeRatio
3. The size of the pie will be always the size of the parent, no matter what value are used as width/height in the Config

**Expected behaviour**
We should respect the `config.width` and `config.height` parameters exposed by the `Partition` component. If we don't want that feature, we should remove them from the API.

**Version:**
 - Elastic Charts: 29.2.0

**Errors in browser console**
N/A

rshen91:
## Issues
Closes #1185 

### Checklist


rshen91:
In discussion with Nick - duplicating efforts closing

PR #1381, opened 2021-09-16
test(xy): scale type improvements

monfera:
## Summary

As a first step for comprehensive type safety, this PR increases TS type safety to the central Cartesian functions/methods/classes and lots of utility and render function dependencies through turning the `Scale` generic on its domain

## Details

There were a couple of strategically placed `any`s in the `Scale` interface, and elsewhere. Though low in number, their propagating effect meant that most of the xy code doesn't really enjoy type safety, there can always be a scale induced `any` leak, so only very specific functions could be considered type safe.

The current PR is a start of work, focusing on `Scale.domain`, `Scale.ticks`, and its positive effects on `Scale` methods and lots of other functions. While there's still a lot of dirt swept under the carpet (just grep `as Scale<`) it's way more explicit, and these assertions can be solved one by one.

So, statically a lot more places where we assert or say `any`, but it's more explicit, closer to the as yet unresolved bits.

There's also a place where the D3 scale is called with a type that's not enabled (`[undefined]` domain); it's now marked explicitly.

## Issues

Under the umbrella of #1303

### Checklist


monfera:
test this

packages/charts/src/chart_types/xy_chart/annotations/line/dimensions.ts
```diff
@@ -164,10 +164,10 @@ function computeXDomainLineAnnotationDimensions(
       }
     } else if (isBandScale(xScale)) {
       if (isHistogramMode) {
-        const padding = (xScale.step - xScale.originalBandwidth) / 2;
+        const padding = (xScale.step - (xScale as ScaleBand<string | number>).originalBandwidth) / 2;
         annotationValueXPosition -= padding;
       } else {
-        annotationValueXPosition += xScale.originalBandwidth / 2;
+        annotationValueXPosition += (xScale as ScaleBand<string | number>).originalBandwidth / 2;
```
markov00, line 170: I think you can remove this assertion because `isBandScale` already define xScale as ScaleBand

packages/charts/src/chart_types/xy_chart/annotations/utils.ts
```diff
@@ -145,22 +145,22 @@ export function computeAnnotationDimensions(
...
-      smallMultipleScales,
-      chartRotation,
-      getAxisStyle,
-      isHistogramModeEnabled,
-    );
+    } else {
+      const dimensions = computeRectAnnotationDimensions(
+        annotationSpec,
+        yScales,
+        xScale as Scale<number>,
```
markov00, line 152: why you are restricting this to `number` only if the passed one could be both?

markov00: You can change `computeRectAnnotationDimensions` to accept `Scale<unknown>` because internally you are calling `isBandScale` that assert the type for you

packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts
```diff
@@ -64,7 +64,7 @@ export function computeRectAnnotationDimensions(
     let xAndWidth: { x: number; width: number } | null = null;
 
     if (isBandScale(xScale)) {
-      xAndWidth = scaleXonBandScale(xScale, x0, x1);
+      xAndWidth = scaleXonBandScale(xScale as ScaleBand<number>, x0, x1);
```
markov00, line 67: this one also seems unnecessary, see my previous comment in file `/xy_chart/annotations/utils.ts `

packages/charts/src/chart_types/xy_chart/annotations/line/dimensions.ts
```diff
@@ -164,10 +164,10 @@ function computeXDomainLineAnnotationDimensions(
...
       }
     } else {
```
nickofthyme, line 172: The `isBandScale` type guard should be able to force this no? maybe need to update `isBandScale` to something like `isBandScale<string|number>`

packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts
nickofthyme, line 67: Same comment as before

packages/charts/src/chart_types/xy_chart/state/selectors/compute_annotations.ts
```diff
@@ -46,7 +46,7 @@ export const computeAnnotationDimensionsSelector = createCustomCachedSelector(
       annotationSpecs,
       settingsSpec.rotation,
       yScales as Map<GroupId, Scale<number>>,
-      xScale,
+      xScale as Scale<string | number>,
```
nickofthyme, line 49: What is the error otherwise?

packages/charts/src/chart_types/xy_chart/rendering/area.ts
```diff
@@ -31,8 +31,8 @@ import {
 export function renderArea(
   shift: number,
   dataSeries: DataSeries,
-  xScale: Scale,
-  yScale: Scale,
+  xScale: Scale<number>,
```
markov00, line 34: this should be `string|number` as every other `xScale`
The scale is used in the `pathGenerator.x` function with the data coming from `dataSeries.data` where the x can be both string or number `DataSeriesDatum.x`

packages/charts/src/scales/index.ts
```diff
@@ -28,16 +28,18 @@ export type ScaleBandType = ScaleOrdinalType;
...
-  domain: D[];
+export interface Scale<T> {
+  domain: T[];
   range: number[];
   /**
    * Returns the distance between the starts of adjacent bands.
    */
   step: number;
-  ticks: () => number[];
+  ticks: () => T[];
```
nickofthyme, line 38: 👍🏼

packages/charts/src/chart_types/xy_chart/rendering/bubble.ts
```diff
@@ -21,8 +21,8 @@ import { MarkSizeOptions } from './utils';
 export function renderBubble(
   shift: number,
   dataSeries: DataSeries,
-  xScale: Scale,
-  yScale: Scale,
+  xScale: Scale<number>,
```
markov00, line 24: `number|string`

packages/charts/src/chart_types/xy_chart/rendering/line.ts
```diff
@@ -30,8 +30,8 @@ import {
 export function renderLine(
   shift: number,
   dataSeries: DataSeries,
-  xScale: Scale,
-  yScale: Scale,
+  xScale: Scale<number>,
```
markov00, line 33: `number|string`

packages/charts/src/chart_types/xy_chart/rendering/points.ts
```diff
@@ -29,8 +29,8 @@ import {
 export function renderPoints(
   shift: number,
   dataSeries: DataSeries,
-  xScale: Scale,
-  yScale: Scale,
+  xScale: Scale<number>,
```
markov00, line 32: `number|string`

packages/charts/src/scales/types.ts
```diff
@@ -23,7 +23,7 @@ export function isLogarithmicScale(scale: Scale<unknown>): scale is ScaleContinu
  * Check if a scale is Band
  * @internal
  */
-export function isBandScale(scale: Scale<unknown>): scale is ScaleBand {
+export function isBandScale(scale: Scale<unknown>): scale is ScaleBand<string | number> {
```
nickofthyme, line 26: I think something like this would work a little nicer and automatically infer the type of `T`

```suggestion
export function isBandScale<T extends string | number>(scale: Scale<T>): scale is ScaleBand<T> {
```

markov00 (commented):
I haven't gone too far, but in general this are the current assumptions for cartesian:
`yScale` is always `Scale<number>`
`xScale` is always `Scale<number|string>`, can be restricted if required but most of the time I don't need to know if the value I'm sending to the scale is a number or a string, I just need to know that the output of the scale is a number

packages/charts/src/chart_types/xy_chart/annotations/line/dimensions.ts
monfera: Indeed, I'll check this specifically and elsewhere; I haven't gotten around to considering type guards for eliminating the new `as`. Plan is, grepping for `as Scale<` and capture as many as possible

packages/charts/src/scales/types.ts
monfera: Thanks, good tip! Held off on turning functions generic, in an attempt to first just remove the top level `any` domain

packages/charts/src/chart_types/xy_chart/annotations/line/dimensions.ts
monfera: Thanks, will check! I haven't added assertions where there was no complaint, but it's possible that a later commit made an earlier assertion unnecessary, indeed a good idea to revisit all

packages/charts/src/chart_types/xy_chart/annotations/utils.ts
monfera: Re the question, it's because it's in a conditional branch, and I didn't want to add potentially unduly restrictive types in the other, unrelated branch in the first round of improvements

packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts
monfera: Yes, thanks for the catch, I need to revisit these (the assertion was done before the function param types were tightened)

packages/charts/src/scales/types.ts
nickofthyme: Right but doing so using `as` could create unexpected runtime errors and making the function generic with default covers both cases.

packages/charts/src/chart_types/xy_chart/rendering/area.ts
monfera: xScale? are we ever linking ordinals with a line? I tightened it to `number` because:
- it unconditionally calls `renderPoints` which expects numeric
- I wasn't aware of using lines for ordinals
- tests let it pass

Just made a quick attempt to relax the `xScale` param of `renderPoints` to also allow strings and it didn't complain, so I can relax it here too, though no idea why I didn't get a complaint for my exclusion of the strings, will look into it

packages/charts/src/chart_types/xy_chart/rendering/points.ts
monfera: Thanks Marco, will revisit these!

packages/charts/src/chart_types/xy_chart/annotations/utils.ts
monfera: I can't change `computeRectAnnotationDimensions` to accept `Scale<unknown>` because calling `limitValueToDomainRange` will have a red squiggly mark (`limitValueToDomainRange` performs a subtraction of domain values, so it must be numeric) and for the same reason it can't be `Scale<number | string>` either

Basically the `computeRectAnnotationDimensions` only seems to work with numeric X, though I haven't tested the runtime, it was enough for me to see the subtraction. Though the subtraction only happens if the domain length is non-zero, I assumed it's always the case

monfera: By extension, the entire `computeRectAnnotationDimensions` only works with a numeric X scale, Marco or Nick lmk if it's not the case

packages/charts/src/chart_types/xy_chart/annotations/line/dimensions.ts
monfera: Indeed, I only later turned the type guard function generic, now it no longer complains there 👯‍♀️

packages/charts/src/chart_types/xy_chart/annotations/line/dimensions.ts
monfera: Done! Looks like I should've first updated the type guard function(s) for fewer false alarms, and it was a tradeoff between me combing through once more, vs getting feedback also from Marco in his evening time 😄

packages/charts/src/chart_types/xy_chart/state/selectors/compute_annotations.ts
monfera: `Scale<unknown>' is not assignable to parameter of type 'Scale<string | number>`

However, now that the 1st round is done, I feel like replacing a couple of `Scale<unknown>`s with Scale<numner | string>` and it should take care of it. Not sure if in this PR or a next one, depending on when this one gets approved 😄

packages/charts/src/chart_types/xy_chart/annotations/line/dimensions.ts
monfera: 0ac3fd257184461392392d94f3ac83114480e9df

packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts
monfera: 0ac3fd257184461392392d94f3ac83114480e9df

packages/charts/src/scales/types.ts
monfera: 💯 % agree, committing your suggestion

packages/charts/src/chart_types/xy_chart/annotations/line/dimensions.ts
monfera: 0ac3fd257184461392392d94f3ac83114480e9df

packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts
monfera: 0ac3fd257184461392392d94f3ac83114480e9df

packages/charts/src/chart_types/xy_chart/rendering/area.ts
monfera: 0ac3fd257184461392392d94f3ac83114480e9df

packages/charts/src/chart_types/xy_chart/rendering/bubble.ts
monfera: 0ac3fd257184461392392d94f3ac83114480e9df

packages/charts/src/chart_types/xy_chart/rendering/line.ts
monfera: 0ac3fd257184461392392d94f3ac83114480e9df

packages/charts/src/chart_types/xy_chart/rendering/points.ts
monfera: 0ac3fd257184461392392d94f3ac83114480e9df

packages/charts/src/chart_types/xy_chart/state/selectors/compute_annotations.ts
monfera: Bit the bullet afaceb6c62ecf4e4e8b3587e00f4a3eac40ca21e

packages/charts/src/scales/types.ts
monfera: 8cd6cda702ab6ee1982c59076b6bfa7e18d7f8c3

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -29,8 +29,8 @@ type BarTuple = {
 export function renderBars(
   orderIndex: number,
   dataSeries: DataSeries,
-  xScale: Scale,
-  yScale: Scale,
+  xScale: Scale<number | string>,
+  yScale: Scale<number | string>,
```
monfera, line 33: It could be tightened to `Scale<number>` but then the caller would need to add an `as` assertion, while there's no harm with the slightly broader type here. In a next PR I'll try to reify the use of type guards so that I can narrow it to `Scale<number>` without an `as`, and can remove the other type assertions this PR added

packages/charts/src/chart_types/xy_chart/state/selectors/compute_small_multiple_scales.ts
```diff
@@ -43,6 +43,10 @@ export function getScale(
   padding: RelativeBandsPadding = DEFAULT_SM_PANEL_PADDING,
 ) {
   const singlePanelSmallMultiple = domain.length <= 1;
-  const defaultDomain = domain.length === 0 ? [undefined] : domain;
-  return new ScaleBand(defaultDomain, [0, maxRange], undefined, singlePanelSmallMultiple ? 0 : padding);
+  return new ScaleBand(
+    domain.length > 0 ? domain : [(undefined as unknown) as number],
```
monfera, line 47: Yes we've been cheating to comply with the D3 scale typedef file and the D3 doc, but now we're clear about it. I'll try to solve it in a next PR, it's only partially addressed in this one

packages/charts/src/scales/scale_band.ts
```diff
@@ -81,15 +81,15 @@ export class ScaleBand implements Scale {
     this.bandwidth = this.d3Scale.bandwidth() || 0;
     this.originalBandwidth = this.d3Scale.bandwidth() || 0;
     this.step = this.d3Scale.step();
-    this.domain = this.d3Scale.domain();
+    this.domain = [...new Set(domain)];
```
monfera, line 84: This property already doesn't have the `undefined` hack for D3 I mentioned in another comment, it's apparently only needed for an undocumented effect in D3, will investigate in the next PR

packages/charts/src/utils/events.ts
```diff
@@ -33,13 +33,6 @@ export function hasDragged(prevProps: DragCheckProps | null, nextProps: DragChec
...
-
-  if (prevLastDrag === null && nextLastDrag !== null) {
-    return true;
-  }
-  if (prevLastDrag !== null && nextLastDrag !== null && prevLastDrag.end.time !== nextLastDrag.end.time) {
-    return true;
-  }
-  return false;
+  const nextLastDrag = nextProps.lastDrag;
+  return nextLastDrag !== null && (prevLastDrag === null || prevLastDrag.end.time !== nextLastDrag.end.time);
```
monfera, line 37: Replacing a hairy control structure and making the logical conditions less redundant

nickofthyme (approved):

PR #1383, opened 2021-09-17
refactor: scale improvements

monfera:
## Summary

Switch to TS 4.4; simplifications; type strength increase via lots of `any` replacements. Replacement and future prevention of coercive `==` and `!=` ops

### BREAKING CHANGE

The public type variations for domains are discontinued, in favor of retaining the single `DomainRange` export, which however now has a mandatory `{min: number, max: number}`. The user can supply `NaN` where a finite min, max or both aren't defined (ie. in place of former omisison or `undefined`).
Some console.warn punctuations changed.

## Details

- All `any`s removed from continuous and band scales; one `any` removed from `Scale`. Some of these `any` fixes needed some new type assertions in more specific, localized places (very low count), future PRs can attempt to solve those too with type guards or generics.
- Lots of code refactoring, optionality removal, parameter removal, function removal, and removal of some ad-hoc types
- 350 lines removed 👯‍♀️ if we don't count the yarn lock growth due to deps bump 

Note to reviewers: 
- please double check the unit test changes if they're in line with the original intent of those cases
- please check the type changes, esp. in place of `any`s, and the downstream type assertions necessitated by `any` removal, because the numerous preexisting (and still extant) `any`s mean that information can't always be learned from other types
- ideas about, or follow-up PRs for the reduction of remaining `any` (eg. in `Scale`) and the ` as ` assertions, necessitated by the removal of the `any` instances are welcome, due to your experience with the Cartesian implementation and use cases within Elastic

### Checklist


monfera:
test this

monfera:
test this

monfera:
test this

monfera:
test this

packages/charts/api/charts.api.md
```diff
@@ -1111,9 +1111,9 @@ export type HistogramModeAlignment = 'start' | 'center' | 'end';
 
 // @public (undocumented)
 export const HistogramModeAlignments: Readonly<{
-    Start: LineAlignSetting;
-    Center: LineAlignSetting;
-    End: LineAlignSetting;
+    Start: HistogramModeAlignment;
+    Center: HistogramModeAlignment;
+    End: HistogramModeAlignment;
```
monfera, line 1116: Not sure why it changed in this specific PR, can look unless someone has a hunch

packages/charts/src/components/portal/tooltip_portal.tsx
```diff
@@ -94,16 +94,12 @@ const TooltipPortalComponent = ({
...
-  const popperSettings = useMemo(
-    () => mergePartial(DEFAULT_POPPER_SETTINGS, settings, { mergeOptionalPartialValues: true }),
-
-    [settings],
-  );
-
+  const base = DEFAULT_POPPER_SETTINGS;
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  const ps = useMemo(() => mergePartial(base, settings, { mergeOptionalPartialValues: true }), [settings, base]);
```
monfera, line 100: `base` got extracted so the `const ps` is in a single line, otherwise `@ts-ignore` isn't effective

packages/charts/src/scales/scale_band.ts
```diff
@@ -60,36 +60,39 @@ export class ScaleBand<T extends number | string> implements Scale<T> {
...
+    // [7] => [7, 7]
+    // [7, 9, ...] => [7, 9, ...]
+    /*
+        const preDomain = [...new Set(inputDomain)];
+        this.domain = [
+          ...preDomain,
+          ...new Array(Math.max(0, 2 - preDomain.length)).fill(preDomain.length > 0 ? preDomain[0] : undefined),
+        ] as [T, T, ...T[]];
+    */
+
```
monfera, line 76: I'm planning to reify the domain as `[T, T, ...T[]]` ie. minimum 2 entries

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
```diff
@@ -106,7 +106,7 @@ export function shapeViewModel(
       ? new ScaleContinuous(
           {
             type: ScaleType.Time,
-            domain: xDomain.domain,
+            domain: xDomain.domain as number[],
```
monfera, line 109: A couple of new assertions were necessitated by the removal of a bunch of `any`, these should eventually be revisited with better type guarding or more generic functions but it's already a win to localize where we skip static type tests

packages/charts/src/scales/scale_continuous.ts
```diff
@@ -476,9 +476,9 @@ export class ScaleContinuous implements Scale<number> {
   ): {
     value: number;
     withinBandwidth: boolean;
-  } | null {
+  } {
     if (data.length === 0) {
-      return null;
+      return { value: NaN, withinBandwidth: false };
```
monfera, line 481: The current `master` has semantics more akin to `value: 0` for the `data.length === 0` case: the question mark will make the `invertWithStep` yielded results go `null`, which then gets `Math.min` / `Math.max`-ed. `null` gets coerced to zero in such cases. No complaints in current `master` because `any`. I think the zero semantics was accidental and more explicit "breaker" value like `NaN` feels better

monfera:
test this

monfera:
test this

monfera:
test this

monfera:
test this

monfera:
test this

monfera:
test this

packages/charts/src/chart_types/xy_chart/domains/nice.ts
```diff
@@ -1,12 +0,0 @@
...
- * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
- * or more contributor license agreements. Licensed under the Elastic License
- * 2.0 and the Server Side Public License, v 1; you may not use this file except
- * in compliance with, at your election, the Elastic License 2.0 or the Server
- * Side Public License, v 1.
- */
-
-/** @internal */
-export function areAllNiceDomain(nice: Array<boolean>) {
-  return nice.length > 0 && nice.every((d) => d);
```
monfera, line 11: Slight change to the meaning: the replacement of this function `!nice.includes(false)` yields `true` with an empty array

monfera:
test this

monfera:
test this

nickofthyme:
@elasticmachine run elasticsearch-ci/docs

nickofthyme:

monfera:
test this

packages/charts/src/scales/scale_band.test.ts
```diff
@@ -41,8 +41,8 @@ describe('Scale Band', () => {
     expect(scale.bandwidth).toBe(25);
     expect(scale.isValueInDomain('a')).toBe(true);
     expect(scale.isValueInDomain('b')).toBe(true);
-    expect(scale.isValueInDomain('z')).toBe(false);
-    expect(scale.isValueInDomain(null)).toBe(false);
+    expect(scale.isValueInDomain(('z' as unknown) as 'a' | 'b' | 'c' | 'd')).toBe(false);
+    expect(scale.isValueInDomain((null as unknown) as 'a' | 'b' | 'c' | 'd')).toBe(false);
```
nickofthyme, line 45: I think we could be looser on using `any` inside tests because unit tests are for the logic and not the types in my eyes. Especially in this case where you are casting something that is obviously not the right type for testing a js edge case.

In this case we could say `@ts-expect-error` with a description to show this was intentional and why.

```ts
    // @ts-expect-error: bad input
    expect(scale.isValueInDomain('z')).toBe(false);
    expect(scale.isValueInDomain(null)).toBe(false);
```

the issue with this is that it applied to the whole line so you don't get type checking outside of the one bad variable. If there were more than one intentional error I think this usage would be good but in this particular case I'm in favor of using `any`.

```suggestion
    expect(scale.isValueInDomain('z' as any)).toBe(false);
    expect(scale.isValueInDomain(null)).toBe(false);
```

packages/charts/src/utils/common.test.ts
```diff
@@ -34,16 +33,6 @@ describe('common utilities', () => {
...
-  test('identity', () => {
-    expect(identity('text')).toBe('text');
-    expect(identity(2)).toBe(2);
-    const a = {};
-    expect(identity(a)).toBe(a);
-    expect(identity(null)).toBe(null);
-    expect(identity(undefined)).toBeUndefined();
-    const fn = () => ({});
-    expect(identity(fn)).toBe(fn);
-  });
```
nickofthyme, line 46: I'd like to keep this.

storybook/webpack.config.js
```diff
@@ -120,5 +120,6 @@ module.exports = async ({ config }) => {
     '@elastic/charts/': path.resolve(__dirname, '../packages/charts/'),
   };
 
+  // eslint-disable-next-line @typescript-eslint/return-await
   return await config;
```
nickofthyme, line 124: remove `async` from function expression too

```suggestion
  return config;
```

nickofthyme (changes requested):
Really incredible changes all around Robert! 🎉 

I have two contentions:

### Loosening null/undefined checks

I see a lot of places where we loosed the type checking from check for (`==`) or against (`!=`) `null` and/or `undefined`. This even occurs on values whose types are neither `null` nor `undefined`. I'm not sure what the use case was for these checks, maybe some js checking, but I'd like to make sure there are no regressions by removing them.

### requiring domain `min` and `max`

I do not think the we should require the consumer of charts to specify the full extents. Before it was one or the other. I think they should be optional on the `Axis.domain` spec and default to `NaN`.

monfera:
Thanks @nickofthyme for checking! 

### Null checks

I'll revisit the places where I switched a `==`/`!==` null check to `===` / `!==`. I grepped these and checked most but not all. Sometimes there was a `== null` check when the value couldn't have been null (just `undefined` or maybe not even that). Likely a result of refactorings over time, which is good (reduction of type variability)

### Domain min/max requirements

I agree that making the prop non-optional is a bit of a departure from our established ways. Charts is consumed by Elastic solutions. The current API feels overly geared toward use cases we increasingly don't cater for (largely outside our control, eg. licensing), with the price of API type complexity that permeates the full depth of our implementation down to render functions (side note, I've long proposed a separation of the API layer to liberate the internals from overly intricate API offerings).

My goals were, in this order:
1. stem all those type variations
2. reduce static and runtime code path divergence by shedding some optionalities or union types; in particular, prevent branching contamination of internals by somewhat arbitrary API complexity 
3. reduce boilerplate (code noise)
4. (minuscule step) move toward a more deliberate, judicious, balanced use for runtime (bundle size etc.) and code source (code readability, and mental space dedicated/required by various aspects)

I'd like to understand why requiring `NaN` in an otherwise required object is burdensome. It's simple to always fill three props. We are not developing the library for individual, ad-hoc library users with wildly different circumstances to roll plots into their data science Jupyter or Observables notebook, so we don't have "retail user" facing requirements. It's likely not as burdensome for the Elastic Developer to use a required property and put `NaN` in it as to read through all the public type permutations. TS is powerful but my feeling is that we're currently on the side of overspending it on union types and optional properties often for no clear gain but of significant cost (while underutilizing TS power via too many `any` with broad downstream reach). I'd like to discuss this as there may be aspects I'm unaware of; my feel is that the developers using our library would also benefit from more regular and less intricately chiselled API choices. Let's work out a more uniform target use of TS, it's over- and underused here and there, depending on the aspect

integration/page_objects/common.ts
```diff
@@ -340,7 +340,7 @@ class CommonPage {
 
       expect(element).toMatchImageSnapshot();
     } catch (error) {
-      throw new Error(error);
+      throw new Error(String(error));
     }
```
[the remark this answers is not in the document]
monfera: Glad to get rid of the try/catch, haven't analyzed its purpose. As we discussed I'd like to get rid of try/catch in the use of scales too, started that work but shelved at the time, it's best to avoid try/catch use in normal chartmaking code as usually there's a better option

packages/charts/src/chart_types/partition_chart/layout/utils/legend_labels.ts
```diff
@@ -29,12 +29,10 @@ function flatSlicesNames(
...
     const formatter = layer?.nodeLabel;
-    let formattedValue = '';
-    if (key != null) {
-      formattedValue = formatter ? formatter(key) : `${key}`;
-    }
+    // todo null check under what circumstances key can be a null, as the type says it can't
+    const formattedValue = key === null ? '' : formatter ? formatter(key) : `${key}`;
     // preventing errors from external formatters
-    if (formattedValue != null && formattedValue !== '' && formattedValue !== HIERARCHY_ROOT_KEY) {
+    if (formattedValue && formattedValue !== HIERARCHY_ROOT_KEY) {
```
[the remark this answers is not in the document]
monfera: For this PR, my goal was mostly, no coercive comparisons. Yet, when viewing some of these, it was appealing to improve the code like here (the single, comapct truthiness check rules out multiple things). In the change you highlighted here, the type safety is at least as strong as before.

I agree there's a real risk of a library emitting unexpected types. Whenever we use a library (whether or not it has an associated type library) we call some shots and live with assumptions about what types may be returned. 

There is value in becoming more runtime-defensive, esp. if we collectively design this. For example, instead of excluding known "bad" values like `null` (and maybe still getting an unexpected `undefined`, `NaN`, whatever) we can use positive checks I've pitched before (and use some in this PR) such as `Number.isFinite()`

I also proposed that we wrap around each dependency we use. For example, D3 scales are a too embedded in Charts. `packages/charts` and even more so, `packages/charts/src` shouldn't use as many node imports for calculations as today, eg. D3 utils, time utils, color utils etc. should be wrapped in simple, compact abstractions outside our `src`, with whatever static and runtime type checks and coercions

integration/server/mocks/@storybook/addon-knobs/index.ts
```diff
@@ -46,16 +46,14 @@ export function text(name: string, dftValue: string, groupId?: string) {
...
+  // the # used for the color knob needs to be escaped on the URL and unescaped here
+  return value === null ? dftValue : unescape(value);
 }
 
 export function array(name: string, dftValues: unknown[], options: any, groupId?: string) {
   const params = getParams();
   const values = [];
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
   // @ts-ignore
   for (const [key, value] of params) {
```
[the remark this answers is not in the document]
monfera: Glad to get rid of it 💃

packages/charts/src/chart_types/partition_chart/layout/viewmodel/scenegraph.ts
```diff
@@ -26,7 +26,7 @@ import { shapeViewModel } from './viewmodel';
 
 function rawTextGetter(layers: Layer[]): RawTextGetter {
   return (node: ShapeTreeNode) => {
-    const accessorFn = layers[node[DEPTH_KEY] - 1].nodeLabel || identity;
+    const accessorFn = layers[node[DEPTH_KEY] - 1].nodeLabel || ((d) => d);
     return `${accessorFn(node.dataName)}`;
```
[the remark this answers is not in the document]
monfera: The function is not much longer than the function name. There's no particular or measurable overhead with it. I originally created the `identity` function but we didn't have a lot of use for it, and I felt an equally long named abstraction for `d => d` plus its noisy imports wasn't worth it. Glad to re-add, lmk

packages/charts/src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -342,31 +349,6 @@ export interface YDomainBase {
...
-export type UpperBoundedDomain = DomainBase & UpperBound;
-/** @public */
-export type CompleteBoundedDomain = DomainBase & LowerBound & UpperBound;
-/** @public */
-export type UnboundedDomainWithInterval = DomainBase;
-
-/** @public */
-export type DomainRange = LowerBoundedDomain | UpperBoundedDomain | CompleteBoundedDomain | UnboundedDomainWithInterval;
 /** @public */
 export type YDomainRange = YDomainBase & DomainRange & LogScaleOptions;
```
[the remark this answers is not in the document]
monfera: See detailed comments below, let's discuss this

packages/charts/src/scales/scale_band.test.ts
monfera: I agree, `any` is absolutely fine by me in tests, in fact usually better

packages/charts/src/utils/common.test.ts
monfera: If we keep the util, it can stay ofc. Btw. it's an example that shows how the coverage and boundaries of unit testing are arbitrary. If we use `d => d` inlined (which is a good choice too), it's well understood by the new code reader, without having to jump to a function def. It also needs no test coverage. It's interesting that a somewhat arbitrary choice like whether or not to extract / export a `d => d` leads to a test case 

Over time, it'd be great to thematically collect utilities, such as, in this case, maybe FP utils, that'd live in `packages`, with their tests

packages/charts/src/scales/scale_band.ts
```diff
@@ -60,36 +60,27 @@ export class ScaleBand<T extends number | string> implements Scale<T> {
...
+    this.domain = (inputDomain.length > 0 ? [...new Set(inputDomain)] : [undefined]) as [T, T, ...T[]];
     this.range = range.slice();
-    if (overrideBandwidth) {
-      this.bandwidth = overrideBandwidth * (1 - safeBarPadding);
-    }
     this.bandwidthPadding = this.bandwidth;
-    // TO FIX: we are assuming that it's ordered
-    this.isInverted = this.domain[0] > this.domain[1];
-    this.invertedScale = scaleQuantize<T>().domain(range).range(this.domain);
+    this.isInverted = inputDomain.length > 1 && this.domain[0] > this.domain[1]; // fixme: we are assuming that it's ordered
```
markov00, line 80: you can make this false by default and fix the domain type to `[T,...T[]]`

markov00 (commented):
Thanks for this great refactoring
I also have some doubts on the required `min`,`max` domain, but I think in Kibana at least, every chart is already consumed with both min and max properties.
We definitely need that API ->to->internals middle-layer to transform API values into an unambiguous extended configuration.

packages/charts/src/components/portal/tooltip_portal.tsx
monfera: This got taken care of via @nickofthyme's suggestion

monfera:
I blindly assumed the need for past `==`/`!=` indeed it's a good idea to revisit these 😄 
image

.eslintrc.js
```diff
@@ -89,11 +89,12 @@ module.exports = {
     'jsx-a11y/mouse-events-have-key-events': 1,
     'jsx-a11y/click-events-have-key-events': 1,
     '@typescript-eslint/member-ordering': 1,
-    eqeqeq: 1,
+    eqeqeq: 2,
 
     /*
      * Standard rules
      */
+    '@typescript-eslint/object-curly-spacing': 0,
```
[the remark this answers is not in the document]
nickofthyme: No action required here btw, just confirming.

packages/charts/src/chart_types/partition_chart/layout/utils/legend_labels.ts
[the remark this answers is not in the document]
nickofthyme: > In the change you highlighted here, the type safety is at least as strong as before

before, not shown in the code diff, was checking `key == null`.

> There is value in becoming more runtime-defensive

Agreed

> I also proposed that we wrap around each dependency we use

Agreed

packages/charts/src/chart_types/heatmap/state/selectors/get_legend_items_labels.ts
```diff
@@ -15,10 +15,11 @@ import { computeLegendSelector } from './compute_legend';
...
-        return { label: `${label}${showLegendExtra ? defaultExtra.formatted : ''}`, depth: 0 };
-      }
-      return { label, depth: 0 };
-    }),
+    legendItems.map(({ label, defaultExtra }) => ({
+      label:
+        defaultExtra && (defaultExtra.formatted ?? null) !== null
+          ? `${label}${showLegendExtra ? defaultExtra.formatted : ''}`
+          : label,
+      depth: 0,
```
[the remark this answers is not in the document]
monfera: Beautiful!

packages/charts/src/utils/domain.ts
```diff
@@ -19,105 +18,61 @@ export type ContinuousDomain = [min: number, max: number];
...
-): string[] | number[] {
-  // TODO: Check for empty data before computing domain
-  if (data.length === 0) {
-    return [0];
-  }
-
-  const domain = data.map(accessor).filter((d) => (removeNull ? d !== null : true));
-  const uniqueValues = [...new Set(domain)];
+export function computeOrdinalDataDomain<T>(data: T[], sorted: boolean, removeNull: boolean): T[] {
+  const uniqueValues = [...new Set(data.filter((d) => (removeNull ? d !== null : true)))];
```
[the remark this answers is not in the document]
monfera: Thanks, I should've noticed the `true` in a branch this when inlining `domain`, will merge

integration/page_objects/common.ts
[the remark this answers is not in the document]
monfera: I looked at it and don't have the courage to tamper with it in this PR, I think @nickofthyme you'd have better intuition about what to do with it

packages/charts/src/chart_types/partition_chart/layout/utils/legend_labels.ts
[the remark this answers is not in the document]
monfera: | before, not shown in the code diff, was checking key == null.

True! I meant the last quoted line, `formattedValue` as a predicate is at least as strong as `formattedValue != null`

packages/charts/src/scales/scale_band.ts
```diff
@@ -35,7 +35,7 @@ export class ScaleBand<T extends number | string> implements Scale<T> {
 
   readonly type: ScaleBandType;
 
-  readonly domain: T[];
+  readonly domain: [T, T, ...T[]];
```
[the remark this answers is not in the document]
monfera: I'll need to check it, having put that type in b/c of some constraint I encountered that demanded two values at least, might have misunderstood some piece of code

packages/charts/src/scales/scale_band.ts
monfera: Oh this was the place that needed two elements 👯‍♀️  I'll set it to `false` (I think it's then not _by default_ as it can't be overridden, but simply, _always_) and see if the type can be made simpler

monfera: e53232128d567f1e2b647834a1d32f2a0972b2ff

monfera:
test this

nickofthyme (approved):
Latest code changes LGTM! Great work Robert!!

markov00 (approved):
Great work here Robert, thanks for taking care of this.
I failed miserably the last time I've tried typing the scale, but you did it an awesome job here!

PR #1386, opened 2021-09-20
test: PR for testing Kibana a-la-carte DO_NOT_MERGE

markov00:
# DO_NOT_MERGE

PR used to test Kibana a la carte


packages/charts/src/renderers/canvas/index.ts
```diff
@@ -37,6 +37,8 @@ export function clearCanvas(ctx: CanvasRenderingContext2D) {
   withContext(ctx, () => {
     ctx.setTransform(1, 0, 0, 1, 0, 0);
     ctx.clearRect(0, 0, ctx.canvas.width, ctx.canvas.height);
+    ctx.fillStyle = 'red';
```
kertal, line 40: This is beautiful, to make it more seem like blood could we switch to r166,g16,b30? maybe playing some 🧛 metal in the background?

Bildschirmfoto 2021-09-20 um 16 08 16

PR #1388, opened 2021-09-20
test: don't import jsdom in node environment

markov00:
I've recently introduced a small change on the dependencies in our jest tests https://github.com/elastic/elastic-charts/pull/1367
After merging that PR, the Jenkins CI was suffering from a missing C++ library, used by the `canvas` package, updated in the mentioned PR.
The issue was that the `canvas` package doesn't need to be called in the VRT environment, but just in the uni test one.
I've discovered that the `canvas` package is only requested when using a `jsdom` environment in jest, but the current VRT tests runs in a `node` environment instead.
I've finally found the root cause: a JSDOM import, in `page_objects/common.ts` that causes the `canvas` lib to be requested and resolved in the jest `node` environment.
I've removed both the import and the unused function that was using that import.

### DIFF
--- a/integration/page_objects/common.ts
+++ b/integration/page_objects/common.ts
@@ -8,7 +8,6 @@
 
 import Url from 'url';
 
-import { JSDOM } from 'jsdom';
 import { AXNode } from 'puppeteer';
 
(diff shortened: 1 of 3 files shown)


monfera (approved):
LGTM!

PR #1389, opened 2021-09-20
feat(axis): add options for linear binary ticks

nickofthyme:
WIP


packages/charts/src/chart_types/xy_chart/utils/get_binary_ticks.ts
```diff
@@ -0,0 +1,56 @@
...
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+/* eslint-disable */
+
+var e10 = Math.sqrt(50),
+    e5 = Math.sqrt(10),
+    e2 = Math.sqrt(2);
```
nickofthyme, line 13: A @markov00 / @monfera Do either of you know what this logic is for? See https://github.com/d3/d3-array/blob/e6b91a3943665836019d631cea932da3c5f8ec82/src/ticks.js

markov00: Not 100% sure, but it looks like this is used to compute the tick increment, using multiple of 2, 5 or 10 depending on the "pure" step (domain extent / count of ticks)

packages/charts/src/scales/scale_continuous.ts
```diff
@@ -262,6 +263,11 @@ type ScaleOptions = Required<LogScaleOptions, 'logBase'> & {
    * Show only integer values
    */
   integersOnly: boolean;
+  /**
+   * scale base used to determine ticks in linear scales
+   * @defaultValue 10
+   */
+  linearBase: number;
```
monfera, line 270: It's a good idea to have a base number like this, though then what's the benefit of also having a more specific, `LinearBinary` scale type? The base pertains to ony the tick generation, not the nature of the scale but there may be some other reason I'm not thinking of now

packages/charts/src/scales/scale_continuous.ts
```diff
@@ -341,8 +353,16 @@ export class ScaleContinuous implements Scale<number> {
...
+      const scaleCopy =
+        type === ScaleType.Linear
+          ? getNiceLinearTicks(
+              this.d3Scale as ScaleContinuousNumeric<PrimitiveValue, number>,
+              desiredTickCount,
+              isBinary ? 2 : linearBase,
+            )
+          : (this.d3Scale as ScaleContinuousNumeric<PrimitiveValue, number>).domain(this.domain).nice(desiredTickCount);
+
+      this.domain = scaleCopy.domain();
```
monfera, line 365: Would be neat to do this on the basis of #1410 as that changes these bits of code here

packages/charts/src/chart_types/xy_chart/utils/get_linear_ticks.ts
```diff
@@ -0,0 +1,115 @@
+/*
```
monfera, line 1: Mentioning here too if others review the PR: this file is closely derived from D3 code, therefore we can't put it under our copyright, we must retain and comply with the original license. 

My understanding is that it's legal to take a D3 file, and modify it, only as long as the original license is kept. Though as mentioned I prefer this were not a source copy/paste but a fork under the `elastic` org (and a PR to the upstream, in the spirit of OS, giving the option of receiving neat functionality to the D3 creator) because that way, its independent, D3 originated test suite can fully run. 

If we don't fork right now, let's put it outside `packages/charts` or certainly outside `src`, I feel we shouldn't have external code copy-pasted into `src` even with the intact, original license, they should be treated like npm dependencies, carefully delineated from code under our license (see our earlier chat on other utils)

Our longer term trend is to not depend on eg. `d3-scale` for Cartesians like we do now, and we have some code already for nice tick generation etc. But that's cleanroom code, not derived from D3 or anything.

IANAL but my firm layman's understanding is that even if we modify some external code fully beyond recognition, and not a single line or variable name remains, it's still derivative work. We take on legal and reputational risk if we take code, modify and remove the license

packages/charts/src/scales/scale_continuous.ts
nickofthyme: Yeah I imitated the `logBase` property. I agree it's a little strange for it to be only affect the ticks.

I wonder if it's worth dismantling the scale classes to extend a base scale or even a continuous scale that each only contains logic for that type be it `Linear`, `Log` or whatever.

packages/charts/src/scales/scale_continuous.ts
nickofthyme: How do you mean?

packages/charts/src/chart_types/xy_chart/utils/get_linear_ticks.ts
nickofthyme: I totally agree, I appreciate you brining it up! I added 4d298a1.

My thought it that we already have code inside `src/` that has adaptive licensing, such as `fast-deep-equal`. This is not ideal and I think we should attempt to move these outside of `src/` in the near future. However, with these changes I see them as temporary and we remove once the code changes are accepted in `d3-x`, however long that takes.

packages/charts/src/scales/scale_continuous.ts
monfera: I meant (yesterday), let's merge #1410 first, and you independently suggested this today, but it's fine either way so I retract the comment

packages/charts/src/scales/scale_continuous.ts
monfera: Wondering about it too. The duplicated axis tick label fix and the time axis improvement both induce changes that render the `Scale` objects more ephemeral. Esp. continuous `Scale` objects (in current `master`) have `desiredTickCount` and a fixed set of ticks immutably embedded in them, and we probably don't want to do D3-like mutation by reinitializing a `Scale` object with a different tick counts

packages/charts/src/scales/scale_continuous.ts
nickofthyme: Oh gotcha, yeah that's ok with me! 👍🏼

storybook/stories/scales/8_linear_binary.story.tsx
```diff
@@ -0,0 +1,70 @@
...
+  return rng(0, 1);
+}
+
+function generateRandomBinary(binaryLength: number) {
+  let binary = '0b';
+  for (let i = 0; i < binaryLength; ++i) {
+    binary += randomByte();
+  }
+  return binary;
+}
```
monfera, line 29:
```suggestion
function generateRandomBinary(bitCount: number) {
  const number = rng(0, 2 ** bitCount);
  return `0b${number.toString(2).padStart(bitCount, '0')}`;
}
```
It was `randomBit` not `randomByte` though this single expression removes the need

monfera: Btw. I don't fully understand the purpose, are we now expressly allowing strings into our API for numbers? Or is it a preexisting capability? I don't know why a user of the library wouldn't just pass a numeric value `rng(0, 2 ** bitCount)` instead of a string, are there some fractional accuracy concerns?

monfera (approved):
LGTM!

Would still like to farm out our copy of D3 derived code so our license is consistent in `charts/` or at least `charts/src` @markov00 seeking your input to this too

I see this PR exercises an ability for Charts to ingest strings for numbers, it's not this PR's making though

Minor code comments inside

packages/charts/src/scales/scale_continuous.ts
monfera: `linearBase` and `logBase` could be united into `base`, in line with what you said. It's always one or the other, and the other PR turns `logBase` to a number. Not necessarily in this PR, just curious if you see any drawback

nickofthyme: Yeah that's true

nickofthyme: I'm going to leave this for now as `logBase` is used widely outside of the scales file.

storybook/stories/scales/8_linear_binary.story.tsx
nickofthyme: Good idea. Changed in 9d25939

PR #1390, opened 2021-09-20
fix(bars): add xAxis tickFormatter check to value labels

### LINKED ISSUE #1387: Wrong tickFormat applied to the value labels
**Describe the bug**
When rendering dataValues on a bar chart, the `tickFormat` is only applied with no rotation is applied.
When a rotation is applied, the `tickFormat` is not considered and the values are rendered without a formatter.

**To Reproduce**
1. Go to https://elastic.github.io/elastic-charts/?path=/story/bar-chart--with-value-label-advanced&globals=theme:light
2. Rotate the chart 90 degree
3. Looking at the code, the `tickFormat` is applied to the left axis, but it still be applied also if the chart is rotated and the driving format axis should be the bottom one (without formatting in the current code)

**Expected behaviour**
The value label should be formatted with the `displayValueSettings.valueFormatter` or, as a fallback, the relative Y axis.

**Screenshots**

 - Elastic Charts: master

**Additional context**
Looks like the `getAxesSpecForSpecId` doesn't take in consideration the chart rotation

**Errors in browser console**
n/a

**Kibana Cross Issues**
https://github.com/elastic/eui/pull/4845#discussion_r708414162

rshen91:
## Summary
The tickFormat on the data values over bars now will correctly display when charts are rotated.

Rotations are accounted for in:
- tooltip headers
- tooltip bodies 
- data values over bar geometries

## Issues
Fixes #1387 

### Checklist


rshen91:
@markov00 I could use some clarification on expected behavior of rotations. In https://elastic.github.io/elastic-charts/?path=/story/bar-chart--with-value-label-advanced&globals=theme:light it seems like the value labels over the bars do respect the tickFormat with rotations. I've added the `Data Value Test` story to try and gain some insight in another example where I was seeing the issue. Let me know if I'm misunderstanding the problem in the `With Label Value Advanced` story.

I have the story working in the latest commit c46db3f  where the chart can be rotated 90 degrees and by toggling the `'turn off special tickFormat'` knob the data values on the bars are changing correctly. However, there's the issue of the ordinal axis and the tickFormat being applied to it leading to the NanK. I would love to make sure I'm implementing the fix correctly and not get into the axes in `axis_utils.ts` unless that's part of the fix that's needed. Is that a separate issue? Thank you!

rshen91:
Closing and opening in a new PR

PR #1391, opened 2021-09-20
docs(playground): fix webpack config

nickofthyme:
## Details

Fix playground build after breaking it in #1231 or #1178 

script now functions as intended...

```
yarn playground
```


rshen91 (approved):
LGTM! Yarn playground is back, baby! I also tested locally bringing in a different story example into playground -- no issues 👍🏻

PR #1392, opened 2021-09-20
fix(theme): allow theme overrides to be accepted

### LINKED ISSUE #1284: Theme overrides not working perfectly
**Describe the bug**
When using a theme override in certain circumstances the first values do **NOT** take precedence and revert to the `baseTheme` value. 

**To Reproduce**
1. Go to https://codesandbox.io/s/happy-violet-xcb3z?file=/src/App.tsx

**Expected behaviour**
In the example codesandbox above, I expect the `crosshair` value to be applied from `theme2` but it is actually being applied from the `baseTheme`.

**Additional context**
Likely related to the `mergeOptionalPartialValues` option on the `mergePartial` utility function.

rshen91:
## Details
The option for `{ mergeOptionalPartialValues: true }` is in most places in the codebase except the couple missed in this PR. 

## Issues
Closes #1284


rshen91:
jenkins test this

nickofthyme:
@rshen91 this is already done in #1410

rshen91:
Ok thanks @nickofthyme I'll close this to avoid the duplication/noise

PR #1396, opened 2021-09-21
fix(interactions): change allowBrushingLastHistogramBin to true

### LINKED ISSUE #1371: Brushing on last bucket in histogram produce wrong results
**Describe the bug**
Brushing on the last bucket of a histogram chart doesn't correctly report the brushed extent

**To Reproduce**
1. Go to https://elastic.github.io/elastic-charts/?path=/story/interactions--brush-selection-tool-on-histogram-time-charts&globals=theme:light
2. Open the Actions panel
3. Brush the last bin on the right
3. The action reports a wrong extent

**Expected behaviour**
The brush event should correctly report a sorted extent

**Screenshots**
see linked kibana issue

 - Elastic Charts: latest/34

**Additional context**

**Errors in browser console**
n/a

**Kibana Cross Issues**
- https://github.com/elastic/kibana/issues/111687

rshen91:
## Summary
This PR changes the default boolean value of allowBrushingLastHIstogramBucket to true instead of false. This setting change more accurately shows the correct value when brushing over the last bucket in a histogram

### BREAKING CHANGE
allowBrushingLastHistogramBucket renamed to allowBrushingLastHistogramBin on the Settings component defaults true and is only applied for histogram type charts. 

## Details
This setting wasn't removed and users are still able to set this to false if they would prefer. The naming of this setting was changed to bin instead of bucket after team discussion. 

## Issues
Closes #1371 

### Checklist


nickofthyme:
> This setting wasn't removed and users are still able to set this to false if they would prefer.

What are your thoughts about removing this setting altogether?

markov00 (changes requested):
Looking at the changes in the test looks like that we aren't properly using the `allowBrushingLastHistogramBucket` with the required constraint that the series needs to be an Histogram (histogramEnabled=true). Instead it looks like we just forget about that and  we use directly `allowBrushingLastHistogramBucket` in the calculation like
``` 
const maxDomainValue = xScale.domain[1] + (allowBrushingLastHistogramBucket ? xScale.minInterval : 0);
```
This create the issue that is visible in the test: in a simple bar chart, with the current default enabled `allowBrushingLastHistogramBucket` we are changing the behavior of the brush. The scope of the `allowBrushingLastHistogramBucket` should only apply to histograms.

I've noticed the same wrong logic applied to the `roundHistogramBrushValues`
we are rounding brush values also to non histogram based charts.

@nickofthyme it looks like that you added those properties, do you remember the reason why these logics are not strictly applied to histogram only charts?

nickofthyme:
> @nickofthyme it looks like that you added those properties, do you remember the reason why these logics are not strictly applied to histogram only charts?

@markov00 I think you are correct, that should only apply to histogram mode.

packages/charts/src/chart_types/xy_chart/state/selectors/on_brush_end_caller.ts
```diff
@@ -136,9 +134,10 @@ function getXBrushExtent(
   histogramMode: boolean,
   xScale: Scale<number>,
   smallMultipleScales: SmallMultipleScales,
+  allowBrushingLastHistogramBin: boolean = true,
```
monfera, line 137: Nit: `allowBrushingLastHistogramBin` is a mandatory argument to `getXBrushExtent` which we can tell from the presence of a parameter that's to the right of it, `seriesSpecs`, which is mandatory (no question mark, no default). So wherever `getXBrushExtent` is used, we already pass something for `allowBrushingLastHistogramBin` otherwise the type checkers would shout. So the default value `= true` can be removed, assuming we are not passing the function `undefined` for this param (this is the only case the `true` default value would be in effect). So it's worth revisiting the caller places, and pass `true` instead of `undefined` if we currently pass `undefined` anywhere. 

This is a very small thing that supports the larger theme of type simplicity: it's simpler to have `boolean` than `boolean | undefined`. Union types like this, and optional parameters should preferably be avoided, unless it leads to convoluted code

rshen91: This is great thanks! 1f5186a

storybook/stories/interactions/10a_brush_selection_bar_hist.story.tsx
```diff
@@ -22,7 +22,7 @@ export const Example = () => (
       onBrushEnd={action('onBrushEnd')}
       rotation={getChartRotationKnob()}
       roundHistogramBrushValues={boolean('roundHistogramBrushValues', false)}
-      allowBrushingLastHistogramBucket={boolean('allowBrushingLastHistogramBucket', false)}
+      allowBrushingLastHistogramBin={boolean('allowBrushingLastHistogramBin', false)}
```
nickofthyme, line 25: nit: maybe we make this the default value

```suggestion
      allowBrushingLastHistogramBin={boolean('allowBrushingLastHistogramBin', true)}
```

storybook/stories/interactions/12_brush_time_hist.story.tsx
```diff
@@ -48,7 +48,7 @@ export const Example = () => {
         onElementClick={action('onElementClick')}
         rotation={getChartRotationKnob()}
         roundHistogramBrushValues={boolean('roundHistogramBrushValues', false)}
-        allowBrushingLastHistogramBucket={boolean('allowBrushingLastHistogramBucket', false)}
+        allowBrushingLastHistogramBin={boolean('allowBrushingLastHistogramBin', false)}
```
nickofthyme, line 51: nit:

```suggestion
        allowBrushingLastHistogramBin={boolean('allowBrushingLastHistogramBin', true)}
```

nickofthyme (approved):
Code changes LGTM. Tested locally on storybook and looks good. 🎉

nickofthyme:
@rshen91 make sure you note the rename here as a breaking change.

storybook/stories/interactions/10a_brush_selection_bar_hist.story.tsx
rshen91: Good catch thank you 9ebadbd

storybook/stories/interactions/12_brush_time_hist.story.tsx
rshen91: 9ebadbd for this fix thank you

packages/charts/src/specs/settings.tsx
```diff
@@ -582,16 +582,16 @@ export interface SettingsSpec extends Spec, LegendSpec {
...
    * A brush selection range of [1.23, 3.6] with a domain of [1, 2, 3]
    *
    * - when true returns [1.23, 3.6]
    * - when false returns [1.23, 3]
    *
-   * @defaultValue false
+   * @defaultValue true
    */
-  allowBrushingLastHistogramBucket?: boolean;
+  allowBrushingLastHistogramBin: boolean;
```
monfera, line 594: Nice removal of an optional here 😍 Also, the Bucket -> Bin change 👍

markov00 (approved):
LGTM!

PR #1402, opened 2021-09-22
refactor(partition): remove config in favor on spec and theme controls

### LINKED ISSUE #1185: Partition chart should pull from theme and spec not config

### LINKED ISSUE #1499: Color contrast on dark theme show invisible link labels
**Describe the bug**
When using the default background color on the dark theme, or black background, the `linkLabel` on a partition chart should be white or at least visible. 

image
[photo link](https://github.com/elastic/elastic-charts/blob/16246550d35810ef7fb6bcf9ccf4118b4d876683/integration/tests/__image_snapshots__/partition-stories-test-ts-axis-stories-theme-dark-should-show-default-text-color-in-dark-theme-1-snap.png)

**To Reproduce**
1. Go to [`24_linked_label.story.tsx`](https://github.com/elastic/elastic-charts/blob/master/storybook/stories/sunburst/24_linked_label.story.tsx)
2. Apply no `textColor` to the `linkLabel` so it uses the default value
3. Change theme to `dark` (not `eui-dark`)
4. Notice the color contrast logic is bad rendering invisible (black) link labels
5. Changing the background color to black causes the same result.

**Expected behaviour**
Color contrast logic returns a sensible color with black background.

nickofthyme:
## Summary

Adds _true_ dark mode to partition charts with configurable theme styles.

### BREAKING CHANGES

The `PartitionSpec.config` prop is removed. All properties have been moved/renamed under new `Theme.partition` options or `PartitionSpec` itself with the following exceptions:

- `PartitionSpec.config.margin` is removed in favor of `Theme.chartPaddings`. This no longer supports ratios.
- `PartitionSpec.config.backgroundColor` is removed in favor of `Theme.background.color`.
- `PartitionSpec.config.width` and `PartitionSpec.config.height` both removed as they were never used.

## Details

Up to now `Partition` charts have had no connection to the chart `Theme` and thus had no canonical inheritance to dark or light theming modes. This PR fixes this connection. `PartitionSpec.config` was broken up and moved into `Theme`, the `PartitionSpec` or deleted, see details in table below.

| Config prop | Moved to `Theme.partition` | Moved to `PartitionSpec` | Unused (removed) | Remarks |
|:-:|:-:|:-:|:-:|:-:|
| `width` |  |  | ✅ | Always overridden by parent dimensions |
| `height` |  |  | ✅ | Always overridden by parent dimensions |
| `margin` | ✅ |  |  | No longer a ratio, use `Theme.chartMargins` or `Theme.chartPaddings` |
| `emptySizeRatio` | ✅ |  |  |  |
| `outerSizeRatio` | ✅ |  |  |  |
| `clockwiseSectors` |  | ✅ |  |  |
| `specialFirstInnermostSector` |  | ✅ |  |  |
| `partitionLayout` |  | ✅ |  | renamed to `layout` |
| `drilldown` |  | ✅ |  |  |
| `fontFamily` | ✅ |  |  |  |
| `circlePadding` | ✅ |  |  |  |
| `radialPadding` | ✅ |  |  |  |
| `horizontalTextAngleThreshold` | ✅ |  |  |  |
| `horizontalTextEnforcer` | ✅ |  |  |  |
| `maxRowCount` |  | ✅ |  |  |
| `fillOutside` |  | ✅ |  |  |
| `radiusOutside` |  | ✅ |  |  |
| `fillRectangleWidth` |  | ✅ |  |  |
| `fillRectangleHeight` |  | ✅ |  |  |
| `fillLabel` | ✅ |  |  |  |
| `fillLabel.valueFormatter` |  |  | ✅ | formatter was always overriden by config default. Use `valueFormatter` on `layers` or `spec`. |
| `linkLabel` | ✅ |  |  |  |
| `linkLabel.valueFormatter` |  |  | ✅ | formatter was always overriden by config default. Use `valueFormatter` on `layers` or `spec`. |
| `backgroundColor` |  |  | ✅ |  |
| `sectorLineWidth` | ✅ |  |  |  |
| `sectorLineStroke` | ✅ |  |  |  |

Changes also add per theme gap stroke color and fix #1499, see demo below...

Screen Recording 2021-12-16 at 04 43 34 PM

## Issues

fix #1185, fix #1499

### Checklist


monfera:
The reason for the config object was/is, automatic generation of config UI: runtime detectable value types, sensible number ranges for sliders and enum strings a UI would conceivably put into a dropdown, and of course booleans for checkboxes. Here's an example:

Oct-28-2019 08-48-01

Over time, some of these powers could be provided to Elastic users for direct tweaks, and would be useful for us too, and for our designers so they can quickly iterate on color, font size, line width, eventual animation parameters and other things.

It's not fleshed out in the code, the thinking became, class objects would be a good representation for the props. For class instances are the only composite data types in JS that are expressively typed in runtime (eg. `instanceof`) and statically (the classes double as TypeScript interfaces) even if I'm not a big fan of classes

nickofthyme:
> The reason for the config object was/is, automatic generation of config UI

@monfera yeah I realize that but this has bleed into the lack of coupling between charts and theme requiring chart users to define options that should be on theme in `config` every time they create a chart, which is not scalable.

I am 100% onboard creating a robust config that not only facilitates an options UI like you show above but also apply validation to provided values at lookup time or when instantiated. This is why I kept all the config meta data. Having a config like this in the future, maybe in playground to start, would be easy to configure such that it feeds config options to spec or theme accordingly.

.eslintrc.js
```diff
@@ -152,14 +152,7 @@ module.exports = {
...
     ],
-    'no-unused-vars': [
-      'error',
-      {
-        vars: 'all',
-        args: 'after-used',
-        ignoreRestSiblings: true,
-      },
-    ],
+    'no-unused-vars': 0,
```
nickofthyme, line 155: There is a ts rule for this

package.json
```diff
@@ -67,8 +67,8 @@
     "@elastic/datemath": "^5.0.3",
     "@elastic/eui": "^38.0.0",
     "@mdx-js/loader": "^1.6.6",
-    "@microsoft/api-documenter": "^7.12.7",
-    "@microsoft/api-extractor": "^7.18.9",
+    "@microsoft/api-documenter": "^7.13.54",
+    "@microsoft/api-extractor": "^7.18.11",
```
nickofthyme, line 71: extractor complained about using different ts version

packages/charts/src/chart_types/partition_chart/specs/index.ts
```diff
@@ -9,25 +9,21 @@
...
+import { FillFontSizeRange, FillLabelConfig } from '../../../utils/themes/partition';
+import { percentFormatter } from '../layout/config';
+import { AnimationConfig, PartitionLayout } from '../layout/types/config_types';
 import { NodeColorAccessor, ShapeTreeNode, ValueGetter } from '../layout/types/viewmodel_types';
 import { AGGREGATE_KEY, NodeSorter, PrimitiveValue } from '../layout/utils/group_by_rollup';
 
-interface ExtendedFillLabelConfig extends FillLabelConfig, FillFontSizeRange {}
+interface ExtendedFillLabelConfig extends FillLabelConfig, FillFontSizeRange {
+  valueFormatter: ValueFormatter;
+}
```
nickofthyme, line 26: Moved to here as the value on the `config` was never used and now the `linkLabel` and `fillLabel` configs are on the `Theme` and I prefer not to have functional parameters on the `Theme`.

packages/charts/src/common/color_calcs.test.ts
```diff
@@ -8,23 +8,11 @@
...
-      expect(fillTextColor(Colors.White.keyword)).toEqual(RGBATupleToString(Colors.Black.rgba));
-    });
-    // test contrast computation
-    it('should return black with yellow/semi-transparent background', () => {
-      expect(fillTextColor(`rgba(255,255,51,0.3)`)).toEqual(RGBATupleToString(Colors.Black.rgba));
-    });
-    it('should use white text for Thailand color', () => {
-      // black for WCAG2, white for WCAG3
-      expect(fillTextColor(`rgba(120, 116, 178, 1)`)).toEqual(RGBATupleToString(Colors.Black.rgba));
-    });
```
nickofthyme, line 27: moved to `packages/charts/src/common/fill_text_color.test.ts`

packages/charts/src/common/fill_text_color.ts
```diff
@@ -11,12 +11,43 @@ import { colorToRgba, RGBATupleToString } from './color_library_wrappers';
...
+    // combine it with white if semi-transparent
+    const bgBlend = combineColors(backgroundRGBA, Colors.White.rgba);
+    const blendedFgBg = combineColors(foregroundRGBA, bgBlend);
+    return RGBATupleToString(highContrastColor(blendedFgBg));
+  }
+
+  // combine it with white if semi-transparent
+  const bgBlend = combineColors(backgroundRGBA, Colors.White.rgba);
+  return RGBATupleToString(highContrastColor(bgBlend));
 }
```
nickofthyme, line 53: @markov00 I am proposing changing this logic to account for a fallback value. If the foreground and background colors are not suitable and there is a fallback color (e.g. color from the theme), then we use that instead of attempting to blend with white on unknown theme. This works the same as before when using none transparent colors.

I also renamed the parameters to be more inline with the `combineColors` naming where the foreground such as pie slice or bar color, can be `null` and the background color (was container color) is consistently the background color, before these were interchangeable.

Finally the background color is defaulted to `transparent` to allow the fallback usage where before this was using a value that was typically defaulted to white from outside this function.

packages/charts/src/components/accessibility/accessibility.test.tsx
```diff
@@ -69,17 +69,15 @@ describe('Accessibility', () => {
         <Partition
           id="spec_1"
           data={mocks.pie}
+          layout={PartitionLayout.treemap}
           valueAccessor={(d: Datum) => d.exportVal as number}
-          valueFormatter={(d: number) => `$${config.fillLabel.valueFormatter(Math.round(d / 1000000000))}\u00A0Bn`}
+          valueFormatter={(d: number) => `$${defaultValueFormatter(Math.round(d / 1000000000))}\u00A0Bn`}
```
nickofthyme, line 74: @monfera should we export this helper funciton from the library?

packages/charts/src/utils/themes/dark_theme.ts
```diff
@@ -230,4 +231,57 @@ export const DARK_THEME: Theme = {
...
+      padding: 0,
+      valueFont: {
+        fontWeight: 400,
+        fontStyle: 'normal',
+        fontVariant: 'normal',
+      },
+    },
+    sectorLineWidth: 1,
+    sectorLineStroke: 'black',
+  },
```
nickofthyme, line 286: We should desire on sensible coloring for dark and light themes.

nickofthyme (commented):
Just a few comments to explain some changes.

monfera:
Thanks Nick, your explanation makes sense, and the UI generation aspect is not something we exercise currently. It's not difficult to bring it back at some point, so shedding it for now is a good call

nickofthyme:
Most if not all partition chart VRT screenshots changed after removing `config.margin` and adding `chartPaddings` and `chartMargins`. I could set all these to `0` for sake of matching the screenshots and update just the padding in a follow up PR if people think that's better.

monfera:
> update just the padding in a follow up PR

I think it'd help a bit, so it's harder to miss some unintended change, though if it's hard to separate, it's fine either way

nickofthyme (commented):
@monfera do you know why the flame and icicle text is bigger? I cannot see any changes that would cause this.

image

nickofthyme:
> I think it'd help a bit, so it's harder to miss some unintended change, though if it's hard to separate, it's fine either way

Ok I reverted the margin and padding changes so we now show the charts as they were, or at least most of them. The caveat to this is the legend top position. I believe before this included the `topGroove` or top margin which is why all the partition chart legends have shifted down.

image

packages/charts/src/common/fill_text_color.ts
markov00: I have few comments on that:
1) instead of this code change, we can always use of the theme `textColor` for both the linked labels and the links lines. This "ensures" that the theme builder must consider the `linkedLabel` color independently from the theme background. If used with a transparent background, the `linkedLabel` should be manually adapted by the user depending on the real background used (within Canvas for example).
2) taking into account what is described in 1) the fillTextColor will then only be used to compute colors within slices/sectors. 
From the code, it looks like that the theme `textColor` is used only when a highly transparent (<0.1 alpha) slice/sector and a highly transparent background are combined. I'm not 100% sure but this doesn't fully solve the problem because a single`textColor` is not always highly contrasting with their background. 
The problem here is more on the "missing" background of the pie/slice/sector. There is a use case (Kibana Canvas) where we don't want to draw a piechart with linked labels with an opaque background because, probably, we want to draw some text/icons around it. In this case, where the `theme.background.color` is transparent, we have a problem. What if, instead of providing a fallback color for the text, we provide an opaque background color for the pie/sector? this color will be used the same way as the container background, but it will not be rendered on the canvas itself, because we rely on the `theme.background.color` or on the actual background color of the chart element.

Combining both 1) and 2) means replace the `textColor` from the `fillLabel` style with something like `opacqueBackgroundColor` or something along these lines, and we then use the function `fillTextColor(foreground,opacqueBackgroundColor)`  only for the filled labels, not for the linked ones

packages/charts/src/chart_types/partition_chart/layout/types/config_types.ts
```diff
@@ -127,14 +30,14 @@ export type EasingFunction = (x: Ratio) => Ratio;
...
 
 /** @public */
-export interface Config extends StaticConfig {
+export interface AnimationConfig {
   /** @alpha */
-  animation: {
+  animation?: {
     duration: TimeMs;
-    keyframes: Array<AnimKeyframe>;
+    keyframes?: Array<AnimKeyframe>;
```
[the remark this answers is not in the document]
nickofthyme: Because there were existing use cases that applied a partial animation object. I could revert this and just default to `[]`.

packages/charts/src/components/accessibility/accessibility.test.tsx
monfera: I think so, though that `defaultFormatter` would benefit from a review by @markov00 and you. I tried to make it sensible but wrote at a time when it was unclear to me if number formatting follows the locale or some option in the theme or somewhere else

packages/charts/src/chart_types/partition_chart/layout/types/config_types.ts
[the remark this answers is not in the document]
monfera: I think we can delete the keyframe based animation initiative, which generated some of the shared gifs during partition development, but not in production use. The only use of animation is in icicle/flame (active in [this story](https://elastic.github.io/elastic-charts/?path=/story/flame-alpha--cpu-profile-flame-chart&globals=theme:light;background:white) only) and it doesn't use keyframe based animation. When the time comes, we can build back better™️

monfera:
@nickofthyme just eyeballing the flame image diffs: it's not just that the font became bigger; also it looks like it's not using variable sized fonts. The idea is that within the minimum and maximum font size, we allow a range of font sizes, visible on the left but not present on the right
image

While sometimes the font sizes—even variable ones—look identical, sometimes there's significant difference, eg. often, smaller fonts in the PR:
image

There seem to be a stroke related regression here: the old one has a stroke color that appears identical with the background (which is a design goal) while in the feature branch, the stroke is visibly darker, easiest to see if you move the slider
image

Some charts shift a bit
image

While the upward movement of the legends looks like a good change, it masks other changes (as in, the reviewer may not do the slider thing, or not attentively, on all the dozens of images and some unexpected changes remain veiled)

The waffle shrank a bit for some reason:
image

All these look OK visually (except the font size issues), it's just I'm uncertain why some things changed the way they did.

The newly added dark background examples look fantastic 🎉

monfera:
@nickofthyme Thanks for your explanation, I now understand that some things changed in line with your expectation, even if the change doesn't look desirable, like making the waffle smaller.

In case of the stroke issue, the chart looks bad if the stroke and background colors do not match, even if there's a reasonable technical explanation for the change. As the stories are also copy/pasteable examples for making specific charts, it'd be better if it was restored to something sensible. Ideally we also avoid story charts that have obvious, or easy to avoid subtle design problems with them

nickofthyme:
@monfera I don't see a grossly different stroke. Can you point out what exactly your are seeing? Thanks.

packages/charts/api/charts.api.md
```diff
@@ -2122,12 +2111,13 @@ export interface Theme {
     colors: ColorConfig;
     // (undocumented)
     crosshair: CrosshairStyle;
-    // (undocumented)
     goal: GoalStyles;
     // (undocumented)
     legend: LegendStyle;
     lineSeriesStyle: LineSeriesStyle;
     markSizeRatio?: number;
+    // Warning: (ae-forgotten-export) The symbol "PartitionStyle" needs to be exported by the entry point index.d.ts
```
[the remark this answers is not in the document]
nickofthyme: 35f1d85

packages/charts/src/components/accessibility/accessibility.test.tsx
nickofthyme: Exported in e29f769

packages/charts/src/common/fill_text_color.ts
```diff
@@ -11,12 +11,43 @@ import { colorToRgba, RGBATupleToString } from './color_library_wrappers';
...
-  return RGBATupleToString(highContrastColor(blendedFbBg));
+  if (backgroundRGBA[3] < TRANSPARENT_LIMIT && !foreground && fallbackColor) return fallbackColor;
+
+  if (foreground) {
+    const foregroundRGBA = colorToRgba(foreground);
+
+    if (backgroundRGBA[3] < TRANSPARENT_LIMIT) {
+      if (foregroundRGBA[3] < TRANSPARENT_LIMIT && fallbackColor) return fallbackColor;
+      // combine it with white if semi-transparent
+      const fgBlend = foregroundRGBA[3] < 1 ? combineColors(foregroundRGBA, Colors.White.rgba) : foregroundRGBA;
```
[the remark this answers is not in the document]
nickofthyme: 599e540

nickofthyme:
@monfera It appears the flame font increase was caused because the flame chart never pulled the max value from the config. Now with it sharing the config the max font value was set to `64` across all partition types and caused the increase. Leaving this for now and will just have to set this as a manual override on the chart if desired. See 37f4349.

packages/charts/src/common/fill_text_color.ts
nickofthyme: @markov00 other than this vrt [diff](https://github.com/elastic/elastic-charts/pull/1402/files?short_path=4c87ef9#diff-4c87ef987868a6351740bdeb02e3af885b68acfc2adc07425d597bf01e5cb167) solving https://github.com/elastic/elastic-charts/issues/1499, it looks like the vrt show no changes to the color fill logic. What about merging this and revisiting the color logic in a future pr?

markov00 (approved):

PR #1404, opened 2021-09-23
chore: cleanup types

nickofthyme:
## Details

- remove unnecessary `// @ts-ignore` comments
- add descriptions to legit `// @ts-ignore` comments
- add [rule](https://github.com/typescript-eslint/typescript-eslint/blob/v4.31.1/packages/eslint-plugin/docs/rules/ban-ts-comment.md) to error on all ts-comments with no description
- remove `global.d.ts` in favor of `declarations/`
- fix type issue related to `jest-extended` and custom matchers
- fix types around `sortSeriesBy` hidden setting
- fix type errors around required `domain` options
- fix type errors around union brush events
- remove deprecated [`msSaveBlob`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/msSaveBlob) usage
- add step to ci to check `yarn typecheck:all` as type errors somehow snuck into `master`

### Eslint changes

Before eslint would **WARN** ⚠️ you when using `// @ts-ignore`, `// @ts-expect-error`, `// @ts-nocheck` or `// @ts-check`. This caused some to ignore the eslint warning but now it is explicit why we are ignoring a given line.

Now it will **ERROR** ❌  when using any of the above _without_ a description, otherwise it will not complain.

Image 2021-09-23 at 12 48 21 PM

### Checklist

### DIFF
--- a/declarations/@types/jest.d.ts
+++ b/declarations/@types/jest.d.ts
@@ -1,0 +1,33 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+export * from 'jest-extended'; // https://github.com/jest-community/jest-extended
+
+/**
+ * Final Matcher type with `this` and `received` args removed from jest matcher function
+ */
+type MatcherParameters<T extends (this: any, received: any, ...args: any[]) => any> = T extends (
+  this: any,
+  received: any,
+  ...args: infer P
+) => any
+  ? P
+  : never;
+
+declare global {
+  namespace jest {
+    interface Matchers<R> {
+      /**
+       * Expect array to be filled with value, and optionally length
+       */
+      toEqualArrayOf(...args: MatcherParameters<typeof toEqualArrayOf>): R;
+    }
+  }
+}
+
+export {}; // ensure this is parsed as a module.
--- a/storybook/stories/area/13_band_area.story.tsx
+++ b/storybook/stories/area/13_band_area.story.tsx
@@ -50,7 +50,11 @@ export const Example = () => {
       />
       <Axis
         id="left"
-        domain={{ fit }}
+        domain={{
+          min: NaN,
+          max: NaN,
+          fit,
+        }}
         title={KIBANA_METRICS.metrics.kibana_os_load[0].metric.title}
         position={Position.Left}
         tickFormat={(d) => Number(d).toFixed(2)}
--- a/storybook/stories/bar/44_test_single_histogram.story.tsx
+++ b/storybook/stories/bar/44_test_single_histogram.story.tsx
@@ -27,6 +27,8 @@ export const Example = () => {
   const formatter = timeFormatter(niceTimeFormatByDay(1));
 
   const xDomain = {
+    min: NaN,
+    max: NaN,
     minInterval: 60000,
   };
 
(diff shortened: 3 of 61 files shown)


storybook/stories/area/13_band_area.story.tsx
```diff
@@ -50,7 +50,11 @@ export const Example = () => {
       />
       <Axis
         id="left"
-        domain={{ fit }}
+        domain={{
+          min: NaN,
+          max: NaN,
+          fit,
+        }}
```
monfera, line 57: Nice catch! For some reason nothing brought it up

storybook/stories/bar/44_test_single_histogram.story.tsx
```diff
@@ -27,6 +27,8 @@ export const Example = () => {
   const formatter = timeFormatter(niceTimeFormatByDay(1));
 
   const xDomain = {
+    min: NaN,
+    max: NaN,
```
monfera, line 31: Wow so many of them 😄

storybook/stories/interactions/17_png_export.story.tsx
```diff
@@ -51,17 +51,12 @@ export const Example = () => {
...
-        document.body.appendChild(link);
-        link.click();
-        document.body.removeChild(link);
-    }
+    const link = document.createElement('a');
+    link.download = fileName;
+    link.href = snapshot.blobOrDataUrl;
+    document.body.appendChild(link);
+    link.click();
+    document.body.removeChild(link);
```
monfera, line 59: IE removal ftw!

storybook/stories/line/2_w_axis.story.tsx
```diff
@@ -26,7 +26,13 @@ const dateFormatter = timeFormatter(niceTimeFormatByDay(1));
...
 export const Example = () => (
   <Chart>
-    <Settings xDomain={{ max: 2 }} baseTheme={useBaseTheme()} />
+    <Settings
+      xDomain={{
+        min: NaN,
+        max: 2,
+      }}
+      baseTheme={useBaseTheme()}
+    />
```
monfera, line 35: It doesn't matter at all in story files and I should be the last person to bring it up with the NaNs, just wanted to mention that in regular source code it's nice to retain a single line for an object or sub-object if props aren't super numerous or deeply nested (in which case the linter breaks the lines anyway)

storybook/stories/area/13_band_area.story.tsx
nickofthyme: Yeah I'm not sure why, which is why I added `yarn typecheck:all` to the ci checks.

storybook/stories/line/2_w_axis.story.tsx
nickofthyme: Yeah I was copy pasting from a multiline so some got push to the next line. We can fix this later.

nickofthyme:
Jenkins test this please

rshen91 (approved):
I like these changes! LGTM! Thanks for going through and cleaning up unneeded @ts-ignore comments; I especially like the change to have comments when using @ts-ignore

declarations/@types/jest.d.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Final Matcher type with `this` and `received` args removed from jest matcher function
+ */
+type MatcherParameters<T extends (this: any, received: any, ...args: any[]) => any> = T extends (
+  this: any,
+  received: any,
+  ...args: infer P
+) => any
+  ? P
+  : never;
```
monfera, line 20: Great, we should use more and more `never`, eg. tying up the loose ends of `switch/case` statements or anything that branches over the full union of something (eg. the ending of `rawChildNodes`)

storybook/stories/area/13_band_area.story.tsx
monfera: Great! What's the local equivalent, just running `yarn typecheck:all`? Does it remove the need for running the API extraction, esp. if the API doesn't change?

storybook/stories/line/2_w_axis.story.tsx
monfera: These are perfectly OK here

monfera (approved):
Great improvement 🎉  My current review is cursory, will report any oddities, if any, down the road

PR #1406, opened 2021-09-23
feat(heatmap): dark mode with theme controls

### LINKED ISSUE #1335: [Heatmap] visualizations don't adjust color values for dark mode

nickofthyme:
## Summary

Screen Recording 2021-09-02 at 02 31 07 PM

### BREAKING CHANGES

The `HeatmapSpec.config` prop is removed. All properties have been moved/renamed under new `Theme.heatmap` options with the following exceptions:

- `HeatmapSpec.margin` was not used, `Theme.chartMargins` should be used in the future.
- `HeatmapSpec.fontFamily` `HeatmapSpec.width` and `HeatmapSpec.height` all removed as they were never used.

## Details

Applies #1356 to `master`

The `Heatmap`, `Partition` and `WordCloud` specs all have no connection to the chart theme and thus have no canonical inheritance to dark mode. This PR fixes this for the `Heatmap` chart and adds some added styles to the text labels as exposed in xy charts. The `HeatmapConfig` was broken down and moved into `Theme` or the `HeatmapSpec` accordingly. We should avoid having a `Config` option to control theme as this not only creates a disconnect but also allows for multiple sources of truth for similar options and another property requirement to toggle for theming between dark and light modes.

## Issues

closes #1335

### Checklist


packages/charts/src/chart_types/heatmap/state/selectors/get_x_axis_right_overflow.ts
```diff
@@ -9,17 +9,26 @@
...
+  [getChartThemeSelector, getHeatmapSpecSelector, getHeatmapTableSelector],
+  (
+    {
+      heatmap: {
+        xAxisLabel: { fontSize, fontFamily, padding, width },
+      },
+    },
+    { timeZone, xAxisLabelFormatter },
+    { xDomain },
+  ): number => {
```
monfera, line 31: Nit: in the name of scrolling and screen space economy, I wouldn't mind more compact objects, eg. 
`{heatmap: { xAxisLabel: { fontSize, fontFamily, padding, width } }`, not sure if here's some formatting option, maybe past `prettier`, which avoids vertical sprawl where there's plenty of room

packages/charts/src/chart_types/heatmap/state/selectors/is_brush_available.ts
```diff
@@ -7,17 +7,17 @@
...
  * The brush is available only if a onBrushEnd listener is configured
  * @internal
  */
 export const isBrushAvailableSelector = createCustomCachedSelector(
-  [getHeatmapConfigSelector, getSettingsSpecSelector],
-  (config, { onBrushEnd }): boolean => {
-    return Boolean(onBrushEnd) && config.brushTool.visible;
+  [getChartThemeSelector, getSettingsSpecSelector],
+  ({ heatmap: { brushTool } }, { onBrushEnd }): boolean => {
+    return Boolean(onBrushEnd) && brushTool.visible;
```
monfera, line 20: Nit: if there's a single return, a fat arrow function is good too, a tad less noise (no curlies, `return`, CRLF)

packages/charts/src/chart_types/heatmap/state/selectors/on_brush_end_caller.ts
```diff
@@ -26,8 +28,13 @@ import { isBrushEndProvided } from './is_brush_available';
  */
 export function createOnBrushEndCaller(): (state: GlobalChartState) => void {
   let prevProps: DragCheckProps | null = null;
-  let selector: Selector<GlobalChartState, void> | null = null;
-  return (state: GlobalChartState) => {
+  let selector: OutputSelector<
```
monfera, line 31: Super optional: while in the neighborhood, it'd be helpful to replace `let`s, sometimes it's easy, or some other approach like `const defaultSelector = ...` ... `return defaultSelector` vs `return createCustomCachedSelector(...)` so basically, manually switching to [static single assignment form](https://en.wikipedia.org/wiki/Static_single_assignment_form)

packages/charts/src/utils/themes/dark_theme.ts
```diff
@@ -230,4 +231,81 @@ export const DARK_THEME: Theme = {
       stroke: 'white',
     },
   },
+  heatmap: {
+    maxRowHeight: 30,
+    maxColumnWidth: 30,
+    brushArea: {
+      visible: true,
+      stroke: '#D3DAE6', // euiColorLightShade,
```
monfera, line 239: At some point in the future, the longwinded theme object ladders could be farmed out so they're external to `src`. Their types are already public for external themeability

monfera (commented):
The code looks good to me! I'd rely on @rshen91 or @markov00 doing proper smoke testing, due to more familiarity with themes. I made only optional code comments

The dark mocks look unusual in that (likely) low values are very salient white against the black background, another palette would be more appropriate, so our examples look more or less OK.

For the record I'm not a big fan of the gridline-free mock where it's hard to read which row a cell is in.

Also in some next PR the highlight could be fixed (I discussed it with @rshen91 earlier)

packages/charts/src/chart_types/heatmap/state/selectors/get_x_axis_right_overflow.ts
markov00: or just deconstruct partially to `{heatmap:{xAxisLabel}}`
and directly access the properties when needed

packages/charts/api/charts.api.md
```diff
@@ -907,8 +907,11 @@ export interface GroupBySpec extends Spec {
 // @public (undocumented)
 export type GroupId = string;
 
+// Warning: (ae-forgotten-export) The symbol "SpecRequiredProps" needs to be exported by the entry point index.d.ts
+// Warning: (ae-forgotten-export) The symbol "SpecOptionalProps" needs to be exported by the entry point index.d.ts
+//
 // @alpha (undocumented)
-export const Heatmap: React_2.FunctionComponent<Pick<HeatmapSpec, 'id' | 'data' | 'colorScale'> & Partial<Omit<HeatmapSpec, 'chartType' | 'specType' | 'id' | 'data'>>>;
+export const Heatmap: React_2.FunctionComponent<SpecRequiredProps_8 & SpecOptionalProps_8>;
```
markov00, line 914: not in this PR but we should find a way to express this in the right way

markov00 (approved):
changes LGTM
We should align and work together with EUI team to introduce the right changes to the theme

packages/charts/api/charts.api.md
nickofthyme: Could you elaborate what your mean here? Would #1421 solve this?

packages/charts/src/chart_types/heatmap/state/selectors/is_brush_available.ts
nickofthyme: Yeah I agree but these selector functions tend to get a little long with implicit returns and linting forces them to 2 lines with is less readable in some cases.

```ts
export const isBrushEndProvided = createCustomCachedSelector([getSettingsSpecSelector], ({ onBrushEnd }): boolean =>
  Boolean(onBrushEnd),
);
```

packages/charts/src/chart_types/heatmap/state/selectors/get_x_axis_right_overflow.ts
nickofthyme: IMO I don't see an issue with this. The line separation clearly shows the hierarchical destructuring, otherwise a single line forces your eyes to match `{}`. If we have a rule to force one way or another that's fine but as of now I'll leave it alone.

packages/charts/src/chart_types/heatmap/state/selectors/on_brush_end_caller.ts
nickofthyme: Yeah I think this could be cleaned up but don't want to create additional changes in this pr.

packages/charts/src/chart_types/heatmap/state/selectors/is_brush_available.ts
nickofthyme: 06240f4

PR #1408, opened 2021-09-23
refactor(goal): remove deprecated config

nickofthyme:
## Summary

Removes deprecated `config` from goal charts. Finalizing changes from #1299.

### BREAKING CHANGE

The `GoalSpec.config` prop is removed. All properties have been moved/renamed under new `Theme.goal` options with the following exceptions:

- `Config.margin` is now controlled by `Theme.chartMargins` and is no longer a margin ratio as before.
- `Config.backgroundColor` is now controlled by `Theme.background.color`, even though it's not yet used.
- `fontFamily` moved into each respective label styles
- `angleStart` and `angleEnd` are moved onto the `GoalSpec` as optional values.
- `sectorLineWidth`, `width` and `height` all removed as they were never used.

## Details

See #1299

### Checklist


monfera (approved):
Nice code trimming 🥩

nickofthyme:

PR #1413, opened 2021-09-28
fix(goal): reduce whitespace for circular charts

### LINKED ISSUE #1240: Reduce wasted space in angular charts
from the original issue: 
> When I choose to display my charts half-circle there is an empty space below that if it didn't exist my chart would appear bigger. I assume that this happens because I have the `labelMajor` and `labelMinor` to empty string. Can I choose that I don't want to be rendered in order to gain this space?

Have the ability to remove empty space and use it for increasing the size of the chart

n/a

**Additional context**

nickofthyme:
## Summary

The whitespace for circular goal charts had been reduced by offsetting the chart center whenever possible. This does not increase the size of the chart but reduces whitespace between the provided title add the chart itself.

Screen Recording 2021-09-28 at 01 06 10 PM

## Details

This is a first step approach to reduce the whitespace only by translating the chart when possible. There are a few limitations of the current goal chart that limit other solutions as well as this one. Namely:

### Limitations

1. The user can set `angleStart` and `angleEnd` to any value thus allowing for nonsensical angle values. Such values are not accounted for in this approach.
2. The chart allows but does not account for a full circle goal chart and assumes the extents of the angle does not collide with the bottom titles. This also assumes the center of the full circle is at the center of the chart, hence why full circles collide with the bottom titles. See gif showing this below.
3. The central titles/labels are assumed to be at the center of the circular chart, thus shifting beyond half the radius is not practical until more complex center placement is allowed.
4. The size of the chart is computed based on hard-coded limiting sizes. This is ideal for now as any translation and resizing would require some iteration to optimize the size given the value of `angleStart` and `angleEnd`.

Screen Recording 2021-09-28 at 01 07 45 PM

### Solution

The solution uses the `angleStart` and `angleEnd` to compute a minimum limiting [sagitta](https://www.mathopenref.com/sagitta.html) with which we compare with the minimum limiting sagitta, assigning the offset as the delta. WHAT!??!?!

We measure the angle from the top (`π/2`) towards the bottom (`3π/2` or `-π/2`). Let's call these `θL` (aka theta-L)  and `θR` for left and Right. Then `θ` is the limiting angle determined by the max of left and right, where theta is greater than `π/2`, see limitation 3. 

Untitled

From here we can now determine the so called **_sagitta_** (`S` in the diagram below) from `θ` and the outer radius of the goal chart.

Untitled 2

Once we have the _sagitta_ for the current angles we need to find the limiting max _sagitta_. This is the max _sagitta_ such that there would be no translation. In the current implementation this an angle of `3π/2` or `θmax`. This was chosen to avoid the title, see limitation 2.

Untitled 3

With these two _sagitta_ values, we can take the difference (aka the whitespace between the chart and the titles) and divide by 2 in order to center the goal chart negating the white space between the chart and titles.

Left - Unshifted | Center - Max/Default | Right - Shifted |

Untitled 4

> Notice the `Smin` on the far left chart is limited to half the radius as mentioned above.

## Issues
closes #1240

### Checklist


monfera (approved):
Looks great 🎉 

This single story needs an update, shorter text or different angles: 
image—unless the purpose of the story is to show that overlap can and does happen, depending on the config

The geometry driven solution is nicer than some ad hoc thing would have been, good call 💪

packages/charts/src/chart_types/goal_chart/layout/viewmodel/geoms.ts
```diff
@@ -390,6 +393,12 @@ export function geoms(
   const maxWidth = abstractGeoms.reduce((p, g) => Math.max(p, get<number>(g.aes, 'lineWidth', 0)), 0);
   const r = 0.5 * referenceSize - maxWidth / 2;
 
+  if (circular) {
+    const sagitta = getMinSagitta(angleStart, angleEnd, r);
+    const maxSagitta = getSagitta((3 / 2) * Math.PI, r);
+    data.yOffset.value = sagitta >= maxSagitta ? 0 : (maxSagitta - sagitta) / 2;
+  }
+
```
monfera, line 401: How is the sometimes considerable stroke width of the arc taken into account?

nickofthyme:
> This single story needs an update, shorter text or different angles:

Yeah I want to make sure the beyond `5π/4` and `-π/4` there is no offset applied since this is the current _max_ value.

> The geometry driven solution is nicer than some ad hoc thing would have been, good call 💪

🙇🏼

packages/charts/src/chart_types/goal_chart/layout/viewmodel/geoms.ts
nickofthyme: Yeah that's true. How is the stroke width able to be set? Also I assume you are referring to the width of the colored bands themselves right?

PR #1418, opened 2021-10-04
fix(debug): add predictable axis labels sorting order

markov00:
## Summary

This commit adds a predictable label sorting order for axis labels in the debug state.
The configured sorting order should always reflect the order starting from the origin of the axis, independently from the rotation or the axis position.
This is the only order that remains constant and predictable.

## Details

These changes are required to fix the functional tests in Kibana, providing a constant order to the axis labels.

### Checklist


packages/charts/src/chart_types/xy_chart/state/selectors/get_debug_state.ts
```diff
@@ -40,43 +44,63 @@ export const getDebugStateSelector = createCustomCachedSelector(
...
+      const sortingOrder = isHorizontalAxis(position)
+        ? rotation === 0 || rotation === 90
+          ? Predicate.NumAsc
+          : Predicate.NumDesc
+        : rotation === 0 || rotation === -90
+        ? Predicate.NumDesc
+        : Predicate.NumAsc;
+      const visibleTicks = [...geom.visibleTicks]
+        .sort(getPredicateFn(sortingOrder, 'position'))
+        .filter(({ label }) => label !== '');
```
monfera, line 88: Let's filter first; in this case there.s no need for sorting something that gets removed, and no need for the `[...geom.visibleTicks]` shallow copy, as the filter creates a fresh array

monfera (approved):
LGTM! One minor optional code comment. Also, the sorting order is conditionally ascending or descending; it may be simpler for both our side and the recipient side to always do the same sorting, eg. `NumAsc`, because the `NumDesc` is arbitrary/artificial anyway. I'm not aware of a reason for complicating "business logic" for such a debug step, and the ascending order is a good standard, as noone cares about the physical screenspace order (imo)

packages/charts/src/chart_types/xy_chart/state/chart_state.debug.test.ts
```diff
@@ -0,0 +1,52 @@
...
+            id: 'left',
+            position: Position.Left,
+            integersOnly: true,
+          }),
+        ],
+        store,
+      );
+      const debugState = getDebugStateSelector(store.getState());
+      expect(debugState.axes?.x[0].labels).toEqual(['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10']);
+      expect(debugState.axes?.y[0].labels).toEqual(['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10']);
```
nickofthyme, line 49: How about `toMatchSnapshot`?

```suggestion
      expect(debugState).toMatchSnapshot();
```

markov00:
> LGTM! One minor optional code comment. Also, the sorting order is conditionally ascending or descending; it may be simpler for both our side and the recipient side to always do the same sorting, eg. `NumAsc`, because the `NumDesc` is arbitrary/artificial anyway. I'm not aware of a reason for complicating the order with "business logic" for such a debug step, and the ascending order is a good standard, as noone cares about the physical screenspace order (imo)

You are right, but I thought it was easier to grasp if we connect that to a "bottom-left" screen space. When we are going to reverse the way we compute the Y position on cartesian we can clean up this code (currently it follows the top-left 0,0 origin, that is actually harder to reason about that a bottom-left one)

packages/charts/src/chart_types/xy_chart/state/chart_state.debug.test.ts
nickofthyme: This would cover other unexpected changes to debug states.

nickofthyme (approved):
LGTM

packages/charts/src/chart_types/xy_chart/state/chart_state.debug.test.ts
markov00: done ba38aa9

packages/charts/src/chart_types/xy_chart/state/selectors/get_debug_state.ts
markov00: you are right, my bad, done in c407bba

PR #1419, opened 2021-10-04
feat(xy): apply the data value formatter to data values over bars

rshen91:
## Details
PR #1390 includes some old commits changing the nature of getAxesBySpecId() but this function is used in multiple places where rotations were already accounted for. This new approach in this PR does not impact the tooltip header or body formatting, rectangular annotations, legend formatting etc. 

### Checklist


rshen91:
@markov00 just wanted to confirm with the with value label story that the rotation in commit 06bef45 seems fine. The data for the non-stacked series isn't whole number data, so I think the formatting is working ok. 
``` 
[
   {x: 0, y: 3.934269, g: 'b'},
   {x: 1, y: 7.662253926666667, g: 'b'},
   {x: 2, y: 7.043862173333342, g: 'b'},
   {x: 3, y: 6.394734540000064, g: 'b'},
   {x: 4, y: 6.059608170666939, g: 'b'},
   {x: 5, y: 0.25, g: 0},
   {x: 6, y: 8, g: 0}
]
```

Let me know what you think, thank you.

packages/charts/src/chart_types/xy_chart/state/utils/utils.ts
```diff
@@ -369,8 +369,10 @@ function renderGeometries(
         mergeOptionalPartialValues: true,
       });
 
-      const { yAxis } = getAxesSpecForSpecId(axesSpecs, spec.groupId);
-      const valueFormatter = yAxis?.tickFormat ?? fallBackTickFormatter;
+      const { xAxis, yAxis } = getAxesSpecForSpecId(axesSpecs, spec.groupId);
+      const valueFormatter = [0, 180].includes(chartRotation)
+        ? yAxis?.tickFormat ?? fallBackTickFormatter
+        : xAxis?.tickFormat ?? fallBackTickFormatter;
```
markov00, line 375: This works, but as specified on the linked issue, the logic should be integrated within the `getAxesSpecForSpecId`.

```
function getAxesSpecForSpecId(axesSpecs: AxisSpec[], groupId: GroupId) {
  return axesSpecs.reduce<{ xAxis?: AxisSpec; yAxis?: AxisSpec }>((result, spec) => {
    if (spec.groupId === groupId && isHorizontalAxis(spec.position)) result.xAxis = spec;
    if (spec.groupId === groupId && isVerticalAxis(spec.position)) result.yAxis = spec;
    return result;
  }, {});
}
```
The current code assigns to the X and Y the axis specs based on their positions.
X and Y are not dependent on the position of the chart, the X should always represent the independent variable of a chart, where the Y represents the dependent variable, irrespective of the chart axis rotation.
When internally we use X and Y in variable names we use the independent, dependent concept.
If `getAxesSpecForSpecId` is used in other part of the chart code in addition to values labels, it's probable that this formatting error appear in other part of the code

markov00 (changes requested):

packages/charts/src/chart_types/xy_chart/state/utils/utils.ts
rshen91: For sure! This function is used in a variety of places:

- the tooltip header
- the tooltip body
- data values over the bars in a bar chart
- in getAnnotationAxis()
- in rectangular annotation dimensions
- in the legend

I'll work on refactoring it so these places don't have to do their current workarounds 🙆🏻

PR #1420, opened 2021-10-07
feat(xy): adaptive tick raster

monfera:
## Summary

PR for lowering the targeted tick count until a compliant tick raster is the result. Compliance criteria are informed by the need to convey truth on the chart, and also by dataviz best practices:
1. All tick labels must fit (none should overlap with any another); 
2. The tick labels should have a preference for a (generally) decimal base, eg. 0-5-10 over 0-4-8-12
3. Nicing, if applies, should be observed
4. Landmark ticks, such as zero, or the extreme tick labels in case of nicing, must be preserved
2. There must be at least two distinct ticks, otherwise the screen projection can't be correlated with the shown data, rendering the chart meaningless
3. All ticks must have a unique label, otherwise there's ambiguity about which represents the true value (if any)
4. The raster should yield even spacing of the tick/gridlines even if screenspace tick label sizes vary
5. Coarser semantic units (eg. hours) should be split into an integral number of smaller units (eg. 1,2,3,4,5,6,10,12,15,20,30mins) - note that due to other constraints, some will be not generated, eg. a cadence of 3 minutes (not just our constraints but also what matured into underlying scale generators like D3, which we can relax if needed in the future)

## Details

The current method relies on tick label culling (and optionally, tick/gridline culling) for avoiding overlap and duplicated tick labels, but the (inherently post hoc) culling of the tick labels doesn't guarantee compliance with the above items. For examples:
- maybe, due to rounding in the formatter, some labels are culled due to duplication, but the surviving one isn't necessarily the one at the actual position of the shown value
- there could be multiple generated ticks, but culling can remove some such that less than two remain
- unusual cadences may be generated, eg. an unusual 3-minute cadence; decimal boundaries can't be prioritized by culling
- variable tick label length (in pixels) can yield an uneven raster, where the distance between any two adjacent ticks can significantly vary
- the zero value, and/or the last tick label, placed expressly by nicing, may be culled

In short, we deactivate the culling of a single raster; instead, we iteratively generate the raster that's closest to the desired tick count, while complying with the rules

`adaptiveTickCount` is set to `false`, ie. old mode, to make all but one VRT images identical with the current snapshots

### Checklist


monfera:
test this

PR #1435, opened 2021-10-19
fix(xy): use round ticks on linear bar charts

markov00:
## Summary

This PR changes the way the X-axis is rendered in linear bar charts.
Instead of computing the ticks as
```
new Array(Math.floor((nicePaddedDomain[1] - nicePaddedDomain[0]) / minInterval) + 1)
            .fill(0)
            .map((_, i) => nicePaddedDomain[0] + i * minInterval);
```
I've converted this code to use the standard `d3.scale.ticks` instead.

This causes a lot of failures in the VRTs because a lot of bar chart uses the Linear scale on the x axis (that should be discouraged if not used in association with the histogram mode or with a controlled number of ticks)

The changes at the screenshot levels are also due to the reverting nature to ordinal scale in the X axis of bar chart and the subsequent removal of the legend extra parameter.

As discussed also with @monfera a bar chart can use the linear scale on the x scale, but its use slightly distorts the nature of the reading: in this case, the bar is only an adornment of a Y value and its width doesn't represent any more an interval in the x scale unless you use the histogram mode that correctly aligns the ticks at the beginning and end of the bar interval.

### Checklist


markov00:
Closing in favour of: #1645 #1646

Issue #1442, opened 2021-10-27
Kibana multi-layer time axis


markov00:
A collection of tasks and issues related to the introduction of the new multi-layer time axis in Kibana.

```[tasklist]
### Tasks
- [ ] https://github.com/elastic/elastic-charts/issues/2158
- [ ] https://github.com/elastic/elastic-charts/issues/2157
- [ ] Improve gridline styles (nicely mark major, minor and ticks)
- [ ] testing: addition of VRT cases for the various time resolutions (quite a few zoom levels from ms to decades, not all covered yet)
- [ ] testing: addition of tests on time zones get handled correctly
- [ ] Lens: try handling the clustered bar time series within the library, or switch to ordinal bar chart within Lens in that specific case
- [ ] https://github.com/elastic/kibana/issues/166240
```

- [x] Verify and fix darkmode (https://github.com/elastic/elastic-charts/pull/1445)
- [x] Don't show tick/grid line for leftmost tick if it's outside the domain (https://github.com/elastic/elastic-charts/pull/1446)
- [x] Show tick value in label, rather than start of the domain for a leftmost tick outside the domain (https://github.com/elastic/elastic-charts/pull/1446)
- [x] Align font sizes in axes (https://github.com/elastic/kibana/pull/116749)
- [x] DRY multi-layer time axis style in Kibana (move the style in `charts` plugin and import when needed) (https://github.com/elastic/kibana/pull/116749)
- [x] TSVB: Disable multi-time axis when the chart is configured with clustered bars (https://github.com/elastic/kibana/pull/116749)
- [x] Lens: ~Hide~ Disable axis tick label orientation configuration when using the multilayer axis with tooltip: "Axis orientation is not compatible with time axis"
- [x] Visualize: align, filter and truncate can be ~removed~ disabled from the UI when using the multiLayerTimeAxis with tooltip "{Featurename} is not compatible with time axis"
- [x] UX issue: sparser ticks/gridlines 2: chart width dependent minimum allowed tick pixel distance
- [x] UX issue: theme integration and ensuring layer and theme dependent tick lengths (or lack thereof), such as grid saliency
- [x] AM/PM --> 24h 

Original source: Details section of https://github.com/elastic/elastic-charts/pull/1430

markov00:
Close as done

PR #1448, opened 2021-10-29
fix(xy): adding missing last tick and other tick improvements

monfera:
## Summary

Changes:
- re-add the last tick line on the right, even if there's no room for the belonging label
- avoid a tick line if it doesn't add info (copying this feature from the prototype) to avoid too many "whiskers" sprouting from the axis line
- limit the number of gridlines to 12 (while the number of ticks is still limited to 50, so there are often more ticks than grids)
- experiment possibility: set HIDE_MINOR_TIME_GRID to true (not used now)

image

## Issues

Partial fulfilment of https://github.com/elastic/elastic-charts/issues/1442

### Checklist

- [ ] ~The `:theme` label has been added and the `@elastic/eui-design` team has been pinged when there are `Theme` API changes~
- [ ] ~New public API exports have been added to `packages/charts/src/index.ts`~
- [ ] ~Unit tests have been added or updated to match the most common scenarios~


markov00 (approved):
Except for a small artifact when hiding ticks (as shown in the above picture), the improvement looks great, tick ratification reduces the gridline noise and without sacrificing too much the readability of the chart.

Screenshot 2021-11-01 at 09 42 09

PR #1452, opened 2021-11-01
fix(theme): merge partial with empty initial partial

### LINKED ISSUE #1284: Theme overrides not working perfectly

nickofthyme:
## Summary

Fix theming override logic to allow initial partial value as empty object (`{}`). 

Before something like the chart below would ignore the `dash` definition because the first partial is empty.

```tsx
  return (
    <Chart>
      <Settings
        baseTheme={useBaseTheme()}
        theme={[
          {},
          {
            crosshair: {
              line: {
                dash: [4, 4],
              },
            },
          },
          EUI_CHARTS_THEME_LIGHT.theme,
        ]}
      />
    </Chart>
  );
};
```

Now the dash is correctly shown.

Screen Recording 2021-11-01 at 06 29 11 PM

## Details

The `mergePartial` logic had two issues...

1. The merge optional values logic only ran if the first partial was not `undefined`. Now it checks for any keys on all partial values at the given level and applies mutations to the `baseClone` as needed, if either the `partial` has applicable keys or any `additionalPartials` have applicable keys.
2. The final mapping over the new `baseClone` was only looking at the keys of the `base` and not the `baseClone`.

Unrelated, deprecates `mergeWithDefaultTheme` and removes all internal usage.

## Issues

Actually fixes #1284 

Related to https://github.com/elastic/kibana/issues/116754 

### Checklist

### DIFF
--- a/packages/charts/api/charts.api.md
+++ b/packages/charts/api/charts.api.md
@@ -1350,7 +1350,7 @@ export function mergeWithDefaultAnnotationLine(config?: Partial<LineAnnotationSt
 // @public (undocumented)
 export function mergeWithDefaultAnnotationRect(config?: Partial<RectAnnotationStyle>): RectAnnotationStyle;
 
-// @public
+// @public @deprecated
 export function mergeWithDefaultTheme(theme: PartialTheme, defaultTheme?: Theme, auxiliaryThemes?: PartialTheme[]): Theme;
 
 // @public (undocumented)
--- a/packages/charts/src/state/selectors/get_chart_theme.ts
+++ b/packages/charts/src/state/selectors/get_chart_theme.ts
@@ -6,8 +6,8 @@
  * Side Public License, v 1.
  */
 
+import { mergePartial } from '../../utils/common';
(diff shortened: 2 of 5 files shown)


markov00 (commented):
I tested the fix locally and works correctly when merging an empty object as the first partial object.
I think this actually doesn't completely fix the issue in Kibana, and it needs to go hand in hand with: https://github.com/elastic/elastic-charts/pull/1453 and https://github.com/elastic/elastic-charts/pull/1447

We need to backport those PRs into `38.0.x` for Kibana 7.16

markov00 (approved):

Issue #1283, opened 2021-08-04
Axis line covers cursor line


nickofthyme:
**Describe the bug**
When using an `AreaSeries` or `LineSeries` the axis line appears to cutoff the cursor line on hover.

**To Reproduce**
1. Go to https://codesandbox.io/s/upbeat-shirley-rjqtm?file=/src/App.tsx

**Expected behaviour**
Cursor line appears above the axis line

**Screenshots**

https://user-images.githubusercontent.com/19007109/128219348-b3ddd428-2b6f-4c12-bd1c-65c4f9726073.mp4

**Additional context**

markov00:
I don't understand why this happen only on some specific cases like yours, and not in examples like https://elastic.github.io/elastic-charts/?path=/story/line-chart--discontinuous-data-points&globals=theme:light

flash1293:
It seems like this is happening for gridlines as well: https://github.com/elastic/kibana/issues/113719#issuecomment-948229126

PR #1453, opened 2021-11-02
fix(interactions): line cursor above the chart, band cursor below

markov00:
## Summary

The effort to fix the visibility of mouse cursors above or below the chart, started in #1447, is now completed thanks to this PR.
I've splitter the crosshair component into two components: one for the mouse cursor line (that needs to be shown above every element on the chart), and the cursor band, which needs to be shown below the chart geometries.
The crossline (the horizontal line) is also extracted and rendered separately above the chart.

related to https://github.com/elastic/kibana/issues/116754

## Details

The current implementation uses the calculated `zIndex` to position the Crosshair component above the chart if we are rendering a line, or below if we are rendering a band.
Due to some changes in the Kibana instance and a possible bug in the calculation of the chart `zIndex` the Crosshair component is always rendered in Kibana with a `zIndex` of 0, causing it to be rendered always below the chart.
A first fix went in thanks to #1447, which removes the background color applied to the `canvas` element making the cursor visible, but the problem is that the visibility is obfuscated by the gridlines.
This PR fixes it removing the need of using a `zIndex` to position the cursor above or below the chart

## Issues

fix https://github.com/elastic/elastic-charts/issues/1283

### Checklist


rshen91 (approved):
Tested locally (using storybook theme light EUI/dark EUI looks good!) LGTM! Such a smart solution breaking the line cursor and line apart! 👍🏻

nickofthyme (approved):

PR #1462, opened 2021-11-04
fix(heatmap): snap time bucket to calendar/fixed intervals

### LINKED ISSUE #1165: Heatmap time buckets should use value from timeScale

markov00:
## Summary

The PR introduced the ability to specify both calendars and fixed intervals as described in Elasticserach [date_histogram aggregation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html)

This fix was required to cover the Lens usage when calendar intervals are used in conjunction with the date_histogram aggs.

 ### BREAKING CHANGE
The `xScaleType` is replaced by the prop `xScale`, which better describes and collects the information related to the passed interval if any.

These are the exposed types as part of the API changes

```ts

interface HeatmapSpec extends Spec {
  ...
  xScale: RasterTimeScale | OrdinalScale | LinearScale;
  ...
}

export interface TimeScale {
  type: typeof ScaleType.Time;
}

export interface RasterTimeScale extends TimeScale {
  interval: ESCalendarInterval | ESFixedInterval;
}

export interface LinearScale {
  type: typeof ScaleType.Linear;
}

export interface OrdinalScale {
  type: typeof ScaleType.Ordinal;
}
interface ESFixedInterval {
  type: 'fixed';
  unit: ESFixedIntervalUnit;
  value: number;
}

interface ESCalendarInterval {
  type: 'calendar';
  unit: ESCalendarIntervalUnit;
  value: number;
}

type ESFixedIntervalUnit = 'ms' | 's' | 'm' | 'h' | 'd';

type ESCalendarIntervalUnit =
  | 'minute'
  | 'm'
  | 'hour'
  | 'h'
  | 'day'
  | 'd'
  | 'week'
  | 'w'
  | 'month'
  | 'M'
  | 'quarter'
  | 'q'
  | 'year'
  | 'y';

```

## Details

The histogram chart needs to discretize the domain if a time scale is passed in the X axis. To correctly discretize based on the domain min, max values, we need to apply a specific rounding formula depending on timezone and type of interval (calendar/fixed). The discretization is required to cover the case of sparse datasets.

## Issues

fix #1165

### Checklist


nickofthyme (approved):
changes LGTM

A side note for the future. I wonder if some of the ES types are available in [`@elastic/elasticsearch`](https://www.npmjs.com/package/@elastic/elasticsearch) that we can leverage.

packages/charts/api/charts.api.md
```diff
@@ -709,6 +709,32 @@ export const entryKey: ([key]: ArrayEntry) => string;
...
+    // (undocumented)
+    type: 'calendar';
+    // (undocumented)
+    unit: ESCalendarIntervalUnit;
+    // (undocumented)
+    value: number;
+}
+
+// @public (undocumented)
+export type ESCalendarIntervalUnit = 'minute' | 'm' | 'hour' | 'h' | 'day' | 'd' | 'week' | 'w' | 'month' | 'M' | 'quarter' | 'q' | 'year' | 'y';
```
monfera, line 723: Not necessarily in this PR, we could eventually depend on types in `elasticsearch.js` as an authoritative source

packages/charts/api/charts.api.md
```diff
@@ -1036,6 +1060,12 @@ export interface HeatmapConfig {
 // @public (undocumented)
 export type HeatmapElementEvent = [Cell, SeriesIdentifier];
 
+// @alpha (undocumented)
+export interface HeatmapNonTimeScale {
```
monfera, line 1064: Minor: A positively stated scale name would work better, if possible. The word "scalar" implies unitlessness but it doesn't look nice next to "Scale", so, maybe, UnitlessScale? 

The word "Heatmap" in the scale name represents tight coupling, because it casts the Scale as a property of (subsidiary to) the chart type Heatmap, while they're at least equal, orthogonal things or even, Heatmap is subsidiary to the Scale. Of course I see how you want to rule out logarithmic etc. scales.

Maybe `UnitlessHeatmapScale` is an OK alternative

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
```diff
@@ -148,14 +120,28 @@ export function shapeViewModel(
...
+          domain: xNumericExtent,
+          range: [0, chartDimensions.width],
+          nice: false,
+        },
+        {
+          desiredTickCount: estimatedNonOverlappingTickCount(chartDimensions.width, config.xAxisLabel),
+          timeZone: spec.xScale.timeZone,
+        },
+      )
+    : null;
```
monfera, line 144: Not for this PR, as it's just going with the preexisting pattern: we should figure out something more comprehensive than generating nulls and then later checking them. All of us incl. @nickofthyme had some type related ideas so maybe we could mine them in the future

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
```diff
@@ -371,19 +358,20 @@ export function shapeViewModel(
...
-    const y1 = chartDimensions.top;
-    const y2 = cellHeight * pageSize;
-    xLines.push({ x1: x, y1, x2: x, y2 });
-  }
+  const xLines = Array.from({ length: xValues.length + 1 }, (d, i) => ({
+    x1: chartDimensions.left + i * cellWidth,
+    x2: chartDimensions.left + i * cellWidth,
+    y1: chartDimensions.top,
+    y2: cellHeight * pageSize,
+  }));
```
monfera, line 366: 👍

monfera (approved):
LGTM! We discussed two improvements:
- long term: possible multilayer axis (after it had been extracted from `xy`)
- shorter term follow-up PR: avoiding tick overlap
image

packages/charts/src/chart_types/heatmap/state/selectors/get_heatmap_table.ts
```diff
@@ -59,26 +59,32 @@ export const getHeatmapTableSelector = createCustomCachedSelector(
...
+    // sort Y values by their predicates
     resultData.yValues.sort(getPredicateFn(ySortPredicate));
 
     return resultData;
   },
 );
+
+function isFiniteNumber(value: number | undefined): value is number {
+  return Number.isFinite(value);
+}
```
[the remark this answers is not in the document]
markov00: I know, I was following your comments on the TS repo and I was thinking that, for now, can solve our cases

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
markov00: Those are highly connected to the generation of the ticks, I can extract that into a function with a simple branch:


```ts
function getXTicks(
  spec: HeatmapSpec,
  config: Config,
  chartDimensions: Dimensions,
  xScale: ScaleBand<string | number>,
  { xValues, xNumericExtent }: HeatmapTable,
  gridHeight: number,
): Array<TextBox> {
  const getTextValue = (
    formatter: Config['xAxisLabel']['formatter'],
    scaleCallback: (x: string | number) => number | undefined | null,
  ) => (value: string | number): TextBox => {
    return {
      text: formatter(value),
      value,
      ...config.xAxisLabel,
      x: chartDimensions.left + (scaleCallback(value) ?? 0),
      y: gridHeight + config.xAxisLabel.fontSize / 2 + config.xAxisLabel.padding,
    };
  };
  if (isTimeXScale(spec.xScale)) {
    const timeScale = new ScaleContinuous(
      {
        type: ScaleType.Time,
        domain: xNumericExtent,
        range: [0, chartDimensions.width],
        nice: false,
      },
      {
        desiredTickCount: estimatedNonOverlappingTickCount(chartDimensions.width, config.xAxisLabel),
        timeZone: spec.xScale.timeZone,
      },
    );
    return timeScale.ticks().map<TextBox>(getTextValue(config.xAxisLabel.formatter, (x) => timeScale.scale(x)));
  }

  return xValues.map<TextBox>((textBox: string | number) => {
    return {
      ...getTextValue(config.xAxisLabel.formatter, xScale)(textBox),
      x: chartDimensions.left + (xScale(textBox) || 0) + xScale.bandwidth() / 2,
    };
  });
}


```

packages/charts/api/charts.api.md
monfera: Part 1: I'm fine with that too, though in terms of syntax, it combines two types of unioning, where it could be just one, eg. `{ type: ordinal | linear | time }` or `xScale: Heatmap | Ordinal | Linear`

Part 2: Neither name is something I really prefer, as both of them tightly bind a scale type to a chart type maybe needlessly, though the first option rolls off the tongue more easily to me. But the best would be, if possible, to not have a chart type name in the name of a scale type. So, something like `TimeScale`, or however we call it in `xy` already

packages/charts/src/utils/chrono/elasticsearch.ts
```diff
@@ -115,30 +115,49 @@ export function timeRange(
   from: number,
   to: number,
   interval: ESCalendarInterval | ESFixedInterval,
-  timeZone?: string,
+  timeZone: string,
 ): number[] {
+  return interval.type === 'fixed'
+    ? fixedTimeRange(from, to, interval, timeZone)
+    : calendarTimeRange(from, to, interval, timeZone);
+}
```
nickofthyme, line 123: 👍🏼

nickofthyme (approved):
LGTM, I don't see any issues after playing with the new story for any of the datasets. Old stories work as before.

packages/charts/src/chart_types/heatmap/state/selectors/get_heatmap_table.ts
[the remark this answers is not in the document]
monfera: Maybe it could be a `value: unknown`, and it can be put into common or utils, as there are many places where I temporarily used a `typeof d === 'number' && Number.isFinite(d)` check where `isFiniteNumber` would be clearer. All places where we do `=== 'number'` is an easy to grep candidate for a future PR. Even the name of this function is perfect 👌

packages/charts/api/charts.api.md
monfera: Just to clarify, as we have an `elasticsearch.ts`, I meant, at some point directly devDepending on https://github.com/elastic/elasticsearch-js for these types. This way, if elasticsearch adds a new unit, and we bump the dependency, we'll get automatic static errors saying that we don't cover the new unit option(s)

PR #1463, opened 2021-11-04
feat(styles): eui Amsterdam theme

### LINKED ISSUE #1461: Amsterdam theme support
Kibana is now using Amsterdam theme by default and overriding styles globally. This is inconsistent with out VRT and has been found to cause style issues in the legend.

use only Amsterdam theme as base and fixup and styles that have a negative impact.

**Kibana Cross Issues**
https://github.com/elastic/kibana/issues/117450

nickofthyme:
## Summary

Move to eui Amsterdam theme. Major changes include:

- slight color differences (copied from eui, since in javascript)
- `:focus` & `:focus-within` states are slightly different. (see gifs below)
- slight dimensional changes

### After

Screen Recording 2021-11-04 at 05 11 32 PM

### Before

Screen Recording 2021-11-04 at 05 09 11 PM

## Details

Chart consumes styles from eui to leverage some variables, mixins, etc. and now that kibana is fully on the new Amsterdam theme I think it's best we fully migrate. This was also causing styles issues related to a different implementation for focus states between the old and the new, which consequently applied both styles. This situation defeats any VRT we have in place.

⚠️ This creates a diff for virtually every VRT screenshot 😭 😔 

cc: @elastic/eui-design

## Issues

fix #1461

### Checklist


cchaos:
👋  Hey @nickofthyme! This is great! I just figured I'd give you a heads up that we will be releasing (another) breaking change soon that will actually make `Amsterdam` the default theme in EUI. This means the import structures will be a bit different affecting this PR.

nickofthyme:
@cchaos Thanks for the heads up. I think the main issue with this change is our VRT and having to check all 900+ diffs. If your breaking change would just mean we have to change some imports but the styles are nearly the same than it shouldn't be an issue. When is this change coming? days, weeks, months?

cchaos:
It's in this PR https://github.com/elastic/eui/pull/5121 which is in it's final reviews. So **maybe** next Tuesday (11/9) or the Tuesday after.

nickofthyme:
Strange diff with `integration/tests/__image_snapshots__/mixed-stories-test-ts-mixed-series-stories-fitting-functions-stacked-charts-as-percentage-area-charts-ordinal-dataset-no-end-value-should-display-correct-fit-for-type-none-1-snap.png`. The latest one looks correct because there is no value for the `test` series at `m`. Thus this should be an orphan point.

image

rshen91 (approved):
LGTM - checked all the vrts thanks for opening/referencing the orphan points. All other screenshots only reflected theme changes

docs/0-Intro/0-Intro.mdx
```diff
@@ -50,11 +50,11 @@ import '@elastic/charts/dist/theme_only_dark.css';
...
-@import '~@elastic/eui/src/themes/eui/eui_colors_light';
-@import '~@elastic/eui/src/global_styling/functions/index';
-@import '~@elastic/eui/src/global_styling/variables/index';
-@import '~@elastic/eui/src/global_styling/mixins/index';
-@import '~@elastic/eui/src/global_styling/reset/index';
+@import '~@elastic/eui/src/themes/eui-amsterdam/eui_amsterdam_colors_light';
+@import '~@elastic/eui/src/themes/eui-amsterdam/global_styling/functions/index';
+@import '~@elastic/eui/src/themes/eui-amsterdam/global_styling/variables/index';
+@import '~@elastic/eui/src/themes/eui-amsterdam/global_styling/mixins/index';
+@import '~@elastic/eui/src/themes/eui-amsterdam/global_styling/reset/index';
```
cchaos, line 57: You can actually import these with a single file

```suggestion
@import '~@elastic/eui/src/themes/eui-amsterdam/eui_amsterdam_globals';
```

packages/charts/src/_eui_imports.scss
```diff
@@ -1,3 +1,3 @@
-@import '~@elastic/eui/src/global_styling/functions/index';
-@import '~@elastic/eui/src/global_styling/variables/index';
-@import '~@elastic/eui/src/global_styling/mixins/index';
+@import '~@elastic/eui/src/themes/eui-amsterdam/global_styling/functions/index';
+@import '~@elastic/eui/src/themes/eui-amsterdam/global_styling/variables/index';
+@import '~@elastic/eui/src/themes/eui-amsterdam/global_styling/mixins/index';
```
cchaos, line 3:
```suggestion
@import '~@elastic/eui/src/themes/eui-amsterdam/eui_amsterdam_globals';
```

storybook/style.scss
```diff
@@ -1,4 +1,4 @@
-@import '~@elastic/eui/dist/eui_theme_light';
+@import '~@elastic/eui/dist/eui_theme_amsterdam_light';
```
cchaos, line 1: Are you sure you want to do this? It will re-compile **all** EUI components which takes a long time. Since you're not adjusting the tokens, I'd suggest importing the compiled CSS instead.

cchaos (commented):
I might be a bit confused about how your docs work or how you expect consumers to consumer charts with/without EUI, but here are some suggestions:

packages/charts/src/_eui_imports.scss
nickofthyme: eaa9032

docs/0-Intro/0-Intro.mdx
nickofthyme: eaa9032

storybook/style.scss
nickofthyme: I'm not sure how to fix this. This is a css file that we import but yet doing so will trigger eui to re-compile all components? I do notice it takes a long time for any change in `scss` files, for the compiler to hot reload the changes.

@markov00 any ideas?

nickofthyme:
@markov00 could you take a look at the style `@import`s? I want to make sure there are not unneeded imports for consuming and usage within charts for the latest eui changes.

storybook/style.scss
nickofthyme: Follow up on this -- We are planning to improve this in a future PR that leverages enhancements from https://github.com/elastic/eui/pull/5674.

Eventually we need to go away from using eui styles directly and create a mechanism for injecting css in js.

nickofthyme:
@markov00 do you think this should be a breaking change?

markov00:
> @markov00 do you think this should be a breaking change?

Changing the colors is actually a breaking change even if it doesn't look like a problematic change (because the colors hue isn't changed, just the saturation a bit)

markov00 (approved):
Changes looks goot to me tested locally and all charts looks great (only the streamgraph is a bit too transparent, but we can fix this in a different PR)

PR #1467, opened 2021-11-05
feat(legend): custom legend width

### LINKED ISSUE #963: Customizable legend width
This is listed as point 9 in the [legend redesign](https://github.com/elastic/elastic-charts/issues/580) issue, but I don't think it has a separate issue to track any updates on it. There is an [equivalent Kibana issue](https://github.com/elastic/kibana/issues/57980) to be able to set the width of legends.

The request is to be able to set the legend width as a pixel value or as a % of available width. Another option would be to show all available text (but I am aware this is more complicated).

nickofthyme:
## Summary

The `legendSize` prop is now available in the `Settings` component to set the exact size of the legend.

Limitations:

- Height and width limited to max of 70% of respective chart dimensions. Prevent squishing the chart area.
- Width limited to min of 30% computed legend width. Prevent squishing the static legend elements.

Screen Recording 2021-11-05 at 11 37 27 AM

> This will override the max dimensioning options on the theme including  `theme.legend.verticalWidth` & `theme.legend.horizontalHeight`.

## Issues

fix #963 

### Checklist


markov00 (approved):
Changes look good to me.
The chosen min/max ratio is fine and allows a nice customizability

nickofthyme:
> The chosen min/max ratio is fine and allows a nice customizability

The only issue i see with adding these limits is if Lens, or anyone, has a value that is outside of these limits, there would be no way for them to know without passing back some callback or them knowing the limits and chart width/height. Say they set the legend width to 1000px but the chart width is only 700px, it would be restricted to 70% of 700px.

I think it's fine for now but to me not ideal.

markov00:
> 

I think it's fine for now, we can expose those ratio in the future through the `theme` if we want

nickofthyme:

Issue #1108, opened 2021-04-06
[Heatmap] the tooltip callout box is centered at the mouse pointer


monfera:
**Describe the bug**
It looks like it's partly a bug (vertical centering) and partly, a less than optimal user experience, or issue, depending on circumstances (horizontal snap with some offset, which causes the occlusion of the very hovered thing):

![image](https://user-images.githubusercontent.com/1548516/113742323-ce315380-9702-11eb-8cb2-d0fcec69932f.png)
The cursor is not visible on the screenshot; it's over Deborah / 1965.

**To Reproduce**
1. Go to https://elastic.github.io/elastic-charts/?path=/story/heatmap-alpha--categorical
2. Hover over cells, at various window widths

**Expected behaviour**
I think that ensuring that the tooltip box is always to the right and below (or maybe above) of the mouse pointer, or even the cell (in this case, with a callout arrow pointing to the cell) would be better

**Checklist**

Delete any items that are not applicable to this issue.

- [ ] every related Kibana issue is listed under `Kibana Cross Issues` list
- [ ] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

markov00:
hei @darnautov are aware of this issue?

darnautov:
thanks for creating an issue, it is not the case with another [story](https://elastic.github.io/elastic-charts/?path=/story/heatmap-alpha--basic) tho. I'll work on a fix 👍

markov00:
This will be fixed in #1163 @darnautov you can see the fix here: https://github.com/elastic/elastic-charts/pull/1163/files#diff-d3b5933a97ba501883dea0379024b3a27b0afbe64480e581cf915c1fd0ef5660
instead of using the center of the cell as anchor point for the tooltip:
```
x1: firstShape.x + chartDimensions.left + firstShape.width / 2,
```
I've changed that to anchor the tooltip around the cell shape:
```
x: firstShape.x + chartDimensions.left,
width: firstShape.width,
y: firstShape.y - chartDimensions.top,
height: firstShape.height,
```

PR #1471, opened 2021-11-09
build: disable commenting on successfully release PRs

markov00:
## Summary

The current `semantic-release` configuration creates a comment on each PR that is released. There is a [bug](https://github.com/semantic-release/github/issues/359) that cause re-commenting on the same PR over and over again if the commit message has the `fix #123`  caption like in https://github.com/elastic/elastic-charts/issues/1108#issuecomment-854838687

This creates a lot of noises in the github notifications.

### DIFF
--- a/.releaserc.js
+++ b/.releaserc.js
@@ -21,7 +21,13 @@ module.exports = {
         execCwd: '.',
       },
     ],
-    '@semantic-release/github',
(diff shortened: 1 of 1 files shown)


nickofthyme (approved):
Wow that's bad!

Issue #300, opened 2019-08-07
a11y


markov00:
**Is your feature request related to a problem? Please describe.**
Chart should be make accessible from users with disabilities.

**Describe the solution you'd like**
The chart should provide:
- alternative text
- keyboard navigation

**Describe alternatives you've considered**
Here are some reference links to accomplish that:
Chart accessibility
- https://www.highcharts.com/blog/post/best-chart-accessibility-practices/
- https://www.elsevier.com/connect/making-charts-accessible-for-people-with-visual-impairments

Canvas accessibility
- https://www.w3.org/Talks/2014/0510-canvas-a11y/#1
- https://developer.paciellogroup.com/blog/2015/02/html5-canvas-sub-dom/
- https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Hit_regions_and_accessibility

**Additional context**
n/a

**Kibana Cross Issues**
- [ ] https://github.com/elastic/kibana/issues/36386
- [ ] https://github.com/elastic/kibana/issues/10726

**Checklist**

- [x] this request is checked against already exist requests
- [ ] every related Kibana issue is listed under `Kibana Cross Issues` list
- [x] `kibana cross issue` tag is associated to the issue if any kibana cross issue is present

markov00:
from @myasonik 

Demo of a simple accessible canvas: https://codepen.io/myasonik/pen/mdbwxNY

Tabbable canvas example I found (with no other info about it): https://www.html5accessibility.com/tests/canvas.html

myasonik:
Just to elaborate a bit further:

Making a truly accessible canvas is... difficult. And currently a lot of the tools to do so seem to be experimental (namely, [hit regions](https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Hit_regions_and_accessibility) ). So, my "accessible" demo above provides a SR alternative to the canvas that can by dynamically generated.

It's not a complete solution for all cases but it's a good place to start from.

markov00:
Tableau narrative: 
https://www.tableau.com/about/blog/2016/8/bringing-power-natural-language-tableau-58325
https://narrativescience.com/products/quill/extensions/tableau/

markov00:
Added dataviz A11y group resources

markov00:
Added few Sarah Fossheim's articles and one from Amy Cesal about writing alt text for datavis

alexstine:
Hello everyone,

[...]

PR #1472, opened 2021-11-09
feat(a11y): xy cartesian data table

rshen91:
## Summary
Cartesian charts now have a data table to provide alternative text descriptions of the chart canvas data. A caption was added to the table to provide more context for users.

For charts where there are more than 20 data points, the table will be similar to the legend. Users can compare the series at the same x interval and click to see more data or previous data. 

For small data sets, users will be able to access the full data set. For charts with small datasets but multiple series, the x values of the points are grouped to provide more a greater idea of comparison between the series. 

- [x] Get the data formatted appropriately for the component (have the full data set formatted vs. how legend items are truncated at last values)
- [x] Fix the show next data for the table
- [x] Small multiple charts
- [x] x Value formatting
- [x] Clean up the caption and any extraneous data for the xy table component (axes or not etc.)
- [x] Clean up keys in the component (no Math.random())
- [x] Limit the comparison by x value behavior for one series or data points that are less than 20 (double check that limit with the current one used in partition charts) 2d399b6
- [x] Clarify the small multiple category specification: the small multiple categories used to describe the vertical and horizontal position of the small multiple needs to be described in the table as columns 8f88313
- [x] Confirm navigation with screen reader/Voice Over - met with Zuhair and Trevor from EUI about usability 
- [x] Add titles from axes to the caption for greater semantic context for users f2d9a35
- [x] Confirmed table is read in Kibana when locally testing
- [x] Write tests
- [x] Add `debugA11y` for developer debugging of the screen reader table

## Details

Unlike the partitions specific data table, the cartesian data table will show a value per series for easier comparison.

- Future refactoring: The current code uses data from Last Values logic: can we find a way to reduce code duplication? Reducing duplication can avoid losing a fix that was deployed only on a particular part of the code and not reported here

## Issues
Part of https://github.com/elastic/elastic-charts/issues/300

### Checklist


rshen91:
Jenkins test this please

packages/charts/src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
```diff
@@ -162,25 +163,25 @@ class XYChartComponent extends React.Component<XYChartProps> {
...
+            width,
+            height,
+          }}
+          // eslint-disable-next-line jsx-a11y/no-interactive-element-to-noninteractive-role
+          role="presentation"
+        >
+          <ScreenReaderSummary />
+          {!debug && <ScreenReaderCartesianTable />}
+        </canvas>
+        {debug && <ScreenReaderCartesianTable />}
```
nickofthyme, line 183: Debug is only used internally. Is this table visibility meant to be toggle-able in kibana or just have the table in the dom?

packages/charts/src/components/accessibility/cartesian_data_table.tsx
```diff
@@ -0,0 +1,138 @@
...
+              </tr>
+            );
+          })}
+        </tbody>
+        <tfoot>
+          <tr>
+            <td colSpan={totalColumns}>
+              <button type="submit" onClick={() => handleMoreData()} tabIndex={-1}>
+                Click to show more data
+              </button>
```
nickofthyme, line 106: Is this necessary? Why not just show all the data with a scroll overflow?

nickofthyme (commented):
Still reviewing just leaving early comments.

packages/charts/src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
rshen91: AFAIK it's so there's the visual table for debugging vs having to navigate through the DOM.

packages/charts/src/components/accessibility/cartesian_data_table.tsx
rshen91: I think it's easier with keyboard navigation to have a button to click through voiceover to get the next set of data but I can explore the scroll overflow idea and report back 👍🏻

nickofthyme: But considering the screen reader case where someone is not able to see the chart and the screenreader is used. Are they gonna have to keep pressing the "more" button? If so maybe we limit it to more only over `100` items or something. What do you think?

nickofthyme: Discussed on zoom

rshen91: Thank you for the zoom! I will make a see previous x value button - that makes a lot of sense thank you!

markov00:
hey @rshen91 is this still a WIP or we can review it?
I see two missing checkbox in the description, do we need to wait for them for the review?

rshen91:
> hey @rshen91 is this still a WIP or we can review it? I see two missing checkbox in the description, do we need to wait for them for the review?

Please feel free to review this thank you!

packages/charts/src/components/accessibility/cartesian_data_table.tsx
```diff
@@ -0,0 +1,318 @@
...
+  const totalColumns: number = isSmallMultiple ? (countOfCol += 3) : countOfCol;
+
+  /**
+   * Data table for small data sets (less than 20 data points) with only one series
+   */
+  const smallDataTableSingleSeries = (
+    <div
+      className={`echScreenReaderOnly ${debug ? 'echScreenReaderOnlyDebug' : ''} echScreenReaderTable`}
+      aria-live="polite"
+      tabIndex={-1}
```
1Copenut, line 102: I'm seeing a few `tabindex="-1"` declarations in the table and would like to learn more about how focus is being managed, and how users will navigate the table and change the data point to previous or next point.

packages/charts/src/components/accessibility/cartesian_data_table.tsx
```diff
@@ -0,0 +1,318 @@
...
+                  <>
+                    <th scope="row" colSpan={2}>
+                      {values[count].smPanelTitle}
+                    </th>
+                  </>
+                )}
+                {isSmallMultiple && cartesianData.smallMultipleTitle.length === 1 && (
+                  <th scope="row">{values[count].smPanelTitle}</th>
+                )}
+                <th scope="row">{label}</th>
```
1Copenut, line 260: Is there a way to conditionally make this a `<th scope="row">` element if the previous `isSmallMultiple` prop is false? Looking at the markup, I could see a situation where we have two TH elements that are scoped to the row. That might cause screen readers to behave unexpectedly.

packages/charts/src/components/accessibility/cartesian_data_table.tsx
```diff
@@ -0,0 +1,318 @@
...
+                  label: boolean | React.ReactChild | React.ReactFragment | React.ReactPortal | null | undefined;
+                  data: boolean | React.ReactChild | React.ReactFragment | React.ReactPortal | null | undefined;
+                  smPanelTitle: boolean | React.ReactChild | React.ReactFragment | React.ReactPortal | null | undefined;
+                },
+                i: any,
+              ) => {
+                return (
+                  <tr key={String(`screen-reader-row--${index} ${i}`)} tabIndex={-1}>
+                    {isSmallMultiple && cartesianData.smallMultipleTitle.length === 2 && (
+                      <td colSpan={2}>{val.smPanelTitle}</td>
```
1Copenut, line 194: Same comment about possibly changing this to `<th scope="row">` for the first column from line#133

packages/charts/src/components/accessibility/cartesian_data_table.tsx
```diff
@@ -0,0 +1,318 @@
...
+        </thead>
+        <tbody>
+          {cartesianData.data.map(({ values }, index) => {
+            {
+              return values.map(
+                (val: { xValue: any; formatted: any; raw: any; smPanelTitle: string | undefined }, i: number) => {
+                  return (
+                    <tr key={String(`screen-reader-row--${index} ${i}`)} tabIndex={-1}>
+                      {isSmallMultiple && cartesianData.smallMultipleTitle.length === 2 && (
+                        <td colSpan={2}>{val.smPanelTitle}</td>
```
1Copenut, line 133: Could we make the first cell a `<th scope="row">` for the small multiple condition? I like tables that scope rows because it adds another layer of richness to the information read back.

1Copenut (changes requested):
@rshen91 Thank you for the work and care you put into this PR! I've got a few items that are questions as much as change requests. Please feel free to take what's relevant and discard what's not.

Holler at me anytime for a quick chat or another review.

zuhairmahd:
I have tested with JAWS and NVDA.  Both screen readers read the table descriptions, but it is important to note that no actual tables are present in the DOM -- I am told this was intended behaviour because the tables are images.

packages/charts/src/components/accessibility/cartesian_data_table.tsx
rshen91: Thank you I think I mixed up the appropriate tab index and fixed in 463fb09. Let me know what you think!

rshen91:
The `GroupBy` component provides the vertical axes titles in small multiples cases such as https://elastic.github.io/elastic-charts/?path=/story/small-multiples-alpha--grid-lines&globals=theme:light vs being available on the axis spec. Opening an issue to track small multiple screen reader data table for the future. https://github.com/elastic/elastic-charts/issues/1557

1Copenut (approved):
👍 LGTM! I like the markup changes for the table rows, having that scoped TH for each body row will be a strong addition. Thank you for the quick turnaround and logging additional issues to track questions.

rshen91:
jenkins test this

packages/charts/src/components/accessibility/partition_data_table.tsx
```diff
@@ -90,7 +90,7 @@ const ScreenReaderPartitionTableComponent = ({
             .slice(0, rowLimit)
             .map(({ panelTitle, depth, label, parentName, valueText, percentage }, index) => {
               return (
-                <tr key={`row--${index}`} ref={rowLimit === index ? tableRowRef : undefined} tabIndex={-1}>
+                <tr key={String(`row--${index}`)} ref={rowLimit === index ? tableRowRef : undefined} tabIndex={-1}>
```
nickofthyme, line 93: This should not be needed.

```suggestion
                <tr key={`row--${index}`} ref={rowLimit === index ? tableRowRef : undefined} tabIndex={-1}>
```

playground/playground.tsx
```diff
@@ -8,8 +8,59 @@
...
+import { Chart } from '../packages/charts/src/components/chart';
+import { ScaleType } from '../packages/charts/src/scales/constants';
+import { AreaSeries, Settings } from '../packages/charts/src/specs';
+// import { Position } from '../packages/charts/src/utils/common';
+import { KIBANA_METRICS } from '../packages/charts/src/utils/data_samples/test_dataset_kibana';
+import { useBaseTheme } from '../storybook/use_base_theme';
 
 export function Playground() {
-  return <Example />;
+  const baseTheme = useBaseTheme();
```
nickofthyme, line 19: Could you restore this file? I think we should just `.gitignore` it but that's a task for later.

packages/charts/src/chart_types/xy_chart/state/selectors/get_screen_reader_data.ts
```diff
@@ -0,0 +1,238 @@
...
+    return ` the ${position} axis is titled ${title}`;
+  });
+  // axesTitles length is currently misleading for small multiple series
+  return `${
+    isSm
+      ? `This chart is a small multiple series.`
+      : `The chart has ${axesTitles.length} axes ${hasTitles.length} axes have titles${
+          hasTitles.length === 0 ? '.' : ':'
+        }`
+  }${getTitles}`;
```
nickofthyme, line 237: Just thought, this looks kinda wonky. It might be easier to read if you pull the first value portion into a variable assignment before.

packages/charts/src/components/__snapshots__/chart.test.tsx.snap
```diff
@@ -68,27 +68,98 @@ exports[`Chart should render the legend name test 1`] = `
                 </Connect(CursorBand)>
                 <Connect(XYChart) forwardCanvasRef={{...}}>
                   <XYChart forwardCanvasRef={{...}} isRTL={false} initialized={true} isChartEmpty={false} debug={true} geometries={{...}} geometriesIndex={{...}} theme={{...}} chartContainerDimensions={{...}} highlightedLegendItem={[undefined]} rotation={0} renderingArea={{...}} chartTransform={{...}} axesSpecs={{...}} perPanelAxisGeoms={{...}} perPanelGridLines={{...}} axesStyles={{...}} annotationDimensions={{...}} annotationSpecs={{...}} panelGeoms={{...}} a11ySettings={{...}} onChartRendered={[Function (anonymous)]}>
-                    <figure aria-labelledby={[undefined]} aria-describedby=\\"chart1--defaultSummary\\">
-                      <canvas dir=\\"ltr\\" className=\\"echCanvasRenderer\\" width={151} height={200} style={{...}} role=\\"presentation\\" />
+                    <figure aria-labelledby={[undefined]} aria-describedby=\\"chart1--defaultSummary\\" aria-live=\\"polite\\">
```
nickofthyme, line 71: I don't know if this is the right place but just to create a thread....

I don't see any changes to the story files so I'm not sure where the overlay is coming from on the the changed vrt screenshots. Also peculiar that they are all annotation stories. Do you know what is causing this?

image

In any case I don't think it is a good idea to change all these screenshot, nor that the accessibility feature is on by default for the story. But I'd be great if we created a new `integration/tests/accessibility_stories.test.ts` test file to capture the desired use cases.

packages/charts/src/components/accessibility/accessibility.test.tsx
```diff
@@ -177,4 +179,137 @@ describe('Accessibility', () => {
       );
     });
   });
+
```
nickofthyme, line 182: 👍🏼

packages/charts/src/components/accessibility/cartesian_data_table.tsx
```diff
@@ -0,0 +1,314 @@
...
+        return formatted.hasOwnProperty(val.xValue)
+          ? formatted[val.xValue].push({
+              label,
+              data: val.formatted ?? val.raw,
+              xValue: val.xValue,
+              smPanelTitle: val.smPanelTitle,
+            })
+          : (formatted[val.xValue] = [
+              { label, data: val.formatted ?? val.raw, xValue: val.xValue, smPanelTitle: val.smPanelTitle },
+            ]);
```
nickofthyme, line 49: This is funky, why are you returning the result of the `push`? Defined to return the new `length` property.

image

The second half of the ternary is also hard to follow with the nested variable assignment. Can we clean this up too?

Is this expected? If not, I'd say just replace this with an if/else block.

packages/charts/src/components/accessibility/cartesian_data_table.tsx
```diff
@@ -0,0 +1,314 @@
...
+  }
+  return {
+    a11ySettings: getA11ySettingsSelector(state),
+    cartesianData: getScreenReaderDataSelector(state),
+    debug: getSettingsSpecSelector(state).debug,
+  };
+};
+
+/** @internal */
+export const ScreenReaderCartesianTable = memo(connect(mapStateToProps)(ScreenReaderCartesianTableComponent));
```
nickofthyme, line 314: Is this meant to always be enabled? Or is there a flag that we can use to determine when to run this extra logic?

nickofthyme (changes requested):
Looking good. Few questions....

packages/charts/src/components/accessibility/partition_data_table.tsx
rshen91: Fixed in 51fd846 thank you!

playground/playground.tsx
rshen91: Nice! Yes, fixed in 51fd846

packages/charts/src/chart_types/xy_chart/state/selectors/get_screen_reader_data.ts
rshen91: Smart - refactored in 51fd846

packages/charts/src/components/accessibility/cartesian_data_table.tsx
rshen91: For now it's always enabled. In testing in Kibana it looks like it's easily avoidable by continuing to tab if screen reader users do not want the information. But I think it makes sense to have it enabled by default, thoughts?

packages/charts/src/components/__snapshots__/chart.test.tsx.snap
rshen91: In line 183 in the xy-chart.tsx file, there's the clause for if debug is specified to show the component visually. The debug mode then adds the 'echScreenReaderOnlyDebug' class in the cartesian data table component making it visually seen for debugging. I'll look into changing this to remove this overlay and add the extra stories so it can show in only the accessibility cases vs debug generally. Thanks!

packages/charts/src/components/accessibility/cartesian_data_table.tsx
rshen91: I tried to replace this with an if/else block but we have a ternary rule that is changing it back. 

For charts that have two series, this function is creating a value as an array if the xValue is not yet in the formatted data type. If the xValue has appeared, then it's added to the array. This makes it smoother in the `smallDataTableMultipleSeries` to show information that is a comparison of the different series on the same xValues vs when there is one series. I'm trying to clean it up a little with some commenting to hopefully clear it up a little.

nickofthyme:
Closing to revisit in the future when accessibility is prioritized.

PR #1476, opened 2021-11-10
refactor(xy): multilayer time axis modularization

monfera:
## Summary

This is to fix instances of overly tight coupling or violation of DRY principles, introduced with or noticed when adding the multilayer time axis.

### Checklist

- [ ] ~The `:theme` label has been added and the `@elastic/eui-design` team has been pinged when there are `Theme` API changes~
- [ ] ~All related issues have been linked (i.e. `closes #123`, `fixes #123`)~
- [ ] ~New public API exports have been added to `packages/charts/src/index.ts`~
- [ ] ~The proper documentation and/or storybook story has been added or updated~


packages/charts/src/chart_types/xy_chart/state/selectors/visible_ticks.ts
```diff
@@ -57,23 +58,21 @@ export function generateTicks(
...
-  const labelFormat = axisSpec.tickFormat ?? fallBackTickFormatter;
   return ticks.map((value) => {
     const domainClampedValue =
       typeof value === 'number' && typeof scale.domain[0] === 'number' ? Math.max(scale.domain[0], value) : value;
     return {
       value,
       domainClampedValue,
-      label: labelFormat(value, tickFormatOptions),
-      axisTickLabel: axisLabelFormat(value, tickFormatOptions),
+      label: tickFormatter(value, tickFormatOptions),
```
monfera, line 74: I still don't like that what's in effect a tooltip value formatter, and also a default (axis) tick label formatter is called `tickFormatter` but fixing it now is not in scope

packages/charts/src/chart_types/xy_chart/state/selectors/visible_ticks.ts
```diff
@@ -57,23 +58,21 @@ export function generateTicks(
...
+      axisTickLabel: labelFormatter(value, tickFormatOptions),
```
monfera, line 75: I still don't like that what's in effect a tooltip value formatter, that merely happens to be a default (axis) tick label formatter is called tickFormatter (and goes into label which evokes axis tick labels moreso than tooltip "labels") but fixing it now is not in scope. Here the change was to use the same variable names as in the other branch, also reusing the formatters already computed there

packages/charts/src/chart_types/xy_chart/state/selectors/visible_ticks.ts
```diff
@@ -377,57 +340,17 @@ function getVisibleTickSets(
...
-        const domainValues = domain.domain; // todo consider a property or object type rename
-        const domainFromS = Number((domain && domainValues[0]) || NaN) / 1000; // todo rely on a type guard or check rather than conversion
-        const extendByOneBin = isX && xDomain.isBandScale && enableHistogramMode;
-        const binWidth = xDomain.minInterval;
-        const domainExtension = extendByOneBin ? binWidth : 0;
-        const domainToS = (((domain && Number(domainValues[domainValues.length - 1])) || NaN) + domainExtension) / 1000;
-        const layers = rasterSelector(notTooDense(domainFromS, domainToS, binWidth, Math.abs(range[1] - range[0])));
-        let layerIndex = -1;
+        const scale = getScale(100); // 10 is just a dummy value, the scale is only needed for its non-tick props like step, bandwidth, ...
+        if (!scale) throw new Error('Scale generation for the multilayer axis failed');
```
markov00, line 344: not for this PR: we reduce that value to 10 as in the comment or less, this is just used for the non-tick props, but the scale constructor initialize the ticks arrays and we waste cycles just for nothing (especially for time axis where the tick computation needs to pass through `momentjs`)

packages/charts/src/chart_types/xy_chart/state/selectors/visible_ticks.ts
markov00: There is definitely a lot of confusion around these formatting functions, we need to refactor and rename them the right way.
In the meantime, I think we can remove one of the two props, in particular, `label` is only assigned and is only read in tests, so we can remove it. As the name implies, it should be the final label, so the formatter needs to be carefully chosen following the fallback chain used previously.
The tests are actually failing because we are not passing the right formatter from what I can see

packages/charts/src/chart_types/xy_chart/state/selectors/visible_ticks.ts
monfera: Added this small change c62c8b8164a30d3951ca43462f5872aea36be061

packages/charts/src/chart_types/xy_chart/state/selectors/axis_tick_formatter.ts
```diff
@@ -0,0 +1,61 @@
...
+    return groupIds.reduce<AxisLabelFormatters>(
+      (acc, groupId) => {
+        const ySpecDataFormatter = (seriesByGroupId[groupId] ?? []).find(({ tickFormat }) => tickFormat)?.tickFormat;
+        const axes = groupAxesByCartesianCoords(axesByGroupId[groupId] ?? [], rotation);
+        axes.x.forEach((spec) => {
+          acc.x.set(spec.id, (v) => (spec?.labelFormat ?? spec?.tickFormat ?? defaultTickFormatter)(v, { timeZone }));
+        });
+        axes.y.forEach((spec) => {
+          acc.y.set(spec.id, (v) =>
+            (spec.labelFormat ?? spec.tickFormat ?? ySpecDataFormatter ?? defaultTickFormatter)(v, {}),
```
markov00, line 43: I've created an issue to track this and possibly refactor the logic in the future https://github.com/elastic/elastic-charts/issues/1479

markov00: I also like to note that this new logic and the exposed `AxisLabelFormatter` function doesn't need anymore the access to the `timeZone` value to correctly call the formatter, but the function already is currying that

packages/charts/src/chart_types/xy_chart/state/selectors/visible_ticks.ts
```diff
@@ -232,224 +213,144 @@ export const getVisibleTickSetsSelector = createCustomCachedSelector(
...
-                  ...raster,
-                  ticks: scale.type === ScaleType.Log ? raster.ticks : nonZeroLengthTicks,
-                },
-                fallbackAskedTickCount,
-              };
-            } else if (atLeastTwoTicks && uniqueLabels.size <= fallbackReceivedTickCount) {
-              // let's remember the smallest triedTickCount that yielded two distinct ticks
-              fallbackReceivedTickCount = uniqueLabels.size;
-              fallbackAskedTickCount = triedTickCount;
+        const fillLayer = (maxTickCountForLayer: number) => {
```
markov00, line 282: There is no big changes here, the `fillLayerTimeslip` and the content of the `if(isMultilayerTimeAxis)` are moved into the `multilayer_ticks.ts` file
The `fillLayer` function has some some small variable renamed

packages/charts/src/chart_types/xy_chart/state/selectors/visible_ticks.ts
markov00: I've fixed this in https://github.com/elastic/elastic-charts/pull/1476/commits/6e488eed702e73591016a32c2966f8198395eaf9

nickofthyme (approved):
Code changes LGTM, tested locally on storybook.

packages/charts/src/chart_types/xy_chart/state/selectors/visible_ticks.ts
```diff
@@ -71,8 +71,8 @@ export function generateTicks(
     return {
       value,
       domainClampedValue,
-      label: tickFormatter(value, tickFormatOptions),
-      axisTickLabel: labelFormatter(value, tickFormatOptions),
+      label: labelFormatter(value),
```
monfera: 😍

packages/charts/src/chart_types/xy_chart/state/selectors/visible_ticks.ts
```diff
@@ -71,8 +71,8 @@ export function generateTicks(
...
+      axisTickLabel: labelFormatter(value),
```
monfera: no concern with the options 😍

markov00:
@monfera @nickofthyme my commits on this PR are moved into #1481 as they are not directly pertinent to this PR context

PR #1478, opened 2021-11-15
fix(partition): rtl text label support (#1433) [38.0.x]

nickofthyme:
Backports the following commits to 38.0.x:
 - fix(partition): rtl text label support (#1433)


nickofthyme:

PR #1493, opened 2021-11-19
fix(xy): remove pointer cursor for onElementOver

### LINKED ISSUE #1437: Interactions: onElementClick or onElementOver
Opening this issue for discussion, in #1428, I went with feature parity with the cartesian charts and have `onElementClick` _or_ `onElementOver` in partition charts lead to the pointer cursor versus default cursor. 

Is it misleading for a hover-only action to have a pointer cursor vs when there is a clickable action? Is there more context in how consumers want their pointers that might answer this? Thanks!

For example the screen recording below has an element _over_ listener however it has no element _click_ listener and it looks like it does.

https://user-images.githubusercontent.com/19007109/138120183-25bcb1bd-4095-441b-a126-193613bbaaa1.mp4

rshen91:
## Summary
The pointer cursor will not be used in over element actions to avoid misleading users.

## Issues
Closes #1437 

### Checklist


markov00 (approved):
Looks good to me

PR #1495, opened 2021-11-22
feat: expose computeRatioByGroups fn

markov00:
## Summary

This PR exposes the function to compute the participation ratio of a value in the total sum of its membership group.
It can be used to compute non-stacked percentage bar charts.
Right now, we only compute such data only for stacked bar charts, but exposing that utility function can also help creating non stacked bar chart.

image

The story was also updated to nicely match the categories used with all the edge cases considered.

## Details

As described in the tsdocs: The ratio is computed using absolute values, for example: 
Product A made a profit of $200, and product B has a loss of $300. In total, the company lost $100 ($200 – $300).
Product A contributed: abs(200) / ( abs(200) + abs(-300) ) * 100% = 40%
Product B contributed: abs(-300) / ( abs(200) + abs(-300) ) * 100% = 60%
Product A and product B contributed 40% and 60% respectively on a total loss of $100.

We don't compute the ratio for non-finite values. In this case, the original non-finite value is returned.
If the sum of the group values is 0, each ratio is considered 0.

## Issues

It enabled the non-stacked percentage bar charts in Kibana https://github.com/elastic/kibana/issues/118525

### Checklist


monfera (approved):
LGTM 🎉 

Some comment changes suggested.

Also, this makes me wonder if there'd be utility to a sign-aware version, such that, with the same setup as in the example, 

- Product A contributes (which I think is an OK word here) -200% of the results
- Product B contributes 300% of the results

The two together contribute 100% of the results. So what's in the PR is a net version, and the above is a gross version

packages/charts/src/utils/data/data_processing.ts
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * The function computes the participation ratio of a value in the total sum of its membership group.
+ * It returns a shallow copy of the input array where each object is augmented with the computed ratio.
+ *
+ * @remarks
+ * The ratio is computed using absolute values.
+ * Product A made a profit of $200, and product B has a loss of $300. In total, the company lost $100 ($200 – $300).
+ * Product A contributed: abs(200) / ( abs(200) + abs(-300) ) * 100% = 40%
+ * Product B contributed: abs(-300) / ( abs(200) + abs(-300) ) * 100% = 60%
+ * Product A and product B contributed 40% and 60% respectively on a total loss of $100.
```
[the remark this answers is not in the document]
markov00: You are right, I've updated it here https://github.com/elastic/elastic-charts/pull/1495/commits/9e63025e8d417140db925216b57fbe47ead6c6e8

markov00:
>Also, this makes me wonder if there'd be utility to a sign-aware version, such that, with the same setup as in the example,
>Product A contributes (which I think is an OK word here) -200% of the results
>Product B contributes 300% of the results
>The two together contribute 100% of the results. So what's in the PR is a net version, and the above is a gross version

@monfera you are right, we can probably add that version in a subsequent PR, in the meantime I've created this issue to track it https://github.com/elastic/elastic-charts/issues/1500

packages/charts/src/chart_types/xy_chart/utils/stacked_series_utils.ts
```diff
@@ -77,10 +78,13 @@ export function formatStackedDataSeriesValues(
...
       // y0 can be considered as always present
-      reorderedArray[xIndex][`${key}-y0`] = y0;
+      reorderedArray[xIndex][`${key}-y0`] = isFiniteNumber(y0) ? (stackAsPercentage ? Math.abs(y0) : y0) : y0;
       // if y0 is available, we have to count y1 as the different of y1 and y0
       // to correctly stack them when stacking banded charts
-      reorderedArray[xIndex][`${key}-y1`] = (y1 ?? 0) - (y0 ?? 0);
+      const nonNullY1 = y1 ?? 0;
+      const nonNullY0 = y0 ?? 0;
+      reorderedArray[xIndex][`${key}-y1`] =
+        (stackAsPercentage ? Math.abs(nonNullY1) : nonNullY1) - (stackAsPercentage ? Math.abs(nonNullY0) : nonNullY0);
```
nickofthyme, line 87: @markov00 is this logic meant to stay? I am removing this to remove support for `y0Accessors` with `stackAccessors` in #1502 please confirm.

nickofthyme: Nevermind, I realized there is a difference between `y0` and `initalY0`

nickofthyme: See https://github.com/elastic/elastic-charts/pull/1502/commits/ef547ea681fe132d9a194542c2bf5c5083c33f77#r765350536

markov00: @nickofthyme the use of the `absolute` value instead of the current one should be used when stacking as percentage

PR #1498, opened 2021-11-22
fix(partition): linkLabel textColor override

### LINKED ISSUE #1469: [Partition] Link label text color parameter passed is ignored
**Describe the bug**
While the `textColor` parameter is still accepted as valid for the `linkLabel` configuration object, its value is overwritten internally to the best "contrast" color against the container background color.

While the aim sounds nice, in Kibana charts embeddable cannot have a background color as Canvas requires them to be styleable somehow, so the background color is currently set to `undefined`/`transparent`. This leads to manually compute the `textColor` programmatically in Lens, and we land on the point above.

**Expected behaviour**
Respect the passed value for `textColor`.

**Kibana Cross Issues**
https://github.com/elastic/kibana/issues/116777

**Checklist**

nickofthyme:
## Summary

The partition chart type now respects the `textColor` defined on the `config.linkLabels` object. This will **NOT** adjust the color based on the background contrast.

Screen Recording 2021-11-22 at 12 28 55 PM

## Details

Added new `ColorVariant` key called `Adaptive` to be used in cases where color should be adaptive depending on the background contrast. Only when the the `linkLabel.textColor` is set to `Adaptive`, is the color contrast determined, otherwise the custom color string is used for all themes.

Also changed the `eachTheme` api to pass `params` without the `&` prefix, to be placed anywhere in the url.

---

Side note: I improved the error handling in the vrt framework. Before we waited for element from selector, if this failed, it would throw a misleading error that didn't help. Now if the element or url fails to load, there is an assertion to check the element exists and if not the test fails but does **NOT** throw an error. If the url fails to load the url will be logged to the console for easier debugging. See example below.

```
    fillLabel textcolor - treemap
      ✕ should show custom red textColor (10008 ms)

Docker Chromium: Shutting down Docker container...
  console.error
    Failed to load url. Check story at:

    	storybook url: http://localhost:9001/?path=/story/treemap--one-layer-2&knob-partitionLayout=sunburst&knob-custom fillLabel.textColor=true&knob-fillLabel.textColor=rgba(255, 0, 0, 0.85)
    	local vrt url: http://host.docker.internal:9002?path=%2Fstory%2Ftreemap--one-layer-2&knob-partitionLayout=sunburst&knob-custom%20fillLabel.textColor=true&knob-fillLabel.textColor=rgba(255%2C%200%2C%200%2C%200.85)&knob-debug=false

      441 |         return true;
      442 |       } catch {
    > 443 |         console.error(`Failed to load url. Check story at: \n\n\t storybook url: ${url}\n\tlocal vrt url: ${cleanUrl}`);
          |                 ^
      444 |         return false;
      445 |       }
      446 |     }

      at CommonPage.<anonymous> (page_objects/common.ts:443:17)
      at step (page_objects/common.ts:51:23)
      at Object.throw (page_objects/common.ts:32:53)
      at rejected (page_objects/common.ts:24:65)
```

## Issues

fix #1469

### Checklist


nickofthyme:
FYI I came across https://github.com/elastic/elastic-charts/issues/1499 while making these changes. So I added a VRT to test that in the future.

markov00 (approved):
Thanks Nick, this change works for me. After removing the `textInvertible` property from the config some release ago, there was no more option to overwrite that computed value.
What do you think if we use the same technique also for the fill_label?

nickofthyme:
> What do you think if we use the same technique also for the fill_label?

Yeah I can add that

nickofthyme:
Added suggested change in 74013ff. @markov00 could you do another review of the added changes please, thanks!

integration/page_objects/common.ts
```diff
@@ -451,19 +465,6 @@ class CommonPage {
...
-   * @param {string} [url]
-   * @param {string} [waitSelector]
-   */
-  async testAccessibilityTree(url: string, waitSelector: string): Promise<AXNode> {
-    await this.loadElementFromURL(url, waitSelector);
-    const accessibilitySnapshot = await page.accessibility.snapshot().then((value) => {
-      return value;
-    });
-    return accessibilitySnapshot;
-  }
```
nickofthyme, line 466: Not used anywhere

integration/page_objects/common.ts
```diff
@@ -426,20 +431,29 @@ class CommonPage {
...
    */
-  async loadElementFromURL(url: string, waitSelector?: string, timeout?: number) {
+  async loadElementFromURL(url: string, waitSelector?: string, timeout?: number): Promise<boolean> {
     const cleanUrl = CommonPage.parseUrl(url);
     await page.goto(cleanUrl);
 
     if (waitSelector) {
-      await this.waitForElement(waitSelector, timeout);
+      try {
+        await this.waitForElement(waitSelector, timeout);
```
nickofthyme, line 440: This method can throw for some reason.

integration/page_objects/common.ts
```diff
@@ -320,23 +318,30 @@ class CommonPage {
...
     options?: ScreenshotElementAtUrlOptions,
   ) {
-    await this.loadElementFromURL(url, options?.waitSelector ?? selector, options?.timeout);
+    try {
+      const success = await this.loadElementFromURL(url, options?.waitSelector ?? selector, options?.timeout);
 
-    if (options?.action) {
-      await options.action();
-    }
+      expect(success).toBe(true);
```
nickofthyme, line 324: Assertions throw errors for some reason, hence wrapping in `try/catch` block.

packages/charts/src/chart_types/partition_chart/layout/viewmodel/viewmodel.ts
```diff
@@ -156,10 +156,13 @@ export function makeQuadViewModel(
     const strokeWidth = sectorLineWidth;
     const strokeStyle = sectorLineStroke;
     const textNegligible = node.y1px - node.y0px < minRectHeightForText;
-    const textColor =
-      !isSunburstLayout && textNegligible
-        ? Colors.Transparent.keyword
-        : fillTextColor(fillColor, containerBackgroundColor);
```
nickofthyme, line 162: @monfera is there a reason for `isSunburstLayout` here?

packages/charts/src/chart_types/partition_chart/layout/viewmodel/viewmodel.ts
```diff
@@ -156,10 +156,13 @@ export function makeQuadViewModel(
...
+
+    const textColor = textNegligible
+      ? Colors.Transparent.keyword
+      : fillLabel.textColor === ColorVariant.Adaptive
+      ? fillTextColor(fillColor, containerBackgroundColor)
+      : fillLabel.textColor;
```
markov00, line 164: @monfera do you remember why the original code was using `!isSunburstLayout && textNegligible` why the non-sunburst layout condition?

integration/tests/partition_stories.test.ts
```diff
@@ -6,6 +6,9 @@
  * Side Public License, v 1.
  */
 
+import { PartitionLayout } from '@elastic/charts/src';
```
markov00, line 9:
```suggestion
import { PartitionLayout } from '@elastic/charts';
```

integration/tests/partition_stories.test.ts
```diff
@@ -19,4 +22,65 @@ describe('Axis stories', () => {
...
+      await common.expectChartAtUrlToMatchScreenshot(
+        `http://localhost:9001/?path=/story/treemap--one-layer2&knob-partitionLayout=${partitionLayout}&knob-custom fillLabel.textColor=true&knob-fillLabel.textColor=rgba(255, 255, 255, 0.3)`,
+      );
+    });
+    it('should show custom black/translucent textColor', async () => {
+      await common.expectChartAtUrlToMatchScreenshot(
+        `http://localhost:9001/?path=/story/treemap--one-layer2&knob-partitionLayout=${partitionLayout}&knob-custom fillLabel.textColor=true&knob-fillLabel.textColor=rgba(0, 0, 0, 0.3)`,
+      );
+    });
+  });
```
markov00, line 85: do we really need all these colors? can we just pass a semi-transparent color and that's it? it should work the same independently of the color value

nickofthyme: I want to check that there is no issue with transparency for each theme.

PR #1514, opened 2021-12-01
feat(heatmap): allow rotation of x axis labels

### LINKED ISSUE #1168: [Heatmap] Rotate or hide overlapping labels on X-axis
**Describe the bug**
For categorical data labels on the X-axis overlap when cells are not wide enough. 

**To Reproduce**
Provide chart data that contains fairly long labels on the `x` axis.
 
**Expected behaviour**
Labels are rotated and trimmed if necessary. 
image

**Screenshots**
image

**Kibana Cross Issues**
- https://github.com/elastic/kibana/pull/97978
- https://github.com/elastic/kibana/issues/61248
- https://github.com/elastic/kibana/issues/34838
- https://github.com/elastic/kibana/issues/211214

rshen91:
## Summary

The `rotation` configuration is available on the `theme.heatmap.xAxisLabel` theme style to rotate the labels of the X axis from 0 to 90 degree.
The rotation will shrink the chart vertically( and occasionally also horizontally) depending on the length of the labels.
Few changes to the x axis labels are also landed in this PR:
- categorical labels overflowing the chart boundaries will play a role in shrinking the chart size.
- the X axis now shows small ticks, center aligned for categorical scale and left aligned for time
- the time axis ticks are aligned to the right of the ticks
- time axis labels instead, will not shrink the chart anymore. The number of labels will be reduced gradually to avoid overlaps and overflows. The right edge of the heatmap will always touch the right edge of the rendering area.

### BREAKING CHANGES

`width`, `align` and `baseline` style properties are removed from the `xAxisLabels` and `yAxisLabels` style of the Heatmap theme.
These properties were not used in the rendering.

### Compute the X axis space

Screenshot 2022-02-15 at 19 18 35

The X axis space is computed as the following:
`categorical`
- measure all the categorical label sizes
- compute the rotated boundingbox
- use the `screenspaceMarkerScaleCompressor` to compute how much we need to shrink the grid width considering the label positions in the domain and their screen size. I also included the Y axis so that we can take also the maximum label width in the calculation.

`time`
- similar to categorical we compute all the label sizes and the rotated bbox
- we compute a first scale compression considering all the label visible. If that step produces no overlaps between labels and no overflows at the chart edges, we use that result if not we iterate over reducing the number of visible ticks each time by skipping ticks with an increasing cadence. The current tick reduction doesn't follow the `multiLayerTimeAxis` but in a future step we can include that.

### Future possible steps

- wrap and/or trim horizontal labels into multiple lines within their grid column size
- reuse the multilayer time axis approach

## Issues

Partially solves #1168  (missing the automatic trimming of horizontal labels) 

### Checklist


packages/charts/api/charts.api.md
```diff
@@ -1167,6 +1167,15 @@ export interface HeatmapConfig {
...
         padding: number;
         formatter: (value: string | number) => string;
+        rotation: number;
+        overflow: 'ellipsis' | false;
+        maxTextLength: {
+            dflt: number;
+            min: number;
+            max: number;
+            documentation: string;
+        };
```
markov00, line 1177: `dflt` and `documentation` here used in the past to generate knobs automatically, but they should not be part of the API IMHO. What do you think?

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
```diff
@@ -418,13 +420,20 @@ function getXTicks(
...
-    };
+    return config.xAxisLabel.overflow
+      ? {
+          ...getTextValue(config.xAxisLabel.formatter, xScale, config.xAxisLabel.overflow)(textBox),
+          x: chartDimensions.left + (xScale(textBox) || 0) + xScale.bandwidth() / 2,
+        }
+      : {
+          ...getTextValue(config.xAxisLabel.formatter, xScale, config.xAxisLabel.overflow)(textBox),
+          x: chartDimensions.left + (xScale(textBox) || 0) + xScale.bandwidth() / 2,
+        };
```
markov00, line 437: I don't see the difference between the two exclusive expressions

packages/charts/src/chart_types/heatmap/renderer/canvas/canvas_renderers.ts
```diff
@@ -115,18 +115,47 @@ export function renderCanvas2d(
...
+                      ctx,
+                      { x: xValue.x, y: xValue.y },
+                      xValue.text,
+                      { ...config.xAxisLabel, align: 'left' },
+                      config.xAxisLabel.rotation,
+                    )
+                  : // no rotations specified
+                    renderText(ctx, { x: xValue.x, y: xValue.y }, xValue.text, config.xAxisLabel),
+              );
+        }),
```
markov00, line 158: I think you can reduce this code if you move the alternate condition within the `forEach` loop and the text left alignment condition within the configuration itself

packages/charts/api/charts.api.md
```diff
@@ -1167,6 +1167,15 @@ export interface HeatmapConfig {
         visible: boolean;
         padding: number;
         formatter: (value: string | number) => string;
+        rotation: number;
+        overflow: 'ellipsis' | false;
+        maxTextLength: {
+            dflt: number;
+            min: number;
```
markov00, line 1174: is the `min` used somewhere?

packages/charts/api/charts.api.md
rshen91: Sounds good to me I'm going to change this to a number vs object 👍🏻

packages/charts/api/charts.api.md
rshen91: Good call - refactoring for clarity 8287a2a

packages/charts/src/chart_types/heatmap/renderer/canvas/canvas_renderers.ts
rshen91: Nice! Good call - let me know what you think in 8287a2a

packages/charts/src/chart_types/heatmap/layout/viewmodel/viewmodel.ts
rshen91: :face_palm: thanks 8287a2a

packages/charts/api/charts.api.md
rshen91: 8287a2a for changes thank you!

rshen91:
jenkins test this

monfera:
If the non-round eg. 36 hour cadence is OK for this PR, let's please link an issue with the iteration 

image

Also, this may be the 1st chart I recall where some rendering happens in the peach border, text on the top and white area on the bottom, seen on the Time and Time Snap stories

integration/tests/heatmap_stories.test.ts
```diff
@@ -58,4 +58,16 @@ describe('Heatmap stories', () => {
       'http://localhost:9001/?path=/story/heatmap-alpha--basic&knob-Show%20x%20axis%20title=true&knob-Show%20y%20axis%20title=true',
     );
   });
+
```
monfera, line 61: If the non-round eg. 36 hour cadence is OK for this PR, let's please link an issue with the iteration 

image

Also, this may be the 1st chart I recall where some rendering happens in the peach border, text on the top and white area on the bottom, seen on the Time and Time Snap stories

integration/tests/heatmap_stories.test.ts
```diff
@@ -58,4 +58,16 @@ describe('Heatmap stories', () => {
...
+  test('rotate categorical axis labels', async () => {
+    await common.expectChartAtUrlToMatchScreenshot(
+      'http://localhost:9001/?path=/story/test-cases--heatmap-label-rotation&globals=theme:light&knob-auto width_y axis=true&knob-width_y axis=50&knob-visible_x axis=true&knob-fontSize_x axis=12&knob-padding_x axis=6&knob-rotation_x axis=45&knob-categorical data=true&knob-show legend=false',
```
monfera, line 64: In Storybook, there's a block for `Heatmap` and a block for `Rotations`. There's no heatmap rotation story in either, they're currently in the nondescript `Test cases` (the entire storybook is test cases). Let's move it to `Heatmap` for discoverability. We should also discuss grouping approaches, such as chart type based vs feature based

markov00: done https://github.com/elastic/elastic-charts/pull/1514/commits/a2d71abe3c259f1b9a9858da3bd5b4809b140b3b

integration/tests/heatmap_stories.test.ts
markov00: I've described that in our DataVis FAQ I will send you the link on slack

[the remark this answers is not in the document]
markov00: With https://github.com/elastic/elastic-charts/pull/1514/commits/ceab0e670f439fc4ab0478188e8d7d3d1743b68f I've introduced the the minimal/optimal rotation that constraint the minimum rotation to that optimal value, avoiding always the overlaps.

monfera (approved):
Good for merging once all tests pass

PR #1554, opened 2022-01-18
fix(deps): remove the ResizeObserver polyfill

markov00:
## Summary
The [Resize Observer API](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) is available in all major browsers and we don't need to polyfill it anymore.

@nickofthyme I've also updated the ech version in the storybook package. I know that it's not used directly, but this can prevent downloading unnecessary libraries if we keep them in sync. Do you know a better way to link a workspace as a dependency without using the specific package version?

### DIFF
--- a/packages/charts/package.json
+++ b/packages/charts/package.json
@@ -46,7 +46,6 @@
     "react-redux": "^7.1.0",
     "redux": "^4.0.4",
     "reselect": "^4.0.0",
-    "resize-observer-polyfill": "^1.5.1",
     "ts-debounce": "^3.0.0",
     "utility-types": "^3.10.0",
     "uuid": "^3.3.2",
--- a/packages/charts/src/components/chart_resizer.tsx
+++ b/packages/charts/src/components/chart_resizer.tsx
@@ -9,13 +9,12 @@
 import React, { RefObject } from 'react';
 import { connect } from 'react-redux';
 import { Dispatch, bindActionCreators } from 'redux';
-import ResizeObserver from 'resize-observer-polyfill';
 import { debounce } from 'ts-debounce';
 
 import { updateParentDimensions } from '../state/actions/chart_settings';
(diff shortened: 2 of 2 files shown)


nickofthyme:
> Do you know a better way to link a workspace as a dependency without using the specific package version?

I was looking into this lately and bazel is obvious if we want to use it which could make linking charts in kibana easier.

The other one I came across was [`nx`](https://www.npmjs.com/package/nx), bascially a stripped down build system for monorepos.

storybook/package.json
```diff
@@ -9,7 +9,7 @@
     "typecheck": "tsc -p ./tsconfig.json --noEmit"
   },
   "dependencies": {
-    "@elastic/charts": "^30.2.0",
+    "@elastic/charts": "latest",
```
nickofthyme, line 12: Do you think a wildcard is more sensible? Your choice.

```suggestion
    "@elastic/charts": "*",
```

yarn.lock
```diff
@@ -10603,7 +10585,8 @@ eslint-module-utils@^2.6.0:
     pkg-dir "^2.0.0"
 
 "eslint-plugin-elastic-charts@link:./packages/eslint-plugin-elastic-charts":
-  version "1.0.0"
+  version "0.0.0"
+  uid ""
```
nickofthyme, line 10589: I don't know why this 💩 happens 🤔

nickofthyme (approved):

storybook/package.json
markov00: I've reverted that change for now

yarn.lock
markov00: change reverted

storybook/package.json
nickofthyme: I think we move to `DEVELOPMENT` or something.

PR #1569, opened 2022-01-25
fix: use window scope to call the ResizeObserver

markov00:
## Summary

JSDOM doesn't have the `ResizeObserver` available in the `global` object but instead is available directly from the `window` object.

Solves Kibana jest test failures [here](https://buildkite.com/elastic/kibana-pull-request/builds/19241)


rshen91 (approved):

markov00:
jenkins test this please

PR #1572, opened 2022-01-29
ci(playwright): e2e framework setup

nickofthyme:
This PR migrates all tests in `e2e/` to the playwright syntax and setups up the playwright test runner.

## What changed?

- adds `yarn test:e2e:server` - starts e2e vrt server as before, except now checks for `tmp/examples.json` before running
- adds `yarn test:e2e` which replaces `yarn test:integration:local` - runs e2e tests in playwright docker container. Accepts any playwright args following command (e.g. `yarn test:e2e --project=chrome`).
- removes the old all-in-one `yarn test:integration` that would start up the e2e web server before running tests. This is not easy with playwright as it does not wait until webpack is finished compiling before starting the tests.


e2e-server/server/webpack.config.js
```diff
@@ -16,11 +16,11 @@ const webpack = require('webpack');
 
 module.exports = {
   entry: '../tmp/index.tsx',
-  mode: 'production',
+  mode: process.env.NODE_ENV === 'production' ? 'production' : 'development',
```
nickofthyme, line 19: Future note to assign `NODE_ENV` to `'production'` in ci for branches, not prs.

e2e/e2e_config.ts
```diff
@@ -6,10 +6,4 @@
...
  */
 
-module.exports = {
-  debug: process.env.DEBUG === 'true',
-  port: process.env.PORT || '9009',
-  hostname: process.env.DEBUG === 'true' ? 'localhost' : 'host.docker.internal',
-  isLocalVRTServer: process.env.LOCAL_VRT_SERVER === 'true',
-  isLegacyVRTServer: process.env.LEGACY_VRT_SERVER === 'true',
-};
+export const environmentUrl = process.env.ENV_URL;
```
nickofthyme, line 9: `PORT` default is assigned in `scripts/test.sh`. The `hostname` is inferred from the environment (i.e. docker or not). These are used to assign `ENV_URL`. If `ENV_URL` is explicitly defined, this will be used as is, disregarding `PORT` and `hostname`.

packages/charts/src/chart_types/heatmap/specs/heatmap.ts
```diff
@@ -95,47 +95,45 @@ export interface HeatmapSpec<D extends BaseDatum = Datum> extends Spec {
...
   props: SFProps<
     HeatmapSpec<D>,
-    keyof typeof buildProps.current['overrides'],
-    keyof typeof buildProps.current['defaults'],
-    keyof typeof buildProps.current['optionals'],
-    keyof typeof buildProps.current['requires']
+    keyof typeof buildProps['overrides'],
+    keyof typeof buildProps['defaults'],
+    keyof typeof buildProps['optionals'],
+    keyof typeof buildProps['requires']
```
markov00, line 133: Thanks!

integration/page_objects/common.ts
```diff
@@ -167,8 +167,8 @@ class CommonPage {
    *
    * @param selector
    */
-  async getBoundingClientRect(selector: string) {
-    return await page.$eval(selector, (element) => {
+  getBoundingClientRect(selector: string) {
+    return page.$eval(selector, (element) => {
```
markov00, line 171: Is that no more async?

e2e/tests/timezone.test.ts
```diff
@@ -6,35 +6,45 @@
...
-    await common.expectChartWithMouseAtUrlToMatchScreenshot(
-      'http://localhost:9001/?path=/story/bar-chart--with-time-x-axis',
-      { left: 80, top: 100 },
-      { screenshotSelector: 'body' },
-    );
+  test.describe('timezone Europe/Rome', () => {
+    const timezoneId = 'Europe/Rome';
+    test.use({ timezoneId });
+    test('use default local timezone Europe/Rome', async ({ page, contextOptions }) => {
+      expect(contextOptions.timezoneId === timezoneId);
```
markov00, line 41: are we testing the test?

integration/page_objects/common.ts
nickofthyme: It is, but the `return await` is unnecessary it always returns a promise.

https://eslint.org/docs/rules/no-return-await#no-return-await

e2e/declarations/@types/jest.d.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+// TODO fix these overrides
+// This file does not take priority of the external jest types in `@types/jest`
+// But the jest-environment-puppeteer override works fine....I'm lost
```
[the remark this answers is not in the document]
nickofthyme: Yeah, anytime you have `jest` or `jest-environment-puppeteer` installed they automatically add their global types to your project. I think this is hysterical so if say `jest` were in `packages/charts/node_modules` that would not pollute `e2e` or other peripheral directories.

The issue with these globals is that writing a test in says `e2e/tests/foo.test.ts`, this file sees `test`, `expect`, `describe`, etc. as global values but really they are not because that project does not use jest. So the IDE or typescript, thinks it's fine to use `jest.test` in `e2e/tests/foo.test.ts` but really that values is undefined and should be using importing `expect`, `test`, etc from `@playwright/test`.

So these declaration files are to override the implicit globals defined in `node_modules/@types`, though this does not work for jest types for some reason that I need to investigate later. It does work for puppeteer types though which is strange.

e2e/tests/timezone.test.ts
nickofthyme: [2744ffa](https://github.com/elastic/elastic-charts/pull/1572/commits/2744ffa617de3f390ba98032a0496095e012de06)

e2e/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "charts-e2e",
+  "description": "Visual regression test runner",
+  "license": "SEE LICENSE IN LICENSE.txt",
```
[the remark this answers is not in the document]
nickofthyme: [2744ffa](https://github.com/elastic/elastic-charts/pull/1572/commits/2744ffa617de3f390ba98032a0496095e012de06)

e2e/declarations/@types/jest.d.ts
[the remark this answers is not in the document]
markov00: thanks, understood

PR #1585, opened 2022-02-09
fix(legend): width with duplicate pie slice labels

### LINKED ISSUE #1584: Bad legend sizing when using 2 identical labels
**Describe the issue**

Using a legend with 2 or more identical labels causes the sizing logic to fail and render a very narrow legend.

**To Reproduce**
1. Go to https://codesandbox.io/s/confident-http-qv2ks?file=/src/App.tsx
2. Notice the legend label is about 1 px width
3. Toggle off the `legendMaxDepth` option and the width with return to the correct value.

**Expected behaviour**
Legend sizes correctly

**Screenshots**

Peek 2022-02-08 15-43

**Kibana Cross Issues**

https://github.com/elastic/kibana/issues/125129

**Checklist**

nickofthyme:
## Summary

fixes a bug returning a zero-width legend for identical legend labels using `legendMaxDepth` option.

### Before 
image

### After
image

## Details

The legend label logic in partition charts was understandably simplified to assume the formatted labels across all depths are unique. However, this simplification runs into issues whenever two or more labels across any layer or any depth are identical. 

https://github.com/elastic/elastic-charts/blob/185ab4733cd2415455664f497bc8a4f0d54fa8b8/packages/charts/src/chart_types/partition_chart/layout/utils/legend_labels.ts#L34-L37

The logic above creates a `keys` `Map` to track the depths across layers and nodes using the formatted label as the key. So in the case where the label is the same for two separate layers (as pictured above), the depth count is incorrectly combined to `2`.

```ts
// current
const legenLabels: LegendItemLabel[] = [
  {
    'label': 'Testing a super duper really long legend',
    'depth': 2
  }
]
```

For the current case, this becomes a problem when limiting the depth using `legendMaxDepth` in which the `depth` count is now incorrectly `2` and thus returns `[]` causing the legend width to be `0`.

https://github.com/elastic/elastic-charts/blob/185ab4733cd2415455664f497bc8a4f0d54fa8b8/packages/charts/src/chart_types/partition_chart/layout/utils/legend_labels.ts#L14-L16

The solution --  prevent label collisions using a key derived from the `arrayNode.path` returning all legend labels as a unique value in the labels array. So the example able would be separated into two where the first is at a `depth` of `1` and the second is at a `depth` of `2`

```ts
// current
const legenLabels: LegendItemLabel[] = [
  {
    'label': 'Testing a super duper really long legend',
    'depth': 2
  }
]

// expected
const legenLabels: LegendItemLabel[] = [
  {
    'label': 'Testing a super duper really long legend',
    'depth': 1
  },
  {
    'label': 'Testing a super duper really long legend',
    'depth': 2
  }
]
```

I confirmed the functionality of the legend highlight strategy is working as expected with this unique label change.

## Issues

fix #1584

### Checklist


packages/charts/src/chart_types/partition_chart/layout/utils/legend_labels.ts
```diff
@@ -8,18 +8,29 @@
...
   return flatSlicesNames(layers, 0, tree).filter(({ depth }) => depth <= legendMaxDepth);
 }
 
+/** @internal */
+export function getArrayNodeKey(arrayNode: ArrayNode): string {
+  return arrayNode[PATH_KEY].reduce<string>((acc, { value, index }) => {
+    if (value === HIERARCHY_ROOT_KEY || value === NULL_SMALL_MULTIPLES_KEY) return acc;
+    return `${acc}(${index}):${value}__`;
+  }, '__');
+}
```
nickofthyme, line 26: Creates keys such as `__(0):1__` and `__(0):1__(0):a__` for example in description.

nickofthyme:
@markov00 can you review and approve this?

packages/charts/src/chart_types/partition_chart/layout/utils/legend_labels.ts
```diff
@@ -32,15 +48,19 @@ function flatSlicesNames(
     const formattedValue = formatter ? formatter(key) : `${key}`;
     // preventing errors from external formatters
     if (formattedValue && formattedValue !== HIERARCHY_ROOT_KEY) {
+      // Node key must be unique for each node in the tree
+      const nodeKey = getArrayNodeKey(arrayNode);
       // save only the max depth, so we can compute the the max extension of the legend
-      keys.set(formattedValue, Math.max(depth, keys.get(formattedValue) ?? 0));
+      depths.set(nodeKey, Math.max(depth, depths.get(nodeKey) ?? 0));
```
markov00, line 54: do we need to compute the `max` depth now since we are creating unique nodeKeys for each node?

storybook/stories/test_cases/test_cases.stories.tsx
```diff
@@ -19,3 +19,4 @@ export { Example as accessibilityCustomizations } from './6_a11y_custom_descript
 export { Example as rtlText } from './7_rtl_text.story';
 export { Example as testPointsOutsideOfDomain } from './8_test_points_outside_of_domain.story';
 export { Example as heatmapAxisLabels } from './9_heatmap_axis_labels.story';
+export { Example as partitionLegendLabels } from './10_partition_legend_labels.story';
```
markov00, line 22: Can you rename this to something more specific like "duplicateLabelsInPartitionLegend" or something along the line?

packages/charts/src/chart_types/partition_chart/layout/utils/legend_labels.ts
```diff
@@ -8,18 +8,34 @@
 
 import { LegendItemLabel } from '../../../../state/selectors/get_legend_items_labels';
```
markov00, line 9: A comment about the screenshots:

- why do you have that screenshot about the heatmap? It doesn't looks good and doesn't pertain to that PR
- missing the scrollbar in the new e2e tests is something we need to solve

nickofthyme: Yeah I can remove the heatmap image, not sure why that was created.

nickofthyme: [3df0720](https://github.com/elastic/elastic-charts/pull/1585/commits/3df0720c08ba51d30e98e4751cf5e63fda9b6936)

storybook/stories/test_cases/test_cases.stories.tsx
nickofthyme: 4eb91963b121b8c302a4d4bcf3cb2488232d01b7

packages/charts/src/chart_types/partition_chart/layout/utils/legend_labels.ts
markov00: @nickofthyme from the previous line comment the nodeKey is unique for each node in the tree, so you probably can rewrite the current line with:
```suggestion
      depths.set(nodeKey, depth);
```

markov00 (approved):
LGTM

nickofthyme:
2 Playwright tests are failing only in the CI. Merging for now and will troubleshoot in a separate PR.

PR #1609, opened 2022-02-24
fix(axis): correct tick alignment in ordinal scale with numeric values

### LINKED ISSUE #1608: [XY] Tick labels overlap for numeric values
**Describe the issue**
I am trying to create a bar (happens for both vertical and horizontal ones) with a numeric field.

image

I have a terms of dayOfweek (number field) on the vertical axis and the tick labels overlap each other.

**To Reproduce**
1. Go to aggregation based XY charts in kibana 
2. Create a horizontal axis on the sample data flights
3. Add an average averageTicket price as metric
4. Add terms on dayOfweek on the X-axis
5. Check the tick labels

**Expected behaviour**
Labels should not overlap each other

**Additional context**
I can't replicate it in Lens

**Errors in browser console**
No errors in the console

**Kibana Cross Issues**
https://github.com/elastic/kibana/issues/126241

**Checklist**

markov00:
## Summary

The tick line and label correctly renders now when using an ordinal scale with numeric values on the x axis.

## Details

The changes were applied in #1446 to remove tick labels that were outside the domain in a multi-layer time axis.
The problem was caused by the fact that the tick position was limited to the min value between the `domain[0]` and the value itself. The `domain[0]`, in a continuous scale is the minimum, but when using an ordinal scale, the `domain[0]` is the first value of the domain and, in a non ordered or just reversed, numerical domain, it can be different than the actual min. This caused the wrong positioning of the labels.

In the VRT test, the change on the data order is desired  to check for regressions.
If you want to double-check locally just replace the changed line to the original version:
```ts
const domainClampedValue = typeof value === 'number' && typeof scale.domain[0] === 'number' ? Math.max(scale.domain[0], value) : value;
```
## Issues

fix #1608 

### Checklist


packages/charts/src/chart_types/xy_chart/state/selectors/visible_ticks.ts
```diff
@@ -72,9 +73,9 @@ export function generateTicks(
   showGrid: boolean,
 ): AxisTick[] {
   const getDirection = getDirectionFn(scale);
+  const isContinuous = isContinuousScale(scale);
   return ticks.map<AxisTick>((value) => {
-    const domainClampedValue =
-      typeof value === 'number' && typeof scale.domain[0] === 'number' ? Math.max(scale.domain[0], value) : value;
+    const domainClampedValue = isContinuous && typeof value === 'number' ? Math.max(value, scale.domain[0]) : value;
```
nickofthyme, line 78: So `typeof scale.domain[0] === 'number'` was redundant?

nickofthyme (approved):

packages/charts/src/chart_types/xy_chart/state/selectors/visible_ticks.ts
markov00: checking if the `scale` is continuous makes the domain a `[number, number]` type.
The main problem here was the fact that, instead of checking for the scale to be continuous we considered the scale continuous if the domain[0] was a number (that can be true also for ordinal scales)

nickofthyme: Got it, thanks!

PR #1621, opened 2022-03-04
feat(goal): expose max sizing limits in theme.goal options

### LINKED ISSUE #1599: [Goal] Resizing limitations.
## Problem description

While trying to replace the `vislib` gauge with the goal from the `elastic-charts`, it has been found the further issue.

The `goal` chart has limitations on its maximum size.

### Current behavior

https://user-images.githubusercontent.com/22456368/154481512-ea9daa67-f9e3-477a-a60a-be0712f85e8d.mov

## Expected behavior

After the resizing of the container, the `goal` chart should be resized without any limitations.

https://user-images.githubusercontent.com/22456368/154482261-dc2641a1-58b0-44bb-bf4f-08ed9f210998.mov

---

 - OS: macOS Big Sur, v11.6.1
 - Browser: chrome, Version 97.0.4692.99 (Official Build) (x86_64)
 - Elastic Charts: 43.1.1

---

Kibana Cross Issues:
- https://github.com/elastic/kibana/issues/118867

nickofthyme:
## Summary

Exposes the sizing limitations for circular and bullet gauge and goal charts. Default values are same as original hard coded values to no functional changes.

```diff
interface Theme {
  ...
  goal: {
      ...
+    maxBulletSize: number;
+    maxCircularSize: number;
+    barThicknessMinSizeRatio: number;
+    baselineArcThickness: number;
+    baselineBarThickness: number;
+    marginRatio: number;
+    maxTickFontSize: number;
+    maxLabelFontSize: number;
+    maxCentralFontSize: number;
+    arcBoxSamplePitch: Radian;
+    capturePad: number;
  }
}
```

## Issues

fix #1599

### Checklist


packages/charts/api/charts.api.md
```diff
@@ -1012,6 +1012,8 @@ export interface GoalStyles {
     majorCenterLabel: Omit<TextStyle, 'padding' | 'fontSize'>;
     // (undocumented)
     majorLabel: Omit<TextStyle, 'padding' | 'fontSize'>;
+    maxBulletSize: number;
+    maxCircularSize: number;
```
markov00, line 1016: I've tested using `Infinity` and it works but there is one aspect that is not considered here:
the tick is rendered as an arc that goes half-degree before the data point and half-degree after the data point, rendering a thick tick when scaling it too much:

Screenshot 2022-03-07 at 10 38 37

https://github.com/elastic/elastic-charts/blob/f1508482c61618c9faf4c08c3348458f0fa00e8a/packages/charts/src/chart_types/goal_chart/layout/viewmodel/geoms.ts#L460-L461

There are also some other parameters to expose in the theme that can be required when the max size is increased (it should be discouraged anyway)

https://github.com/elastic/elastic-charts/blob/f1508482c61618c9faf4c08c3348458f0fa00e8a/packages/charts/src/chart_types/goal_chart/layout/viewmodel/geoms.ts#L20-L30

nickofthyme: > the tick is rendered as an arc that goes half-degree before the data point and half-degree after the data point, rendering a thick tick when scaling it too much:

True but that is an issue I see with the current implementation of the chart as a whole, not just for the extreme use case. To me We should render the ticks as is and extend the end geometry half tick past. Should I open an issue for this?

nickofthyme: > There are also some other parameters to expose in the theme that can be required when the max size is increased.

You are saying to move and expose all those constants on the `Theme`?

markov00: >To me We should render the ticks as is and extend the end geometry half tick past. Should I open an issue for this?

Ok let's open a different issue for now

>You are saying to move and expose all those constants on the Theme?

I think most of them goes hand in hand with the referenceSizeCaps variables so they should be tweaked together. @monfera to the rescue here

markov00: hey @nickofthyme, I've chatted with @monfera, we agree that we can move all these variables into the theme.

nickofthyme: See 57ef9a9

markov00 (approved):
LGTM

PR #1628, opened 2022-03-16
feat(annotations): animated focus states for hovered annotation

### LINKED ISSUE #1625: [Annotations] Reduce opacity of non-hovered annotations
Large number of annotations can clutter the visualization and could be hard to loo at the chart when there are too many of them 

Frame 10

In order to improve chart readability when interacting with annotations, it would be great to reduce opacity of annotations that are not selected/hovered (image shows a test with 0.1 opacity)

Frame 11
_[images are made purely to explain the concept]_

nickofthyme:
## Summary

`LineAnotations` now have a fade out hover/focus state to reduce opacity for non-hovered annotations. The motivation for this is to reduce the noice of surrounding annotations.

Screen Recording 2022-03-29 at 01 06 06 PM

See demo above [here](https://ech-e2e-ci--pr-1628-np4k32ms.web.app/?path=/story/annotations-rects--styling&knob-showLineAnnotations=true)

## Details

If any `LineAnotations` are actively hovered, the opacity of the hovered annotations is set to `theme.sharedStyle.highlighted` while non-hovered opacity is set to `theme.sharedStyle.unhighlighted`. If no annotations are actively hovered all opacities are set to `theme.sharedStyle.default`.

> The code for `RectAnotations` is commented out until we determine a good solution UX. When enabled there is no separation between `LineAnotations` and `RectAnotations` where the hover state would act independently between the two types of annotation. See 18b903f for reverting these changes.

It is possible to use the same animated `prop` string across multiple different render functions. However, this can be problematic when these multiple values start to diverge from each other in which case the animation logic will not be able to know which value to trust. To prevent this I built in a check to flag this as a development error if and when this occurs.

```ts
const heightA = aCtx.getValue()('bar-height', 10);
const heightB = aCtx.getValue()('bar-height', 20);
```

image

#### Stress test

See demo below [here](https://ech-e2e-ci--pr-1628-np4k32ms.web.app/?path=/story/annotations-rects--styling&globals=theme:light&knob-debug=&knob-SeriesType=line&knob-Scale%20type=linear&knob-chartRotation=0&knob-annotation%20zIndex=0&knob-rect%20border%20stroke%20width=0&knob-rect%20border%20stroke%20color=rgba(250,%200,%200,%201)&knob-fill%20color=rgba(255,%200,%200,%201)&knob-annotation%20opacity=0.5&knob-fade%20out%20on%20hover=true&knob-has%20custom%20tooltip%20render=&knob-y-domain%20axis%20is%20Position.Left=true&knob-x-domain%20axis%20is%20Position.Bottom=true&knob-hide%20tooltips=&knob-showLineAnnotations=true&knob-annotation%20count=80)

https://user-images.githubusercontent.com/19007109/160674092-6db82446-5acc-4b67-a6ae-fde64c95f27e.mp4

## Issues

close #1625

### Checklist


nickofthyme:

gvnmagni:
Thanks Nick, It looks great! If I can I would add a little comment. I'm afraid this is my fault since I wasn't probably very clear in the beginning and I apologise for this.

I just wanted to ask if we can adjust a couple of details:
- when we over the start or end of a rect annotations I would suggest to highlight the whole group (start + end + inner rect)
- same when we hover the inner rect itself, I'd suggest to highlight the entire group

In both cases what's currently shown in the tooltip is correct, the only different things would be the opacity of elements

Do you think it could work? Again, I'm sorry I haven't been clear in the first moment

markov00:
> * when we over the start or end of a rect annotations I would suggest to highlight the whole group (start + end + inner rect)

@gvnmagni we don't have the concept of groups in annotation. What was showing the the gif was just a set of ungrouped line annotation and rect annotation, all of them are a part of each other.

nickofthyme:
Yeah @gvnmagni they are separate 😄 

The deployment finally succeeded so you can take a look [here](https://ech-e2e-ci--pr-1628-np4k32ms.web.app/?path=/story/annotations-rects--styling&globals=theme:light&knob-debug=&knob-chartRotation=0&knob-annotation%20zIndex=0&knob-rect%20border%20stroke%20width=1&knob-rect%20border%20stroke%20color=#e5e5e5&knob-fill%20color=#e5e5e5&knob-annotation%20opacity=0.5&knob-has%20custom%20tooltip%20render=&knob-y-domain%20axis%20is%20Position.Left=true&knob-x-domain%20axis%20is%20Position.Bottom=true&knob-hide%20tooltips=&knob-showLineAnnotations=true) if you'd like.

packages/charts/src/utils/themes/merge_utils.ts
```diff
@@ -31,6 +31,7 @@ export const DEFAULT_ANNOTATION_RECT_STYLE: RectAnnotationStyle = {
   strokeWidth: 0,
   opacity: 0.25,
   fill: '#FFEEBC',
+  fadeOut: false,
```
nickofthyme, line 34: @markov00 Was your intent to make this `false` by default?

packages/charts/src/chart_types/xy_chart/state/selectors/get_highlighted_annotation_ids_selector.ts
```diff
@@ -0,0 +1,45 @@
...
+import { getHighlightedGeomsSelector } from './get_tooltip_values_highlighted_geoms';
+
+const getHoveredDOMElement = (state: GlobalChartState) => state.interactions.hoveredDOMElement;
+
+/** @internal */
+export const getHighlightedAnnotationIdsSelector = createCustomCachedSelector(
+  [getHoveredDOMElement, getMultipleRectangleAnnotations, getAnnotationSpecsSelector, getHighlightedGeomsSelector],
+  (hoveredDOMElement, rectAnnotationTooltips, specs, highlightedGeoms): string[] => {
+    // TODO: Remove when annotation tooltip is itegrated into main tooltip
+    if (highlightedGeoms.length > 0) return [];
```
nickofthyme, line 25: This check is to prevent annotation fading when annotation is behind an actively hovered geometry element.

nickofthyme:
Working on transitioning line fade now...

packages/charts/src/utils/themes/merge_utils.ts
markov00: I don't know actually, I think it depends on the final outcome: if, with the transition we achieved a nice and smooth fadein/out of the various elements I'm ok changing this to true.
The `fadeOut` name is also very bad :D

storybook/preview.ts
```diff
@@ -7,15 +7,9 @@
...
-import { preloadIcons } from './preload_icons';
 import { StoryWrapper } from './story_wrapper';
 
 import './style.scss';
 
-if (process.env.VRT) {
-  preloadIcons();
-  document.querySelector('html')?.classList.add('disable-animations');
-}
-
```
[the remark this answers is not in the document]
markov00: Thank you!

markov00:
@nickofthyme I've pulled the PR and did a bunch of changes to see how it looks like with a real opacity animation (with a small initial delay) and it looks great.
I'm only afraid of possible performance issues since we are doing full re-rendering of everything on every animation frame where instead we just need two layers to be rendered.
I think the whole animation function needs to be carefully crafted, but we are aiming in the right direction!


https://user-images.githubusercontent.com/1421091/160099977-01f56462-9ea4-4dfc-804d-f7cd06c8ab7e.mp4



https://user-images.githubusercontent.com/1421091/160099950-2562ba96-3af9-431e-a456-79a40c68accd.mp4

nickofthyme:
@markov00 could you review this again when you get time?

Since your review I have added...

- timing animation logic on top of delays
- timing functions to match css transition easing functions
- added the animation options to the respective style config
- removed the `fadeOut` option for rects. I wasn't sure if this logic was only cuz we didn't have a way to animate the rect annotations. If you want to keep this option to allow splitting of hover states between annotations I could add it back but the previous logic would have impacted other non animation logic.

Still need add some tests but other than that it ready. 👍🏼

packages/charts/src/chart_types/xy_chart/renderer/canvas/bars.ts
```diff
@@ -35,8 +37,17 @@ export function renderBars(
...
+        bars.forEach((barGeometry, i) => {
+          const { x, width, color, seriesStyle: style, seriesIdentifier } = barGeometry;
+          const height = aCtx.getValue({ initialValue: 0, duration: 400 })(
+            `bar-${i}-${seriesIdentifier.key}-height`,
+            barGeometry.height,
+          );
+          const y = aCtx.getValue({ initialValue: barGeometry.y + barGeometry.height, duration: 400 })(
+            `bar-${i}-${seriesIdentifier.key}-y`,
+            barGeometry.y,
+          );
```
nickofthyme: @markov00 after fiddling around with the `initialValue` logic, and solving a timing bug, I was able to animate the bar height across all bar just with just this code. Since it's rendered wrt the canvas coordinate reference it requires animation both `height` and `y` simultaneously. See 0b28444 but I reverted these changes for this pr.

Apr-06-2022 16-22-50

> granted there would need to be tweak to this logic for stacked bars but it demonstrates the idea.

packages/charts/src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
```diff
@@ -135,7 +140,7 @@ class XYChartComponent extends React.Component<XYChartProps> {
       const width = (vertical ? renderingArea.height : renderingArea.width) + CLIPPING_MARGINS * 2;
       const height = (vertical ? renderingArea.width : renderingArea.height) + CLIPPING_MARGINS * 2;
       const clippings = { x: -CLIPPING_MARGINS, y: -CLIPPING_MARGINS, width, height };
-      renderXYChartCanvas2d(this.ctx, this.devicePixelRatio, clippings, this.props);
+      renderXYChartCanvas2d(this.ctx, this.devicePixelRatio, clippings, this.props, this.animationState);
```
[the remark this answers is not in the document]
nickofthyme: See b0547f2 & 8b7dc42

packages/charts/src/chart_types/xy_chart/renderer/canvas/animations/index.ts
```diff
@@ -0,0 +1,87 @@
...
+        return {
+          getValue: getAnimatedValueFn(t),
+        };
+      }
+
+      renderFn(getAnimationContext(0));
+
+      animationState.rafId = window.requestAnimationFrame((epochStartTime) => {
+        const anim = (t: number) => {
+          const elapsed = moment.duration(t - epochStartTime).asMilliseconds();
```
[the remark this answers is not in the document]
nickofthyme: Good call I forgot about this line. 429d4d6

packages/charts/src/chart_types/xy_chart/renderer/canvas/animations/animation.ts
```diff
@@ -0,0 +1,128 @@
...
+    this.target = initial;
+    this.current = initial;
+    this.delay = typeof options?.delay === 'string' ? AnimationSpeed[options.delay] : options?.delay ?? 0;
+    this.duration = typeof options?.duration === 'string' ? AnimationSpeed[options.duration] : options?.duration ?? 0;
+    this.timeFunction = options?.timeFunction ?? TimeFunction.linear;
+    this.snapValues = options?.snapValues ?? [];
+  }
+
+  isActive(t: number) {
+    if (isNil(this.initial) || isNil(this.target) || this.initial === this.target) {
```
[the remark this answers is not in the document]
nickofthyme: 429d4d6

packages/charts/src/chart_types/xy_chart/renderer/canvas/animations/animation.ts
```diff
@@ -0,0 +1,128 @@
...
+  private delay: number; // ms
+  private duration: number; // ms
+  private timingFn: TimingFunction = () => NaN;
+
+  constructor(initial: AnimatedValue, options: AnimationOptions = {}) {
+    this.initial = initial;
+    this.target = initial;
+    this.current = initial;
+    this.delay = typeof options?.delay === 'string' ? AnimationSpeed[options.delay] : options?.delay ?? 0;
+    this.duration = typeof options?.duration === 'string' ? AnimationSpeed[options.duration] : options?.duration ?? 0;
```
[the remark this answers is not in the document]
nickofthyme: Good point but checking only `.hasOwnProperty` doesn't cast out the `number` type causing more issues. I just did this instead (see 429d4d6)...

```ts
this.delay = (typeof options?.delay === 'string' ? AnimationSpeed[options.delay] : options?.delay) ?? 0;
```

packages/charts/src/chart_types/xy_chart/annotations/tooltip.ts
```diff
@@ -73,15 +73,15 @@ export function computeMultipleRectAnnotationTooltipState(
     .filter(isRectAnnotation)
     .sort(({ zIndex: a = Number.MIN_SAFE_INTEGER }, { zIndex: b = Number.MIN_SAFE_INTEGER }) => a - b);
   return sortedAnnotationSpecs.reduce<AnnotationTooltipState[]>((acc, spec) => {
-    const annotationDimension = annotationDimensions.get(spec.id);
+    const annotationDimension = annotationDimensions.get(spec.id) as AnnotationRectProps[];
     if (!spec.hideTooltips && annotationDimension) {
```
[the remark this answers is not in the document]
nickofthyme: 429d4d6

packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts
```diff
@@ -242,3 +243,16 @@ function getOutsideDimension(style: AxisStyle): number {
...
+
+/**
+ * @internal
+ */
+export function getAnnotationRectPropsId(
+  specId: string,
+  datum: RectAnnotationDatum,
+  index: number,
+  verticalValue?: any,
+  horizontalValue?: any,
```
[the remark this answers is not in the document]
nickofthyme: 429d4d6

integration/tests/annotations_stories.test.ts
```diff
@@ -19,6 +19,41 @@ describe('Annotations stories', () => {
...
+        });
+      });
+      it('should fade all other annotations when rect annotation is hovered', async () => {
+        await common.expectChartWithMouseAtUrlToMatchScreenshot(getUrl(90), {
+          bottom: 90,
+          left: 300,
+        });
+      });
+    });
+  });
```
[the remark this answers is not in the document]
nickofthyme: I guess it's not that critical. 429d4d6

packages/charts/src/chart_types/xy_chart/renderer/canvas/bars.ts
nickofthyme: Though it does tend to struggle with the `High data volume` story.

Apr-06-2022 18-04-15

packages/charts/src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
```diff
@@ -100,6 +104,7 @@ class XYChartComponent extends React.Component<XYChartProps> {
     super(props);
     this.ctx = null;
     this.devicePixelRatio = window.devicePixelRatio;
+    this.animationState = { rafId: NaN, pool: new Map() };
```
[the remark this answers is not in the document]
nickofthyme: From our zoom call the consensus is to keep the current approach as is and possibly improve it or migrate to, or use in tandem, an event-driven animation mechanism in the future.

packages/charts/src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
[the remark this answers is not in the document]
markov00: Should be enough, thanks!

packages/charts/src/chart_types/xy_chart/renderer/canvas/bars.ts
markov00: Thanks for the try, it looks great! We will anyway limit the animation to charts with a max number of geometries to animate and we just don't animate dense chart like that one!

markov00:
>From our zoom call the consensus is to keep the current approach as is and possibly improve it or migrate to, or use in tandem, an event-driven animation mechanism in the future.

A [POC](https://github.com/elastic/elastic-charts/pull/1628#discussion_r844434775) for animated bars height is available here: https://github.com/elastic/elastic-charts/commit/0b28444cb43ad3dedcc03624444516bf0e10adca

image

markov00 (commented):
It looks great for me, I just want to wait for the final decision about the default configuration (animation on or off by default)

packages/charts/src/chart_types/xy_chart/state/selectors/get_highlighted_annotation_ids_selector.ts
```diff
@@ -20,7 +20,8 @@ const getHoveredDOMElement = (state: GlobalChartState) => state.interactions.hov
 export const getHighlightedAnnotationIdsSelector = createCustomCachedSelector(
   [getHoveredDOMElement, getMultipleRectangleAnnotations, getAnnotationSpecsSelector, getHighlightedGeomsSelector],
   (hoveredDOMElement, rectAnnotationTooltips, specs, highlightedGeoms): string[] => {
-    // TODO: Remove when annotation tooltip is itegrated into main tooltip
+    // TODO: Remove when annotation tooltip is integrated into main tooltip
+    // This check is to prevent annotation fading when annotation is behind an actively hovered geometry element
```
nickofthyme, line 24: 👍🏼

nickofthyme (commented):
Changes LGTM, thanks for the cleanup!

packages/charts/api/charts.api.md
```diff
@@ -2210,13 +2244,13 @@ export interface ShapeTreeNode extends TreeNode, SectorGeomSpecY {
 }
 
 // @public (undocumented)
-export interface SharedGeometryStateStyle {
+export interface SharedGeometryStateStyle<S extends CSSProperties = GeometryStateStyle> {
```
monfera, line 2247: nice!

packages/charts/src/chart_types/xy_chart/renderer/canvas/annotations/index.ts
```diff
@@ -26,24 +29,51 @@ interface AnnotationProps {
...
-        renderRectAnnotations(ctx, annotation as AnnotationRectProps[], rectStyle, rotation, renderingArea);
+        renderRectAnnotations(
+          ctx,
+          aCtx,
+          annotation as AnnotationRectProps[],
+          rectStyle,
+          getHoverParams,
+          rotation,
+          renderingArea,
+        );
```
monfera, line 76: PR-unrelated nit: the `if` may not be necessary, as only two things are conditionalized: the 4th argument (style) and the render function itself. So it could be 
```
const render = isLine ? renderLineAnnotations : renderRectAnnotations;
const styler = isLine ? mergeWithDefaultAnnotationLine : mergeWithDefaultAnnotationRect;
render(ctx, aCtxi, annotation, styler(spec.style), getHoverParams, rotation, renderingArea);
```

packages/charts/src/chart_types/xy_chart/renderer/canvas/animations/animation.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  private setTimingFn() {
+    const scalar = this.target - this.initial;
+    this.timingFn = (t) => {
+      const multiplier = TimingFunctions[this.timeFunction](t);
+      return this.initial + scalar * multiplier;
+    };
+  }
+
+  getValue(t: TimeMs): AnimatedValue {
```
monfera, line 95: Could use a more informative name at the expense of little added verbosity, like `valueAtTime` (no need for `get`)

packages/charts/src/chart_types/xy_chart/renderer/canvas/animations/animation.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  isActive(t: TimeMs) {
+    if (!isFiniteNumber(this.initial) || !isFiniteNumber(this.initial) || this.initial === this.target) {
+      return false;
+    }
+
+    return t - this.delay < this.duration;
+  }
+
+  next(value: AnimatedValue) {
```
monfera, line 76: Could use a more telling name and/or method doc

monfera:
It's great that a demo is linked, and that most of the implementation detail of the animations is placed in its special area. 

It doesn't seem to have any `xy` specific thing in the animation engine. It also don't seem to depend on anything `xy`, which is super nice 💪 🎉  There seems to be no obstacle to moving the `animation` files outside `xy` or even, `charts`

The current demo makes me wish for these styling changes (some may just be due to example values):

1. The demo tweening is too slow, something like 150-250ms should be a maximum length for the duration+delay summed; being slow makes the feature harder to discover as well
2. The reversal of the animation (mouse leave) is sudden, it's unpleasant to use it due to the stark contrast to the tweening
3. As there are only a couple of annotations, I couldn't test the behavior of rapid hovering over a bunch of items, could you add a knob that adds like several dozens of them randomly?
4. (may have a reason) Hovering over a rectangle leads to different strokes on the left vs right of the rectangle

We discussed that the LERP would be a more appropriate animation model for this use case than tweening, A bunch of union types and indirections may emerge when LERP is added, if we even get around to doing it (priorities). Also, userland code is sticky. Once it's in Kibana, it'll be harder to switch; BREAKING change, user habituation etc. Wouldn't it be easier to switch to LERP now than later? Has someone got a preference, or future use case for tweening? If it's about implementation simplicity: LERP is a bit simpler. If it's about doing some animation as a first step: that's done, and it's easier to add/change for the better UX now than later. Happy to chat about LERP vs tweening on a call

nickofthyme:
> It doesn't seem to have any `xy` specific thing in the animation engine. It also don't seem to depend on anything `xy`, which is super nice 💪 🎉

❤️ Exactly! I thought about replacing the 🔥 / 🧊 drilldown animation but thought it best for another PR, but it is very much possible to replace that logic.

> There seems to be no obstacle to moving the animation files outside xy or even, charts

Yup we could definitely move it to the top and use it anywhere. Probably best to do that in another PR.

> 1. The demo tweening is too slow, something like 150-250ms should be a maximum length for the duration+delay summed; being slow makes the feature harder to discover as well

Happy to reduce this

> 2. The reversal of the animation (mouse leave) is sudden, it's unpleasant to use it due to the stark contrast to the tweening

Yeah this was discussed during the weekly to have it _snap_ back when the value in un-highlighted. But this can also change to transition both value changes. You can change this in the demo by setting the `snap values` to empty, see [link](https://ech-e2e-ci--pr-1628-np4k32ms.web.app/?path=/story/annotations-rects--styling&globals=theme:light&knob-enabled_Animations=true&knob-delay%20(ms)_Animations=300&knob-duration%20(ms)_Animations=300&knob-time%20function_Animations=linear&knob-debug=&knob-SeriesType=line&knob-Scale%20type=linear&knob-chartRotation=0&knob-annotation%20zIndex_Styles=0&knob-rect%20border%20stroke%20width_Styles=1&knob-rect%20border%20stroke%20color_Styles=#e5e5e5&knob-fill%20color_Styles=#e5e5e5&knob-annotation%20opacity_Styles=0.5&knob-has%20custom%20tooltip%20render_Styles=&knob-y-domain%20axis%20is%20Position.Left_Styles=true&knob-x-domain%20axis%20is%20Position.Bottom_Styles=true&knob-hide%20tooltips_Styles=&knob-showLineAnnotations_Styles=true&knob-annotation%20count_Styles=6).

Screen Recording 2022-04-13 at 08 10 40 AM


> 3. As there are only a couple of annotations, I couldn't test the behavior of rapid hovering over a bunch of items, could you add a knob that adds like several dozens of them randomly?

There is a knob on the demo to increase the annotation count. See [link](https://ech-e2e-ci--pr-1628-np4k32ms.web.app/?path=/story/annotations-rects--styling&globals=theme:light&knob-enabled_Animations=true&knob-delay%20(ms)_Animations=300&knob-duration%20(ms)_Animations=300&knob-time%20function_Animations=linear&knob-snap%20values_Animations[0]=1&knob-debug=&knob-SeriesType=line&knob-Scale%20type=linear&knob-chartRotation=0&knob-annotation%20zIndex_Styles=0&knob-rect%20border%20stroke%20width_Styles=1&knob-rect%20border%20stroke%20color_Styles=#e5e5e5&knob-fill%20color_Styles=#e5e5e5&knob-annotation%20opacity_Styles=0.5&knob-has%20custom%20tooltip%20render_Styles=&knob-y-domain%20axis%20is%20Position.Left_Styles=true&knob-x-domain%20axis%20is%20Position.Bottom_Styles=true&knob-hide%20tooltips_Styles=&knob-showLineAnnotations_Styles=true&knob-annotation%20count_Styles=80).

Screen Recording 2022-04-13 at 08 08 57 AM


4. (may have a reason) Hovering over a rectangle leads to different strokes on the left vs right of the rectangle

This is caused by the tooltip snapping to the closest value on either side of the mouse position.

Screen Recording 2022-04-13 at 08 12 33 AM

packages/charts/src/chart_types/xy_chart/renderer/canvas/annotations/index.ts
nickofthyme: Agreed, though the `annotations` would need to be cast differently.

packages/charts/src/chart_types/xy_chart/renderer/canvas/animations/animation.ts
nickofthyme: d012145

packages/charts/src/chart_types/xy_chart/renderer/canvas/animations/animation.ts
nickofthyme: d012145

nickofthyme:
Based on the original logic, whenever an actively transitioning animation is interrupted, the current value is captured and used as the initial value on the next transition. This could be problematic because this warps the timing function of the next transition. For example, if we have a timing curve to transition opacity between `1` and `0`, `500ms` delay and `2000ms` duration. The current approach would look something like the chart below, where the dashed line is the complete timing curve if left uninterrupted.

Animation interupt - Google Sheets 2022-04-15 at 10 14 56 AM

Notice by taking the current at the time of interrupt as the next initial, the timing function is much shallower than desired. This also included the delay of `500ms` following the interruption.

In a perfect world the transition is uninterrupted on the way up and down which would produce a time curve that would look something like this...

Animation interupt - Google Sheets 2022-04-15 at 10 14 17 AM

The ideal scenario would be to still use the current value as the next initial value but use the shape of the time curve as if the value was never interrupted. To do so we capture the current value and the previous target value, we set the previous target the the new initial, as if it completed a full transition, which creates the correct time curve shape. Then we use the current value to determine the required time offset for the curve to align properly, this time without a delay. So this new approach could have a time curve that looks something like that below, where the dash line indicate the full uninterrupted shape of the time curves.

Animation interupt - Google Sheets 2022-04-15 at 10 15 15 AM

So visually this is a little harder to see but a demo of this would be to hover over a `LineAnnotation` marker and watch the `RectAnnotation` opacity start at `1` the start moving towards `0.25`, get interrupted and go back to `1` along the proper time curve.

Screen Recording 2022-04-15 at 08 30 28 AM

> Notice the opacity values in the console

monfera:
Regarding the last tween chart, it'd be great to smooth (avoid the jump discontinuity of the first derivative and a ) by doing something like this, in this PR or a future one:
image

monfera:
Thanks @nickofthyme I just checked the demo again for look and feel, it's a vast improvement, seems slick and polished 💅

packages/charts/src/chart_types/xy_chart/renderer/canvas/animations/animation.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+    // if (this.debug && count++ % 30 === 0) console.log(t, this.delay, this.duration, this.timeOffset);
+
+    return t - this.delay + this.timeOffset < this.duration;
+  }
+
+  /**
+   * Sets the target value to trigger tweening of value
+   */
+  setTarget(value: AnimatedValue) {
```
nickofthyme, line 88: This is the old `next` method.

nickofthyme:
> Regarding the last tween chart, it'd be great to smooth (avoid the jump discontinuity of the first derivative and a ) by doing something like this, in this PR or a future one: image

Agreed, this could be a good future enhancement, the discontinuity is hardly perceivable but would be a better curve for sure.

markov00 (approved):
Looks great! thanks Nick for all the improvements

nickofthyme:
> Looks great! thanks Nick for all the improvements. We should write some documentation on how animation works so we don't lose all this

Added 34a099f

nickofthyme:
@markov00 did we decide on whether to enable this by default?

markov00:
> @markov00 did we decide on whether to enable this by default?

Let's turn it on by default, but when making a PR in Kibana let's find every place using LineAnnotation or RectAnnotation, let's add the animation style with the `enable: true` so that everybody can know how to turn it off if they don't want it.

monfera (commented):
👯‍♀️

PR #1637, opened 2022-03-31
feat(goal): auto generated linear ticks

### LINKED ISSUE #1238: Auto-compute nice ticks for linear scales based on range
related to TSVB/Visualize gauge replacement

The ability to automatically compute a set of nice rounded ticks based on a range when using a linear bullet graph with no range bands

from original the issue: e.x. if `ticks` prop is not defined to detect automatically based on an interval (currently it renders default ticks from 0 to 100)
  
  image

Using a manual configuration for the ticks, in this case, can be tedious and error-prone.

**Additional context**
related to TSVB/Visualize gauge replacement

nickofthyme:
## Summary

Adds automated linear ticks for goal/gauge charts when to `ticks` value is provided. Adds a new `nice` prop to force ticks to be _niced_ given the specified domain. The autogenerated ticks are independent of the values provided for `bands`.

### Before

When no `ticks` prop is provided we just default to this...

Image 2022-03-30 at 10 05 25 PM

### After
Now whenever `ticks` prop is not defined when compute the sensible linear ticks.

image

## Details

This PR adds linear ticks generation from the provided domain via a `ScaleContinuous`.

See demo story [here](https://ech-e2e-ci--pr-1637-rqyolfud.web.app/?path=/story/goal-alpha--auto-linear-ticks&globals=theme:light)

## Issues

closes #1238

### Checklist


packages/charts/src/chart_types/goal_chart/layout/viewmodel/viewmodel.ts
```diff
@@ -38,11 +38,24 @@ export function shapeViewModel(spec: GoalSpec, theme: Theme, chartDimensions: Di
...
-    ([min, max], value) => [Math.min(min, value), Math.max(max, value)],
-    [Infinity, -Infinity],
-  );
+  const [lowestValue, highestValue] = [
+    base,
+    ...(target ? [target] : []),
+    actual,
+    ...bands,
+    ...(spec.ticks ?? []),
+  ].reduce(([min, max], value) => [Math.min(min, value), Math.max(max, value)], [Infinity, -Infinity]);
```
markov00, line 49: You can use a function called `extent` in `common/math`

packages/charts/src/chart_types/goal_chart/layout/viewmodel/viewmodel.ts
```diff
@@ -38,11 +38,24 @@ export function shapeViewModel(spec: GoalSpec, theme: Theme, chartDimensions: Di
...
+
+  const ticks =
+    spec.ticks ??
+    new ScaleContinuous({
+      type: 'linear', // TODO allow other scale types
+      domain: [lowestValue, highestValue],
+      range: [0, chartDimensions.width], // TODO apply range based on chart rotation
+      nice,
```
markov00, line 57: Nicing can't be applied only for ticks. Nicing changes the domain of the chart increasing it to the next nice/rounded value.

I think we can't introduce that feature if we don't also change the domain used by the chart scale.

packages/charts/src/chart_types/goal_chart/layout/viewmodel/viewmodel.ts
```diff
@@ -38,11 +38,24 @@ export function shapeViewModel(spec: GoalSpec, theme: Theme, chartDimensions: Di
     bandLabels,
     angleStart,
     angleEnd,
+    nice,
   } = spec;
-  const [lowestValue, highestValue] = [base, ...(target ? [target] : []), actual, ...bands, ...ticks].reduce(
-    ([min, max], value) => [Math.min(min, value), Math.max(max, value)],
-    [Infinity, -Infinity],
-  );
+  const [lowestValue, highestValue] = [
```
markov00, line 43: this was used before to compute the domain min and max values. But now, adding the automatic tick generation a user can't use the bullet graph without bands (the minimal version of it) because the band is used to define a maximum on the scale.
I think we need to rethink the domain definition and introduce a real `domain` prop that will describe the `min, max` values to be used in the chart scale.
The rest should stay:
- the `base` should stay the same (I can have a negative-to-positive bullet with a base on 0)
- the `target` should stay the same (it's a comparative metric that should be within the bullet domain)
- both `bands` and `ticks` can be auto-generated if needed.a

Optional property can be `domainClip: boolean` that can describe the fact that the `actual` value can be clipped within the domain ranges (the bar should not exceed the min, max domain values) or it can overflow the domain values (and in this case the domain can be: extended with that actual value, or we can find an alternative way to describe the fact that value exceed the domain max or min)
@monfera to the rescue

packages/charts/src/chart_types/goal_chart/layout/viewmodel/viewmodel.ts
```diff
@@ -38,11 +38,24 @@ export function shapeViewModel(spec: GoalSpec, theme: Theme, chartDimensions: Di
...
+    ...bands,
+    ...(spec.ticks ?? []),
+  ].reduce(([min, max], value) => [Math.min(min, value), Math.max(max, value)], [Infinity, -Infinity]);
+
+  const ticks =
+    spec.ticks ??
+    new ScaleContinuous({
+      type: 'linear', // TODO allow other scale types
+      domain: [lowestValue, highestValue],
+      range: [0, chartDimensions.width], // TODO apply range based on chart rotation
```
markov00, line 56: we don't need a real range here, `[0,1]` is good enough

packages/charts/src/chart_types/goal_chart/layout/viewmodel/viewmodel.ts
```diff
@@ -38,11 +38,24 @@ export function shapeViewModel(spec: GoalSpec, theme: Theme, chartDimensions: Di
...
+    base,
+    ...(target ? [target] : []),
+    actual,
+    ...bands,
+    ...(spec.ticks ?? []),
+  ].reduce(([min, max], value) => [Math.min(min, value), Math.max(max, value)], [Infinity, -Infinity]);
+
+  const ticks =
+    spec.ticks ??
+    new ScaleContinuous({
```
markov00, line 53: The default number of desired ticks is 10, this value is actually too much for a standard bullet/gauge.
We can configure it to something around 3 max.
This needs to be applied until we don't have the mechanism to reduce the number of ticks when they overlap.

markov00 (changes requested):

packages/charts/src/chart_types/goal_chart/layout/viewmodel/viewmodel.ts
nickofthyme: reverted in 4a3d090

packages/charts/src/chart_types/goal_chart/layout/viewmodel/viewmodel.ts
nickofthyme: 4a3d090

packages/charts/src/chart_types/goal_chart/layout/viewmodel/viewmodel.ts
nickofthyme: f1857d2

packages/charts/src/chart_types/goal_chart/layout/viewmodel/viewmodel.ts
nickofthyme: Ok sounds good. I'm not sure how we should compute the `bands` if they don't supply them. But if they _do_ supply a custom band values we should still use them in the `extents` calculations right?

markov00: I think also `bands` could be marked as `optional`.
In that way we can have:
| `ticks` prop | `bands` prop | ticks | bands |
| ---- | ---- |  ---- |  ---- | 
| undefined | undefined | generated | use bands computed from generated ticks |
| undefined | [] | generated | no band rendered |
| [] | undefined | no ticks | generate bands similar to tick generation | 
| [] | [] | no ticks | no bands |

> But if they do supply a custom band values we should still use them in the extents calculations right?

I think the `extent` should be always limited within the `min,max` of the `domain` prop

packages/charts/src/chart_types/goal_chart/layout/viewmodel/viewmodel.ts
markov00: nice dynamic ticks calculation nick!

yea, using `ticks: number | number[]` will probably work. Robert is not a fan of union type, but I see more confusion having that as a separate prop:
```
ticks?: []
desiredTickCount?: number
```
It is less clear how the two interact with each other then the union type proposal

packages/charts/src/chart_types/goal_chart/layout/viewmodel/viewmodel.ts
markov00: I've just read the next comment about the `ticks: number | number[]` and the approach can probably be attached to `bands` too

nickofthyme: OK I'll take a stab at it and we can tweak it from there.

So should the `domain` be required? Or should we find the extents from the `base`, `target` and `actual` only when the `domain` is not provided?

packages/charts/src/chart_types/goal_chart/layout/viewmodel/viewmodel.ts
nickofthyme: Ok I added this in ed843b6

packages/charts/src/chart_types/goal_chart/layout/viewmodel/viewmodel.ts
nickofthyme: Ok I took a stab at this in ed843b6. See the deployed story [here](https://ech-e2e-ci--pr-1637-rqyolfud.web.app/?path=/story/goal-alpha--auto-linear-ticks).

| Criteria  | Example |
| - | - |
| No `domain.min` nor `domain.max`, uses computed extents `min` and/or `max`. Computed `min` here contolled by `base` value of `100` and the `max` is controlled by the `target` of `260`. Both `bands` and `ticks` are `undefined` and autogenerated. See [example](https://ech-e2e-ci--pr-1637-rqyolfud.web.app/?path=/story/goal-alpha--auto-linear-ticks&globals=theme:light&knob-actual=200&knob-angleEnd%20(%CF%80)=-0.25&knob-angleStart%20(%CF%80)=1.25&knob-auto%20domain%20max=true&knob-auto%20domain%20min=true&knob-auto%20generate%20bands=true&knob-auto%20generate%20ticks=true&knob-base=100&knob-target=260&knob-reverse=&knob-domain%20min=0&knob-domain%20max=400&knob-bands[0]=200&knob-bands[1]=250&knob-bands[2]=300). |  |
| `domain.min = 0` and `domain.max = 400`, uses specified `min` and `max`. Both `bands` and `ticks` are `undefined` and autogenerated. See [example](https://ech-e2e-ci--pr-1637-rqyolfud.web.app/?path=/story/goal-alpha--auto-linear-ticks&globals=theme:light&knob-actual=200&knob-angleEnd%20(%CF%80)=-0.25&knob-angleStart%20(%CF%80)=1.25&knob-auto%20domain%20max=&knob-auto%20domain%20min=&knob-auto%20generate%20bands=true&knob-auto%20generate%20ticks=true&knob-base=100&knob-target=260&knob-reverse=&knob-domain%20min=0&knob-domain%20max=400&knob-bands[0]=200&knob-bands[1]=250&knob-bands[2]=300). |  |
| The `bands` are set to `[0, 200,250,300]`. The `ticks` are auto computed driven by the `bands` domain. See [example](https://ech-e2e-ci--pr-1637-rqyolfud.web.app/?path=/story/goal-alpha--auto-linear-ticks&globals=theme:light&knob-actual=200&knob-angleEnd%20(%CF%80)=-0.25&knob-angleStart%20(%CF%80)=1.25&knob-auto%20domain%20max=true&knob-auto%20domain%20min=true&knob-auto%20generate%20bands=&knob-auto%20generate%20ticks=true&knob-base=100&knob-target=260&knob-reverse=&knob-domain%20min=0&knob-domain%20max=400&knob-bands[0]=0&knob-bands[1]=%20200&knob-bands[2]=250&knob-bands[3]=300). |  |
| The `ticks` are set to `[0, 100, 200, 300]`. The `bands` are auto computed driven by the `ticks` domain. See [example](https://ech-e2e-ci--pr-1637-rqyolfud.web.app/?path=/story/goal-alpha--auto-linear-ticks&globals=theme:light&knob-actual=200&knob-angleEnd%20(%CF%80)=-0.25&knob-angleStart%20(%CF%80)=1.25&knob-auto%20domain%20max=true&knob-auto%20domain%20min=true&knob-auto%20generate%20bands=true&knob-auto%20generate%20ticks=&knob-base=100&knob-target=260&knob-reverse=&knob-domain%20min=0&knob-domain%20max=400&knob-ticks[0]=0&knob-ticks[1]=%20100&knob-ticks[2]=%20200&knob-ticks[3]=%20300). |  |
| The `bands` are set to `[]`. The `ticks` are auto computed driven by the `base` and `target`. No bands are rendered. See [example](https://ech-e2e-ci--pr-1637-rqyolfud.web.app/?path=/story/goal-alpha--auto-linear-ticks&globals=theme:light&knob-actual=200&knob-angleEnd%20(%CF%80)=-0.25&knob-angleStart%20(%CF%80)=1.25&knob-auto%20domain%20max=true&knob-auto%20domain%20min=true&knob-auto%20generate%20bands=&knob-auto%20generate%20ticks=true&knob-base=100&knob-target=260&knob-reverse=&knob-domain%20min=0&knob-domain%20max=400&knob-ticks[0]=0&knob-ticks[1]=%20100&knob-ticks[2]=%20200&knob-ticks[3]=%20300). |  |
| When `domain.min` and `domain.max` are specified, in this case `domain = { min: 0, max: 300 }`. The values outside of the explicit domain are filtered from `ticks` and `bands` values. With `ticks` set to `[-100, 0, 200, 300, 400]` and `bands` set to `[-100, 0, 200, 300, 400]`, both the `-100` and `400` values will be excluded from the rendering for both `ticks` and `bands`. See [example](https://ech-e2e-ci--pr-1637-rqyolfud.web.app/?path=/story/goal-alpha--auto-linear-ticks&globals=theme:light&knob-actual=200&knob-angleEnd%20(%CF%80)=-0.25&knob-angleStart%20(%CF%80)=1.25&knob-auto%20domain%20max=&knob-auto%20domain%20min=&knob-auto%20generate%20bands=&knob-auto%20generate%20ticks=&knob-base=100&knob-target=260&knob-reverse=&knob-domain%20min=0&knob-domain%20max=300&knob-bands[0]=-100&knob-bands[1]=%200&knob-bands[2]=%20200&knob-bands[3]=%20300&knob-bands[4]=%20400&knob-ticks[0]=-100&knob-ticks[1]=%200&knob-ticks[2]=%20100&knob-ticks[3]=%20200&knob-ticks[4]=%20300&knob-ticks[5]=%20400). |  |

@markov00 Let me know if you think any these cases should have a different behavior.

markov00: To summarize what we decided:
1) the `domain` is mandatory with a min and max, both of them finite numbers (if non-finite, we can consider 0 as min and 1 as max).
2) an element in `ticks` that falls outside the domain is filtered out
3) an element in `bands` that falls outside the domain is filtered out
4) auto-ticks and auto-bands are computed based on the `domain`
5) `base`, `target` and `actual` are limited within the domain. If they fall outside, we still use these values but we limit the rendering within the domain. In particular `target` will not be rendered at all, `base` will be clamped to the min value of the domain and the `actual` value will be rendered only within the domain, with its extremes clamped to the min/max of the domain.

packages/charts/src/chart_types/goal_chart/specs/index.ts
```diff
@@ -45,8 +45,18 @@ export interface GoalSpec extends Spec {
...
   target?: number;
   actual: number;
-  bands: number[];
-  ticks: number[];
+  /**
+   * array of descrete band intervals or aproximate number of desired bands
+   */
+  bands?: number | number[];
+  /**
+   * array of descrete tick values or aproximate number of desired ticks
```
markov00, line 53:
```suggestion
   * Array of discrete tick values or approximate number of desired ticks
```

packages/charts/src/chart_types/goal_chart/specs/index.ts
```diff
@@ -45,8 +45,18 @@ export interface GoalSpec extends Spec {
   base: number;
   target?: number;
   actual: number;
-  bands: number[];
-  ticks: number[];
+  /**
+   * array of descrete band intervals or aproximate number of desired bands
```
markov00, line 49:
```suggestion
   * array of discrete band intervals or approximate number of desired bands
```

packages/charts/src/utils/common.ts
```diff
@@ -645,3 +645,9 @@ export function stripUndefined<R extends Record<string, unknown>>(source: R): R
     return acc;
   }, {} as R);
 }
+
+/**
+ * Returns filter callback for values between a min and max
+ * @internal
+ */
+export const isBetween = (min: number, max: number) => (n: number) => n <= max && n >= min;
```
markov00, line 653: is not clear from the name nor the docs if the value is included in the range or excluded.
A possible suggestion can be:
```
isValueInInterval = (min, max, closed = true) => (n: number) => closed ? min >= n && n <= max : min > n && n < max;
```

storybook/stories/goal/26_auto_linear_ticks.story.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  const angleStart = start * Math.PI;
+  const angleEnd = end * Math.PI;
+
+  return (
+    <Chart>
+      <Settings baseTheme={useBaseTheme()} />
+      <Goal
+        id="spec"
+        subtype={subtype}
```
markov00, line 44: Can we make this dynamic and use bullet by default? I think we should push more for bullet when possible

storybook/stories/goal/26_auto_linear_ticks.story.tsx
```diff
@@ -0,0 +1,66 @@
...
+        actual={actual}
+        ticks={ticks}
+        bands={bands}
+        domain={{ min, max }}
+        angleStart={reverse ? angleEnd : angleStart}
+        angleEnd={reverse ? angleStart : angleEnd}
+        tickValueFormatter={({ value }: BandFillColorAccessorInput) => String(value)}
+        labelMajor=""
+        labelMinor=""
+        centralMajor="280 MB/s"
```
markov00, line 56: this should be dynamic and linked to the actual value

packages/charts/src/utils/common.ts
nickofthyme: ec2a590

storybook/stories/goal/26_auto_linear_ticks.story.tsx
nickofthyme: 6e23e03

storybook/stories/goal/26_auto_linear_ticks.story.tsx
nickofthyme: 6e23e03

packages/charts/src/chart_types/goal_chart/layout/viewmodel/viewmodel.ts
nickofthyme: Ok I think I captured all the points above in e9dc2f1, thanks for the summary 🙏. 

> 4. auto-ticks and auto-bands are computed based on the domain

Just to be clear, the ticks/bands are only computed from the domain when left `undefined` right?

packages/charts/src/chart_types/goal_chart/specs/index.ts
```diff
@@ -37,6 +37,12 @@ export type BandFillColorAccessor = (input: BandFillColorAccessorInput) => Color
 /** @alpha */
 export type GoalLabelAccessor = LabelAccessor<BandFillColorAccessorInput>;
 
+/** @alpha */
+export interface GoalDomainRange {
+  min: number;
+  max: number;
```
markov00, line 43: Can you please specify, in a comment that the min and max should be finite numbers?

packages/charts/src/chart_types/goal_chart/specs/index.ts
```diff
@@ -45,8 +51,18 @@ export interface GoalSpec extends Spec {
...
+  /**
+   * array of discrete band intervals or approximate number of desired bands
+   */
+  bands?: number | number[];
+  /**
+   * Array of discrete tick values or approximate number of desired ticks
+   */
+  ticks?: number | number[];
+  /**
+   * Domain extents of goal chart. Overrides computed extents.
```
markov00, line 63: We don't compute an extent, but we take the domain for granted.
We can say that the chart limits every other value to this domain

storybook/stories/goal/26_auto_linear_ticks.story.tsx
```diff
@@ -0,0 +1,65 @@
...
+        labelMinor={subtype === GoalSubtype.Goal ? '' : `${actual} MB/s`}
+        centralMajor={`${actual} MB/s`}
+        centralMinor=""
+      />
+    </Chart>
+  );
+};
+
+Example.parameters = {
+  markdown: `Leaving \`ticks\` and/or \`bands\` as \`undefined\` will automatically generate linear values given the computed or specified domain.
```
markov00, line 63:
```suggestion
  markdown: `Leaving \`ticks\` and/or \`bands\` as \`undefined\` will automatically generate linear values given the specified domain.
```

markov00 (approved):
Everything looks good now, thanks for all the updates.
I've put some minor comments about the docs/tsdocs, everything else is great!

packages/charts/src/chart_types/goal_chart/layout/viewmodel/viewmodel.ts
markov00: >the ticks/bands are only computed from the domain when left undefined right?

yes, when `ticks` and/or `bands` are `undefined` we want to automatically compute the ticks for them based on the provided domain.

storybook/stories/goal/26_auto_linear_ticks.story.tsx
nickofthyme: 94393c4

packages/charts/src/chart_types/goal_chart/specs/index.ts
nickofthyme: 94393c4

PR #1645, opened 2022-04-12
fix(axis): ticks generation for linear scale on bar charts

markov00:
## Summary

We introduced a logic into tick generation to avoid generating denser ticks when rendering bar charts. It also reduce the number of ticks generated in this case, rendering only a set of ticks at nicer rounded intervals.

See screenshot changes for a comparison of what has changed.

Replace #1435

## Details

I have introduced a tick rarefication logic in the continuous scale that works only for bar charts. Nothing has changed for line/area as suggested by @monfera https://github.com/elastic/elastic-charts/pull/1435#issuecomment-951326514

**TO REVIEWERS**: do you consider this as a breaking change?

### Checklist


packages/charts/src/scales/scale_continuous.ts
```diff
@@ -148,22 +148,22 @@ export class ScaleContinuous implements Scale<number> {
...
+            scaleOptions.timeZone,
+            scaleOptions.bandwidth === 0 ? 0 : scaleOptions.minInterval,
+          )
+        : type === ScaleType.Linear
+        ? getLinearNonDenserTicks(
+            nicePaddedDomain,
+            scaleOptions.desiredTickCount,
+            this.linearBase,
+            scaleOptions.bandwidth === 0 ? 0 : scaleOptions.minInterval,
+          )
```
nickofthyme, line 164: So we always want this to be applied, correct? I think that's fine so long as we considering the `minInterval`.

packages/charts/src/scales/scale_continuous.ts
```diff
@@ -260,6 +266,23 @@ function getTimeTicks(desiredTickCount: number, timeZone: string, domain: number
...
+  domain: number[],
+  desiredTickCount: number,
+  base: number = 2,
+  minInterval: number,
+): number[] {
+  const start = domain[0];
+  const stop = domain[domain.length - 1];
+  let currentCount = desiredTickCount;
+  let ticks = getLinearTicks(start, stop, desiredTickCount, base);
+  while (ticks.length > 2 && currentCount > 0 && ticks[1] - ticks[0] < minInterval) {
```
nickofthyme, line 279: 😱 `while...`

nickofthyme (approved):

packages/charts/src/chart_types/xy_chart/utils/get_linear_ticks.ts
```diff
@@ -30,7 +30,7 @@ const e5 = Math.sqrt(10);
 const e2 = Math.sqrt(2);
```
nickofthyme, line 30: > do you consider this as a breaking change

I would say no on the breaking change, that said the next release will be breaking anyway from https://github.com/elastic/elastic-charts/commit/5437d8e985ecc8cfa32651ccc97fdf7b9da84b01.

packages/charts/src/scales/scale_continuous.ts
markov00: Yep, the idea is to always apply that for linear scale where we points have "bandwidth", meaning linear scale for bar charts. the check `scaleOptions.bandwidth === 0 ? 0 : scaleOptions.minInterval` actually make the call return with the desired tick count if we are on a line/area chart, and instead iterate the reduction for bar charts

packages/charts/src/scales/scale_continuous.ts
markov00: I know, do you prefer a recursive function? :D

PR #1646, opened 2022-04-12
fix(axis): use correct desired tick count based on axis type

markov00:
## Summary

Fix the calculation of the max desired tick count for the X axis.

From this, on master, with the X axis ticks set to `ticks={3}`
Screenshot 2022-04-12 at 18 37 53

To
Screenshot 2022-04-12 at 18 37 42

## Details

Break out from https://github.com/elastic/elastic-charts/pull/1435

### Checklist


packages/charts/src/chart_types/xy_chart/state/selectors/get_api_scale_configs.ts
```diff
@@ -54,19 +53,17 @@ export function getScaleConfigsFromSpecs(
...
-  const isHorizontalChart = isHorizontalRotation(settingsSpec.rotation);
-
   // x axis
-  const xAxes = axisSpecs.filter((d) => isHorizontalChart === isHorizontalAxis(d.position));
-  const xTicks = xAxes.reduce<number>((acc, { ticks = X_SCALE_DEFAULT.desiredTickCount }) => {
+  const xAxesSpecs = axisSpecs.filter((spec) => isXDomain(spec.position, settingsSpec.rotation));
+  const maxTickCountForXAxes = xAxesSpecs.reduce<number>((acc, { ticks = X_SCALE_DEFAULT.desiredTickCount }) => {
     return Math.max(acc, ticks);
-  }, X_SCALE_DEFAULT.desiredTickCount);
+  }, -Infinity);
```
nickofthyme, line 60:
```suggestion
  }, Number.MIN_SAFE_INTEGER);
```

Didn't @monfera say something about using `Number.MIN_SAFE_INTEGER` over `Infinity`?

packages/charts/src/chart_types/xy_chart/state/selectors/get_api_scale_configs.ts
```diff
@@ -80,16 +77,19 @@ export function getScaleConfigsFromSpecs(
...
   const y = Object.keys(scaleConfigsByGroupId).reduce<ScaleConfigs['y']>((acc, groupId) => {
-    const axis = yAxes.find((yAxis) => yAxis.groupId === groupId);
-    const desiredTickCount = axis?.ticks ?? Y_SCALE_DEFAULT.desiredTickCount;
-    const scaleConfig = scaleConfigsByGroupId[groupId];
-    const customDomain = customDomainByGroupId.get(groupId);
+    const maxTickCountYAxes = yAxisSpecs.reduce<number>((maxTickCount, yAxis) => {
+      return yAxis.groupId === groupId
+        ? Math.max(maxTickCount, yAxis.ticks ?? Y_SCALE_DEFAULT.desiredTickCount)
+        : maxTickCount;
+    }, -Infinity);
```
nickofthyme, line 86:
```suggestion
    }, Number.MIN_SAFE_INTEGER);
```

Same here

nickofthyme (approved):

packages/charts/src/chart_types/xy_chart/state/selectors/get_api_scale_configs.ts
markov00: I think he was referring to the usage of non finite number when sorting:
```
const BIG_NUMBER = Number.MAX_SAFE_INTEGER; // the sort comparator must yield finite results, can't use infinities

```

markov00: I think we could also simply put a `-1` there and check for a positive value when assigning the value to the ScaleConfig

nickofthyme: Cool, either way is fine.

PR #1649, opened 2022-04-14
fix(deps): update typescript to 4.6.3

markov00:
## Summary

The PR updates Typescript dependencies to the latest version and updated a bunch of other related dependencies to reduce the number of ts versions on our yarn.lock file

### Checklist


storybook/package.json
```diff
@@ -2,15 +2,15 @@
...
+  "version": "0.0.0",
+  "private": "true",
   "scripts": {
     "start": "VRT=true start-storybook -s ../public -p 9001 -c . --ci --no-version-updates",
     "build": "rm -rf ../.out && build-storybook -s ../public -c . -o ../.out",
     "build:firebase": "build-storybook -s ../public -c . -o ../e2e-server/public",
     "typecheck": "tsc -p ./tsconfig.json --noEmit"
   },
   "dependencies": {
-    "@elastic/charts": "^30.2.0",
```
nickofthyme, line 13: Removing this will always error on the [`no-unresolved`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-unresolved.md) eslint rule. I think the best approach is either use a [local path](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#local-paths) or just `"*"`. I think `"*"` would be easiest cuz you might have issues with the local path.

I want to make this nicer at some point where each workspace functions on its own.

nickofthyme (approved):
Just the one comment on fixing the linting errors.

storybook/package.json
markov00: if we use `*` the dependency needs to come from somewhere, instead, we don't want to download a dependency, we just want to link this dependency to the source code coming from the workspaces.
It is not an issue with eslint actually. I was able to use this PR nicely in my local environment (after cleaning also the node_modules to be entirely sure about the effect). Is seems that the CI instead can't properly pick up the jest alias for the dependency.

Issue #1365, opened 2021-09-08
Accessibility: more legible axis tick labels and other textual chart elements


monfera:
**Is your feature request related to a problem? Please describe.**
Increase font legibility and disambiguation in Charts examples and Kibana, by default. Problems:
- too small font size for axis tick labels in practice (in Kibana and in Charts examples)
- zeros are not crossed; default fonts (in Kibana and in Charts examples) with their default settings are not as suitable for data visualization in our typically technical domain
- not enough contrast with the default grey, and I haven't encountered justification for NOT upping axis tick label legibility via maximum contrast

**Describe the solution you'd like**

- larger fonts as defaults, eg. size 11 instead of size 10
- black font (on white background) for axis tick labels instead of the current medium grey
- disambiguated and accessibility driven font choice

**Describe alternatives you've considered**

Letting Kibana decide on the font anyway. While the capability exists, the themes specify 
- a bit too small font esp. for axis tick labels
- WAY too faint fonts for axis tick labels—it should be much closer to black (on light background) than is now, even if the user hasn't set high contrast preferences or accessibility controls

Canvas2d [can't currently use](https://github.com/whatwg/html/issues/3571) most font options in this space while there's [work in progress](https://github.com/whatwg/html/issues/5617). So, we have these non-exclusive options:
- wait for the Canvas2d standard to grow into modern a11y and other font config needs, and for the main browsers to adopt these (you can help by starring the linked gh issues and following the conversation)
- switch to font rendering with HTML instead of, or as a complement to Canvas2d, permitting more font options
- advocate for fonts which treat character face disambiguation as a key objective even by default; with our typically high technical usage involving lots of alphanumeric codes in ES, it should be the default

Personally I think we should do all these, and using more legible fonts out of the box is the smallest, quickest change in `elastic-charts` so our examples may play with font cc @cchaos @ghudgins 

**Additional context**
The below started as a response to @nickofthyme's question on why putting the Hyperlegible font in chart our examples:

Being able to bring in other font types need not be for specificity. For example, Canvas users are more likely style their charts [beyond the usual fonts](https://www.elastic.co/blog/elasticon-kibana-canvas-story-elasticoffee) and it's neat to have font flexibility, test drive odd sized fonts etc. 

![image](https://user-images.githubusercontent.com/1548516/132453410-4b56f3cd-5cbf-4093-8f50-34b8cadc3f87.png)

[...]

PR #1664, opened 2022-05-03
feat: flame graph with WebGL

monfera:
## Summary

Status: 

- [x] API (column vectors)
- [x] scaling to OOM 1m nodes
- [x] click to focus
- [x] reset (click to empty area)
- [x] show adjacent nodes on the left and right side for context
- [x] text placement into visible part even if box is partly out of view
- [x] text truncation with no ellipsis to preserve info
- [x] proportional fonts to maximize visible text length
- [x] styling (color, gap, rounding)
- [x] API for passing two different datasets for tweening between them (for baseline use, `position0`/`position1` etc. should get the reference to the same typed array)
- [x] horizontal zoom (other than click to focus) with drag and cmd-wheel
- [x] vertical zoom (other than click to focus)
- [x] xy zoom/pan
- [x] specify row height, instead of fitting all rows into the available vertical space
- [x] wheel zooming with the mouse rather than touchpad interacts with the tooltip and highlight: eliminate the flashing 
- [x] fix for the mid-tween chart exit animation glitch
- [x] avoid text flashing on resize
- [x] rerender (text and box sharpening) upon changing display resolution or dragging from one display to another (changing devicePixelRatio)
- [x] rerender (text and box edge sharpening) upon exercising a browser pinch zoom
- [x] variable font size (supralinear with box height)
- [x] provide a margin around the chart, or some other means for the ability to zoom from the side exactly (@markov00's  suggestion)—achieved by zooming from edge if cursor is within 16 CSS pixels of the edge
- [x] to avoid a jump glitch, reset drag start when mouse exits the chart mid drag (short term); allow drag to continue to work outside the chart (medium term)
- [x] deactivate the hover tooltip over the minimap area
- [x] ~~open new issue for a future PR: add render counter and debug info onto DOM element attributes; accessibilitiy with keyboard navigation for all interactions~~ added https://github.com/elastic/elastic-charts/issues/1671

### Checklist


packages/charts/api/charts.api.md
```diff
@@ -555,6 +564,9 @@ export function computeRatioByGroups<T extends Record<string, unknown>>(data: T[
 // @public (undocumented)
 export type ContinuousDomain = [min: number, max: number];
 
+// @public
+export type ControlProviderCallback = (controlName: string, controlFunction: (...args: unknown[]) => void) => void;
```
monfera, line 568: Not yet used in the code

packages/charts/api/charts.api.md
```diff
@@ -57,6 +57,14 @@ export interface AngleFromTo {
 // @public
 export type AnimatedValue = number;
 
+// @public (undocumented)
+export interface AnimationConfig {
```
monfera, line 61: Just an extraction

packages/charts/api/charts.api.md
```diff
@@ -952,6 +964,39 @@ export type FitConfig = {
...
+    columnarData: ColumnarViewModel;
+    // (undocumented)
+    controlProviderCallback: ControlProviderCallback;
+    // (undocumented)
+    specType: typeof SpecType.Series;
+    // (undocumented)
+    valueAccessor: ValueAccessor<D>;
+    // (undocumented)
+    valueFormatter: ValueFormatter;
+    // Warning: (ae-forgotten-export) The symbol "ValueGetterFunction" needs to be exported by the entry point index.d.ts
```
monfera, line 994: More work is needed for actual formatting, see todo

packages/charts/api/charts.api.md
```diff
@@ -1662,7 +1708,7 @@ export type PartialTheme = RecursivePartial<Theme>;
 // Warning: (ae-forgotten-export) The symbol "PartitionSpec" needs to be exported by the entry point index.d.ts
 //
 // @public
-export const Partition: <D extends BaseDatum = any>(props: SFProps<PartitionSpec<D>, "chartType" | "specType", "animation" | "layout" | "valueAccessor" | "clockwiseSectors" | "specialFirstInnermostSector" | "drilldown" | "maxRowCount" | "fillOutside" | "radiusOutside" | "fillRectangleWidth" | "fillRectangleHeight" | "valueFormatter" | "valueGetter" | "percentFormatter" | "topGroove" | "smallMultiples" | "layers", never, "data" | "id">) => null;
+export const Partition: <D extends BaseDatum = any>(props: SFProps<PartitionSpec<D>, "chartType" | "specType", "animation" | "layout" | "valueAccessor" | "valueFormatter" | "valueGetter" | "clockwiseSectors" | "specialFirstInnermostSector" | "drilldown" | "maxRowCount" | "fillOutside" | "radiusOutside" | "fillRectangleWidth" | "fillRectangleHeight" | "percentFormatter" | "topGroove" | "smallMultiples" | "layers", never, "data" | "id">) => null;
```
monfera, line 1711: Looks like a reorder

packages/charts/api/charts.api.md
```diff
@@ -2249,6 +2295,8 @@ export interface ShapeTreeNode extends TreeNode, SectorGeomSpecY {
     // (undocumented)
     value: number;
     // (undocumented)
+    vmIndex: number;
```
monfera, line 2298: I'll remove the added `vmIndex` props as vestigial

packages/charts/src/common/event_handler_selectors.ts
```diff
@@ -50,15 +52,16 @@ export const getOnElementOutSelector = (prev: { pickedShapes: number | null }) =
   if (!settings.onElementOut) {
     return;
   }
-  const nextPickedShapes = pickedShapes.length;
+  const nextPickedShapes =
+    pickedShapes.length > 0 && pickedShapes[0].length > 0 && pickedShapes[0][0].vmIndex !== 0 ? 1 : 0;
```
markov00, line 56: can you add a small comment on that one? since vmIndex is only used for flamegraph is difficult to understand what this is doing here in a `common` code

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -0,0 +1,357 @@
...
+  }
+
+  private inTween = (t: DOMHighResTimeStamp) =>
+    this.drilldown.timestamp + this.props.animationDuration + TWEEN_EPSILON_MS >= t;
+
+  componentDidMount = () => {
+    /*
+     * the DOM element has just been appended, and getContext('2d') is always non-null,
+     * so we could use a couple of ! non-null assertions but no big plus
+     */
```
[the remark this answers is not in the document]
monfera: It's one of the millions of TypeScript blind spots, which is why I added the comment. In a future TS version, maybe it'll be smart enough. The web standards guarantee that it be a non-null

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -0,0 +1,357 @@
...
+    e.stopPropagation();
+    const hovered = this.getHoveredDatumIndex(e);
+    const prevHoverIndex = this.hoverIndex >= 0 ? this.hoverIndex : NaN; // todo instead of translating NaN/-1 back and forth, just convert to -1 for shader rendering
+    if (hovered) {
+      this.hoverIndex = Number.isNaN(hovered.datumIndex) ? DUMMY_INDEX : hovered.datumIndex;
+      if (Number.isFinite(hovered.datumIndex) && !Object.is(this.hoverIndex, prevHoverIndex)) {
+        this.props.onElementOver([{ vmIndex: hovered.datumIndex }]); // userland callback
+      }
+      this.setState({});
+      this.drawCanvas(); // todo use setState properly which would also trigger drawCanvas
```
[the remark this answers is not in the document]
monfera: Thanks Marco, good call! It used to be separate (to allow component change like tooltip removal) but not anymore. Will avoid the duplication

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -0,0 +1,357 @@
...
+    animationDuration: flameSpec?.animation.duration ?? 0,
+    chartDimensions: state.parentDimensions,
+    a11ySettings: getA11ySettingsSelector(state),
+    tooltipRequired: getTooltipType(settingsSpec) !== TooltipType.None,
+
+    // mandatory charts API protocol; todo extract these mappings once there are other charts like Flame
+    onElementOver: settingsSpec.onElementOver ?? (() => {}),
+    onElementClick: settingsSpec.onElementClick ?? (() => {}),
+    onElementOut: settingsSpec.onElementOut ?? (() => {}),
+    onRenderChange: settingsSpec.onRenderChange ?? (() => {}), // todo eventually also update data props on a local .echChartStatus element: data-ech-render-complete={rendered} data-ech-render-count={renderedCount} data-ech-debug-state={debugStateString}
```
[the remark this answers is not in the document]
monfera: Indeed! I meant, it'll be needed for Lens and similar apps but afaik isn't blocking the initial Observability use case

packages/charts/src/chart_types/flame_chart/render/animated_draw.ts
```diff
@@ -0,0 +1,71 @@
...
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+import { LabelAccessor } from '../../../utils/common';
+import { AnimationState, ContinuousDomainFocus, GLResources } from '../types';
+import { drawFrame } from './draw_a_frame';
+
+const linear = (x: number) => x;
+const easeInOut = (alpha: number) => (x: number) => x ** alpha / (x ** alpha + (1 - x) ** alpha);
```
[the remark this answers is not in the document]
monfera: History: it comes from the standalone prototype, and I'm planning to replace tweening with a recurrence relation, so it'll be gone

packages/charts/src/chart_types/flame_chart/render/draw_canvas.ts
```diff
@@ -0,0 +1,78 @@
...
+      ctx.beginPath();
+      ctx.rect(x, y - zoomedRowHeight * cssHeight, width, zoomedRowHeight * cssHeight);
+      if (textColor !== lastTextColor) {
+        // as we're sorting the iteration, the number of color changes (API calls) is minimized
+        ctx.fillStyle = textColor;
+        lastTextColor = textColor;
+      }
+      ctx.save();
+      ctx.clip();
+      ctx.fillText(label, x + TEXT_PAD_LEFT, y - ROW_OFFSET_Y * rowHeight * cssHeight);
```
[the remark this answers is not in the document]
monfera: The `maxWidth` argument ensures that the text isn't spilling out of the specified width, but it does so by shrinking the text 🤣 Which makes it a strong contender for the "least uselful element of the Web API" 😄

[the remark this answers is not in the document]
markov00: oh really? ...that so bad!

packages/charts/src/chart_types/flame_chart/render/ensure_webgl.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  const readPixelXY: PickFunction = (x, y) => {
+    if (gl) {
+      bindFramebuffer(gl, GL_READ_FRAMEBUFFER, pickTexture.target());
+      const pixel = readPixel(gl, dpr * x, textureHeight - dpr * y);
+      const found = pixel[0] + pixel[1] + pixel[2] + pixel[3] > 0;
+      const datumIndex = found
+        ? pixel[3] + 256 * (pixel[2] + 256 * (pixel[1] + 256 * pixel[0])) - GEOM_INDEX_OFFSET
+        : NaN;
+      return Number.isNaN(datumIndex) ? NaN : datumIndex;
```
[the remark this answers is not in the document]
monfera: 4559c0a1d210a559df1703fc7e417c0dde3d9ee3

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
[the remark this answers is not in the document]
monfera: 4af121e01c6ff1f496fbae241d96e1a525c3b731

packages/charts/src/chart_types/flame_chart/flame_api.ts
```diff
@@ -0,0 +1,66 @@
...
+    keyof typeof buildProps['overrides'],
+    keyof typeof buildProps['defaults'],
+    keyof typeof buildProps['optionals'],
+    keyof typeof buildProps['requires']
+  >,
+) {
+  const { defaults, overrides } = buildProps;
+  useSpecFactory<FlameSpec<D>>({ ...defaults, ...stripUndefined(props), ...overrides });
+  return null;
+};
```
nickofthyme, line 66: You mentioned this being a little complex and you are correct 😅. The simpler approach is to use `specComponentFactory` directly which simply takes the overrides and defaults and handles the rest for you.

However, using `specComponentFactory` does not allow you to use a top-level generic type, hence the crazy type magic. In this case you are using `D` to pass as a generic to `FlameSpec`, if that is not needed you can you the simpler approach.

I am also now noticing that the inside call the `useSpecFactory` could be simplified a little.

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -0,0 +1,598 @@
...
+    this.pickTexture = NullTexture;
+    this.glResources = NULL_GL_RESOURCES;
+    this.glCanvasRef = createRef();
+    this.animationRafId = NaN;
+    this.prevT = NaN;
+    this.currentFocus = focusRect(this.props.columnarViewModel, props.chartDimensions.height, 0, -Infinity);
+    this.targetFocus = { ...this.currentFocus };
+    this.hoverIndex = NaN;
+    this.pointerX = -10000;
+    this.pointerY = -10000;
```
nickofthyme, line 176: What is the significance of this value? And not being `Infinity`?

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -0,0 +1,598 @@
...
+
+  componentWillUnmount() {
+    this.props.containerRef().current?.removeEventListener('wheel', this.preventScroll);
+  }
+
+  componentDidUpdate = () => {
+    if (!this.ctx) this.tryCanvasContext();
+    this.ensurePickTexture();
+    this.drawCanvas(); // eg. due to chartDimensions (parentDimensions) change
+    // this.props.onChartRendered() // creates and infinite update loop
```
nickofthyme, line 232: @monfera could you take a quick look at this to see what could fix this? Not a blocker but just wanted to make you aware of it.

packages/charts/src/chart_types/flame_chart/render/draw_canvas.ts
```diff
@@ -0,0 +1,126 @@
...
+  const minRectWidthForTextInCssPix = minTextLengthCssPix + TEXT_PAD_LEFT + TEXT_PAD_RIGHT;
+  const minRectWidth = minRectWidthForTextInCssPix / cssWidth;
+  const textColor = 'black'; // todo it could come from config / theme or automatic decision like in other charts
+
+  // text rendering
+  ctx.save();
+  ctx.textAlign = 'left';
+  ctx.textBaseline = 'middle';
+  ctx.scale(dpr, dpr);
+  ctx.font = `${fontSize}px "Atkinson Hyperlegible"`;
```
nickofthyme, line 52: Do we want to hardcode this?

packages/charts/src/chart_types/partition_chart/renderer/dom/highlighter.tsx
```diff
@@ -282,8 +282,7 @@ export const DEFAULT_PROPS: HighlighterProps = {
 /** @internal */
 export function highlightSetMapper(geometries: QuadViewModel[], foci: IndexedContinuousDomainFocus[]) {
   return (vm: ShapeViewModel): HighlightSet => {
-    const { index } = vm;
-    const { innerIndex } = vm;
+    const { index, innerIndex } = vm;
```
nickofthyme, line 285: 😌

packages/charts/src/common/kingly.ts
```diff
@@ -0,0 +1,626 @@
+/*
```
nickofthyme, line 1: Once we get the package logic sorted to where we can have multiple packages, this seems like a good candidate to be its own package.

storybook/main.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
   stories: ['./stories/**/*.(stories|story).ts?(x)'],
   addons: [
     '@storybook/addon-actions',
-    '@storybook/addon-docs',
+    // '@storybook/addon-docs',
```
nickofthyme, line 13: What was this issue this was posing?

storybook/style.scss
```diff
@@ -127,7 +127,7 @@ body {
   position: relative;
   z-index: 500;
   box-sizing: border-box;
-  background-color: blanchedalmond;
+  /* background-color: blanchedalmond; */
```
nickofthyme, line 130: 😱 So is this official @markov00?

nickofthyme (approved):
Looks great Robert!

Thanks for walking me through the code at a top-level that helped me wrap my head around this code much easier.

I tested locally and had the chart disappear on me once when clicking around but haven't been able to reproduce it. Also turned off webGL and the difference is night 🌃  and day 🏙️!

storybook/stories/bar/54_functional_accessors.story.tsx
```diff
@@ -31,11 +32,16 @@ export const Example = () => {
   const splitAccessorFn: AccessorFn = (d) => d.g2;
   splitAccessorFn.fieldName = text('split fn name', '') || undefined;
 
-  const onElementClick: ElementClickListener = ([[, { key }]]) => action('clicked series key')(key);
+  const onElementClick = ([[, { key }]]: XYChartElementEvent[]) => action('clicked series key')(key);
```
nickofthyme, line 35: @markov00 We need to come up with a better API for this, the union types are starting to diverge too much. I think a simple fix would just be to change the events array to a record per type, so like...

```
interface ClickEventElementMap {
  xy: XYChartElementEvent[];
  partition: PartitionElementEvent[];
  flame: FlameElementEvent[];
  heatmap: HeatmapElementEvent[];
  wordCloud: WordCloudElementEvent[];
}
export type ElementClickListener = (elements: ClickEventElementMap) => void;
```

Not saying in this PR!

storybook/style.scss
monfera: I pitched this in our channel a couple of days ago, not sure yet where Marco stands on this issue. The upshot is that the blanchedalmond seems to have served its purpose, and feels like it's no longer worth having it, due to the visual interference it causes. It snuck into the PR by accident. Happy to either remove the line or remove the asterisks. Alternatively we could put a 1px grey border around all the charts (like the WebGL flame, or fainter) which is unlikely to be mistaken with chart rendering

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
monfera: It was just a copy-paste offscreen value from someplace else at a time when switching it would've been tangential. Will revise

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
monfera: The component update triggers a redux action(!) and that eventually cascades into another component update. See our channel convo with Marco on May 2 to my question earlier (search for `CHART_RENDERED`)`. This circularity was the motivation. 

Upshot: I feel like the proper fix would be the demotion of the `CHART_RENDERED` action, even if the current circularity could be technically fixed. Because it is not an action (open to prepared arguments). We seemed to arrive at a consensus that over time its action-ness could be phased out. As there's no current need for this callback for this chart, I didn't want to spend time fixing it for the old way, and doing the new way, across all chart types, is not for this PR

packages/charts/src/chart_types/flame_chart/render/draw_canvas.ts
monfera: Oh totally. Not. Even if I love it on accessibility grounds. I put this on the todo list in the Description above, probably for a follow-up PR.   

Even when it happens, I'd like to keep using this font for this mock because it's a disambiguated font, here are my [couple of lines](https://github.com/elastic/elastic-charts/issues/1365) on this topic (maybe the Canvas2d handling is caniuse-able by now?)

nickofthyme:
I updated the puppeteer screenshots from my intel mac (https://github.com/elastic/elastic-charts/pull/1664/commits/77e78f04b6bf606311625b79ac4041bbd865a0c4). There are 2 stories with errors in code preventing them from rendering, see:

1. https://ech-e2e-ci--pr-1664-8xu76vsj.web.app/?path=/story/sunburst--no-slice
2. https://ech-e2e-ci--pr-1664-8xu76vsj.web.app/?path=/story/sunburst--total-zero

storybook/main.js
monfera: Marco made a mock with 110k data points, which we use for testing, but it's not in the repo due to its size. I can send it to you if you want to try it. There's load time slowdown if the addon-docs remains active. Though the `//` should probably be deleted before merge, lmk

storybook/stories/bar/54_functional_accessors.story.tsx
monfera: We could also have a broader chat on what my mind calls "elastic-charts protocol". The intent is a level of uniformity, which is sensible. Yet it's also a set of somewhat vaguely delineated internal conventions that don't have documented reasons afaik. For some of these things, [composition > convention](https://clojurefun.wordpress.com/2012/08/17/composition-over-convention/).

packages/charts/src/chart_types/flame_chart/flame_api.ts
monfera: Most of the broilerplate and maybe some other few hundred lines wouldn't be needed if our entire spec lived in JSON, right? Complexity whose main purpose is to make Charts appear as a React component with other Charts React components as children. If I understand, it has no technical or functional purpose, it's cosmetics, which was useful at the beginning but might no longer be a net positive. IIRC it's not just delivered and executed code, type complications etc. but there are some consequences like only probabilistically knowing we got to the end of spec processing

tl;dr when seeing `...Factory`, I start to wonder if the root cause could be addressed. I'd like to learn about current or future Charts use cases of anything with `Factory` in its name where it is fundamental or beneficial for the end user, rather than as a way of addressing implementation choices

packages/charts/src/common/event_handler_selectors.ts
monfera: I reverted the `vmIndex` for now in f3b61427

packages/charts/src/chart_types/flame_chart/flame_api.ts
monfera: Deep linked this convo in https://github.com/elastic/elastic-charts/issues/1671 so this can be marked as resolved

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
monfera: Deep linked it in https://github.com/elastic/elastic-charts/issues/1671 so this can be resolved

packages/charts/src/chart_types/flame_chart/render/draw_canvas.ts
monfera: Deep linked it in https://github.com/elastic/elastic-charts/issues/1671 so it can be resolved

storybook/stories/bar/54_functional_accessors.story.tsx
monfera: I deep linked it from https://github.com/elastic/elastic-charts/issues/1671 so it can be marked as resolved here

.prettierignore
```diff
@@ -10,3 +10,5 @@ coverage/
 **/tmp/
```
markov00, line 10: Two VRTs comments:
- there are two snapshots that should be removed: the pie charts with the `2-snap` suffix
- any ideas about this 0 bytes change for the annotation chart?

markov00 (approved):
It looks awesome! Great job Robert!
I've tested it locally on mac on Safari, Chrome and Firefox. Everything works flawlessly and smoothly.
I've added a small possible feature to #1671 an alternative to the minimap with scrollbars, just and idea, and we can discuss in a later stage

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -0,0 +1,638 @@
...
+
+  private datumAtXY: PickFunction = (x, y) =>
+    this.glContext ? colorToDatumIndex(readPixel(this.glContext, x, y)) : NaN;
+
+  private updatePointerLocation(e: { clientX: number; clientY: number }) {
+    if (!this.props.forwardStageRef.current || !this.ctx) return;
+    const box = this.props.forwardStageRef.current.getBoundingClientRect();
+    this.pointerX = e.clientX - box.left;
+    this.pointerY = e.clientY - box.top;
+  }
```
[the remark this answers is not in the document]
monfera: Thanks, I added it to the new issue

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -0,0 +1,638 @@
...
+    const style: CSSProperties = {
+      width,
+      height,
+      top: 0,
+      left: 0,
+      padding: 0,
+      margin: 0,
+      border: 0,
+      position: 'absolute',
+      cursor: this.hoverIndex >= 0 ? 'pointer' : DEFAULT_CSS_CURSOR,
```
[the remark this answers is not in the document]
monfera: Great idea, added it to the issue, though maybe it can catch this PR too

monfera:
@markov00 addon-docs reactivated in b7fe94bf

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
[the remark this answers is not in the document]
monfera: Added it 50940dc5

.prettierignore
markov00: solved in https://github.com/elastic/elastic-charts/pull/1664/commits/568f7434540bda44e018dafa8cfcbacc7da2fbc1

PR #1672, opened 2022-05-16
fix: runtime error with `process.env` in src

nickofthyme:
## Summary

Quick fix for https://github.com/elastic/elastic-charts/commit/0bbb809132897dc691f1a71c676c46ec91aa59df#r73626789.

When using `process.env` in `packages/charts/src` the node process may not always exist.

### DIFF
--- a/packages/charts/src/chart_types/xy_chart/renderer/canvas/animations/index.ts
+++ b/packages/charts/src/chart_types/xy_chart/renderer/canvas/animations/index.ts
@@ -12,7 +12,7 @@ import { Logger } from './../../../../../utils/logger';
 import { AnimatedValue, Animation, AnimationOptions, AnimationState } from './animation';
 
 // TODO find a better way to do this when we have an actual build process
-const DISABLE_ANIMATIONS = (process && process.env && process.env.VRT) === 'true'
+const DISABLE_ANIMATIONS = (process && process.env && process.env.VRT) === 'true';
 
 /**
  * Function used to animate values from within a render context.


spalger (approved):
Thanks!

markov00 (approved):
LGTM

packages/charts/src/chart_types/xy_chart/renderer/canvas/animations/index.ts
```diff
@@ -11,7 +11,8 @@ import { debounce } from 'ts-debounce';
 import { Logger } from './../../../../../utils/logger';
 import { AnimatedValue, Animation, AnimationOptions, AnimationState } from './animation';
 
-const DISABLE_ANIMATIONS = Boolean(process.env.VRT);
+// TODO find a better way to do this when we have an actual build process
+const DISABLE_ANIMATIONS = (process && process.env && process.env.VRT) === 'true'
```
nickofthyme, line 15: @markov00 I think this is the best solution until we get a full-blown build system in place for the library.

The issue is that webpacks `EnvironmentPlugin` just does an inline value replacement but this fails when used with optional chaining.

```ts
// before bundling or without webpack
const value1 = process.env.VRT;
const value2 = process?.env?.VRT;

// after bundling
const value1 = "true";
const value2 = (_a = process === null || process === void 0 ? void 0 : process.env) === null || _a === void 0 ? void 0 : _a.VRT;
```

In such case `value2 = undefined` and the vrt [tests fail](https://kibana-ci.elastic.co/job/elastic+elastic-charts+pr-vrts/3483/gcsObjects/) because the animations are no longer disabled.

This line now gets converted to 👇🏼 which still works for our CI checks. 🤞🏼 

```ts
const DISABLE_ANIMATIONS = (process && process.env && 'true') === 'true'
```

markov00: Ok, good for me, but please double check with what exists today in the kibana-ci stats: https://github.com/elastic/kibana-ci-stats/pull/501/files
because it seems the `VRT` field is a boolean not a string

playground/webpack.config.js
```diff
@@ -74,7 +74,7 @@ module.exports = {
       filename: 'index.html',
       favicon: '../public/favicon.ico',
     }),
-    new webpack.EnvironmentPlugin({ RNG_SEED: null, VRT: false }),
+    new webpack.EnvironmentPlugin({ RNG_SEED: null, VRT: 'false' }),
```
markov00, line 77: See https://github.com/elastic/kibana-ci-stats/pull/501/files

packages/charts/src/chart_types/xy_chart/renderer/canvas/animations/index.ts
nickofthyme: Spencer just added that to prevent the error. I changed them all to strings as the environment variable set in the command line would never be a boolean.

playground/webpack.config.js
nickofthyme: Yeah either way the default is `false` unless the value is a string `'true'`.

Issue #1673, opened 2022-05-17
[bubble] can't pick all data points with mouse


markov00:
**Describe the issue**
Under specific circumstances, I can't mouse over on every data point on a bubble chart. The tooltip doesn't appear.

**To Reproduce**
1. Go to [http://192.168.8.123:9001/?path=/story/bubble-chart-alpha--multiple&globals=theme:light](http://192.168.8.123:9001/?path=/story/bubble-chart-alpha--multiple&globals=theme:light)
2. Move the mouse over various points
3. Most datapoints are not hoverable, it becomes worst when increasing the number of points shown

**Expected behaviour**
I should be able to pick every data points on the bubble chart with the mouse pointer.

**Screenshots**

https://user-images.githubusercontent.com/1421091/168792896-7d496547-97d1-4dcd-b93a-61c5da9ea8b2.mov

**Version (please complete the following information):**
 - OS: mac monterey
 - Browser: chrome/safari/firefox
 - Elastic Charts: master

**Errors in browser console**
If you click on the `debug` knob the following error message appear:
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'delaunay')
    at Voronoi.render (index.ts:933:1)
    at renderers.ts:154:1
    at withContext (index.ts:32:1)
    at renderers.ts:137:1
    at index.ts:51:1
    at Array.forEach (<anonymous>)
    at renderLayers (index.ts:51:1)
    at renderers.ts:56:1
    at withContext (index.ts:32:1)
    at render (renderers.ts:34:1)
```

**Checklist**

- [x] The proper **chart type** label has been added (e.g. `:xy`, `:partition`)
- [x] The proper **feature** labels have been added (e.g. `:interactions`, `:axis`)

PR #1674, opened 2022-05-17
fix(bubble): tooltip visibility on hover

nickofthyme:
## Summary

Fixes a bug in bubble chart that prevented the hovered values from correctly being active.

### Before

Screen Recording 2022-05-17 at 09 31 57 AM

### After

Screen Recording 2022-05-17 at 09 30 01 AM

## Details

The issue is that the bubble chart was being filled on the x domain and this created bad `NaN` values in the Delaunay index and excluded those values from the tooltip list.

I'm not sure what actually caused this as I don't see or think of any recent (<16m) changes to the problematic files. Maybe the `NaN` scale changes just brought the issue to light but I don't think that is what caused this.

## Issues

#1673

### Checklist


packages/charts/src/chart_types/xy_chart/renderer/canvas/renderers.ts
```diff
@@ -145,17 +145,6 @@ export function renderXYChartCanvas2d(
...
-            const renderer = geometriesIndex.triangulation([0, 0, width, height])?.render;
-            if (isCanvasRenderer(renderer)) {
-              ctx.beginPath();
-              ctx.translate(left, top);
-              ctx.setLineDash([5, 5]);
-              renderer(ctx);
-              ctx.lineWidth = 1;
-              ctx.strokeStyle = 'blue';
-              ctx.stroke();
-            }
```
nickofthyme, line 158: @markov00 This fixes the `debug` error.

packages/charts/src/chart_types/xy_chart/utils/fill_series.ts
```diff
@@ -62,11 +56,6 @@ export function fillSeries(
...
 
-function isXFillNotRequired(spec: BasicSeriesSpec, groupScaleType: ScaleType, isStacked: boolean) {
-  const onlyNoFitAreaLine = (isAreaSeriesSpec(spec) || isLineSeriesSpec(spec)) && !spec.fit;
-  const onlyContinuous =
-    groupScaleType === ScaleType.Linear ||
-    groupScaleType === ScaleType.LinearBinary ||
-    groupScaleType === ScaleType.Time;
-  return onlyNoFitAreaLine && onlyContinuous && !isStacked;
+function isXFillRequired(spec: BasicSeriesSpec, isStacked: boolean): boolean {
+  return (isAreaSeriesSpec(spec) || isLineSeriesSpec(spec)) && (isStacked || !!spec.fit);
```
markov00, line 60: I don't think this is the correct way to fix the problem and to me it looks like it doesn't fully respect the original constraints making a general change on how to fill missing values.
I know is a bit convoluted with all those double negations, but the inverted representation you are doing doesn't fully cover all the aspects that where there before.

packages/charts/src/chart_types/xy_chart/rendering/points.ts
```diff
@@ -112,7 +112,7 @@ export function renderPoints(
       const isInYDomain = yScale.isValueInDomain(valueAccessor(datum));
       indexedGeometryMap.set(pointGeometry, geometryType);
```
markov00, line 113: What you can don here instead is to check if the `datum` was filled before updating the geometry map with 
```ts
if(!isDatumFilled(datum)) indexedGeometryMap.set(pointGeometry, geometryType);
```

packages/charts/src/chart_types/xy_chart/renderer/canvas/renderers.ts
markov00: more then fix it avoids the error :D, I like btw the voronoi debug, there is definitely some issue with the closures or something along that blocks that renderer.
I'm fine commenting this out until we don't fix the issue if you don't mind

markov00 (changes requested):
The fix should not change the current existing fill behaviour of line/area charts.
I proposed a solution that I tested locally and that could work fine without changing the that behaviour

packages/charts/src/chart_types/xy_chart/renderer/canvas/renderers.ts
nickofthyme: Oh yeah I forgot about the voronoi debug 🤦‍♂️. I'll add it back and look at the closure.

packages/charts/src/chart_types/xy_chart/utils/fill_series.ts
nickofthyme: That's fine but in general why are we defaulting to fill x values? By saying `isXFillNotRequired` defines only when it is skippable, that way everytime we add new chart types, we would need to update this fill it is not needed.

markov00: I've reverted back this to the current master, this PR https://github.com/elastic/elastic-charts/pull/1688 will take care of fixing this correctly

packages/charts/src/chart_types/xy_chart/utils/fill_series.ts
```diff
@@ -62,11 +61,11 @@ export function fillSeries(
   });
 }
 
-function isXFillNotRequired(spec: BasicSeriesSpec, groupScaleType: ScaleType, isStacked: boolean) {
-  const onlyNoFitAreaLine = (isAreaSeriesSpec(spec) || isLineSeriesSpec(spec)) && !spec.fit;
+function isXFillRequired(spec: BasicSeriesSpec, groupScaleType: ScaleType, isStacked: boolean): boolean {
```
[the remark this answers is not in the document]
markov00: I'm going to fix this in a different PR: https://github.com/elastic/elastic-charts/pull/1688

markov00 (approved):
I've refactored the fix to push the solution more into the spatial index context rather than an an higher level.
Please @nickofthyme check my comments on this page and review the changes

packages/charts/src/chart_types/xy_chart/renderer/canvas/renderers.ts
```diff
@@ -145,13 +145,12 @@ export function renderXYChartCanvas2d(
...
-
-            const renderer = geometriesIndex.triangulation([0, 0, width, height])?.render;
-            if (isCanvasRenderer(renderer)) {
+            const voronoi = geometriesIndex.triangulation([0, 0, width, height]);
+            if (voronoi) {
               ctx.beginPath();
               ctx.translate(left, top);
               ctx.setLineDash([5, 5]);
-              renderer(ctx);
+              voronoi.render(ctx);
```
nickofthyme, line 153: 👍🏼

packages/charts/src/chart_types/xy_chart/utils/indexed_geometry_spatial_map.ts
```diff
@@ -42,15 +42,17 @@ export class IndexedGeometrySpatialMap {
   set(points: PointGeometry[]) {
     this.maxRadius = Math.max(this.maxRadius, ...points.map(({ radius }) => radius));
     const { pointGeometries } = this;
-    points.forEach((p) => pointGeometries.push(p));
+    points.forEach((p) => {
+      if (isFiniteNumber(p.y)) pointGeometries.push(p);
+    });
```
nickofthyme, line 47: Nice!

packages/charts/src/chart_types/xy_chart/rendering/points.ts
```diff
@@ -112,7 +112,7 @@ export function renderPoints(
...
       // use the geometry only if the yDatum in contained in the current yScale domain
-      if (y !== null && yDefined(datum, valueAccessor) && isInYDomain && !isDatumFilled(datum)) {
+      if (isFiniteNumber(y) && yDefined(datum, valueAccessor) && isInYDomain && !isDatumFilled(datum)) {
```
nickofthyme, line 115: I don't think this check is necessary at all cuz canvas just ignores `NaN` values.

```suggestion
      if (yDefined(datum, valueAccessor) && isInYDomain && !isDatumFilled(datum)) {
```

nickofthyme (commented):
Can't approve my own PR but changes all look good.

markov00 (approved):

Issue #1671, opened 2022-05-15
Flame: chart protocol completion and other todo


monfera:
Chart protocol:
- [ ] debug DOM data generation
- [ ] accessibility layer generation
- [ ] signaling of chart rendered

Interactions / examples:
- [ ] placement and design for affordances (reset, search, direct manipulation interaction modifier keys)
- [ ] reset (via input/callback eg. for a reset button)
- [ ] text/regex search facility
- [ ] keyboard navigation and accessibility
- [ ] focus rectangle: move, move border, resize on minimap
- [ ] box select and focus (horizontal, vertical, both) on focus layer

Chart configurability
- [ ] actually make use of the `valueFormatter` in the charts API
- [ ] make use of the font family from the charts theme

Cross platform / browser
- [ ] triage the Ubuntu render issue mentioned by @thomasdullien (we've been awaiting further info/collab on it as we have no issue reproduction yet)
- [ ] Cross-browser issues: big browser/pinch zoom disappearance (FF, Safari); vertical zoom jumpy text length (Safari)

Arising from https://github.com/elastic/elastic-charts/pull/1664

PR #1678, opened 2022-05-23
feat(flame): minimap pan

monfera:
## Summary

minimap pan

- Adds minimap pan interaction
- Minor: the focus area drag is now more precise (it considers margins) so dragging a long way still keeps the mouse pointer glued to the exact slippy paper location established when starting the drag

Clears an item in https://github.com/elastic/elastic-charts/issues/1671

### Checklist


packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -265,6 +267,10 @@ class FlameComponent extends React.Component<FlameProps> {
     // this.props.onChartRendered() // creates and infinite update loop
   };
 
+  private pointerInMinimap = (x: number, y: number) =>
+    x === clamp(x, this.getMinimapLeft(), this.getMinimapLeft() + this.getMinimapWidth()) &&
+    y === clamp(y, this.getMinimapTop(), this.getMinimapTop() + this.getMinimapHeight());
+
```
nickofthyme, line 273: If not too difficult, I think it would be nice to set the `cursor` to `move` when hovering the minimap, but not outside the minimap obviously (ignore this in the gif example).

Screen Recording 2022-05-23 at 01 33 05 PM

nickofthyme (approved):
Looks great! nice reuse of code. I just had one suggestion but other than that it works as I'd expect.

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
monfera: Thanks for spotting it! It was supposed to work on both charts, but a perf optimization in the preceding PR causes to sometimes not make it appear, but sometimes it does, also on the minimap (likely missing a setState somewhere). Btw. it's the grab hand rather than the "move" wind rose. Let's discuss the shape and the fix for the next PR if that's OK, as the regression is not introduced with this PR and @gvnmagni and @markov00 should chime in too. I'd like to make a stream of small PRs instead of keeping them alive for days

monfera: Also, the minimap pan currently works even when initiating it from outside the focus rectangle. That's something to discuss as it has pros&cons. A pro is adherence to Fitts's law: even if the focus rectangle is small, in an already potentially small minimap, it can be super tedious to precisely click on, just for a pan jolt that can be a simple, pleasing interaction. Discussing all interactions, and the need for it (like resize the rect in the minimap) is for a future PR

nickofthyme: Yeah with a really tiny focus area it would not be practical. The only preferred UI IMO would be a [`hitSlop`](https://reactnative.dev/docs/touchablewithoutfeedback#hitslop) value which always allows an easy handle on even the smallest areas. Happy to discuss.

monfera: I added an item to the bottom of https://github.com/elastic/elastic-charts/issues/1671 and detailed the pertinent line in the Interactions block (on potential future resize or box select in the minimap)

PR #1679, opened 2022-05-24
build(react): allow `react@^17` as peer dependency

### LINKED ISSUE #1638: Compatibility with React 17
We are heavy ElasticSearch and elastic-charts users. Great library, guys!

Just one major downside: `package.json` is locked to `"react": "^16.0.0"`, which means we can't upgrade to React 17 (required for many latest libraries, including Material UI v5).

Any challenges with making this `"^16.0.0|^17.0.0"`?

nickofthyme:
## Summary

Allow consumers of `@elastic/charts` to use `react@^17`.

## Details

Updates `peerDependency` to match that of `@elastic/eui` [deps](https://github.com/elastic/eui/blob/8de2483b06056c548a2617e67cca348f9e1883dc/package.json#L253-L254). I don't see any issues, ~~running PR checks now to confirm~~ vrt tests pass with only slight diffs on advanced annotation story, see job [here](https://kibana-ci.elastic.co/job/elastic+elastic-charts+pull-request/3535/).

image

With using the charts library from source inside storybook and the e2e-server applications, yarn workspace hoists this in a difficult way such that having more that one version of `react` is a little tricky, particularly with using react hooks and having 2 versions of react. I think we could get around this in some way but didn't want to put time into it just yet.

@markov00 Since it would take some time to setup a periodic check for both versions (i.e. `^16` & `^17`) what are your thoughts on supporting just the version of react from kibana until they upgrade? Then upgrading with them? In the meantime we allow react `^17` with the caveat that we do not develop specifically against that version of react, but know it to generally work as expected.

Removes implicit children typings per type [changes](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-typescript-definitions).

## Issues

closes #1638 

### Checklist


markov00:
>@markov00 Since it would take some time to setup a periodic check for both versions (i.e. ^16 & ^17) what are your thoughts on supporting just the version of react from kibana until they upgrade? Then upgrading with them? In the meantime we allow react ^17 with the caveat that we do not develop specifically against that version of react, but know it to generally work as expected.

@nickofthyme I'm fine stick with the Kibana React version

markov00 (approved):
Thanks @nickofthyme for the checks and the update!

homeyjd:
Would love to use these changes asap! When will this PR+merge hit the NPM release channel?

nickofthyme:
@homeyjd apologies for the delay. I made one last fix based on the changes in 18 and will merge and release this today.

damikun:
What about React 18?

PR #1682, opened 2022-05-25
feat(flame): context loss handling

monfera:
## Summary

This PR adds context loss handling to the WebGL flame chart. 

Web standards do not guarantee that rendering resources aren't lost. Rendering resources can be lost for various reasons, such as an overloaded system (even due to other applications), graphics driver error, or too many contexts. In fact, a context loss may even happen with Canvas2d. But it's more important for WebGL, where a context can be lost even simply having too many distinct WebGL canvases on a page (more than 16).

The PR contains other minor refactoring, such as pulling the font type from the default font, and updating `kingly.ts` to the version that can run more WebGL charts that are in the pipeline. For example, via configurable polygon vertex winding order.

### Checklist

### DIFF
--- a/packages/charts/src/chart_types/flame_chart/flame_chart.tsx
+++ b/packages/charts/src/chart_types/flame_chart/flame_chart.tsx
@@ -889,10 +889,38 @@ class FlameComponent extends React.Component<FlameProps> {
 
     this.ensurePickTexture();
 
-    if (this.glContext && this.glResources === NULL_GL_RESOURCES) {
-      this.glResources = ensureWebgl(this.glContext, Object.keys(this.props.columnarViewModel));
-      uploadToWebgl(this.glContext, this.glResources.attributes, this.props.columnarViewModel);
-      testContextLoss(this.glContext);
+    const initializeGL = (gl: WebGL2RenderingContext) => {
+      this.glResources = ensureWebgl(gl, Object.keys(this.props.columnarViewModel));
+      uploadToWebgl(gl, this.glResources.attributes, this.props.columnarViewModel);
+    };
+
+    const restoreGL = (gl: WebGL2RenderingContext) => {
+      initializeGL(gl);
+      this.drawCanvas();
+    };
+
+    if (glCanvas && this.glContext && this.glResources === NULL_GL_RESOURCES) {
+      glCanvas.addEventListener('webglcontextlost', (event) => event.preventDefault(), false); // we could log it for telemetry etc todo add the option for a callback
+      glCanvas.addEventListener(
+        'webglcontextrestored',
+        () => {
+          // browser trivia: the duplicate calling of ensureContextAndInitialRender and changing/resetting the width are needed for Chrome and Safari to properly restore the context upon loss
+          // we could log context loss/regain for telemetry etc todo add the option for a callback
+          if (!glCanvas || !this.glContext) return;
+          restoreGL(this.glContext);
+          const widthCss = glCanvas.style.width;
+          const widthNum = parseFloat(widthCss);
+          glCanvas.style.width = `${widthNum + 0.1}px`;
+          window.setTimeout(() => {
+            glCanvas.style.width = widthCss;
+            if (this.glContext) restoreGL(this.glContext);
+          }, 0);
+        },
+        false,
+      );
+
+      initializeGL(this.glContext);
+      // testContextLoss(this.glContext);
     }
   };
 }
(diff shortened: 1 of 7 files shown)


packages/charts/src/chart_types/flame_chart/render/draw_canvas.ts
```diff
@@ -52,7 +53,7 @@ export const drawCanvas2d = (
   ctx.textAlign = 'left';
   ctx.textBaseline = 'middle';
   ctx.scale(dpr, dpr);
-  ctx.font = `${fontSize}px sans-serif`;
+  ctx.font = `${fontSize}px ${DEFAULT_FONT_FAMILY}`;
```
markov00, line 56: you can probably reuse the `cssFontShorthand` to be sure about the right param positioning

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -297,9 +296,8 @@ class FlameComponent extends React.Component<FlameProps> {
     if (!this.isDragging(e)) {
       e.stopPropagation();
       this.updatePointerLocation(e);
-      if (this.pointerInMinimap(this.pointerX, this.pointerY)) return;
       const hovered = this.getHoveredDatumIndex(e);
-      const prevHoverIndex = this.hoverIndex >= 0 ? this.hoverIndex : NaN; // todo instead of translating NaN/-1 back and forth, just convert to -1 for shader rendering
+      const prevHoverIndex = this.hoverIndex >= 0 ? this.hoverIndex : NaN;
```
markov00, line 300: > todo instead of translating NaN/-1 back and forth, just convert to -1 for shader rendering

Was the proposal, in the deleted comment, already implemented?

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -896,9 +894,39 @@ class FlameComponent extends React.Component<FlameProps> {
...
+    };
+
+    const restoreGL = (gl: WebGL2RenderingContext) => {
+      initializeGL(gl);
+      this.pickTexture = NullTexture;
+      this.ensureTextureAndDraw();
+    };
+
+    if (glCanvas && this.glContext && this.glResources === NULL_GL_RESOURCES) {
+      glCanvas.addEventListener('webglcontextlost', (event) => event.preventDefault(), false); // we could log it for telemetry etc todo add the option for a callback
```
markov00, line 909: just out of curiosity, what does the `event.preventDefault` do in this case?

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -896,9 +894,39 @@ class FlameComponent extends React.Component<FlameProps> {
...
+      glCanvas.addEventListener(
+        'webglcontextrestored',
+        () => {
+          // browser trivia: the duplicate calling of ensureContextAndInitialRender and changing/resetting the width are needed for Chrome and Safari to properly restore the context upon loss
+          // we could log context loss/regain for telemetry etc todo add the option for a callback
+          if (!glCanvas || !this.glContext) return;
+          restoreGL(this.glContext);
+          const widthCss = glCanvas.style.width;
+          const widthNum = parseFloat(widthCss);
+          glCanvas.style.width = `${widthNum + 0.1}px`;
```
markov00, line 919: are you doing that because using the same width value will not trigger a repaint?

markov00 (commented):
After triggering the context lost, in the storybook the chart highlight is now very bright, and when start searching, faded out elements are nearly blacked out 
Screenshot 2022-05-26 at 10 12 45

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
monfera: Yes, already did most of the planned transformation so it's currently good, might iterate on it in the future

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
monfera: The default behavior is that nothing is going to happen, and the `webglcontextrestored` event won't be raised

packages/charts/src/chart_types/flame_chart/render/draw_canvas.ts
monfera: 076dc72dfbe9e2243ca3a94a09cfc745093330ea

PR #1688, opened 2022-05-27
fix(cartesian): fill missing datapoints only for stacked area charts

### LINKED ISSUE #1683: Band area doesn't show up with `fit` set for certain range of area
**Describe the issue**

**To Reproduce**

Can be reproducible with https://codesandbox.io/s/recursing-euclid-rjr3x0:

1. Go to[ 'src/data.ts' ](https://codesandbox.io/s/recursing-euclid-rjr3x0?file=/src/data.ts)
2. Go to line 11, comment `fit: "linear"` back in to the data
3. See that the `Expected bounds` green area band disappears

**Expected behaviour**
The `Expected bounds` green area band should not disappear. As a matter of fact setting `fit: "none"` doesn't fix the issue either.

**Screenshots**

https://user-images.githubusercontent.com/43350163/170393115-3ac8149b-ea46-4e37-b527-d8b147038d28.mov

 - Browser: [e.g. chrome, safari]
 - Elastic Charts: [e.g. 1.0.1]

**Additional context**

**Errors in browser console**

**Kibana Cross Issues**

**Checklist**

- [ ] Every related Kibana issue is listed under `Kibana Cross Issues` list and the `kibana cross issue` tag is applied

markov00:
## Summary
This PR fixes an issue caused by a mixed configuration with a data fill functionality, the fit functionality within a band chart.

There are multiple factors that are causing the rendering error in #1683:
- the fit functionality doesn't work for band charts yet, it applies the fit function only to the `y1` values
- we are wrongly filling charts with null data points on every missing x value depending on some wrongly configured constraints.
- when rendering a band both y values needs to exist to be considered a defined value. If only one is missing, then that band section is considered not defined and will not be rendered.

The mix of these factors cause the issue linked below.

On master:
Screenshot 2022-05-27 at 11 07 39

On this PR:
Screenshot 2022-05-27 at 11 09 16

## Details
The current dataset filling operation can be described as the following:
- collect every available value along the X axis across all series
- if the filling operation is required (a constraint previously defined by negating the `isXFillNotRequired` method) then we loop through the data array, adding a partial data point (with null Y values) on every missing x datapoint.

I've chosen to fix the current policy on filling data sets with empty X values expressed in the `isXFillNotRequired` method.
I've decided to apply the filling policy only to stacked area charts. The fill policy should be required only for that case because:
- stacked area charts need to perfectly juxtapose areas on top of each other. If we miss a data point in one area,  the area above or below can overlap or not completely touch the nearest area, causing the chart to be wrongly rendered creating a bad data representation.
- bar chart don't need this fill policy at all.
- line charts can clearly overlaps, we don't stack line charts so that policy is completely useless.
- non stacked area charts, behaves like line charts, so to the policy doesn't need to be applied.

## Issues

fix #1683

### Checklist

### DIFF
--- a/packages/charts/src/chart_types/xy_chart/utils/fill_series.ts
+++ b/packages/charts/src/chart_types/xy_chart/utils/fill_series.ts
@@ -6,29 +6,25 @@
  * Side Public License, v 1.
  */
 
-import { ScaleType } from '../../../scales/constants';
 import { DataSeries } from './series';
-import { BasicSeriesSpec, isLineSeriesSpec, isAreaSeriesSpec } from './specs';
+import { isAreaSeriesSpec } from './specs';
 
 /**
  * @internal
  */
-export function fillSeries(
-  dataSeries: DataSeries[],
-  xValues: Set<string | number>,
-  groupScaleType: ScaleType,
-): DataSeries[] {
+export function fillSeries(dataSeries: DataSeries[], xValues: Set<string | number>): DataSeries[] {
   const sortedXValues = [...xValues.values()];
   return dataSeries.map((series) => {
     const { spec, data, isStacked } = series;
 
-    const noFillRequired = isXFillNotRequired(spec, groupScaleType, isStacked);
-    if (data.length === xValues.size || noFillRequired) {
+    if (!isStacked || !isAreaSeriesSpec(spec)) {
       return {
         ...series,
         data,
       };
     }
+    // fill x value only to stacked area charts. All other charts doesn't need that data processing
+    // it used to correctly stack areas where a datapoint is not present on one of the stacked area dataset.
     const filledData: typeof data = [];
     const missingValues = new Set(xValues);
     for (let i = 0; i < data.length; i++) {
@@ -60,13 +56,4 @@ export function fillSeries(
       data: filledData,
     };
   });
-}
-
-function isXFillNotRequired(spec: BasicSeriesSpec, groupScaleType: ScaleType, isStacked: boolean) {
-  const onlyNoFitAreaLine = (isAreaSeriesSpec(spec) || isLineSeriesSpec(spec)) && !spec.fit;
-  const onlyContinuous =
-    groupScaleType === ScaleType.Linear ||
-    groupScaleType === ScaleType.LinearBinary ||
-    groupScaleType === ScaleType.Time;
-  return onlyNoFitAreaLine && onlyContinuous && !isStacked;
 }
--- a/packages/charts/src/chart_types/xy_chart/state/utils/utils.ts
+++ b/packages/charts/src/chart_types/xy_chart/state/utils/utils.ts
@@ -134,7 +134,7 @@ export function computeSeriesDomains(
   const xDomain = mergeXDomain(scaleConfigs.x, xValues, fallbackScale);
 
   // fill series with missing x values
-  const filledDataSeries = fillSeries(dataSeries, xValues, xDomain.type);
+  const filledDataSeries = fillSeries(dataSeries, xValues);
 
   const seriesSortFn = getRenderingCompareFn((a: SeriesIdentifier, b: SeriesIdentifier) => {
     return defaultXYSeriesSort(a as DataSeries, b as DataSeries);
--- a/storybook/stories/area/22_fit_and_fill_band_area_chart.story.tsx
+++ b/storybook/stories/area/22_fit_and_fill_band_area_chart.story.tsx
@@ -1,0 +1,78 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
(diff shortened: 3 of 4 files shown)


packages/charts/src/chart_types/xy_chart/utils/series.test.ts
```diff
@@ -289,7 +289,7 @@ describe('Series', () => {
     const stackedValues = formatStackedDataSeriesValues(dataSeries, xValues, 'bar');
     expect(stackedValues.map(matchOnlyDataSeriesLegacySnapshot)).toMatchSnapshot();
   });
-  test('Can stack simple dataseries with scale to extent', () => {
+  test('should not fill NA data points when stacking bars', () => {
```
markov00, line 292: Test renamed and snapshot fixed

packages/charts/src/chart_types/xy_chart/utils/series.test.ts
```diff
@@ -390,29 +391,26 @@ describe('Series', () => {
     const { formattedDataSeries } = computeSeriesDomainsSelector(store.getState());
     expect(formattedDataSeries.map(matchOnlyDataSeriesLegacySnapshot)).toMatchSnapshot();
   });
-  test('Can stack simple dataseries with scale to extent with y0', () => {
+  test('should automatically fill NA data points when stacking areas', () => {
```
markov00, line 394: Test replaced with a more useful one

packages/charts/src/chart_types/xy_chart/utils/fill_series.ts
```diff
@@ -6,29 +6,25 @@
...
 
-    const noFillRequired = isXFillNotRequired(spec, groupScaleType, isStacked);
-    if (data.length === xValues.size || noFillRequired) {
+    if (!isStacked || !isAreaSeriesSpec(spec)) {
       return {
         ...series,
         data,
       };
     }
+    // fill x value only to stacked area charts. All other charts doesn't need that data processing
```
nickofthyme, line 26:
```suggestion
    // fill x value only for stacked area charts. All other charts don't need this data processing
```

packages/charts/src/chart_types/xy_chart/utils/fill_series.ts
```diff
@@ -61,12 +57,3 @@ export function fillSeries(
     };
   });
 }
-
-function isXFillNotRequired(spec: BasicSeriesSpec, groupScaleType: ScaleType, isStacked: boolean) {
-  const onlyNoFitAreaLine = (isAreaSeriesSpec(spec) || isLineSeriesSpec(spec)) && !spec.fit;
-  const onlyContinuous =
-    groupScaleType === ScaleType.Linear ||
-    groupScaleType === ScaleType.LinearBinary ||
-    groupScaleType === ScaleType.Time;
```
nickofthyme, line 70: Do you recall why this was a condition before?

packages/charts/src/chart_types/xy_chart/utils/fill_series.ts
```diff
@@ -6,29 +6,25 @@
...
-  groupScaleType: ScaleType,
-): DataSeries[] {
+export function fillSeries(dataSeries: DataSeries[], xValues: Set<string | number>): DataSeries[] {
   const sortedXValues = [...xValues.values()];
   return dataSeries.map((series) => {
     const { spec, data, isStacked } = series;
 
-    const noFillRequired = isXFillNotRequired(spec, groupScaleType, isStacked);
-    if (data.length === xValues.size || noFillRequired) {
+    if (!isStacked || !isAreaSeriesSpec(spec)) {
```
nickofthyme, line 20: I agree with this but it seems we are missing a condition that causes regressive changes in the VRT screenshots. I think filling is not required on `LineSeries` for stacking but it is required for the separate series to know what values they are missing, or do they just act independently.

I think these are wrong 👇🏼.

image

image

nickofthyme (changes requested):

packages/charts/src/chart_types/xy_chart/utils/fill_series.ts
markov00: Actually no, it also looks strange because we don't consider log/sqrt as continuous scales here

packages/charts/src/chart_types/xy_chart/utils/fill_series.ts
markov00: The fact is that the there is no missing Y values on those series. If a Y value is marked as `null`, then I can clearly describe it with a dashed line and apply an interpolation function. But if that doesn't happen,  then the line should be considered continuous, even if it has less or a different set of X values.
I'm preparing some docs/slides to describe that behaviour, because I think I've also found another issue in our current code

nickofthyme: Discussed idea of separate series when _independent_ otherwise when series are coupled they could fill empty values.

PR #1693, opened 2022-06-01
fix(annotations): animated fade trigger

nickofthyme:
## Summary

Adds `trigger` option to specify which triggers to act of thus allowing the user to opt-in to only the animations they desire for a given `Spec`.

BREAKING CHANGES: moves the animation configuration from the `AnnotationSpecStyles` to the `AnnotationSpec` and removes existing prop on the styles.

Screen Recording 2022-06-02 at 09 08 14 PM

## Details

Removed `animation` prop on `AnnotationSpecStyle`.

### Checklist


nickofthyme:
@markov00 Still need to iron out a few kinks with the marker hover style but would still like to get your feedback, especially on naming and the API changes.

packages/charts/src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -836,6 +837,19 @@ export interface LineAnnotationDatum<D = any> {
...
+export const AnnotationAnimationTrigger = Object.freeze({
+  FadeOnFocusingOthers: 'FadeOnFocusingOthers' as const,
+});
+/** @public */
+export type AnnotationAnimationTrigger = $Values<typeof AnnotationAnimationTrigger>;
+
+/** @public */
+export interface AnnotationAnimations {
+  triggers?: AnnotationAnimationTrigger[];
+  options?: AnimationOptions;
```
markov00, line 850: I'm a bit concerned about this choice: the type name is `AnnotationAnimations` but we have a single `AnimationOptions` configuration that partially describes the animation (it describes basically duration and delay) but the same configurations are applied to multiple "triggers"
A trigger is a key that describes the actual animation and what should trigger it `fade` `on` `focusing others` and at this stage is fine for me, but I don't understand why we have single option per multiple triggers. If we have something like "blink on highlight" we probably want a different configuration, don't we?

I see this being rewritten as:

```ts
type LineAnnotationSpec {
animations?: Array<{trigger: Trigger, options: Options}>;
}
```

In this way, each trigger can have its own options set

nickofthyme: 691f6a5

packages/charts/src/utils/themes/theme.ts
```diff
@@ -713,6 +713,8 @@ export interface LineAnnotationStyle {
   line: StrokeStyle & Opacity & Partial<StrokeDashArray>;
   /**
    * Animation options
+   *
+   * @deprecated use `animation.options` prop on annotation specs
```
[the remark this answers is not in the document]
nickofthyme: 691f6a5

packages/charts/src/chart_types/xy_chart/utils/specs.ts
```diff
@@ -836,6 +837,19 @@ export interface LineAnnotationDatum<D = any> {
   header?: string;
 }
 
+/** @public */
+export const AnnotationAnimationTrigger = Object.freeze({
+  FadeOnFocusingOthers: 'FadeOnFocusingOthers' as const,
```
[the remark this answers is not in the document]
nickofthyme: Based on zoom chat I think many triggers can be linked to one animated key/value.

.gitignore
```diff
@@ -5,6 +5,7 @@ test/spec/screenshots/diff/**/*.png
 test/spec/screenshots/screen/**/*.png
 test/failure-screenshots/**/*.png
 
+.env
```
markov00, line 8: not related to that change:
I'm seeing an issue where the first line marker always goes back to full opacity when moving the mouse over other elements

https://user-images.githubusercontent.com/1421091/171808778-6539e803-521b-491e-9a71-6f12f8ddc328.mov

nickofthyme: Are you talking about the crosshair? That is separate from the annotation.

nickofthyme: Strange, I can't reproduce this locally on Chrome. Are you still seeing this? If so what browser?

Screen Recording 2022-06-03 at 11 50 37 AM

markov00: I removed the cache and looks correct now, strange that's the first time I see a cache problem like that

packages/charts/src/chart_types/xy_chart/utils/specs.ts
markov00: Looks good to me now! Thanks

packages/charts/src/utils/themes/theme.ts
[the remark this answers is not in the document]
markov00: Perfect

markov00 (approved):
Changes looks good to me.
The issues with the line marker were solved by clearing the cache of the deployment page

nickofthyme:
Confirmed these changes fix issues with `markerBody` mentioned in https://github.com/elastic/kibana/pull/131642#discussion_r881509220

Screen Recording 2022-06-06 at 07 30 44 PM

PR #1694, opened 2022-06-01
feat(flame): pulsate node on click focus and search hit focus

monfera:
## Summary

As the chart zooms and pans, it's useful to briefly pulsate the activated node.

hit pulsate

It's a short pulsation for click focus, because
- the user knows where he/she clicked, it's a direct interaction
- the node will already be present in the current visible area

It's a long pulsation for the search hit focus, because
- the user doesn't know where it'll arrive
- there is no visual indicator for what the actual hit is
- esp. with the current, suboptimal pan (no zoom out/in) it takes time for the hit to arrive into the visible area

A more permanent, additional marker (eg. magenta border around the node, which stays until the next interaction not including hovering) is planned for a separate PR. It's a former item titled _Cycling through matches should have a more salient style on the actively selected element_ in the linked issue.

## Details

Partial fulfilment of https://github.com/elastic/elastic-charts/issues/1671

### Checklist

### DIFF
--- a/packages/charts/src/chart_types/flame_chart/shaders.ts
+++ b/packages/charts/src/chart_types/flame_chart/shaders.ts
@@ -47,7 +47,9 @@ const uniformDefs = /* language=GLSL */ `
     float rowHeight1;
     float t; // 0: start position; 1: end position
     float cornerRadiusPx;
-    int hoverIndex;
+    float hoverIndex;
+    float wobbleIndex;
+    float wobble;
     bool pickLayer;
   };
 `;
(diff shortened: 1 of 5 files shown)


packages/charts/src/chart_types/flame_chart/shaders.ts
```diff
@@ -47,7 +47,9 @@ const uniformDefs = /* language=GLSL */ `
     float rowHeight1;
     float t; // 0: start position; 1: end position
     float cornerRadiusPx;
-    int hoverIndex;
+    float hoverIndex;
```
markov00, line 50: why it's now a float? is due to the Buffer offset calculation that doesn't change much for such types?

markov00 (approved):
Looks good to me, the wobble effect looks better than the on/off one

packages/charts/src/chart_types/flame_chart/shaders.ts
monfera: There seems to be a compat issue with integers in uniform buffers, I'll have to investigate. Integers are new with WebGL2 and less reliable. It might also solve a problem encountered on a Debian linux system

PR #1695, opened 2022-06-02
fix(flame): cross browser issues

monfera:
## Summary

Fixes:
- the Safari Metal shader compiler has a bug (causing no render at all) that needed a 1-liner workaround, see the [1st commit](https://github.com/elastic/elastic-charts/pull/1695/commits/599db56f4cd240aa43e8bdd1c81127ab97f340e5).
- click focusing with certain browser zoom levels (eg. 110% in Chrome) and pinch zoom stopped working, due to a regression introduced with these [lines](https://github.com/elastic/elastic-charts/pull/1682/files#diff-f017ed451ccf2474f26b0e480ee6ade71500e04765029c93afd6925249d205bfR365-R371) - thanks @ghudgins for the report

The PR also contains 
- visual improvements to the blinking focus indicator: 
  - when clicking on a node, or advancing to another text search hit, the pulsation makes full periods, so the end of the pulsation is smooth
  - the pulsation length is now the same for click focus and search hit focus
- the tooltip won't show up during the pulsation (therefore, approx. during the focus animation)
- a test for congruent column array lengths

## Details

The workaround is that some operation had to be done with `wobble` in the `true` branch, see the first commit. All these worked:
- `float(wobble)` (it's already a float, of course)
- `wobble + 0.0`

To improve on the node blinking, I needed to switch to `1.0 - wobble` which is good too

Note: there are other less glaring cross browser issues under investigation.

### Checklist

### DIFF
--- a/packages/charts/src/chart_types/flame_chart/flame_chart.tsx
+++ b/packages/charts/src/chart_types/flame_chart/flame_chart.tsx
@@ -871,6 +871,10 @@ class FlameComponent extends React.Component<FlameProps> {
         this.animationRafId = window.requestAnimationFrame(anim);
       } else {
         this.prevT = NaN;
+        this.currentFocus.x0 = this.targetFocus.x0;
+        this.currentFocus.x1 = this.targetFocus.x1;
+        this.currentFocus.y0 = this.targetFocus.y0;
+        this.currentFocus.y1 = this.targetFocus.y1;
       }
     };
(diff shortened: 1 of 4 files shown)


markov00 (approved):
Looks good to me, tested in Safari, Firefox and Chrome.
Whenever Playwright VRTs will be enabled by default we can also cross-test those changes in different browsers

PR #1669, opened 2022-05-10
fix(deps): update dependency d3-scale to v3.3.0

watson:
I'm performing this upgrade to get a newer version of `d3-color` which is a direct dependency of `d3-scale`.

I think this is the only commit to d3-scale v2.0.0 with potentially breaking changes: https://github.com/d3/d3-scale/commit/78193364284f56c4bdfcf4e8a768d558b4bb379b

I think that we should be ok with that in this package, but hopefully CI will let us know.

## To Reviewers

This PR also fixes the issue listed here: https://github.com/elastic/elastic-charts/issues/1095
The new d3 scale version project a datapoint to the center of its range if the domain min === domain max see https://github.com/d3/d3-scale/issues/117

Issue #1095, opened 2021-03-30
Better handling of linear scale with a single data point and no domain


markov00:
**Is your feature request related to a problem? Please describe.**
If the dataset contains only a single datum, we can't compute a complete domain, the min and max are the same.
The datum is rendered at the beginning of the relative axis, making it difficult to read and easily understand.
We cannot even extend the domain automatically by a +- static value, because the domain depends on the current metric unit, and we can't assume our static value is within the context range.

**Describe the solution you'd like**
For such specific use case, where an axis uses a linear scale and the domain is computed with a single value, we should change the `range` of the scale from `0, width` to `width/2, width/2` so that the point is rendered exactly in the center of the axis.

**Describe alternatives you've considered**
Extend the domain by a static value `[datum_value - static, datum_value + static]` but computing a valid value to add and subtract is non-trivial and some heuristics needs to be computed before doing that.

**Additional context**
Originally requested screenshot
![](https://user-images.githubusercontent.com/52681807/109020351-bdef7880-766e-11eb-8506-b0cc525c35fb.png)

**Kibana Cross Issues**
- https://github.com/elastic/kibana/issues/92631
- https://github.com/elastic/kibana/issues/108442

markov00:
we should/can consider also the case with a dataset of two data points. In this case both points are on the respective edges and this can be wrongly understood by the reader. A domain padding is probably what we are looking for in this case. Not sure if this is a consumer configuration or an internal behaviour

markov00:
closed by https://github.com/elastic/elastic-charts/pull/1696

PR #1696, opened 2022-06-03
refactor(scales): use specific scales instead of generics

markov00:
## Summary

This PR is the replacement of https://github.com/elastic/elastic-charts/pull/1669 and contains:
- an update of d3-scale to `v3.3.0`
- Refactoring of the scale types, making a clear distinction between a categorical and continuous scales (this helped the types updated caused by the new d3-scale version and fixed some tech debt)

### Checklist


packages/charts/src/chart_types/xy_chart/axes/timeslip/multilayer_ticks.ts
```diff
@@ -40,7 +40,7 @@ export function multilayerAxisEntry(
   extendByOneBin: boolean,
   range: [number, number],
   timeAxisLayerCount: any,
-  scale: Scale<string | number> | ScaleContinuous, // fixme it's only the latter for now
+  scale: ScaleBand | ScaleContinuous, // fixme it's only the latter for now
```
monfera, line 43: Would this be a good opportunity to adhere to the code comment in this row, or would it be too tedious now? Or maybe the thinking is, let's keep it for the future when `ScaleBand` is indeed supported

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -49,8 +49,8 @@ export function renderBars(
   const isInvertedY = yScale.isInverted;
   return dataSeries.data.reduce((barTuple: BarTuple, datum) => {
     const xScaled = xScale.scale(datum.x);
-    if (!xScale.isValueInDomain(datum.x) || Number.isNaN(xScaled)) {
-      return barTuple; // don't create a bar if not within the xScale domain
+    if (Number.isNaN(xScaled)) {
+      return barTuple; // a NaN value means the datum is not within the scale range
```
monfera, line 53: not within the scale domain?

packages/charts/src/chart_types/xy_chart/rendering/points.ts
```diff
@@ -109,7 +107,8 @@ export function renderPoints(
         panel,
         orphan,
       };
-      const isInYDomain = yScale.isValueInDomain(valueAccessor(datum));
+      const yValue = valueAccessor(datum);
+      const isInYDomain = !isNil(yValue) && yScale.isValueInDomain(yValue);
```
monfera, line 111: Can the accessor function be typed so it only yields `NaN`? And maybe `yScale.isValueInDomain` already returns `false` for everything that is `!Number.isFinite(yValue)`

packages/charts/src/chart_types/xy_chart/state/selectors/get_cursor_band.ts
```diff
@@ -71,6 +73,9 @@ function getCursorBand(
   // external pointer events takes precedence over the current mouse pointer
   if (isValidPointerOverEvent(xScale, externalPointerEvent)) {
     fromExternalEvent = true;
+    if (isNil(externalPointerEvent.x)) {
+      return;
+    }
```
monfera, line 78: Is it not redundant? My drive is the preferable avoidance of:
- code lines
- redundancy / repetition
- branching
- more than the necessary return counts from a function

I guess it wasn't added for the sole purpose of saving on some cycles (avoid execution of row 79) because that wouldn't be worth it, but maybe even slightly deeper causes could be solved

packages/charts/src/chart_types/xy_chart/state/selectors/get_cursor_band.ts
```diff
@@ -86,14 +91,18 @@ function getCursorBand(
...
       return;
     }
   }
   const { horizontal, vertical } = smallMultipleScales;
-  const topPos = vertical.scale(pointerPosition.verticalPanelValue) || 0;
-  const leftPos = horizontal.scale(pointerPosition.horizontalPanelValue) || 0;
+  const topPos =
+    (!isNil(pointerPosition.verticalPanelValue) && vertical.scale(pointerPosition.verticalPanelValue)) || 0;
+  const leftPos =
+    (!isNil(pointerPosition.horizontalPanelValue) && horizontal.scale(pointerPosition.horizontalPanelValue)) || 0;
```
monfera, line 105: Can we do without the `isNil` checks? It's used in a number of places, which suggests that maybe the "type leak" can be contained at a point much closer to its origination

packages/charts/src/chart_types/xy_chart/state/selectors/get_elements_at_cursor_pos.ts
```diff
@@ -46,6 +48,10 @@ function getElementAtCursorPosition(
   { chartDimensions }: ChartDimensions,
 ): IndexedGeometry[] {
   if (isValidPointerOverEvent(scales.xScale, externalPointerEvent)) {
+    if (isNil(externalPointerEvent.x)) {
+      return [];
+    }
+
```
monfera, line 54: Same here

packages/charts/src/chart_types/xy_chart/state/selectors/get_elements_at_cursor_pos.ts
```diff
@@ -54,14 +60,16 @@ function getElementAtCursorPosition(
     // TODO: Handle external event with spatial points
     return geometriesIndex.find(externalPointerEvent.x, { x: -1, y: -1 });
   }
-  const xValue = scales.xScale.invertWithStep(orientedProjectedPointerPosition.x, geometriesIndexKeys);
-  if (!xValue) {
+  const xValue = isContinuousScale(scales.xScale)
+    ? scales.xScale.invertWithStep(orientedProjectedPointerPosition.x, geometriesIndexKeys as number[]).value // TODO FIX this
+    : scales.xScale.invert(orientedProjectedPointerPosition.x);
+  if (isNil(xValue) || Number.isNaN(xValue)) {
```
monfera, line 66: And here

packages/charts/src/chart_types/xy_chart/state/selectors/get_projected_scaled_values.ts
```diff
@@ -24,13 +25,15 @@ export const getProjectedScaledValues = createCustomCachedSelector(
       return;
     }
 
-    const xValue = xScale.invertWithStep(x, geometriesIndexKeys);
+    const xValue = isContinuousScale(xScale)
+      ? xScale.invertWithStep(x, geometriesIndexKeys as number[]).value // TODO fix this cast
+      : xScale.invert(x);
```
monfera, line 30: Why is it that there was no need for a ternary here and there is now?

monfera: Also, if the `invertWithStep` needs the `geometriesIndexKeys`, and —I'm gonna make a big assumption here— `geometriesIndexKeys` is not going to change, AND `xScale` is an object of our liking, why can't this scale object receive `geometriesIndexKeys` as a parameter for its constructor function?

packages/charts/src/chart_types/xy_chart/state/selectors/get_tooltip_position.ts
```diff
@@ -36,8 +37,14 @@ export const getTooltipAnchorPositionSelector = createCustomCachedSelector(
...
-    const topPos = vertical.scale(projectedPointerPosition.verticalPanelValue) || 0;
-    const leftPos = horizontal.scale(projectedPointerPosition.horizontalPanelValue) || 0;
+    const topPos =
+      (!isNil(projectedPointerPosition.verticalPanelValue) &&
+        vertical.scale(projectedPointerPosition.verticalPanelValue)) ||
+      0;
+    const leftPos =
+      (!isNil(projectedPointerPosition.horizontalPanelValue) &&
+        horizontal.scale(projectedPointerPosition.horizontalPanelValue)) ||
+      0;
```
monfera, line 47: Same

packages/charts/src/chart_types/xy_chart/state/selectors/get_tooltip_values_highlighted_geoms.ts
```diff
@@ -100,6 +101,9 @@ function getTooltipAndHighlightFromValue(
   let tooltipType = getTooltipType(settings);
   if (isValidPointerOverEvent(scales.xScale, externalPointerEvent)) {
     tooltipType = getTooltipType(settings, true);
+    if (isNil(externalPointerEvent.x)) {
+      return EMPTY_VALUES;
+    }
```
monfera, line 106: Same

packages/charts/src/chart_types/xy_chart/state/selectors/on_pointer_move_caller.ts
```diff
@@ -46,16 +48,16 @@ function getPointerEvent(
   if (x === -1 || y === -1) {
     return { chartId, type: PointerEventType.Out };
   }
-  const xValue = xScale.invertWithStep(x, geometriesIndexKeys);
-  if (!xValue) {
+  const xValue = isContinuousScale(xScale) ? xScale.invertWithStep(x, geometriesIndexKeys).value : xScale.invert(x);
+  if (isNil(xValue)) {
     return { chartId, type: PointerEventType.Out };
   }
```
monfera, line 54: Same

packages/charts/src/chart_types/xy_chart/state/utils/utils.ts
```diff
@@ -353,8 +353,8 @@ function renderGeometries(
 
     const { stackMode } = ds;
 
-    const leftPos = smHScale.scale(ds.smHorizontalAccessorValue) || 0;
-    const topPos = smVScale.scale(ds.smVerticalAccessorValue) || 0;
+    const leftPos = (!isNil(ds.smHorizontalAccessorValue) && smHScale.scale(ds.smHorizontalAccessorValue)) || 0;
+    const topPos = (!isNil(ds.smVerticalAccessorValue) && smVScale.scale(ds.smVerticalAccessorValue)) || 0;
```
monfera, line 357: Same

packages/charts/src/scales/scale_continuous.ts
```diff
@@ -165,21 +176,22 @@ export class ScaleContinuous implements Scale<number> {
         : (d3Scale as D3ScaleNonTime).ticks(scaleOptions.desiredTickCount);
 
     this.domain = nicePaddedDomain;
-    this.project = (d: number) => d3Scale(d);
-    this.inverseProject = (d: number) => d3Scale.invert(d);
+    // Returning NaN means that the value is projectable/invertible within the domain or range
+    this.project = (d: number) => d3Scale(d) ?? NaN;
```
monfera, line 180: OK so the continuous scale projection will not yield anything other than `number` (it's also true from the type signature of `this.project`

packages/charts/src/scales/scale_continuous.ts
```diff
@@ -165,21 +176,22 @@ export class ScaleContinuous implements Scale<number> {
...
+    this.inverseProject = (d: number) => d3Scale.invert(d) ?? NaN;
   }
 
-  scale(value?: PrimitiveValue) {
+  scale(value?: PrimitiveValue): number {
```
monfera, line 184: Under what scenarios can this get an `undefined` or a non-`number`?

packages/charts/src/scales/scale_continuous.ts
```diff
@@ -232,15 +244,13 @@ export class ScaleContinuous implements Scale<number> {
...
   }
 
-  isSingleValue() {
+  isSingleValue(): boolean {
     return this.isSingleValueHistogram || isDegenerateDomain(this.domain);
   }
 
-  isValueInDomain(value: number) {
-    return this.domain[0] <= value && value <= this.domain[1];
+  isValueInDomain(value: unknown): boolean {
```
monfera, line 251: A code change elsewhere may indeed lead to a situation where we ask a function like `isValueInDomain` to do a bit more than before. Yet it feels possible to (eventually?) statically ensure that the continuous scale won't ever be called with something other than a number, making this change unnecessary. If this is not the case, I'm interested in what prevents this, as we type-control user supplied accessor functions

packages/charts/src/state/selectors/is_external_tooltip_visible.ts
```diff
@@ -26,7 +28,12 @@ export const isExternalTooltipVisibleSelector = createCustomCachedSelector(
     computeChartDimensionsSelector,
   ],
   ({ externalPointerEvents }, hasExternalEvent, pointer, { xScale }, { chartDimensions }): boolean => {
-    if (!pointer || pointer.type !== PointerEventType.Over || externalPointerEvents.tooltip?.visible === false) {
+    if (
+      !pointer ||
+      pointer.type !== PointerEventType.Over ||
+      isNil(pointer.x) ||
+      externalPointerEvents.tooltip?.visible === false
+    ) {
```
monfera, line 36: Another `isNil`

packages/charts/src/chart_types/xy_chart/rendering/points.ts
monfera: A most likely inconsequential point: `isNil` doesn't guarantee finiteness, though I suppose there's no known risk of infinities (and the scale can handle that, as `<=` etc. work with infinities

packages/charts/src/scales/scale_band.ts
```diff
@@ -63,56 +50,54 @@ export class ScaleBand<T extends number | string> implements Scale<T> {
...
+
+    this.minInterval = 0; // FIXED doesn't exist in reality
+    this.project = (d) => d3Scale(d) ?? NaN;
+    this.inverseProject = (d) => invertedScale(d);
   }
 
-  scale(value?: PrimitiveValue) {
-    const scaleValue = this.d3Scale(stringifyNullsUndefined(value));
-    return typeof scaleValue === 'number' && Number.isFinite(scaleValue) ? scaleValue : NaN; // fixme when TS improves
+  scale(value: string | number): number {
```
monfera, line 79: Would be neat to extract out the numerous appearances of `string | number`, specifically in places where it semantically means, domain value of a band scale? Ie. the extracted type would be something like `ScaleBandDomainValue`, so not just a drop-in for all kinds of `string | number` (oh when will TS have nominal types)

monfera: One more note if we want to really take TS seriously (of which I'm not sure):
The domain of discrete functions, such as `ScaleBand` `scale`, is `string | number` or let's call it
```
type  ScaleBandDomainValue = string | number
```
But the [natural domain](https://en.wikipedia.org/wiki/Domain_of_a_function#Natural_domain) of the scale function is much more restricted: maybe it's `1 | 2 | 3 | 4`.

So the "proper" TS thing to do (again, debatable value as optimal TS-ness boundary is subjective and varies widely) would be to make the chart, and the scale generic on the natural domain. This would also make a lot of current runtime checks unnecessary

packages/charts/src/scales/scale_band.ts
```diff
@@ -63,56 +50,54 @@ export class ScaleBand<T extends number | string> implements Scale<T> {
...
 
   isSingleValue() {
     return this.domain.length < 2;
   }
 
-  isValueInDomain(value: T) {
-    return this.domain.includes(value);
+  isValueInDomain(value: unknown) {
+    return this.domain.includes(value as string | number);
   }
```
monfera, line 102: ... so, assuming that the chart / scale is generic on the natural domain, this function—if still necessary—can become an `is` predicate, because it'd decide whether an unknown value is in the natural domain or not

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
markov00: Yea, sorry, I bring it back after checking the tests

packages/charts/src/chart_types/xy_chart/rendering/points.ts
markov00: You are right, `isValueinDomain` already check the type of its input, so that `isNil` is not required.
I've slightly changed to check for finiteness inside the `isValueInDomain` so we can catch multiple edge cases https://github.com/elastic/elastic-charts/pull/1696/commits/5b512a6c99e60af2978d74ba347a4ddfc6ee4c89

packages/charts/src/chart_types/xy_chart/state/selectors/get_cursor_band.ts
markov00: This is what we discussed last week: where should I put the check for values that cannot be scaled?
currently, the scale function is typed as the following:
```
scale(value: string | number): number {
```
that means that the input check needs to be done before the `scale`   call.

I can probably intercept a bit earlier, but it should be checked anyway. that data arrives from an external event, we can change its type to avoid such null check completely, but this cause effects to the overall chain: a breaking change needs to be issued, kibana needs to be updated and everything that breaks needs to be fixed, and probably that null check needs to be applied on every line sending that event to the chart.

I see your point, but IMHO that doesn't yield an immediate benefit.

packages/charts/src/chart_types/xy_chart/state/selectors/get_cursor_band.ts
markov00: Yes it could, but not in this PR. As mentioned in https://github.com/elastic/elastic-charts/pull/1696#discussion_r889979775 the check is required since `scale` can't accept null values.
I can probably bring the `nil` check at an earlier phase, but it will be just moved somewhere else, I don't see much the ability to remove it completely.

packages/charts/src/chart_types/xy_chart/state/selectors/get_elements_at_cursor_pos.ts
markov00: probably, but there is a difference here: a categorical scale return `undefined` if the value can't be inverted, but a numerical/continuous scale return `NaN`. We should check for both here or try to be very smart and be sure that we will never invert a value that is not invertible but blocking it at an earlier stage.
That basically moves the checks about a value invertibility at an earlier stage, but leaves the usage of `invert` to "malicious" usage. I can't be sure that no one of us will never call `invert` with an non-invertible value.
I prefer to add a check, rather then leave the door open to bugs

packages/charts/src/chart_types/xy_chart/state/selectors/get_projected_scaled_values.ts
markov00: I've reintroduced the `invertWithStep` here https://github.com/elastic/elastic-charts/pull/1696/commits/c03e76de605f30e5d7db4116825de91c7b9598f5
As already described, the return type of `invert` is different between a categorical scale and continuous one, so I need to add the subsequent checks to verify one or the other case.

packages/charts/src/chart_types/xy_chart/state/selectors/on_pointer_move_caller.ts
markov00: addressed the ternary here https://github.com/elastic/elastic-charts/commit/c03e76de605f30e5d7db4116825de91c7b9598f5 but the check remain

packages/charts/src/scales/scale_continuous.ts
markov00: actually should be none. This was the original signature, since I've changed the one used in the categorical scale I'm going to change also this one to accept only number

packages/charts/src/scales/scale_continuous.ts
markov00: >as we type-control user supplied accessor functions

never trust your users :D 
As always, this requires an extensive rework of our external API, causing breaking changes to flow down to Kibana at least. We can also add a guard layer right after the API, but this is actually a big work that needs to be done and I don't consider it as a big priority right now, it doesn't yield to a massive gain in performance, probably an improved code readability but I don't see it as a potential blocking issue. Happy to change my mind

packages/charts/src/scales/scale_band.ts
markov00: And how would you define that generic domain from the runtime? I think we couldn't reach that, at least not in the environment we are working on, where the values can't be predicted. 
In a static visualization environment, where I perfectly know my data, this can be done, but I have doubts this can be applied to data coming from "databases"

packages/charts/src/scales/scale_continuous.ts
monfera: I agree it's not a big priority now. I've just been pondering this as responses like this are informative. The "clean" ends of the spectrum, ie. 
- typescript-handled enumerations for categorical data, relying on generics and static assurances
- standard javascript, with runtime checks
feel straightforward to me.

We're sometimes OK relying on static assurances, and that users of the library adhere to the spec, and sometimes we don't want to trust users, and the boundary feels arbitrary (driven by past/present priorities and historical code path, rather than something deeper than that).

I still agree it's not the most urgent thing, and that it's probably easier to desire static category assurances than shaping our code toward it. 

So maybe we can agree that it'd be the right thing to do (if we continue toward idiomatic TS use) but we don't prioritize it, or we could discuss why certain things fall the "static assurances" way while other things fall the "don't trust your users" way (and sometimes, but not always, both)

Either way it's not for this PR, just something to keep learning about

packages/charts/src/scales/scale_band.ts
monfera: Good question, not sure, let me think about it (pls. set it to resolved for the purpose of this PR). One interesting path is the one in Canvas and the expression language (these have an analogous challenge), though I don't have all the details rn so I can't say if it'd solve it. 

While we don't know the data ahead of time, we know that we can't have a legend for what is not a category; a user needs to pass on a label formatter for the given categories; and that our code will only encounter the used categories, avoiding checks that permeate the code now.

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
markov00: bring it back in fbd370edcf255e264def417aa87f808075efe66e

packages/charts/src/scales/scale_continuous.ts
markov00: I have double-checked this: is similar to the `isNil` check we have here and there.
The problem is: the X scale can be either a categorical one or a continuous one. This means that when, in the code, I need to scale an X value, I need to check the used scale type and the value passed into the scale function.
If I change the scale to receive only numbers, everywhere I have an union type of scales, I need to check for the scale type and the value type before calling the scale.
If, instead, I use `scale(value?: PrimitiveValue)` I'm moving the check within the scale function and alleviate the rest of the code from those checks.
The only way to solve this is to apply a generic that specifies the scale type to each function that uses that scale and be sure to pass this along from the beginning of the call graph that generates the scale.
Doable, but out of the scope right now, for the complexity.

packages/charts/src/chart_types/xy_chart/axes/timeslip/multilayer_ticks.ts
markov00: done in d21ce76ec6d9b0a3774ef1b08834a86efd3d5e42

monfera (approved):

nickofthyme:
@markov00 so did you want to link #1095 to this PR and close the issue?

PR #1697, opened 2022-06-03
feat(flame): keyboard shortcuts for the text field

monfera:
## Summary

When using the search field, or even after a manual pan drag on the chart, the following keyboard shortcuts work:
- Enter: next hit
- Shift Enter: previous hit
- Escape: clear search field

## Details

Partial fulfilment of keyboard accessibility in https://github.com/elastic/elastic-charts/issues/1671

### Checklist


markov00 (approved):
Thank you Robert, I've tested it locally and works as expected.
One major thing: I think we should not work too much within the input/search part of the flamegraph. This should go hand in hand with EUI a11y and other conventions we have in Kibana.
This is good anyway to have at least the prev/next hit functionalities setup and expose them relatively easily

PR #1709, opened 2022-06-10
fix(flame): no focus border

monfera:
## Summary

Removes the keyboard focus border that interferes with the zoom/pan focus indicators on the top and left. A future PR needs to address chart focusability including adding the missing keyboard interactions for navigation. Also contains a wordcloud story fix ht Marco

## Details

In current `master`, the issue is reproducible best in Chrome, when clicking on the large chart, then hitting the Escape or some other key. A keyboard focus line on top and left appear, coniciding with the rendered zoom/pan focus lines

### Checklist

### DIFF
--- a/packages/charts/src/chart_types/flame_chart/flame_chart.tsx
+++ b/packages/charts/src/chart_types/flame_chart/flame_chart.tsx
@@ -695,7 +695,7 @@ class FlameComponent extends React.Component<FlameProps> {
             onKeyPress={this.handleEnterKey}
             onKeyUp={this.handleEscapeKey}
             onWheel={this.handleWheel}
-            style={style}
+            style={{ ...style, outline: 'none' }}
             // eslint-disable-next-line jsx-a11y/no-interactive-element-to-noninteractive-role
             role="presentation"
           />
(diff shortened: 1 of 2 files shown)


packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -695,7 +695,7 @@ class FlameComponent extends React.Component<FlameProps> {
             onKeyPress={this.handleEnterKey}
             onKeyUp={this.handleEscapeKey}
             onWheel={this.handleWheel}
-            style={style}
+            style={{ ...style, outline: 'none' }}
```
markov00, line 698: More than a fix is a temporary patch, because the canvas needs to be focusable due to the enter/esc key press to navigate and having a clear signal that where the focus is can help a lot the navigation.
Fine for now anyway.

This, solves the black outline, but doesn't solves a problem that still exists: pressing two time the tab, to move from the canvas focus to the input focus we are pushing the canvas up by 1 or 2 pixels losing the top bar that shows where I am zooming in the overall chart 

https://user-images.githubusercontent.com/1421091/173093113-2581fca8-b22f-4041-aa22-9e19e042b162.mov

monfera: Thanks, good to know! Keyboard navigation between the main chart and form elements is still unsupported, and as we discussed, the controls would live outside the chart proper. Still, I'll check if there's a better way for the next PR

markov00:
jenkins test this please

PR #1710, opened 2022-06-10
fix: return always a full projection in multilayerAxisEntry

### LINKED ISSUE #1592: `maxLabelBboxHeight` can be undefined
A kibana user experienced an error where the server-side rendering of a dashboard in PDF the following error in the console appeared:  

```
Cannot read properties of undefined (reading 'maxLabelBboxHeight')
```

~We should find the possible root cause of this undefined value causing a blank chart rendered~

As notified in https://github.com/elastic/kibana/issues/134025 the issue appears when resizing a multi-layer time axis to a small size, causing the error to surface

markov00:
## Summary

When resizing too much a time series chart with a multilayer time axis, the function that computes the ticks can receive a a zero-length array of `layers`.  Typescript probably doesn't pick up correctly a mis-typed reduce and it propagated through without getting noticed causing an un-recoverable error as reported in the [linked issue](#https://github.com/elastic/elastic-charts/issues/1592) and the Kibana one.

This PR fixes it by adding the correct return type to the function and fixing the reduce function initialization.

The timeslip story is also updated with a switch to show the legend simplifying the reproducibility of the issue.

## Issues

fix #1592

### Checklist

### DIFF
--- a/packages/charts/src/chart_types/xy_chart/axes/timeslip/multilayer_ticks.ts
+++ b/packages/charts/src/chart_types/xy_chart/axes/timeslip/multilayer_ticks.ts
@@ -9,8 +9,7 @@
 import { Scale, ScaleContinuous } from '../../../../scales';
 import { XDomain } from '../../domains/types';
 import { AxisLabelFormatter } from '../../state/selectors/axis_tick_formatter';
-import { GetMeasuredTicks } from '../../state/selectors/visible_ticks';
-import { AxisTick } from '../../utils/axis_utils';
+import { GetMeasuredTicks, Projection } from '../../state/selectors/visible_ticks';
 import { rasters, TimeBin, TimeRaster } from './rasters';
 
 const MAX_TIME_TICK_COUNT = 50; // this doesn't do much for narrow charts, but limits tick count to a maximum on wider ones
@@ -42,7 +41,7 @@ export function multilayerAxisEntry(
   timeAxisLayerCount: any,
   scale: Scale<string | number> | ScaleContinuous, // fixme it's only the latter for now
   getMeasuredTicks: GetMeasuredTicks,
-) {
+): Projection {
   const rasterSelector = rasters({ minimumTickPixelDistance: 24, locale: 'en-US' }, xDomain.timeZone);
   const domainValues = xDomain.domain; // todo consider a property or object type rename
   const domainFromS = Number(domainValues[0]) / 1000; // todo rely on a type guard or check rather than conversion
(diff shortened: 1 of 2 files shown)


packages/charts/src/chart_types/xy_chart/axes/timeslip/multilayer_ticks.ts
```diff
@@ -63,8 +62,8 @@ export function multilayerAxisEntry(
       fallbackAskedTickCount: NaN,
     };
   };
-  return layers.reduce(
-    (combinedEntry: { ticks: AxisTick[] }, l: TimeRaster<TimeBin>, detailedLayerIndex) => {
+  return layers.reduce<Projection>(
```
monfera, line 65: Nice, totally worth the couple of added lines for the initial value

packages/charts/src/chart_types/xy_chart/axes/timeslip/multilayer_ticks.ts
```diff
@@ -85,8 +84,9 @@ export function multilayerAxisEntry(
       }
 
       return {
-        ...entry,
         ...combinedEntry,
+        ...entry,
+        scale,
```
monfera, line 89: `entry` has the `scale` already (OK to keep it)

monfera (approved):

packages/charts/src/chart_types/xy_chart/axes/timeslip/multilayer_ticks.ts
```diff
@@ -97,6 +97,16 @@ export function multilayerAxisEntry(
...
+      ticks: [],
+      labelBox: {
+        maxLabelBboxWidth: 0,
+        maxLabelBboxHeight: 0,
+        maxLabelTextWidth: 0,
+        maxLabelTextHeight: 0,
+        isHidden: true,
+      },
+      scale,
+    }, // this should turn into a full Projection
```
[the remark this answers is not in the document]
markov00: :D yes

packages/charts/src/chart_types/xy_chart/axes/timeslip/multilayer_ticks.ts
markov00: true, and also combinedEntry :)

markov00: done https://github.com/elastic/elastic-charts/pull/1710/commits/ba7bddfb309e30216358bedf69638ef2ee5911f6

PR #1713, opened 2022-06-15
chore: pick up perf and refactor commits as patch release

markov00:
## Summary

This PR updates the semantic-release commit analyzer to pick up bot the `perf` and the `refactor` commit prefix as patch releases. 

@nickofthyme we should also change the current semantic-pull-request app, because is no longer maintained and seems not working anymore (see this PR title for example), to this https://github.com/Ezard/semantic-prs#readme that has the same API, wdyt?


nickofthyme (approved):
Yeah I think these changes are fine. Did you want to bump `semantic-release` per #1707 in this PR too?

markov00:
> Yeah I think these changes are fine. Did you want to bump `semantic-release` per #1707 in this PR too?

yep why not!


> Yeah I think that is fine with me. Do you think this bot would fix the issues mentioned in #1471?

No, the bug was not caused by the `semantic-pull-request` github app, but by the github plugin of semantic-release

PR #1719, opened 2022-06-21
feat(flame): navigation and controls

monfera:
## Summary

Adds navigation, using a nav history queue:
- Back
- Forth
- Home

Adds exposed control callbacks:
- `resetFocus`
- `focusOnNode`
...more to come

### Checklist


packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -223,8 +223,20 @@ class FlameComponent extends React.Component<FlameProps> {
...
+    const { controlProviderCallback } = this.props;
+    if (controlProviderCallback.resetFocus) {
+      controlProviderCallback.resetFocus(() => this.resetFocus());
+    }
+    if (controlProviderCallback.focusOnNode) {
+      controlProviderCallback.focusOnNode((nodeIndex: number) => {
+        if (this.navQueue[this.navIndex] !== nodeIndex) this.navQueue.splice(++this.navIndex, Infinity, nodeIndex);
+        this.focusOnNode(nodeIndex);
+      });
+    }
```
markov00, line 237: If, for some reason, the user updates the props linking the callbacks in a later stage than the initial mounting event, they will never receive those controls. 
We should probably link those at the mount/update stage

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -420,23 +453,16 @@ class FlameComponent extends React.Component<FlameProps> {
...
-        this.wobbleIndex = hovered.datumIndex;
-        this.prevT = NaN;
-        this.hoverIndex = NaN; // no highlight
-        this.setState({});
-        this.props.onElementClick([{ vmIndex: hovered.datumIndex }]); // userland callback
+        const { datumIndex } = hovered;
+        if (this.navQueue[this.navIndex] !== datumIndex) this.navQueue.splice(++this.navIndex, Infinity, datumIndex);
+        console.log(this.navQueue.join(' > '), this.navIndex);
+        this.focusOnNode(datumIndex);
+        this.props.onElementClick([{ vmIndex: datumIndex }]); // userland callback
```
markov00, line 465: not in this PR, but I was wondering if we need to call this an async way, like in a different requestAnimationFrame. This is just because we can "finish" our rendering part and detach the event from the event handling

monfera: It is already in an event handler, so it gets activated asynchronously. I don't currently see a reason for making the interaction asynchronous, as it's not clear what it would buy, yet it'd definitely involve more code, and maybe somewhat less predictable execution, because another event handler may asynchronously trigger something else, so a game with rAF queues, ordering, cancellation etc. might start. As it is, there's no extra code involved

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
monfera: Thanks for the suggestion, added the binding for `componentDidUpdate` too, addreseed in 90ab4d778bded583558d3768d451101eaf40ffd4 and commented for the API in 2316bdd8764021fffb241b2741a77dbb5a82d12c

storybook/stories/icicle/04_cpu_profile_gl_flame.story.tsx
```diff
@@ -48,11 +49,10 @@ const columnarData = {
   size1: size,
 };
 
-export const Example = (
-  // should check why it's not a good idea; in the meantime:
-  // eslint-disable-next-line unicorn/no-object-as-default-parameter
-  props?: { controlProviderCallback: ControlProviderCallback },
-) => {
+let resetFocusControl: ResetFocusControl = () => {}; // initial value
+let focusOnNodeControl: FocusOnNodeControl = () => {}; // initial value
```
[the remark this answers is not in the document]
monfera: Done in 90ab4d778bded583558d3768d451101eaf40ffd4 made possible by rebinding in the update

markov00 (approved):
Changes looks good to me, tested inside the deployment, works as expected

PR #1730, opened 2022-06-27
fix(events): remove async callbacks from code

markov00:
## Summary

Working at #1729 leads to an interesting issue regarding the asynchronous callback.
In particular, we were calling the `onChartRendered` and `onProjectionAreaChanged` callbacks in a new macro task thanks to a `requestAnimationFrame`.

The reasons to avoid this way of handling callbacks are:
- give the freedom to the user to choose how they want to execute/queue the task in their own environment
- events caused by mouse events are already async by nature
- those callbacks should be used only for small side effects, like triggering a focus, a slight change in the state (like disabling a loading spinner) etc the user should not attach long-running sync methods to these events.

This PR removes the use of RAF in favor of sync calls.

This solves a side issue in the composable story where the chart render-complete was not called correctly.


nickofthyme (approved):
Excellent!! Works great. Thanks for jumping on this so fast! Any idea why we wrapped those in a RAF?

PR #1748, opened 2022-07-11
feat(flame): Add zoom and pan to the navigation history

markov00:
## Summary

This adds the ability to go back and forth to the last zoomed/panned area in the screen.
In particular, the navigation history works the following way:
- a pan/zoom action clears the most recent items in the history starting from the current navigation item and replaces those with an item describing the last visible area after a zoom/pan.
- if a node click action is performed after a pan/zoom action, the pan/zoom history item is replaced by the click action.


markov00:
buildkite update screenshots

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -6,6 +6,14 @@
...
  */
 
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
```
monfera, line 16: Was the image change intended? In case of the initial render, the "home" interaction is a noop, which are often shown as grayed out affordances. Btw. this copyright notice shows up as a change, I don't immediately see what changed in it

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -229,6 +241,9 @@ class FlameComponent extends React.Component<FlameProps> {
     this.bindControls();
     this.currentFocus = { ...this.targetFocus };
 
+    // Initialize nav queue with the root element
+    this.navigator = new NavBtnControlledZoomPanHistory({ ...this.getFocusOnRoot(), index: 0 });
```
monfera, line 245: Nit: let's spell out `Button`, not worth saving three characters when the name is otherwise nice and self explanatory (therefore on the long side). I'm adding so much value!

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -238,21 +253,56 @@ class FlameComponent extends React.Component<FlameProps> {
...
   private bindControls() {
     const { controlProviderCallback } = this.props;
     if (controlProviderCallback.resetFocus) {
       controlProviderCallback.resetFocus(() => this.resetFocus());
     }
     if (controlProviderCallback.focusOnNode) {
       controlProviderCallback.focusOnNode((nodeIndex: number) => {
-        if (this.navQueue[this.navIndex] !== nodeIndex) this.navQueue.splice(++this.navIndex, Infinity, nodeIndex);
+        const rect = focusRect(this.props.columnarViewModel, this.props.chartDimensions.height, nodeIndex);
+        this.addToNav({ ...rect, index: nodeIndex });
```
monfera, line 298: Why did the need for splicing (or analogous) go away? The user may have performed a bunch of undos, currently in the middle of the history, it's possible to redo some or all of what was undone. So the now "future" part of the history needs to be erased

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -757,17 +812,15 @@ class FlameComponent extends React.Component<FlameProps> {
...
+                this.navBackward();
               }}
               style={{ display: 'none' }}
             />
           </label>
           <label
             title="Reset"
             style={{
-              color: this.isAtHomePosition() ? 'darkgray' : 'black',
+              color: 'black',
```
monfera, line 823: OK so it looks like it was intentional

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -777,8 +830,6 @@ class FlameComponent extends React.Component<FlameProps> {
               type="button"
               tabIndex={0}
               onClick={() => {
-                this.lastNavLeft = false;
-                if (this.isAtHomePosition()) return;
```
monfera, line 781: Not sure about this change, unless there's an equivalent mechanism behind `this.resetFocus` that avoids writing into the history a possible repetition of home button clicks (or more generally, repetition of anything—it's sensible to filter for this centrally)

packages/charts/src/chart_types/flame_chart/navigation.ts
```diff
@@ -0,0 +1,198 @@
...
+    if (!this.canNavBackward()) {
+      return;
+    }
+    this.navIndex--;
+    return this.current();
+  }
+
+  canNavForward() {
+    return this.navIndex < this.navQueue.length - 1;
+  }
```
monfera, line 47: Most of these are one-liners, maybe they could be single line fat arrows, even without empty rows between the one-liners, though whichever level of compactness you prefer

packages/charts/src/chart_types/flame_chart/navigation.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  lastInQueue() {
+    return this.navIndex === this.navQueue.length - 1;
+  }
+}
+
+/**
+ * Only click
+ * Adds click even in middle of history
```
monfera, line 89: even or event?

packages/charts/src/chart_types/flame_chart/navigation.ts
```diff
@@ -0,0 +1,198 @@
...
+export class ClickStrategy1 extends NavigationStrategy {
+  add(toAdd: NavRect) {
+    if (isZoomPanNav(toAdd)) {
+      return;
+    }
+    if (this.current()?.index !== toAdd.index) {
+      this.navQueue.splice(++this.navIndex, 0, toAdd);
+    }
+  }
+}
```
monfera, line 101: OK so `add` and `addToNav` can remove too. Maybe we can dream up other names for these, I'll add it here or while we chat if something comes to mind. In its general form, the user is branching in the multiverse 😄 

Also it feels like it'd be great to have revealing names for the strategies, instead of indexing with Strategy1, Strategy2

packages/charts/src/chart_types/flame_chart/navigation.ts
```diff
@@ -0,0 +1,198 @@
...
+        }
+      } else {
+        // add only if differ from current
+        if (this.current()?.index !== toAdd.index) {
+          this.navQueue.splice(++this.navIndex, Infinity, toAdd);
+        }
+      }
+    }
+  }
+}
```
monfera, line 157: When such level of branching is necessary, maybe it's a good time to consider representing things with, and deriving options through a state diagram. Could be a something like `xstate` (an example can be visualized [here](https://stately.ai/viz)), or our own more compact utility (or in this order, so worth wrapping the lib). Not necessarily in this PR, though nested `if`s are often fragile with insidious edge cases

monfera (commented):
This is an initial source review, will do local testing today esp. if no significant change is deemed necessary

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
markov00: I removed that duplicated header, don't know why that was added.

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
markov00: done in https://github.com/elastic/elastic-charts/pull/1748/commits/14ec6ebd9a9a25bbc962b6dc995f3ce6ad754ab5

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
markov00: Yes, it was intentional, the idea here is to have a reset button always available even if I'm at the home position. This will allow us to reset the history anytime, even at the root

packages/charts/src/chart_types/flame_chart/navigation.ts
markov00: done in https://github.com/elastic/elastic-charts/pull/1748/commits/841f78bb66d79f00e7e55cc601f13dfb112a7ca7

packages/charts/src/chart_types/flame_chart/navigation.ts
markov00: Done [here](https://github.com/elastic/elastic-charts/pull/1748/commits/4e19704c6ee682573d1f8c42b9e763ce80774cb9) together with a bit more renaming/updating comments

packages/charts/src/chart_types/flame_chart/navigation.ts
markov00: I removed the `addToNav` and a few other redundant methods from `flame_chart.tsx` in https://github.com/elastic/elastic-charts/pull/1748/commits/9f43b5d475a545d19b836b9e722bba03474d60fa

packages/charts/src/chart_types/flame_chart/navigation.ts
markov00: Agree, I was also considering that when I was implementing this method. I rarely played with FSM but I think this is a good candidate for such implementation.

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
markov00: The strategy was slightly adjusted from the one requested and described, I can chat about that offline

monfera (approved):

PR #1749, opened 2022-07-12
chore(deps): remove canvas dependency

markov00:
## Summary

This PR removes the `canvas` dev dependency.
That dependency was required to run jest tests in order to mock canvas dependencies. There was a few problems with that: 
- it doesn't fully work with Apple Silicon (because it needs a native build and they were not published 
in https://github.com/Automattic/node-canvas)
- the used capabilities were limited to the text measurement, that was even not precise because depends on the running environment.
- I've seen some issues with jest test and the release CI steps in Buildkite  https://github.com/elastic/elastic-charts/commit/7c1c8935cb22939f033ac0bf09c6fbd7bb77ea0a

Due to the previous facts, the PR removes the dependency, updates the jest snapshots (with a slight changes in the measured output) and adds a more general mock for the `getContext` function.

We can live without a precise text measurement calculation because we have enough VRTs tests to check and support that case.
### Checklist


packages/charts/src/utils/bbox/canvas_text_bbox_calculator.test.ts
```diff
@@ -8,7 +8,7 @@
 
 import { withTextMeasure } from './canvas_text_bbox_calculator';
 
-describe('CanvasTextBBoxCalculator', () => {
+describe.skip('CanvasTextBBoxCalculator', () => {
```
nickofthyme, line 11: Do we still need to skip this?

nickofthyme (approved):

packages/charts/src/utils/bbox/canvas_text_bbox_calculator.test.ts
markov00: I think we can completely remove that test.
If I unskip it, the resulting size will be zero and it doesn't have any meaning actually....

markov00:
buildkite test this

packages/charts/src/utils/bbox/canvas_text_bbox_calculator.test.ts
markov00: I removed the test [a463334](https://github.com/elastic/elastic-charts/pull/1749/commits/a463334fb2292f841deb6d1c87e3e8dbfd3dc025)

PR #1751, opened 2022-07-12
use typeof check to check for `process` variable

spalger:
## Summary

https://github.com/elastic/elastic-charts/pull/1672 fixed an issue where `process.env.VRT` was directly referenced in the code, which isn't available outside of specific environments. This broke ci-stats, which is broken again. It would seem that the module moved to a new chunk and in that chunk strict-mode is enabled which is now throwing a runtime error just by referencing `process`.

image

This uses `typeof` to check for the `process` variable before doing additional accesses which will resolve to `undefined` in strict environments instead of throwing a runtime error.


nickofthyme (approved):
LGTM, Thanks for taking care of this! ❤️‍🔥

spalger:
You bet, thanks!

PR #1753, opened 2022-07-13
ci: move buildkite to new datavis agent queue

nickofthyme:
Move all buildkite jobs to `datavis-n2-2` agent `queue`.


nickofthyme:
Yeah the pipeline upload step is configured in https://buildkite.com/elastic/datavis-ci/steps. I updated that to the new `queue` just now.

markov00 (approved):

PR #1760, opened 2022-07-20
ci: fix skip logic for dependabot

nickofthyme:
## Summary

When `dependabot` and others, update a PR it could update only the `yarn.lock` and not the touch the `package.json`. This fix triggers the step checks correctly in such a case. Also the `minimatch` options ignore dot paths (i.e. `.buildkite/`) by default, this now respects changes to dot paths by default.


markov00 (approved):
Looks sensible, thanks!

PR #1765, opened 2022-07-28
fix(time): render correctly 0-24 hourly clock

markov00:
## Summary

When rendering a multi-layer time axis with ticks every hour, the formatted, that should use a 24 hours representation, is showing the hours with the following sequence:  22 - 23 - 24 - 01 - 02 - 03 etc as seen in the screenshot below
Screenshot 2022-07-28 at 14 46 55

This is caused by the way `Intl.DateTimeFormat` handles the configured option `hour12:false`. After researching a bit, it looks like we can replace that with the `hourCycle: h23` that represent the hours from 00:00 to 23:59.

Screenshot 2022-07-28 at 14 47 44

To test it you can use the following URL locally: [storybook](http://localhost:9001/?path=/story/area-chart--timeslip&globals=theme:light&knob-Bin%20width%20in%20ms%20(0:%20none%20specifed)=0&knob-Minor%20grid%20lines=true&knob-Shift%20time=0&knob-Stretch%20time=2.4&knob-Time%20zoom=120&knob-layerCount=3) or replace the host with the one from this hosted storybook

## Details

See https://stackoverflow.com/questions/65604554/intl-datetimeformat-returns-an-hour-over-24 and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

### Checklist


monfera (approved):
Nice solution!

PR #1782, opened 2022-08-10
feat(tooltip): pinned tooltip with actions

### LINKED ISSUE #1857: Cursor remains visible while brushing
**Describe the issue**

When brushing is enabled and the tooltip is set to snap the cursor line or band is visible while brushing.

#### Line

Screen Recording 2022-11-01 at 12 44 00 PM

#### Bar

Screen Recording 2022-11-01 at 12 45 25 PM

nickofthyme:
## Summary

Adds sticky tooltip interaction whenever `Tooltip.actions` are defined. See demo story [here](https://ech-e2e-ci--pr-1782-tvisv7qq.web.app/?path=/story/interactions--brush-selection-tool-on-time-charts&globals=theme:light).

Features:
- Right-click to stick tooltip
- Select series on stuck tooltip
- Disables stickiness for empty tooltips
- trigger stickiness from tooltip actions or flag
- trigger actions with selected series
- global click listener to dismiss pinned tooltip
- Clear pinned tooltip on resize, <kbd>esc</kbd> or outside click
- limit scroll height to 5 items
- scroll nearest elements into view

Chart types with context menu:

- [x] `cartesian charts` [story example](https://ech-e2e-ci--pr-1782-aq31oo2m.web.app/?path=/story/components-tooltip--cartesian-charts)
- [x] `partition charts` - [story example](https://ech-e2e-ci--pr-1782-aq31oo2m.web.app/?path=/story/components-tooltip--partition-charts)
- [x] `heatmap` - [story example](https://ech-e2e-ci--pr-1782-aq31oo2m.web.app/?path=/story/components-tooltip--heatmap-chart)
- [x] `flamegraph` - [story example](https://ech-e2e-ci--pr-1782-aq31oo2m.web.app/?path=/story/components-tooltip--flamegraph)

TODOs (moved to https://github.com/elastic/elastic-charts/issues/1859)
- [x] Configure what is passed back to user (heatmap & flame)
- [x] Block tooltip render when brushing (xy)
- [ ] Define if we need a different prompt if no actions are available but you can pin the tooltip to show all series.
 

https://user-images.githubusercontent.com/19007109/185769263-d0e4b4c1-ffb4-49e4-9963-0944c74e2cb1.mp4

## Testing in kibana

To test against this PR on kibana update the `package.json` to point to the `tooltip-actions-dist` branch.

```jsonc
// package.json
{
  "dependencies": {
    "@elastic/charts": "https://gitpkg.now.sh/nickofthyme/elastic-charts/packages/charts?tooltip-actions-dist"
  }
}
```

Then run bootstrap with `--no-validate` flag...

```
yarn kbn bootstrap --no-validate
```

See kibana example/testing/demo at https://github.com/elastic/kibana/pull/141503

## Details

Modifies the `interactions` redux state to add tooltip stuck/selection state and account for the stuck state and disable existing interactions like drag and click while stuck.

The <kbd>esc</kbd> option to cancel the brush interaction was added a while ago but was a little loose when adding and removing the keyboard event listeners.

> I have noticed that some of the tooltip interactions are tightly coupled to the redux store/selectors. This might become problematic as we are adding event listeners within `packages/charts/src/components/chart_container.tsx` that may not be implemented inside the reducers. For example, we had a `keyup` listener that was added but the reducer that it triggers may not actually do any logic.
> 
> We should have a better way to communicate interactions to the component that does not require as much redundant checks or logic.

### Scroll into view demo

https://user-images.githubusercontent.com/19007109/187324616-175be9cb-d09c-40e1-bde0-3c2341313983.mp4

This feature works by scrolling to the first highlighted tooltip value. Some improvements on this logic could be...

- Some type of collapsing on groups for charts with hierarchical structures to always show highlighted values.
- Collapse values in between until tooltip is pinned.
- Currently the order of tooltip items is arbitrary but if we change this to be nearest-first we could have a better interaction. Currently, the closest is not element is not always scrolled to view, as seen at the end of the recording above where the cursor is closest to the line point but the bar values is focused.
- Shuffle the items to allow showing all highlighted values. Might be a bad idea

### Issues

fix #1857

### Checklist


monfera (commented):
Nice work! It's a WIP so all these below may be on your list already:

There are some statefulness issues, for example:
- activating the tooltip, then, as the next step, trying to activate it elsewhere just clears the pinned tooltip
- even after clearing (right-click on another bar) it's not possible to pin the tooltip, it requires that the mouse first be moved, probably to populate x/y values (right-click DOM interactions also pass on coordinates, so I think there should be no requirement for a mousemove)
- activating the tooltip, then left-clicking on another bar (or pretty much anything else, except, apparently, the Esc key or another right-click tooltip activation attempt) doesn't clear the pinned tooltip

It'd feel odd to maintain the pinned status if the user clicks (eg. left-clicks) elsewhere. A click is a strong expression of intent, whether activating something else or cancelling a current modality. Our interaction design for this should be such that it's super easy to get out of the pinned mode. So, starting with "literally everything gets the user out of pinned mode" then remove hover from that list (I think only a subsequent hover should leave the tooltip pinned)

Wondering if the pinned tooltip needs the common ☒ button on the top of the tooltip as another means for cancellation. While also observing Fitts law (ie. the unpin hotspot can/should be larger than the visible ☒ or pin symbol)

(Maybe it's done already) It'd also be great to link to an interaction design document. The current way doesn't just pin the tooltip, it in effect establishes a strong, new modality that overrides everything. Worth comparing it to the more fleeting persistence of such tooltips in Tableau. In any case, it's worth making interaction design choices explicit, and supported with comparisons and arguments

Minor: I'd also rename the PR, because it's more of a pinned thing than a sticky one, and whether the "tooltip" has actions or not is secondary, so I'd just call it "pinned tooltip" or similar. "Sticky" also has the issue that its past participle is "stuck" and that has other connotations (unwanted behavior, like a stuck button)

markov00:
Agree with Robert, this only partially is related to the tooltip, this is more a [contextual menu](https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event) than a sticky/pinned tooltip.

For the interaction/cancellation/clearing it should behave exactly as a right-click context menu in the browser: if you click elsewhere or in one item of the menu,  the menu will disappear immediately

monfera (changes requested):
I'd like to ask for a refinement in interaction design, the context menu should be more fleeting. Examples:
- I hit the escape key, nothing happens (should clear the thing, I think on any key but definitely on escape, as users expect that, even outside accessibility)
- I can resize the chart area and the tooltip remains (should clear)
- I can visit other browser tabs and applications, interact there and it remains (should clear)

The current concept has unsolved aspects of interaction design, which is expected as the PR is itself also the prototype (may or may not be the most efficient this way). For example, the user can pin the floating, modal context view (it's not a tooltip or a context menu, and I prefer truthful names) and it remains there indefinitely. But the association with the data weakens. It's not obvious which data point the context view relates to. For example, a little arrow triangle at the appropriate part of the context view perimeter could more persistently link the context view to the data point.

I'm also interested in the accessibility concept for this. As it's quite novel, it's not exactly trivial. It'd be great to prototype that before the functionality is released, as there's some chance that the accessibility concept will alter the baseline interaction design. It does not need to be in this PR, just designed before release.

I also got an error when pinning and then trying to scroll with the scrollbar between the chart area and the right-side storybook panel (I wanted to test if scrolling clears the thing): 
image

Another issue seems to be related to the fact that canvas resizing is minimized, to avoid thrashing the GL and Canvas2d contexts 60/120 times a second, it's much more expensive than a simple rerender. This chart resizes the canvases in 256 pixel increments. Upon resize, the context view remains in place, losing its adjacency with the data point, then jumps at once when the 256 pixel canvas resize threshold is reached. Though this could be preempted by the above ask of making it more fleeting, such as removing the context view upon resize.

There's also an avoidable warning when pinning:
image

packages/charts/src/chart_types/flame_chart/internal_chart_state.ts
```diff
@@ -31,7 +31,12 @@ export class FlameState implements InternalChartState {
   getLegendExtraValues = () => new Map<SeriesKey, LegendItemExtraValues>();
   getPointerCursor = () => DEFAULT_CSS_CURSOR;
   getTooltipAnchor = () => ({ x: 0, y: 0, width: 0, height: 0 });
-  isTooltipVisible = () => ({ visible: false, isExternal: false });
+  isTooltipVisible = () => ({
+    visible: false,
+    isExternal: false,
+    displayOnly: false,
+  });
+
```
monfera, line 39: Nit, this "null value" would fit in one line, blending in with the rest of the file, which is intentionally devoid of real content, all these initial values aren't super informative anyway

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -6,7 +6,7 @@
  * Side Public License, v 1.
```
nickofthyme, line 6: > - activating the tooltip, then, as the next step, trying to activate it elsewhere just clears the pinned tooltip
> - even after clearing (right-click on another bar) it's not possible to pin the tooltip, it requires that the mouse first be moved, probably to populate x/y values (right-click DOM interactions also pass on coordinates, so I think there should be no requirement for a mousemove)
> - activating the tooltip, then left-clicking on another bar (or pretty much anything else, except, apparently, the Esc key or another right-click tooltip activation attempt) doesn't clear the pinned tooltip

These are all good points, I updated the code to fix them or add these interactions.

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -6,7 +6,7 @@
...
  */
```
nickofthyme, line 8: > modal context view (it's not a tooltip or a context menu, and I prefer truthful names) - @monfera 

> this only partially is related to the tooltip, this is more a [contextual menu](https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event) than a sticky/pinned tooltip @markov00 

What's the significance of the naming of here? For all intents and purposes it just an extension of our tooltip regardless of the trigger. We are not opening a modal we are pinning/freezing/(insert synonymous verb here) the existing tooltip.

Is this naming for external purposes or internal as well?

nickofthyme, line 8: > It's not obvious which data point the context view relates to. For example, a little arrow triangle at the appropriate part of the context view perimeter could more persistently link the context view to the data point.

This is a good _point_ but I think this decision falls on @gvnmagni, that can be added in a later PR

nickofthyme, line 8: > I'm also interested in the accessibility concept for this. As it's quite novel, it's not exactly trivial. It'd be great to prototype that before the functionality is released, as there's some chance that the accessibility concept will alter the baseline interaction design. It does not need to be in this PR, just designed before release.

There is currently no accessibility related geoms to even trigger the tooltip. As far as accessibility when the tooltip is open we can improve that with help from the accessibility peeps.

nickofthyme, line 8: > I also got an error when pinning and then trying to scroll with the scrollbar between the chart area and the right-side storybook panel (I wanted to test if scrolling clears the thing):

Yeah this is related to the `scrollIntoView` logic that sounds like we are going to remove for a different approach.

nickofthyme, line 8: > Another issue seems to be related to the fact that canvas resizing is minimized, to avoid thrashing the GL and Canvas2d contexts 60/120 times a second, it's much more expensive than a simple rerender. This chart resizes the canvases in 256 pixel increments. Upon resize, the context view remains in place, losing its adjacency with the data point, then jumps at once when the 256 pixel canvas resize threshold is reached. Though this could be preempted by the above ask of making it more fleeting, such as removing the context view upon resize.

This is beyond my understanding, we could zoom about this and find a solution.

nickofthyme, line 8: > activating the tooltip, then, as the next step, trying to activate it elsewhere just clears the pinned tooltip

This is as designed, as we demoed the chrome context menu. When you open the context menu, then click somewhere else no action is taken requiring a second `click`, or whatever, before it applies that action.

nickofthyme, line 8: > even after clearing (right-click on another bar) it's not possible to pin the tooltip, it requires that the mouse first be moved, probably to populate x/y values (right-click DOM interactions also pass on coordinates, so I think there should be no requirement for a mousemove)

@markov00 what do you think about this?

Basically should we activate the hover when the tooltip is unpinned, or wait for the user to move their mouse?

Screen Recording 2022-09-22 at 07 44 40 AM

nickofthyme, line 8: > It'd feel odd to maintain the pinned status if the user clicks (eg. left-clicks) elsewhere. A click is a strong expression of intent, whether activating something else or cancelling a current modality. Our interaction design for this should be such that it's super easy to get out of the pinned mode. So, starting with "literally everything gets the user out of pinned mode" then remove hover from that list (I think only a subsequent hover should leave the tooltip pinned)

Yup good points, we changed this to be easily dismissible and less _strong_ 💪🏼 .

nickofthyme, line 8: > Wondering if the pinned tooltip needs the common ☒ button on the top of the tooltip as another means for cancellation. While also observing Fitts law (ie. the unpin hotspot can/should be larger than the visible ☒ or pin symbol)

This was decided as not necessary

nickofthyme (commented):
@monfera thanks for the reviews, here's some direct replies to your comments.

packages/charts/src/chart_types/flame_chart/internal_chart_state.ts
nickofthyme: See f7b05f6

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
nickofthyme: A similar interaction is used in the current flame chart for `click` action. Notice when you click the chart focuses the selected element but the cursor lands on a value and the position is not updated until the cursor is moved.

Screen Recording 2022-09-22 at 10 20 45 AM

markov00: I think we should separate the concept, and consider tooltip only what is visible at hover and we can call context menu what we call now pinned tooltip.
In the long run that pinnet tooltip could be a real context menu that appear when you click on an axis, on a title, on a white space and this is no longer a tooltip, but a real menu.

markov00: IMHO when you are inspecting and using a BI tool you never stop your mouse from moving. It could annoy someone for now, but is definitely a minimal defect for now.
I've double-checked the standard OS behaviour and what happens is:
- if you click a button/link when a context menu is open, the button/link doesn't trigger
- if you right-click on a different place when you already have a context menu open then the context menu just moves to the new right-clicked position
- if you have a context menu is open and you move over a section/button/link/image with a default onHover behavior (like show a tooltip) the tooltip is shown when you click there and make the context menu disappear.

We probably should follow that behavior, but for now, I will put this adjustment as one of the last items on the priority list

markov00: No necessary, exactly, this is a non standard  behaviour for context menu.

packages/charts/src/components/tooltip/components/tooltip_wrapper.tsx
```diff
@@ -9,22 +9,76 @@
...
+      return (
+        <button
+          className="echTooltip__action"
+          key={i}
+          title={typeof reason === 'string' ? reason : undefined}
+          disabled={Boolean(reason)}
+          onClick={() => {
+            onTooltipPinned(false, true);
+            // timeout used to close tooltip before calling action
+            setTimeout(onSelect, 0, selected);
```
markov00, line 57: If I'm not wrong the `onSelect` takes 2 arguments, the selected and the all items array.
ESLint is also shouting on that line

packages/charts/src/chart_types/heatmap/state/selectors/tooltip.ts
```diff
@@ -63,6 +66,7 @@ export const getTooltipInfoSelector = createCustomCachedSelector(
             value: `${shape.datum.y}`,
             formattedValue: spec.yAxisLabelFormatter(shape.datum.y),
             datum: shape.datum,
+            displayOnly: true,
```
markov00, line 69: By default this and the X value should be selectable and the cell value not.

Since both axes' values don't have a color we can probably use something like a neutral color. In @gvnmagni's animations the filters are just actions on the values directly and nothing was selectable, but in this first step we can just have them selectable
Screenshot 2022-09-23 at 18 34 26

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -487,29 +576,65 @@ class FlameComponent extends React.Component<FlameProps> {
...
+  };
+
+  private handleContextClick = (e: MouseEvent) => {
+    e.preventDefault(); // prevent browser context menu
+
+    window.addEventListener('keyup', this.handleKeyUp);
+    window.addEventListener('click', this.handleContextClose);
+    window.addEventListener('visibilitychange', this.handleContextClose);
+
+    this.onTooltipPinned(!this.tooltipPinned);
```
markov00, line 611: I think we should mark the tooltip pinned only if we are over a node, if not it cause an issue:
if you try to click on a white area, the tooltip doesn't appear anymore until you press esc or click somewhere

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -368,40 +402,84 @@ class FlameComponent extends React.Component<FlameProps> {
...
+
+  private unpinTooltip(redraw = false) {
+    this.pinnedPointerX = NaN;
+    this.pinnedPointerY = NaN;
+    this.tooltipPinned = false;
+    this.tooltipSelectedSeries = [];
+    this.updateHoverIndex();
+    if (redraw) {
+      this.smartDraw();
+    }
```
markov00, line 420: I see the pain of rewriting this logic :(

packages/charts/src/components/tooltip/components/tooltip_table_body.tsx
```diff
@@ -37,18 +47,45 @@ export const TooltipTableBody = <D extends BaseDatum = Datum, SI extends SeriesI
...
+  const scrollToTarget = debounce((i) => {
+    const target = tableBodyRef.current?.querySelector(`#${getRowId(i)}`);
+    if (!target) return;
+    target.scrollIntoView({ behavior: 'smooth', block: 'center', inline: 'start' });
+  }, 100);
+  useEffect(() => {
+    if (!ready || targetRowIndex === -1 || props.pinned) return;
+    scrollToTarget(targetRowIndex);
+  }, [scrollToTarget, ready, targetRowIndex, props.pinned]);
+
```
markov00, line 64: We should get rid of this and use the other visibility method defined:
- show up to 5 series
- if more then 5, then show only the highlighted ones (up to 5)
- everything else is shown in the context menu

This mean that the context menu is available not only if there are actions but also if there are more then 5 series

nickofthyme: Yes working on this now.

I think _five_ is arbitrary since the items can overflow and increase the height size of each row independently.

image

gvnmagni: good point

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
gvnmagni: I genuinely lack some context I'm afraid! 😄

packages/charts/src/components/tooltip/components/tooltip_wrapper.tsx
nickofthyme: See 06cc1ac

flash1293:
It seems like if there is a custom tooltip bit, it won't render the "right click to stick" message anymore:
Screenshot 2022-09-29 at 18 18 25

It still works though:
Screenshot 2022-09-29 at 18 18 29


Maybe I'm using it wrong

nickofthyme:
@flash1293 Nope you are correct, I am working on adding support for the `customTooltip` option.

flash1293:
Not sure whether this is some kind of bug with my way of loading this branch, but the actions are not typed: 
Screenshot 2022-09-30 at 09 15 45

flash1293:
I tested a bit more and the current api is a bit limiting. There can only be one set of actions for a chart no matter what value the user clicked. This is an assumption that holds true most of the time, but not necessarily always.

Also due to this setup it's not possible to only fetch some information if the user is actually opening the tooltip which might be necessary at some point (we could work around it right now, but not forever) - for example we might want to disable an action if a certain saved object is created already, but we need to check with an additional request. We can only send this request if the user actually pinned the tooltip - right now there is no place to do this

What I'm suggesting is instead of a prop which takes a list of actions, having a prop which takes a component to render the actions on its own

e.g. like this:
old:
```
  actions: {[ label: "abc", onSelect: (values) => handleClick(values) ]}
```

new:

```
import { TooltipAction } from '@elastic/charts';

function TooltipActions(values) {
  useEffect(() => {
    // some async stuff
  })
  return <>
    <TooltipAction label="abc" onClick={() => handleClick(values)}>
  <>
}

// ...
actions: {TooltipActions}
```

it will also come in handy later for having more sophisticated interactivity in there instead of just a list of links to click (I remember some mockups with mini-forms in there).

Let me know what you think.

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
nickofthyme: @gvnmagni I think the ask here is should we have a carrot UI on the tooltip to point to the selected element(s).

Image 2022-10-11 at 4 19 52 PM

packages/charts/src/components/tooltip/components/tooltip_table_body.tsx
nickofthyme: Fixed by UI to truncate values until pinned to a fixed height and configurable number of max elements, this would default to 5 (+ ½) items in the tooltip.

nickofthyme:
@monfera I'm getting an issue where pinning the tooltip then quickly clicking outside of the chart, causes the chart to go blank on the next render and following renders.

Screen Recording 2022-10-11 at 05 10 44 PM

I believe it is somehow related to `contextRestoreHandler` specifically the code in the `setTimeout` below. Increasing this timeout makes it easier to trigger the blank chart (i.e. time to pin and blur chart). Any thoughts on why this maybe the case?

https://github.com/elastic/elastic-charts/blob/accb50a790c1785c8b691e66f52587a88611dd66/packages/charts/src/chart_types/flame_chart/flame_chart.tsx#L1192-L1195

markov00:
> 98f5436

Hey @nickofthyme this https://github.com/elastic/elastic-charts/pull/1782/commits/98f5436fafacba186c1a582b53f89b1e1816c045 should solve the issue. You where registering a mousemove listener when handling the context menu click. This cause the `handleMouseDragMove` to be called making the glitch to appear. You can better reproduce the problem by right-click on a node and by dragging outside the context menu.

nickofthyme:
@gvnmagni These changes are nearly code complete, just fixing up the code checks and doing a final check on screenshots. Please take a look at let me know if there are any changes you'd like to make. I am missing a few colors for the dark mode namely the tooltip hover highlighter color and the action text color. If they are the same that's fine just want to make sure as the figma colors don't match any of the eui color values [here](https://eui.elastic.co/#/theming/colors/values).

nickofthyme:
buildkite update vrt

gvnmagni:
Thank you Nick! I believe @markov00 is already working on both the colors and will be fixed soon!

gvnmagni (approved):

nickofthyme:
buildkite update vrt

nickofthyme:
buildkite update vrt

packages/charts/src/components/tooltip/components/_tooltip.scss
```diff
@@ -223,10 +148,7 @@ $tableRowHoverSelectedColor: tintOrShade($euiFocusBackgroundColor, 0, 10%);
     position: relative;
 
     &--highlighted {
-      .echTooltip__tableCell:not(.echTooltip__colorCell),
-      .echTooltip__colorCell .echTooltip__colorStrip--spacer {
-        background-color: transparentize($euiColorLightShade, 0.5);
-      }
+      background-color: transparentize($euiColorLightShade, 0.5);
```
nickofthyme, line 151: This does not work because the `__tableRow` has `display: contents` for sizing the grid columns. So it must be applied to the children.

```suggestion
      .echTooltip__tableCell:not(.echTooltip__colorCell),
      .echTooltip__colorCell .echTooltip__colorStrip--spacer {	
        background-color: transparentize($euiColorLightShade, 0.5);	
      }
```

markov00:
> ## Deployments - [f4d0656](https://github.com/elastic/elastic-charts/commit/f4d065613cede38c85797c5214ddccb1f8b5c7a5)
> * Storybook ([link](https://ech-e2e-ci--pr-1782-aq31oo2m.web.app))
> * e2e server ([link](https://ech-e2e-ci--pr-1782-aq31oo2m.web.app/e2e))
> * Playwright report ([link](https://ech-e2e-ci--pr-1782-aq31oo2m.web.app/e2e-report))

The changes on the VRTs from master looks good to me and only related to:
- the added tooltip border
- a tooltip rendered with the prompt if the list of values where too long

markov00:
buildkite update screenshots

nickofthyme:
buildkite update vrt

nickofthyme:
buildkite update vrt

nickofthyme:
buildkite test this

nickofthyme:
buildkite test this

nickofthyme:
buildkite update vrt

nickofthyme:
buildkite update vrt

nickofthyme:
buildkite update vrt

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
markov00: moved to https://github.com/elastic/elastic-charts/issues/1859

markov00: added to https://github.com/elastic/elastic-charts/issues/1859

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
markov00: done

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
markov00: added to https://github.com/elastic/elastic-charts/issues/1859

markov00 (changes requested):
Everything looks good expect for the fact that the `control+click` on mac doesn't works as expected. It initially pins the tooltip but a subsequent mouse movement will unpin the tooltip.
`control+click` should follow the same behaviour of the `secondary click` on mac (usually the two finger click)

nickofthyme:
> Everything looks good expect for the fact that the `control+click` on mac doesn't works as expected. It initially pins the tooltip but a subsequent mouse movement will unpin the tooltip. `control+click` should follow the same behaviour of the `secondary click` on mac (usually the two finger click)

Good catch, fixed in 0367fed

markov00 (approved):
Everything looks good to me, amazing work Nick!

PR #1791, opened 2022-08-18
feat(xy): collapsing time into a single hh:mm[:ss] layer

monfera:
## Summary

- Collapses time into a single hh:mm[:ss] layer
- The Timeslip alpha chart could therefore switch to a maximum of two labeled time axis layers (suggested for all other charts)
- The minimum gridline pitch for a new unlabeled grid layer to appear is about 25% wider, so the chart/grid is slightly less dense
- Added the weekly (week start) unlabeled raster into the starting raster set (see new gridlines with dailies)
- Made the weekly labels (day of the start of the week) more spaced out to reduce their horizontal density
- Switched to long month names for all resolutions in the bottom (context) labeled layer for consistency
- Timeslip alpha: Deeper maximum zoom so normal sized charts can reach the individually labeled milliseconds level
- Factored out numerous common constants, formatters and layer cascading (object spreads) for more code DRYness and easier future maintenance
- Minor type and other code improvements

image

### Checklist


monfera:
buildkite update vrt

markov00:
Hi @monfera it looks great in my opinion. I'm wondering if we can add intermediate steps between the jumps, for example:
- between days and 6-hours, add 12 hours raster
- between 6-hour and hour, add a 3 hours raster
- between hourly and quarter hours, add a half-hour raster 
- between minutes and quarter minutes, add an half-minute raster

There could be a bit more raster to add, but can be considered to add a more "sparse" representation.
Thanks to the already existing unlabelled raster, understanding the time in between two labelled rasters is easy and you have to add up to 5 to one date/time to get the right value, looks simple enough for everybody and also reduces a but he labelling noise around, WDYT?

monfera:
@markov00 I think some or all the rasters can be added, but it'd be great to discuss and implement it in a next PR, as it's in my understanding, lower priority than, and separable from the merger of the layers: 
- The full absorption of this current PR will involve a change from 3 to 2 layers. So it'd be useful to get a head start on that. so it's available for developers using charts
- From a risk management PoV it's nice to decouple the two PRs
- The addition of new rasters doesn't fundamentally change the overall logic such as collapsing and labeled layer count, so the next PR would be a more incremental one. As we have no user feedback for that, there's less of a time pressure for trying them (deployed in a new PR) and discussing them from a design PoV

monfera:
@markov00 btw. I was planning to update the 3-layer stories to 2-layer stories in a separate PR too, so the effect of individual changes can be more incrementally assessed, but I can add it to this PR if you think it's better that way

markov00:
@monfera can we say that we are going to deprecate completely the 3 layers implementation? or do you like to keep that configuration and let's the developers decide?

monfera:
@markov00 Yes I think it's best to switch from where it is 3 layers today to 2 layers, because the reason for the 3 layers was the rather granular layer labeling, which we consolidated with the layer collapse. You might want to circulate it among call sites and PMs, though it's not in the way of this specific PR, as layer count is externally specified

packages/charts/src/chart_types/xy_chart/axes/timeslip/rasters.ts
```diff
@@ -358,116 +363,113 @@ export const rasters = ({ minimumTickPixelDistance, locale }: RasterConfig, time
...
       ...hourFormat,
       minute: 'numeric',
       timeZone,
     }).format,
-    minorTickLabelFormat: (d) =>
-      `${new Intl.DateTimeFormat(locale, {
-        minute: 'numeric',
-        timeZone,
-      }).format(d)}'`,
+    minorTickLabelFormat: (d) => `${minutesFormatter.format(d).padStart(2, '0')}`, // what DateTimeFormat doing?
```
markov00, line 390: Hey Robert, if minutesFormatter is
```
const minutesFormatter = new Intl.DateTimeFormat(locale, { hour: '2-digit', minute: '2-digit', timeZone, hourCycle });
```
the padding method can't be applied because you still have both hour and minute as the output of the formatted. Am I missing something?

markov00 (approved):
It looks great to me.
I've added just one comment to surface before merging

packages/charts/src/chart_types/xy_chart/axes/timeslip/rasters.ts
monfera: Thanks Marco for the catch, the `padStart` doesn't do anything there, removed in 0b5c332223852874b331bfa0a72e5c952ccd9c76

PR #1828, opened 2022-09-22
fix(style): import the correct legacy reset css

markov00:
## Summary

We completely miss some reset CSS properties (like global font-family) that are available in the EUI legacy reset.
This PR will bring it back and restore the required style when importing elastic-charts only in a project without EUI.

## Details

I tried to import elastic-charts in an external project and the font used in the Metric viz wasn't the correct one.
With this PR I was able to bring back the fonts on such external project

### DIFF
--- a/packages/charts/src/_reset.scss
+++ b/packages/charts/src/_reset.scss
@@ -1,79 +1,8 @@
 // Used as base reset used with EuiProvider. TODO: remove need for this file export.
 // https://github.com/elastic/eui/blob/master/src/global_styling/reset/reset.ts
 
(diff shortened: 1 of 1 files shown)


nickofthyme (approved):

nickofthyme (commented):
Have you tested this locally on the dev server with changes to `.scss` files? To see if it takes forever to update the compiled css?

markov00:
> Have you tested this locally on the dev server with changes to `.scss` files? To see if it takes forever to update the compiled css?

Actually no, I haven't tested this way. Will do

PR #1842, opened 2022-10-10
fix(tooltip): heatmap visible tooltip option

nickofthyme:
## Summary

Match behavior of `externalPointerEvents.tooltip.visible` option on heatmap. Currently, heatmap external band will only show when this option is set to `true`.

The current heatmap tooltip is only visible when both a x and y values are valid, hence not on external events. This could be improved upon in the future to add a list of categories for a given `x` value. Showing just the x value would not serve any purpose.

### Checklist

### DIFF
--- a/packages/charts/src/chart_types/heatmap/state/selectors/get_cursor_band.ts
+++ b/packages/charts/src/chart_types/heatmap/state/selectors/get_cursor_band.ts
@@ -7,7 +7,7 @@
  */
 
 import { Rect } from '../../../../geoms/types';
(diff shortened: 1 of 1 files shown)


markov00 (approved):
Correct, cursor band and tooltip are two separate things.
We can in case add an external cursor visibility option if we want, but this is good for now

nickofthyme:
> We can in case add an external cursor visibility option if we want

Yeah but that's implicitly enabled by calling `dispatchExternalPointerEvent`

PR #1846, opened 2022-10-12
docs: move documentation to GH Wiki

markov00:
## Summary

Our documentation is a bit scattered around. I've moved all our markdown file into the [GH wiki](https://github.com/elastic/elastic-charts/wiki) added an initial version of a TOC with a bit of hierarchy.

In this way we can just direct people to the GH wiki to find all the written docs instead of having a mix of markdown in the `root` folder, in the `wiki` folder and in the GH wiki.

When ready and public we can also move the docs to the new doc system


nickofthyme (approved):
Love it!

PR #1860, opened 2022-11-07
feat(flame): node size and position tweening

monfera:
## Summary

Completes support for tweening node positions between `position0` -> `position1` and `size0` -> `size1`

The behavior is unchanged  if `columns.position0 === columns.position1 && columns.size0 === columns.size1`

This change allows the following functionality:

With node identities unchanged (meaning that the same node count is present, and the i-th node in both the `*0` and `*1` columns means the same entity: either the position, or size, or both can change. The tweening is currently linear, set to 500ms (as it can be a sweeping change).

## Details

If nodes would mismatch, for example, some nodes go away, while others are entered, with generally different node count between the start and target state:

1. Use the larger of the node count in all arrays; the superfluous entries will just get the appropriate position but their size should be set to zero, so it has no visible effect
2. Set the `size0` of nodes that enter to zero; set the `size1` of nodes that exit to zero
3. Rerender as usual; this will perform the tweening
4. If further updates may happen, it's worth rerendering once again, so that `size1` / `position1` go into both the `*0` and the `*1` columns. This will not cause a visible change, but it'll be easier to perform a similar enter/exit/change cycle in the future

The equality check of the vectors is done via array identity for performance reasons, rather than comparing values one by one in multiple, possibly very large vectors.

Color tweening isn't currently supported, but it'd be possible by adding a new column by the analogy of the position and size handling. Color tweening can be linear initially, as it's very brief anyway (easy to do it with `mix`) and for best results, color tweening in another color space would be the icing on the cake, assuming it leads to a noticeable difference.

### Checklist

### DIFF
--- a/storybook/stories/icicle/04_cpu_profile_gl_flame.story.tsx
+++ b/storybook/stories/icicle/04_cpu_profile_gl_flame.story.tsx
@@ -10,7 +10,16 @@ import { action } from '@storybook/addon-actions';
 import { boolean, button } from '@storybook/addon-knobs';
 import React from 'react';
 
-import { Chart, Datum, Flame, Settings, PartialTheme, FlameGlobalControl, FlameNodeControl } from '@elastic/charts';
+import {
+  Chart,
+  Datum,
+  Flame,
+  Settings,
+  PartialTheme,
+  FlameGlobalControl,
+  FlameNodeControl,
+  ColumnarViewModel,
+} from '@elastic/charts';
 import columnarMock from '@elastic/charts/src/mocks/hierarchical/cpu_profile_tree_mock_columnar.json';
 import { getRandomNumberGenerator } from '@elastic/charts/src/mocks/utils';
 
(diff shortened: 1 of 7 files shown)


storybook/stories/icicle/04_cpu_profile_gl_flame.story.tsx
```diff
@@ -36,16 +45,16 @@ const paletteColorBrewerCat12 = [
...
-const columnarData = {
+const columnarData: ColumnarViewModel = {
   label: columnarMock.label.map((index: number) => columnarMock.dictionary[index]), // reversing the dictionary encoding
   value: new Float64Array(columnarMock.value),
   // color: new Float32Array((columnarMock.color.match(/.{2}/g) ?? []).map((hex: string) => Number.parseInt(hex, 16) / 255)),
   color: new Float32Array(
     columnarMock.label.flatMap(() => [...paletteColorBrewerCat12[pseudoRandom(0, 11)].map((c) => c / 255), 1]),
   ),
-  position0: position, // new Float32Array([...position].slice(1)), // try with the wrong array length
+  position0: position.map((p, i) => (i % 2 === 0 ? 1 - p - size[i / 2] : p)), //.map((p, i) => (i % 2 === 0 ? 1 - p - size[i / 2] : p)), // new Float32Array([...position].slice(1)), // try with the wrong array length
```
monfera, line 55: I'll delete the comment duplication but don't want to spend CI electricity on such a small change

nickofthyme (approved):
LGTM, I'd like to attempt to migrate this implementation to the animation framework I implemented for the annotation animations at some point, in the future.

monfera:
@nickofthyme Thanks for the review. It makes sense to unify the disparate, locally optimized, minimal animations indeed. For example, even if there are 10 charts on a dashboard that animate, it'd be preferable to have them share a `rAF` loop across charts, a bit like what's possible with `d3-timer`. It'd make sense for this animation microlibrary to exist on its own, with its own API etc. living outside the `charts` directory as it's nothing to do with charts—in theory it could animate UI elements or illustrative SVG graphics too. I'm not a big fan of the alternative (internal integration, which doesn't buy too much but makes the chart next to impossible to separate in the future, due to the tight coupling) so ideally, first the animation tools become abstracted away behind an API outside `charts`, then a bunch of code can go.

It's useful to bear in mind the various types of animations:
- some animations may not ever terminate
- the same chart may need multiple animations which are decoupled from one another (still allowing for riding the same `rAF` callback)
- there are curve based animations (tweening over a specific duration) and recurrence based animations (physics, like generating the next value from the current value, delta time and maybe some state, only stopping if movement became close to zero)

markov00 (approved):
LGTM, thanks for enabling such feature!

Issue #1868, opened 2022-11-10
Flamegraph not rendering on FF 106.0.5 Intel Mac 


markov00:
**Describe the issue**
The webgl flamegraph is not rendering on FF version 106.0.5 on Intel Mac.
This is the returned error:

![image](https://user-images.githubusercontent.com/1421091/201157890-a464f963-054b-40b2-bad8-59d8028bd37d.png)

**To Reproduce**
1. Go to [flamegraph storybook example](https://elastic.github.io/elastic-charts/?path=/story/flame-alpha--cpu-profile-gl-flame-chart&globals=theme:light)
2. No flamegraph rendered
3. Open the console and see the error about Uniform location null

**Version (please complete the following information):**
 - OS: macOS
 - Browser: Firefox 106.0.5
 - Elastic Charts: 50.x

**Additional context**
Works fine in Chrome on Intel Mac

**Errors in browser console**
```
Error: Whoa, uniform location null (name: webgl_gl_VertexID_Offset, type: 5124) not found
WebGL warning: validateProgram: Implemented as a no-op on Mac to work around crashes. 2
Error: Whoa, uniform location null (name: webgl_gl_VertexID_Offset, type: 5124) not found
```

PR #1869, opened 2022-11-10
fix: workaround firefox amd macbook regression webgl_gl_VertexID_Offset

monfera:
## Summary

Fixes an apparent Firefox regression on Mac OS with Intel CPUs with a Firefox version on or before 106.0.5 but after 103.0.2. The browser issue itself is potential reportable browser infosec vulnerability, see the details.

The PR also converts most thrown errors to `Logger.warn` to be less draconian. Reporting still happens but erroneous or inconsequential issues won't prevent rendering.

Fixes https://github.com/elastic/elastic-charts/issues/1868

## Details

A Firefox version on or before 106.0.5 but after 103.0.2 introduced reporting `webgl_gl_VertexID_Offset` as a uniform, when querying `gl.ACTIVE_UNIFORMS` by `gl.getProgramParameter(program, GL.ACTIVE_UNIFORMS))`. While we use an *attribute*, called `gl_VertexID` that's implied into existence just by referencing it in the shader code (eg. it consumes from the attrib count limit), this issue looks like GPU driver info leaking into userland. (In this, it's a possible browser infosec issue, as it allows fingerprinting at least)

### Checklist

### DIFF
--- a/packages/charts/src/common/kingly.ts
+++ b/packages/charts/src/common/kingly.ts
@@ -171,12 +171,12 @@ export const createCompiledShader = (
   source: string,
 ): WebGLShader => {
   const shader = gl.createShader(shaderType);
-  if (!shader) throw new Error(`Whoa, shader could not be created`); // just appeasing the TS linter
+  if (!shader) throw new Error(`kinGLy exception: shader could not be created`); // just appeasing the TS linter
   gl.shaderSource(shader, source);
   gl.compileShader(shader);
   if (GL_DEBUG && !gl.getShaderParameter(shader, GL.COMPILE_STATUS) && !gl.isContextLost()) {
     const shaderTypeName = shaderType === GL.VERTEX_SHADER ? 'vertex' : 'fragment';
-    Logger.warn(`Whoa, compilation error in a ${shaderTypeName} shader: ${gl.getShaderInfoLog(shader)}`);
+    Logger.warn(`kinGLy exception: compilation error in a ${shaderTypeName} shader: ${gl.getShaderInfoLog(shader)}`);
   }
   return shader;
 };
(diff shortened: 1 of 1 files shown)


packages/charts/src/common/kingly.ts
```diff
@@ -175,7 +176,7 @@ export const createCompiledShader = (
   gl.compileShader(shader);
   if (GL_DEBUG && !gl.getShaderParameter(shader, GL.COMPILE_STATUS) && !gl.isContextLost()) {
     const shaderTypeName = shaderType === GL.VERTEX_SHADER ? 'vertex' : 'fragment';
-    throw new Error(`Whoa, compilation error in a ${shaderTypeName} shader: ${gl.getShaderInfoLog(shader)}`);
+    Logger.warn(`Whoa, compilation error in a ${shaderTypeName} shader: ${gl.getShaderInfoLog(shader)}`);
```
nickofthyme: I thought this was a native webGL error. lol _Whoa_!

packages/charts/src/common/kingly.ts
```diff
@@ -192,18 +193,18 @@ export const createLinkedProgram = (
   gl.attachShader(program, vertexShader);
   gl.attachShader(program, fragmentShader);
   if (GL_DEBUG && gl.getProgramParameter(program, GL.ATTACHED_SHADERS) !== 2)
-    throw new Error('Did not manage to attach the two shaders');
+    Logger.warn('Whoa, did not manage to attach the two shaders');
```
nickofthyme: Much better wording! 😄

nickofthyme (approved):
LGTM, tried testing on Firefox `105.0.3 (64-bit)` on intel mac but could not reproduce.

packages/charts/src/common/kingly.ts
monfera: Sorry, the subsequent commit replaces these standard prefixes with slightly less emotive wording

markov00 (approved):

PR #1887, opened 2022-11-24
feat(tooltip): animate context menu prompt

markov00:
## Summary

This PR adds the slideIn animation for the tooltip prompt.
The prompt is now visible everytime you stop the mouse on the chart for more then 1.5s.

https://user-images.githubusercontent.com/1421091/203832959-7badc5ba-2fc4-4bf3-b7b9-362d96ebf213.mov

### Checklist


nickofthyme:
buildkite test this

packages/charts/src/components/tooltip/components/tooltip_prompt.tsx
```diff
@@ -13,7 +13,7 @@ import { TooltipDivider } from './tooltip_divider';
 /** @internal */
 export function TooltipPrompt({ children }: PropsWithChildren<{}>) {
   return (
-    <div className="echTooltipPrompt">
+    <div className="echTooltipPrompt slideDown" key={Math.random()}>
```
dej611, line 16: Is this supposed to be unique?

markov00: Yes, this gives us the ability to "reset" the animation starting time at any mouse change. Without that, the prompt animation will start (with its delay) the first time the tooltip is created (your mouse enter the chart area) and will never reset until your mouse will leave the chart area. 
Our idea is to make the prompt appear only if you stop moving for a certain amount of time (tooltip delay) but we want to reset that whenever you move your mouse (hiding again the prompt until the next time you stop your mouse).

The other mechanism was to reset an Interval on every `mousemove` but I thought this could be relatively more cumbersome than the `math.random key` method. If you have other suggestions please don't hesitate to comment

nickofthyme (approved):
LGTM, checked both action and max item cases. I think I'd be nice if the animation delay grew over time. Fast in the beginning, for easier discoverability, then slower after users know the deal. Could be a user flag but just pontificating here 🤔

markov00:
> 

Interesting, definitely a nice enhancement: it could be driven, at kibana level, by:
- number of time the right-click was used (more use less frequent the prompt is shown) 
- the time since the kibana has been upgraded to that version (after 1 or 2 months since the upgrade we can increase the animation delay (we can actually do that programmatically the release after this).
There could be some pro and cons about that but is something we should definitely check.

markov00:
buildkite update screenshots

markov00:
buildkite update screenshots

nickofthyme:
@markov00 can you check the screenshots again? It looks like two of the screenshots updated in 7afde47 shifted enough to change the effect of the click interaction.

image

PR #1888, opened 2022-11-28
chore: update to uuid v9 for better exports support

spalger:
Bumps the uuid version so that uuid does not trigger loading an ESM version via jest. This wasn't previously a problem for Kibana because we had previously implemented a custom resolver for Jest. I'm trying to remove it but the uuid dep in `@elastic/charts` is breaking that branch.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -30,14 +30,4 @@ module.exports = {
       },
     ],
   },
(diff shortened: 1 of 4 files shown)


markov00 (approved):
LGTM thanks for taking care of this!

Issue #862, opened 2020-10-08
Allow a configurable title of the legend


timroes:
To make timelion in Kibana use elastic-charts we'd need to be able to have the "header"/"title" of the legend be configurable. This currently shows the active hovered time when you hover over the chart. We'd need this either built into elastic-chart, or make the title configurable, so we can build this implementation on the Kibana side:

![screenshot-20201008-100941](https://user-images.githubusercontent.com/877229/95455981-ec461300-096e-11eb-8e60-aa492ca076fb.png)

**Kibana cross issue:** https://github.com/elastic/kibana/issues/79984

markov00:
I've taken a deeper look at Timelion and I think we can live without that feature for the following reasons:
- when moving the mouse over the chart, Timelion doesn't show the tooltip, but the legend act as a tooltip: the header is the current cursor position along the x-axis + every series current value. Our tooltip works exactly in the same way at the moment.
- keep consistency with other tools using elastic-charts: I think it's preferable to use the same tooltip all over the place, instead of having Timelion charts with a Legend-Tooltip but Lens one having Legend + Tooltip
- in a near future, the tooltip and the legend can be designed as a unique and interchangeable component, and at that point, we can have a fixed legend/tooltip like today in Timelion. I prefer to work on this together with all the legend and tooltip redesign effort #580 #615 

@timroes are you ok if we change the way Timelion legend/tooltip works and align that to the available behavior of elastic-charts? The need to render the legend within the chart will remain and we will work on that

stratoula:
@markov00 as this legend appears on hover as the tooltip I am fine with that and I agree, let's be consistent

markov00:
Ok good, I'm closing this ticket as we are not going to implement the requested feature. 
If required, please feel free to reopen it

yannbolliger:
Would be awesome to have customisable legend titles.

PR #1889, opened 2022-12-02
feat(legend): add custom legend component

### LINKED ISSUE #561: Make the legend extra parameter customizable
The current legend extra parameter is used to display the latest value on the data series, or the hovered data point value. We should make this configurable to show other values or statistics, like the total value, averages etc

We should move the legend props to it's own `<Legend>` component and allow the user to customize in some way the possible extra parameter

n/a

**Additional context**
continuation of https://github.com/elastic/elastic-charts/issues/246

**Kibana Cross Issues**
- https://github.com/elastic/kibana/issues/19485
- https://github.com/elastic/kibana/issues/16096
- https://github.com/elastic/kibana/issues/10524
- https://github.com/elastic/kibana/issues/6279

**Checklist**

yannbolliger:
## Summary

The settings now have a new prop `customLegend` which lets users provide there own legend React component that is rendered with the chart state into the legend container.

## Details

Analogously to the `customTooltip` in the `Tooltip`, the `customLegend` takes a React component which will receive the state of the standard legend as props. This allows users to provide their own legend implementation. The legend is still rendered inside of the `div.echLegendListContainer`.

## Issues

Fixes https://github.com/elastic/elastic-charts/issues/862 because it is now feasible with a custom component as well as fixes #561.

### Checklist


nickofthyme:
buildkite test this

packages/charts/src/components/legend/legend.tsx
```diff
@@ -164,6 +171,7 @@ const mapStateToProps = (state: GlobalChartState): LegendStateProps => {
     size: getLegendSizeSelector(state),
     items: getLegendItemsSelector(state),
     extraValues: getLegendExtraValuesSelector(state),
+    tooltipHeader: getInternalTooltipInfoSelector(state)?.header,
```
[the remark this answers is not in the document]
yannbolliger: Thanks for the question. As you might have seen this is not yet very beautiful.
What I'd need here is the timestamp _aka_ the x-value of where the mouse pointer is hovering, indeed.
So, I should probably create a `currentXvalueSelector` or something the like?

packages/charts/src/specs/settings.tsx
```diff
@@ -365,6 +368,21 @@ export type LegendPositionConfig = {
...
 };
 
+/**
+ * The props for {@link CustomLegend}
+ * @public
+ */
+export interface CustomLegendProps extends Omit<LegendItemProps, 'item'> {
+  items: LegendItem[];
+  pointerValue?: PointerValue;
+}
```
markov00, line 378: The `LegendItemProps` is marked as `@internal` type, and this cause the API error you see on the CI.
To fix this, the easiest way is to mark `LegendItemProps` and all its child types as `@public`, but this is not the best solution because we are exposing as part of the API something that we use mostly internally.
Our philosophy is to limit as much as possible what we are exposing to minimize breaking changes.
In this case you probably just want to expose only a few props of the `LegendItemProps` type like:
```
extraValues
mouseOverAction
mouseOutAction
```

The `mouseOverAction` and `mouseOutAction` are two redux action types that we probably don't want to expose. Is better to redefine them as:
```
onLegendItemOverActon: (path: LegendPath) => void
onLegendItemOutAction: () => void
```
to minimize the exposure of internal implementation details.

WDYT?

markov00 (commented):
I've took the chance to push a commit to:
- fix the legend style (avoiding the box shadow on the legend title)
- add a small notice about always using the `legendSize` when using the `customLegend` (it can probably be enforced with typescript somehow, because if not the auto-legend-size will kick in and the available size for the legend will not reflect the actual custom legend size

packages/charts/src/specs/settings.tsx
yannbolliger: Sounds good to me 👍 
Will implement the feedback in a next commit.

yannbolliger: It would be nice, to still be able to hide/make transparent all the series except the one of the currently hovered legend-item, so if possible we'd like to keep the actions in the props. Of course, we could rename them or something...

yannbolliger:
buildkite update screenshots

nickofthyme:
buildkite update screenshots

nickofthyme:
Sorry @yannbolliger the ci should permit you to run the **_buildkite update screenshots_** command when the `ci:approved ✅` label is added to the pr.

nickofthyme:
buildkite test this

markov00:
@nickofthyme @yannbolliger I'm taking a loot at that added selector in the Legend component. It will triggers a react component rerendering (most of the time will not change the DOM but still execute the legend component and diff the results) on every mouse move. I prefer to avoid that if possible. I think we need to improve the current `getTooltipInfo` to avoid this recomputation as much as possible. I will be back soon with a solution to that

yannbolliger:
> that added selector in the Legend component. It will triggers a react component rerendering (most of the time will not change the DOM but still execute the legend component and diff the results) on every mouse move. 

We could connect only the custom legend component (and not the regular one) via Redux to this additional selector.

markov00:
> > that added selector in the Legend component. It will triggers a react component rerendering (most of the time will not change the DOM but still execute the legend component and diff the results) on every mouse move.
> 
> We could connect only the custom legend component (and not the regular one) via Redux to this additional selector.

That could work, is anyway just a workaround, the output of such selector, is always the same if you hover over the same geometry but you move slightly your mouse

yannbolliger:
OK, so I changed the PR to only provide the `pointerValue` from Redux to the custom legend component. This should fix what you were concerned about @markov00. (Even though my testing showed that the legend is re-rendered when the cursor moves even on the `main` branch...)

markov00 (approved):
Tested locally, looks good to me.
Thanks @yannbolliger for the changes.
@nickofthyme Yann was right, we are will rerendering the legend independently from the changes in this PR. The legend selectors are firing too on each mouse move, is better to double-check this and fix it on our selectors.

PR #1899, opened 2022-12-12
fix: tooltip white border on custom tooltips

### LINKED ISSUE #1901: Remove the white border for custom tooltip
**Describe the issue**
The tooltip has a 1px border by default (that is useful to split the optional color stripe from the chart background).
This is applied to the container of the tooltip.

When rendering a custom tooltip, the white border remains visible creating the following issue: a border on every custom tooltip and in particular a 1x1 border pixel when rendering a null tooltip (in the case when we just want to show the cursor line and no tooltip)

Screenshot 2022-12-12 at 17 54 57

**To Reproduce**
1. Go to [the custom legend PR preview](https://ech-e2e-ci--pr-1889-0j6h522w.web.app/?path=/story/legend--custom-legend&globals=theme:light)
2. Move your mouse around
3. See the white dot floating near the cursor

**Expected behaviour**
I'm expecting to see a such border only with our exposed custom tooltip components.

 - Elastic Charts: 51.x

**Additional context**

I see some possible solutions:
- find a better way to put such white border at the left of the color stripe
- create a `TooltipContainer` component that needs to be used to contain our exposed tooltips
- find a way to force such border when using the color-strip

markov00:
## Summary

Adds a `style` prop to the `TooltipSpec` to enable overriding _select_ styles on the main tooltip wrapper element.

## Details

Permits a users with custom tooltips to control the styles of the `echTooltip` wrapper element, this contains both the custom tooltip and the ech-controlled actions UI below.

An alternative approach would be to add a prop to hide and show the tooltip. Either solution would solve the problem, but the `styles` approach allows for future expansion of user style adjustments.

## Issues

fix #1901

### Checklist


nickofthyme:
buildkite test this

nickofthyme:
buildkite update vrt

nickofthyme:
buildkite update vrt

nickofthyme (approved):

markov00:
@nickofthyme I don't like too much this solution. I don't want to push this to the user. This should be a problem of our tootlip and the user should not need to disable it to use a custom one

packages/charts/api/charts.api.md
```diff
@@ -2804,7 +2804,7 @@ export function toEntries<T extends Record<string, string>, S>(array: T[], acces
 export type ToggleSelectedTooltipItemCallback = (item: TooltipValue<any, SeriesIdentifier>) => any;
```
nickofthyme, line 2805: Ok, another alternative is to move the border back to the list element on the left side only where I had it originally. This would look like...

image

image

markov00: the problem is the one shown in the partition: the first and last element doesn't get the nice rounded border...

markov00:
I'm going to close this in favour of #1915

PR #1906, opened 2022-12-16
fix: don't account for label overflow with multi-layer time axis

markov00:
## Summary

When using the multi-layer time axis we should not consider label overflows as part of the dynamic axis size calculation.
This PR address that giving the time axis the full size is achievable without considering label overflows.

**Before** (note the space at the right of the chart, it could happen also at the left)
Screenshot 2022-12-16 at 15 42 50

**After**

Screenshot 2022-12-16 at 15 42 29

## Details

The axis size is computed initially depending on its content. We have a poor algorithm that is used to compute the max overflow a chart can have: it is actually defined as the half (because categorical labels are centered on tick) of the longest label (considering that a label center can be placed at the far end of a chart).
Unfortunately this is not a good choice, and cause numerous problem.
In this PR I'm addressing this problem only for the multilayer time axis that are self contained and no label will overflow the axis.

## Issues

It finally fixes the problem described here https://github.com/elastic/kibana/issues/138390 by removing any remaining margins

### Checklist


markov00:
buildkite update screenshots

nickofthyme (approved):
So much _less_ whitespace! 😍

PR #1915, opened 2022-12-29
feat: tooltip container

### LINKED ISSUE #1901: Remove the white border for custom tooltip

markov00:
## Summary

This PR changes a bit the way tooltip components needs to be used. It introduce the `TooltipContainer` container element that needs to be used as the parent element of a new custom tooltip that uses our Tooltip components.

The PR also:
- exposes the `useTooltiopContext()` hook to be able to pick the necessary info down the custom component tree
- the custom tooltip is not completely separated by the customization of footer/header/body. 
  - The former, the custom tooltip, should leave on its own and should handle by its self what was previously done internally like: actions/table body limits etc. This because we don't want to pollute custom components with elements that are not naturally built for them (like actions on a fully custom tooltip).
  - the latter, customization of body, footer, header, is instead done within the Tooltip internals and brings all the advantages. You **should** use our Tooltip Components to customize those parts. This customization helps us applying actions to a tooltip that has the same look-and-feel that we expect.
- fixes the `TooltipTable` number cell to align the number to the `left` of the cell by default    

This PR also fixes a small bug about a tooltip style (border-radius and white border) applied to custom tooltip too: #1901

### BREAKING CHANGE
You must wrap your `customTooltip` components with a `TooltipContainer` if you are using our `Tooltip` components. The `actions`, if available, are no longer applied to a `customTooltip`.

## Issues

fix #1901 

### Checklist


markov00:
buildkite update screenshots

packages/charts/src/components/tooltip/components/tooltip_table_body.tsx
```diff
@@ -82,7 +82,7 @@ function getCellStyles<D extends BaseDatum = Datum, SI extends SeriesIdentifier
   style,
   type,
 }: TooltipTableColumn<D, SI>): TooltipCellStyle {
-  const textAlign: TooltipCellStyle['textAlign'] = type === 'number' ? 'left' : type === 'text' ? 'right' : undefined;
+  const textAlign: TooltipCellStyle['textAlign'] = type === 'number' ? 'right' : type === 'text' ? 'left' : undefined;
```
markov00, line 85: this was an existing issue: in a table, number should be aligned to the right and text to the left

markov00:
@nickofthyme I need your review here

packages/charts/api/charts.api.md
```diff
@@ -2902,9 +2935,19 @@ export interface TooltipSpec<D extends BaseDatum = Datum, SI extends SeriesIdent
     actionsLoading: string | ComponentType<{
         selected: TooltipValue<D, SI>[];
     }>;
+    body: 'default' | 'none' | ComponentType<{
+        items: TooltipValue<D, SI>[];
+        header: TooltipValue<D, SI> | null;
+    }>;
     customTooltip?: CustomTooltip<D, SI>;
```
nickofthyme, line 2942: Are you planning to deprecate this at some point with these changes?

packages/charts/src/components/tooltip/components/tooltip_wrapper.tsx
```diff
@@ -72,7 +72,7 @@ export const TooltipWrapper = <D extends BaseDatum = Datum, SI extends SeriesIde
       onClick={(e) => e.stopPropagation()} // block propagation of tooltip click
       onKeyPress={(e) => e.stopPropagation()} // block propagation of tooltip click
     >
-      {renderComplexChildren(children)}
+      {children}
```
nickofthyme, line 75: So this is now handled in the `TooltipContainer` correct?

packages/charts/src/components/tooltip/components/tooltip_table.tsx
```diff
@@ -55,14 +54,11 @@ export const TooltipTable = <D extends BaseDatum = Datum, SI extends SeriesIdent
...
-  const { theme, maxItems, ...rest } = useTooltipContext<D, SI>();
-
   if ('children' in props) {
-    const { gridTemplateColumns } = props;
+    const { gridTemplateColumns, maxHeight } = props;
     const classes = classNames('echTooltip__table', className, {
       'echTooltip__table--noGrid': !gridTemplateColumns,
     });
-    const maxHeight = props.maxHeight ?? theme.maxTableHeight;
     return (
```
nickofthyme, line 62: We don't need this fallback?

packages/charts/src/components/tooltip/components/tooltip_table.tsx
```diff
@@ -71,22 +67,18 @@ export const TooltipTable = <D extends BaseDatum = Datum, SI extends SeriesIdent
       </div>
     );
   }
-
   const columns = props.columns.filter(({ hidden }) => {
     return !(typeof hidden === 'boolean' ? hidden : hidden?.(props.items) ?? false);
   });
 
-  const { items, pinned = false, onSelect, selected = [] } = { ...rest, ...props };
+  const { items, pinned = false, onSelect, selected = [] } = { ...props };
```
nickofthyme, line 74: 😱

packages/charts/src/components/tooltip/tooltip.tsx
```diff
@@ -164,6 +173,7 @@ export const TooltipComponent = <D extends BaseDatum = Datum, SI extends SeriesI
     ...(info?.values?.map?.(({ label }) => label) ?? []),
     info?.header?.label ?? '',
   ]);
+  const textDirectionality = isMostlyRTL ? 'rtl' : 'ltr';
 
   const columns: TooltipTableColumn<D, SI>[] = [
     {
```
nickofthyme, line 179: It would be nice to remove the scroll bar styles when the tooltip is not `pinned`. Cuz it looks scrollable when it is not.

Components Tooltip - Actions With Custom Tooltip ⋅ Storybook 2023-01-23 at 9 22 18 AM

packages/charts/src/components/tooltip/tooltip.tsx
```diff
@@ -214,52 +224,107 @@ export const TooltipComponent = <D extends BaseDatum = Datum, SI extends SeriesI
...
+            >
+              {TooltipCustomHeader === 'none' ? null : TooltipCustomHeader === 'default' ? (
+                <TooltipHeader header={info.header} formatter={headerFormatter} />
+              ) : (
+                <TooltipHeader>
+                  <TooltipCustomHeader items={info.values} header={info.header} />
+                </TooltipHeader>
+              )}
+
+              {headerBottomDividerVisibility && <TooltipDivider />}
```
nickofthyme, line 298: Nice!

storybook/stories/components/tooltip/1_composed_list.story.tsx
```diff
@@ -10,14 +10,14 @@ import { action } from '@storybook/addon-actions';
 import { boolean, number, select } from '@storybook/addon-knobs';
 import React from 'react';
```
nickofthyme, line 12: **Note: this is old behavior, not related to your changes**

I noticed when we use the `TooltipTable` we have the user handle the `onSelect` method. This story just logs a storybook action. But if the charts consumer does this they now need to handle the selection logic. Do you think we should...

1. Handle the selection always and just send the clicked action `onSelect`.
2. Default to handling the selection but allow consumer to opt-out and handle the logic themselves via `useTooltiopContext`.
3. Keep it as is and always require the consumer to handle the selection logic manually.

Screen Recording 2023-01-23 at 09 49 01 AM

nickofthyme (approved):
Thanks for the ping, I lost this PR in my notifications.

> You must wrap your `customTooltip` components with a `TooltipContainer`

Is there a way to check or enforce this, seems like this could be another blind spot for developer. Maybe we could have a `try`/`catch` to the `useTooltiopContext` to look for an empty `Provider`.

---

Tested changes locally on Chrome and all functionality is working correctly. Just a few comments and questions.

packages/charts/api/charts.api.md
markov00: are you referring to the `customTooltip`? if so no I don't think we need to deprecate it

packages/charts/src/components/tooltip/components/tooltip_table.tsx
markov00: I don't even know what it means... I completely forgot what is this. Am I correct if I fix it this way?
```suggestion
  const { items, pinned = false, onSelect, selected = [] } = props;
```

packages/charts/src/components/tooltip/tooltip.tsx
markov00: I know, I was on doubt on that. Because it shows you a bit more that there is something more there, and only with right-click you can discover it

storybook/stories/components/tooltip/1_composed_list.story.tsx
markov00: Good question, I think we always need to handle the selection because that's they way actions, prompts etc works naturally: what is selected and what not is passed into their rendering function

packages/charts/src/components/tooltip/components/tooltip_table.tsx
nickofthyme: Ok perfect 👍🏼

packages/charts/src/components/tooltip/components/tooltip_table.tsx
nickofthyme: I read this wrong initially thinking it was a useless destructure.

I think this dual spread is still needed unless `props` contains `items`, `pinned`, `onSelect` and `selected`, which in this case it does but they can be undefined.

The previous logic was spreading the `rest` values from the `useTooltipContext` as defaults if they were not defined in the `props`.

So I'd say revert this to the old way and add back the `useTooltipContext` above.

```suggestion
  const { items, pinned = false, onSelect, selected = [] } = { ...rest, ...props };
```

storybook/stories/components/tooltip/1_composed_list.story.tsx
nickofthyme: Yeah, I don't see a good use case for them having to update the selection on click. That would always toggle a given value's selection. I'll create a new issue.

packages/charts/src/components/tooltip/tooltip.tsx
nickofthyme: FYI the `Total of X Categories` is a custom `footer` just for this story and not standard.

https://github.com/elastic/elastic-charts/blob/af1461d0c25e84ac2169d239f0fbc8b3fc4fd542/storybook/stories/components/tooltip/4_actions_with_custom_tooltip.story.tsx#L85

markov00: done in [a998979](https://github.com/elastic/elastic-charts/pull/1915/commits/a998979b6e0893deecaddaa84af4c0ee50ec27a0)

packages/charts/src/components/tooltip/components/tooltip_table.tsx
markov00: done in https://github.com/elastic/elastic-charts/pull/1915/commits/d5f88e642c4240d918102d6b095472953295814d
Only `selected` and `pinned` are available in the `tooltipContext`. `onSelect` is not in the context, `items` will be overwritten by props anyway.

nickofthyme (approved):

PR #1951, opened 2023-01-30
feat(metric): new size breakpoints

markov00:
## Summary
Add a bunch of new breakpoints for metric size xl, xxl, xxxl to handle bigger screens and panels in accordance with @gvnmagni


nickofthyme (approved):
`xxxl` 😮. Breakpoints LGTM, tested locally on chrome.

markov00:
> `xxxl` 😮. Breakpoints LGTM, tested locally on chrome.

I know, but it make sense if you really are looking for such big charts in a big 4meter x 4meter dashboard on a wall :D

PR #1956, opened 2023-02-07
fix(partition): show partitions with zero value on legend 

markov00:
## Summary

This PR fixes the missing presence of the partitions with a zero value in the legend.
There are cases where the legend should show all non-negative and finite values passed to the pie chart, for example: when we have the overall count of `positive` and `negative` replies to a question and one of the two scores to zero.
The renderer should not show a zero slice, but the legend, for consistency, should display both of them.

## Details

The fix is implemented in pie charts and treemap. It is not required for flamegraph because the legend works just fine (and in most case is not a good practice to show).
For the pie chart, we just filter negative and non-finite-values from the data.
For treemap instead, we also need to fix the calculation to avoid infinite loops.

## Issues

This is a preliminary fix for https://github.com/elastic/kibana/issues/149681.
A subsequent fix within Lens is required to convert `null` values to `zero`s where necessary

## Notes to reviewers

The VRT changes are due to:
- the legend is now visible in some pre-existing  screenshots
- the legend for pie/donut is sorted hierarchically even if the legend is flat (see BREAKING CHANGES)
- The legend in VTRs that use a custom sorting function is now sorted also by value in a descending order due to the sorted nature of the hierarchical tree model used to generate the legend.
- some charts use a different color schema due to a missing parameter that can't be passed with the new `fillColor` arguments. In particular, we can't easily color by "angle" because we don't expose anymore that information, but it can be probably computed somehow as a workaround from the data itself. I haven't pursued that exercise in the code and I just considered using a simplified color function instead.

## BREAKING CHANGES

The `Layer.fillColor` function now accepts the following signature: `(key: Key, sortIndex: number, node: ArrayNode, tree: HierarchyOfArrays) => string;`. This exposes a similar set of information but allows us more control over the internals to define/generate the legend.
The legend for pie/donut/mosaic/treemap charts is sorted in a hierarchical way even if the legend is flat. The original behavior wasn't explicit enough to be clearly understood. Now, the default sorting order is predictable even under a flat legend.
We kept the original behaviors for flame/icicle/waffle charts due to their nature: flame/icicle usually have too deeply nested hierarchy and is hard to understand in a simple list; the waffle rendering nature instead makes a linear top/bottom natural ordering of the elements. 
The default highlight strategy, used to highlight partition elements when hovering over the legend item has being changed to `LegendStrategy.Path` because seems the most aligned with the hierarchical nature of the legend sorting

### Checklist


packages/charts/src/chart_types/partition_chart/partition.test.tsx
```diff
@@ -129,131 +129,7 @@ describe('Retain hierarchy even with arbitrary names', () => {
...
-            { index: 2, value: 'C' },
-            { index: 1, value: 'B' },
-          ],
-          depth: 1,
-          label: 'B',
-          seriesIdentifiers: [{ key: 'B', specId: 'spec1' }],
-          keys: [],
-        },
-      ]);
+      expect(computeLegendSelector(store.getState())).toMatchSnapshot();
```
markov00, line 132: replaced by snapshots (the only real change is that the legend also returns the `defaultExtra` value

packages/charts/src/chart_types/partition_chart/layout/utils/highlighted_geoms.ts
```diff
@@ -85,7 +85,7 @@ export const LegendStrategy = Object.freeze({
 /** @public */
 export type LegendStrategy = Values<typeof LegendStrategy>;
 
-const defaultStrategy: LegendStrategy = LegendStrategy.Key;
+const defaultStrategy: LegendStrategy = LegendStrategy.Path;
```
markov00, line 88: This method seems way more useful as it shows always the path from the root to the current selected node

markov00:
buildkite update screenshots

markov00:
buildkite update screenshots

packages/charts/src/specs/settings.tsx
```diff
@@ -417,6 +417,7 @@ export interface LegendSpec {
   /**
    * Limit the legend to the specified maximal depth when showing a hierarchical legend
    */
+  //TODO this is not the max depth, but the number of level shown: 0 none, 1 first, 2 up to the second etc
   legendMaxDepth: number;
```
nickofthyme, line 421:
```suggestion
  /**
   * Limit the legend to the specified maximal depth when showing a hierarchical legend
   * 
   * TODO: this is not the max depth, but the number of level shown: 0 none, 1 first, 2 up to the second etc
   */
  legendMaxDepth: number;
```

Can you put this comment in the TDDoc comment?

storybook/stories/legend/10_sunburst.story.tsx
```diff
@@ -125,22 +123,23 @@ export const Example = () => {
             groupByRollup: (d: Datum) => d.sitc1,
             nodeLabel: (d: any) => productLookup[d].name,
             shape: {
-              fillColor: (d: ShapeTreeNode) => discreteColor(colorBrewerCategoricalStark9, 0.7)(d.sortIndex),
+              fillColor: (key, sortIndex) => discreteColor(colorBrewerCategoricalStark9, 0.7)(sortIndex),
             },
```
nickofthyme, line 127: This is a big improvement in the API!

packages/charts/src/chart_types/partition_chart/state/selectors/compute_legend.ts
```diff
@@ -8,27 +8,132 @@
...
+                smAccessorValue: tree.smAccessorValue,
+                specId: aItem.item.seriesIdentifiers[0].specId,
+                key: aItem.item.seriesIdentifiers[0].key,
+              } as SeriesIdentifier,
+              {
+                smAccessorValue: tree.smAccessorValue,
+                specId: bItem.item.seriesIdentifiers[0].specId,
+                key: bItem.item.seriesIdentifiers[0].key,
+              } as SeriesIdentifier,
+            )
```
nickofthyme, line 63: is the `as` necessary here? If so maybe we can tweak the `settings.legendSort` type signature to be flexible, something like...

```ts
export type SeriesCompareFn<SI extends SeriesIdentifier> = (siA: SI, siB: SI) => number;
```

packages/charts/src/components/legend/legend_item.tsx
```diff
@@ -235,8 +237,3 @@ export class LegendListItem extends Component<LegendItemProps, LegendItemState>
     );
   }
 }
-
-/** @internal */
-export function renderLegendItem(item: LegendItem, props: Omit<LegendItemProps, 'item'>, index: number) {
-  return <LegendListItem key={`${index}`} item={item} {...props} />;
-}
```
nickofthyme, line 242: 👍🏼

nickofthyme (approved):
Reviewed code changes and all LGTM. Gonna poke around in storybook in the morning.

dej611 (commented):
Should the total zero case still show the legend? I've tweaked the story locally and nothing happens: is that by design?
Screenshot 2023-02-28 at 11 39 07

I've tested locally also all other partition chart types with the legend enabled and could not find other issues 👍

packages/charts/src/chart_types/partition_chart/state/selectors/compute_legend.ts
markov00: Oh yea, definitely better

packages/charts/src/chart_types/partition_chart/state/selectors/compute_legend.ts
```diff
@@ -8,27 +8,132 @@
...
+): { item: LegendItem; node: ArrayNode }[] {
+  if (tree.length === 0) {
+    return legendItems;
+  }
+
+  for (const [key, node] of tree) {
+    const layer = layers[depth - 1];
+    const formatter = layer?.nodeLabel ?? ((d) => `${d}`);
+
+    const fill = layer?.shape?.fillColor ?? 'rgba(128, 0, 0, 0.5)';
```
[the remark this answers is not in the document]
markov00: done in https://github.com/elastic/elastic-charts/pull/1956/commits/177e63a2fa8f85bde960c7209bb7ccbc2cc01bde

packages/charts/src/chart_types/partition_chart/state/selectors/compute_legend.ts
```diff
@@ -8,27 +8,132 @@
...
+  HierarchyOfArrays,
+  PATH_KEY,
+} from '../../layout/utils/group_by_rollup';
+import { isLinear } from '../../layout/viewmodel/viewmodel';
+import { Layer, PartitionSpec } from '../../specs';
+
+function compareLegendItemNames(aItem: LegendNode, bItem: LegendNode): number {
+  const a = aItem.item.label;
+  const b = bItem.item.label;
+  return a < b ? -1 : a > b ? 1 : 0;
```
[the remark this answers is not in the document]
markov00: done in https://github.com/elastic/elastic-charts/pull/1956/commits/177e63a2fa8f85bde960c7209bb7ccbc2cc01bde

packages/charts/src/chart_types/partition_chart/state/selectors/compute_legend.ts
```diff
@@ -8,27 +8,132 @@
...
+                smAccessorValue: tree.smAccessorValue,
+                specId: bItem.item.seriesIdentifiers[0].specId,
+                key: bItem.item.seriesIdentifiers[0].key,
+              } as SeriesIdentifier,
+            )
+        : undefined;
+      const useHierarchicalLegend = isHierarchicalLegend(flatLegend, legendPosition);
+      const { valueFormatter } = specs[0];
+      const items = walkTree(specs[0].id, useHierarchicalLegend, valueFormatter, tree.tree, specs[0].layers, 0);
+      return [...items.values()]
```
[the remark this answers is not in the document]
markov00: Good catch (I was using previously a Set and this was left over thanks)  https://github.com/elastic/elastic-charts/pull/1956/commits/177e63a2fa8f85bde960c7209bb7ccbc2cc01bde

markov00:
> Should the total zero case still show the legend? I've tweaked the story locally and nothing happens: is that by design?

@dej611 good question: with all slices with zero values, I believe the chart is correctly not showing (because it breaks a bit the Mathematics if I'm showing that I have 4 categories with a value of zero shown as 25% of a whole, that is zero). But I see a reason to show probably the legend with this info. I will check what is blocking the renderer, but I assume that is a more complex thing to fix with the current internal

dej611:
> > Should the total zero case still show the legend? I've tweaked the story locally and nothing happens: is that by design?
> 
> @dej611 good question: with all slices with zero values, I believe the chart is correctly not showing (because it breaks a bit the Mathematics if I'm showing that I have 4 categories with a value of zero shown as 25% of a whole, that is zero). But I see a reason to show probably the legend with this info. I will check what is blocking the renderer, but I assume that is a more complex thing to fix with the current internal

I can see how the rendering pipeline could be hard to tweak in order to show the all-0 legend. I would be ok even to split this edge case out and track it into a separate issue.

nickofthyme (approved):
After testing changes on the firebase PR deployment all changes appear to function correctly.

packages/charts/src/chart_types/partition_chart/state/selectors/compute_legend.ts
markov00: I've created an issue to track this https://github.com/elastic/elastic-charts/issues/1982
It complicates this PR because we need to pass the SI from the Settings component and I think we will find anyway some blockers of this method because we don't have those types in the internal selectors

dej611 (approved):

markov00:
> I can see how the rendering pipeline could be hard to tweak in order to show the all-0 legend. I would be ok even to split this edge case out and track it into a separate issue.

@dej611  I've created the following issue to track this enhancement https://github.com/elastic/elastic-charts/issues/1983

packages/charts/src/specs/settings.tsx
markov00: done in https://github.com/elastic/elastic-charts/pull/1956/commits/18913e0ccbb8549d45a624502e59fa854c76e116

markov00:
buildkite update screenshots

PR #1958, opened 2023-02-09
fix(xy): respect `integersOnly` prop on axis component

### LINKED ISSUE #1920: `integersOnly` is no longer working
**Describe the issue**
The `integersOnly` prop in `Axis` component is no longer working.

**To Reproduce**
1. Go to https://codesandbox.io/s/small-hooks-7gxxv5?file=/src/App.tsx
2. The Y axis shows non-integers ticks even if the `integersOnly` prop is enabled

**Expected behaviour**
With `integersOnly` the associated axis should show only integers ticks

**Screenshots**
Screenshot 2023-01-03 at 12 58 20

 - OS: all
 - Browser: all
 - Elastic Charts: since version [38.0.0](https://github.com/elastic/elastic-charts/blob/main/CHANGELOG.md#3800-2021-10-15)

**Additional context**
The problem was introduced in version [38.0.0](https://github.com/elastic/elastic-charts/blob/main/CHANGELOG.md#3800-2021-10-15). In version 37 it was working fine (tested with the example in codesandbox).
It looks like the code is no more using `integersOnly` to remove non-integers tick values. Probably due to the [changes in the scales](https://github.com/elastic/elastic-charts/pull/1410)

There was no screenshot testing this feature or it just wen't unnoticed when we switched to playwright

A test case was added in https://github.com/elastic/elastic-charts/pull/1922

**Errors in browser console**
N/A

**Kibana Cross Issues**
N/A

**Checklist**

dej611:
## Summary

Continuous axis can now render integers-only values when requested.

Before:
Screenshot 2023-02-09 at 09 27 29

After:
Screenshot 2023-02-09 at 09 25 46

## Issues

fix #1920 

### Checklist


packages/charts/src/scales/scale_continuous.ts
```diff
@@ -166,14 +172,18 @@ export class ScaleContinuous {
...
+            type === ScaleType.Linear
+              ? getLinearNonDenserTicks(
+                  nicePaddedDomain,
+                  scaleOptions.desiredTickCount,
+                  this.linearBase,
+                  scaleOptions.bandwidth === 0 ? 0 : scaleOptions.minInterval,
+                  scaleOptions.integersOnly,
+                )
+              : (d3Scale as D3ScaleNonTime).ticks(scaleOptions.desiredTickCount),
+          );
```
markov00, line 186: If you are already filtering within the `getLinearNonDenserTicks` why do you want also to wrap the output into  `filterValues` and filter it again? 

I see this applied to the non-linear scale, but it is a waste doing filtering 2 times for linear ones.

You can probably avoid the `filterValues` function completely, keep the filtering within the  `getLinearNonDenserTicks` function and apply a filter only of the non-linear one.

packages/charts/src/scales/scale_continuous.ts
```diff
@@ -290,7 +301,7 @@ function getLinearNonDenserTicks(
     currentCount--;
     ticks = getLinearTicks(start, stop, currentCount, base);
   }
-  return ticks;
+  return integersOnly ? [...ticks.filter((v) => v % 1 === 0)] : ticks;
```
markov00, line 304: you don't the spread operator here
```suggestion
  return integersOnly ? ticks.filter((v) => v % 1 === 0) : ticks;
```

packages/charts/src/scales/scale_continuous.ts
```diff
@@ -63,6 +63,12 @@ const defaultScaleOptions: ScaleOptions = {
 };
 
 const isUnitRange = ([r1, r2]: Range) => r1 === 0 && r2 === 1;
+const filterValues = ({ integersOnly }: Pick<ScaleOptions, 'integersOnly'>, values: number[]) => {
+  if (!integersOnly) {
+    return values;
+  }
+  return values.filter((v) => Number.isInteger(v));
+};
```
markov00, line 71: even if this can be removed as suggested in another comment, I'd like to highlight one thing here: the `filterValues` function name is not really clear about what is filtering. Something like `integerOnly` or something else to make it explicit what you are filtering.

packages/charts/src/scales/scale_continuous.ts
markov00: or even reusing the same functionality used before:
```suggestion
  return integersOnly ? ticks.filter(Number.isInteger) : ticks;
```

packages/charts/src/scales/scale_continuous.test.ts
```diff
@@ -202,6 +212,18 @@ describe('Scale Continuous', () => {
...
     });
+
+    it('should return integers ticks only when requested', () => {
+      const scale = new ScaleContinuous(
+        {
+          type: ScaleType.Linear,
+          domain: [0, 2],
+          range: [0, 3],
+        },
+        { integersOnly: true },
```
markov00, line 223: Is better to force manually the desired number of ticks to a number like 10 (even if this is the default), because in the future, if we change that default to something like 3 this test can pass even if the `integersOnly` doesn't work

packages/charts/src/scales/scale_continuous.ts
dej611: I thought at that at first, passing only the `integersOnly` flag, but then decided to pass the full `scaleOptions` to have a more generic filtering value function. If a new flag will be introduced then the same function can be reused.
No strong opinion hold here. Can refactor to be more specific at will.

markov00: The scope of the fix is really limited and I don't see the need for a generic filtering function here at the cost of readability, the function is called `filterValues` and to me it looks a bit too generic to be quickly read. A more specific name in this case could simplify the code a bit.

We can also avoid the function completely since the number of ticks is relatively limited (up to 10/15 ticks) and thus calling always a filter function doesn't impact the performance: 
```.filter(scaleOptions.integersOnly ? Number.isInteger : () => true)````

dej611:
buildkite update screenshots

markov00 (approved):

PR #1979, opened 2023-02-27
docs: fix enum knob types

nickofthyme:
## Summary

Internal changes to the `knobs.tsx` file used in storybook and VRT.

## Details

This PR reduces some tech debt related to our storybook knob/interaction code.

The problem is that we use knobs all around storybook and often replicate the same code to generate knobs across all of our story code. This was helped by the use of the `knobs.tsx` utils file especially when using `getKnobFromEnum` function to generate knobs for any given charts enum to be used as a storybook `select` knob.

Until now the `getKnobFromEnum` function was typed very loosely to facilitate speed of development. I recently noticed that we use very similar `select`s across `stories/`.  In some cases this is because the property only accepts a partial union of the full enum, think `ScaleType` used as `yScaleType` which will throw a type error.

All that being said, the changes in this PR include...
- Separating the `knobs.tsx` file into a `knobs/` directory for better organization
- Strongly type the `getKnobFromEnum` to account for optional `undefined` and `include` and `exclude` options to selectively control the options per story **WITH** accurate type return values.
- Create a `getKnobsFnFromEnum` to simplify creating custom knobs for enums, with the added benefit of the strong typings from `getKnobFromEnum`.
- Export all custom knob utils from `knobs/index.ts` as `customKnobs`. This allows for much easier discoverability and usage of the different knobs available.

### Example changes before & after

Single import, many knob options

```diff
- import { getChartRotationKnob, arrayKnobs } from '../../utils/knobs';
+ import { customKnobs } from '../../utils/knobs';

  export const Example = () => {
-   const rotation = getChartRotationKnob();
+   const rotation = customKnobs.enum.rotation();
-   const data = arrayKnobs('data values', [2.5, 7.2]);
+   const data = customKnobs.array('data values', [2.5, 7.2]);
  }

```

No more using `as XX`
```diff
- const paddingUnit = getKnobsFromEnum(
-   'Domain padding unit',
-   DomainPaddingUnit,
-   DomainPaddingUnit.DomainRatio as DomainPaddingUnit,
- );
+ const paddingUnit = customKnobs.fromEnum(
+   'Domain padding unit',
+   DomainPaddingUnit,
+   DomainPaddingUnit.DomainRatio,
+);
```

Simple `include`/`exclude` option with dynamically accurate types.

```diff
- const yScaleType = select<typeof ScaleType.Linear | typeof ScaleType.Log | typeof ScaleType.Sqrt>(
-   'Scale Type',
-   {
-     Linear: ScaleType.Linear,
-     Log: ScaleType.Log,
-     Sqrt: ScaleType.Sqrt,
-   },
-   'linear',
- );
+ const yScaleType = customKnobs.enum.scaleType(
+   'Scale type',
+   ScaleType.Linear,
+   { include: ['Linear', 'Log', 'Sqrt'] },
+ );
```

https://user-images.githubusercontent.com/19007109/221714303-ccb479d6-19da-4bef-81ae-24d8701d5551.mp4


markov00 (approved):
Nice cleanup here, I love these new custom knobs

nickofthyme:
> There are probably just a set of missing/wrongly named knobs that case some vrts to fail

Yeah it's failing on the `CurveType`, probably not handling `number`s. I'll take a look.

nickofthyme:
buildkite update screenshots

nickofthyme:
@markov00, somehow these stories were using a `curve` value of `1` but the VRT picked it up as `0`. That's where all these updated screenshots are from in 0fc91ab. I'm double check them to be sure.

image

PR #1989, opened 2023-03-07
fix(tooltip): custom tooltip header context

### LINKED ISSUE #1988: [Tooltip] `customTooltip` provides incorrect `header` value
**Describe the issue**

When using the `customTooltip` prop on the `Tooltip` spec, the `props` passed to the custom component are incorrect providing the wrong vertical position of the cursor.

https://codesandbox.io/s/vigilant-villani-5w6xxs?file=/src/App.js

**Expected behaviour**

The `customTooltip` is provided the `header` for the actively hovered element.

**Screenshots**

Screen Recording 2023-03-06 at 03 04 03 PM

nickofthyme:
## Summary

The `header` of the `TooltipInfo` used to pull the tooltip value from the first/top value in the list of tooltip values. This behavior was unorthodox and unreliable. We have changed the `TooltipInfo` type to better represent the `header` and `values`.

**BREAKING CHANGES:** The `header` property of `TooltipInfo` type was simplified to `PointerValue` as to include only relevant properties. This change is propagated to all other types using `header` as a `TooltipValue`. The `TooltipInfo.values` used to conditionally pass only highlighted `TooltipValue`s when using a `customTooltip` and now _always_ passes all `values`.

## Issues

fixes #1988 

### Checklist


markov00:
So checking the code I think we should:
- in cartesian charts, the`tooltip.values`  should always return the full set of values under the same X value (this doesn't happen if we apply the fix in this way)
- the `header` usually doesn't represent exactly the top, highlighted element, but just the first on the list. We usually don't care about that because what we use on the header of the tooltip is generally the X value that is available in the `formattedValue` (`const value = isHeader ? x : y;` in `packages/charts/src/chart_types/xy_chart/tooltip/tooltip.ts`)
The `label` value instead is "wrong" actually and we can probably replace it with an empty string.
Also the `datum` is wrong, because we should just extract the current x value. So probably, in the customTooltip situation, we could return just a null header and we could give the user che choice to extract the information they need from the values.
Alternatively we can return a `TooltipValue` with no `datum` and empty `label`, or a label that reflect the x axis title.

packages/charts/api/charts.api.md
nickofthyme, line 1: @markov00 I made the changes based on your comment.

Now...

- All `values` are provided whenever a `customTooltip` is used
- The tooltip `header` values are treated as `PointerValue` not `TooltipValue`, so no `label`, `datum` and all the other properties that are specific only to the `values`.

packages/charts/src/chart_types/xy_chart/tooltip/tooltip.ts
```diff
@@ -65,27 +65,41 @@ export function formatTooltip(
...
+  axisSpec?: AxisSpec,
+): PointerValue {
+  const tickFormatOptions: TickFormatterOptions | undefined = spec.timeZone ? { timeZone: spec.timeZone } : undefined;
+  const tickFormatter = axisSpec?.tickFormat ?? defaultTickFormatter;
+
+  return {
+    valueAccessor: accessor,
+    value: x,
+    formattedValue: tickFormatter(x, tickFormatOptions),
+  };
```
markov00, line 104: Looking at the unit tests, it seems that the `valueAccessor` is not the `xAccessor` but the yAccessor.
Exposing the accessor is also a bit confusing: we could have 2 different series with different xAccessors, so we don't know what accessor we are referring to. Probably is better just to return the raw value and the formatted one, wdyt?

nickofthyme: Good catch, I removed the `valueAccessor` or the formatted `header` objects, see c265029

I assume you wanted to just remove the property value in the `header` and not the `valueAccessor` from `PointeValue` or `TooltipValue` types, correct? If your idea was to remove `valueAccessor` altogether I think that might break things in partition charts, but I don't see anyone using the `valueAccessor` in kibana.

markov00: perfect, thanks, yes I was referring to remove the valueAccessor from the header, not from the PointerValue type

packages/charts/src/components/tooltip/components/tooltip_header.tsx
```diff
@@ -8,30 +8,30 @@
...
 
-type TooltipHeaderProps<
-  D extends BaseDatum = Datum,
-  SI extends SeriesIdentifier = SeriesIdentifier,
-> = PropsOrChildrenWithProps<{
-  header: TooltipValue<D, SI> | null;
-  formatter?: TooltipValueFormatter<D, SI>;
+/** @public */
+export type TooltipHeaderProps<D extends BaseDatum = Datum> = PropsOrChildrenWithProps<{
+  header: PointerValue<D> | null;
```
markov00, line 18: can we just make it optional? instead of `null`?

markov00 (approved):
Changes looks good to me, remember to mark this as breaking change

packages/charts/src/components/tooltip/components/tooltip_header.tsx
nickofthyme: Sure but this would create a type error when using `TooltipInfo.header`, requiring additional code to align the types, something like...

```html
<TooltipHeader header={info.header ?? undefined} formatter={headerFormatter} />
```

I propose we simply allow `undefined` on the `TooltipHeader.header` prop type. See XXX

markov00: Ok, sure go on with that

PR #2003, opened 2023-03-23
docs: remove legacy unused docs

markov00:
## Summary

This PR removes the docs, created years ago and never published.
It also removes the storybook docs addon that was used for these docs.


stratoula:
I love PRs that remove code 😍

markov00:
> I love PRs that remove code 😍

hehe, this actually removes docs but those are legacy docs :D

storybook/main.js
```diff
@@ -10,7 +10,6 @@ module.exports = {
   stories: ['./stories/**/*.(stories|story).ts?(x)'],
   addons: [
     '@storybook/addon-actions',
-    '@storybook/addon-docs',
     '@storybook/addon-knobs',
```
nickofthyme, line 13: This needs to stay, the `docs` addon is required for the `storysource` to work.

PS: @stratoula another reason I loath storybook, took me DAYS to figure this out back in the day!

storybook/parameters.ts
```diff
@@ -24,9 +24,6 @@ export const storybookParameters: Parameters = {
     selectedPanel: process.env.NODE_ENV === 'development' ? SB_KNOBS_PANEL : SB_SOURCE_PANEL,
   },
   previewTabs: {
-    'storybook/docs/panel': {
-      hidden: true,
-    },
```
nickofthyme, line 29: I think we put this back too

nickofthyme (changes requested):
Everything looks good regarding the `/docs` removal. But the `docs` addon needs to stay if we still want the story source tab to work.

storybook/main.js
stratoula: I feel your pain Nick :D

markov00: oh really??? come on storysource, you can handle that yourself.... you don't need everything else

nickofthyme (approved):

Issue #1953, opened 2023-02-01
[metric] render trend with string value


markov00:
Add the ability to render the trend even if the value is a `string`.

The user should be careful of describing what the trend in the background means to avoid confusions.

<img width="306" alt="Screenshot 2023-02-01 at 13 07 32" src="https://user-images.githubusercontent.com/1421091/216038423-4e492113-86e0-4810-858f-49cbc6cbfc10.png">

PR #2011, opened 2023-04-03
feat(metric): trend with string value

markov00:
## Summary
This PR adds the ability to show a text, rather than a value, when rendering a metric with a trend in the background.

image

## Details

This is a different Datum type, with a `value: string` and no `valueFormatter` since the string doesn't need an additional treatment.

## Issues

fix https://github.com/elastic/elastic-charts/issues/1953

### Checklist


markov00:
buildkite update screenshots

nickofthyme (approved):
LGTM, tested on pr deployment, works well with truncating long text. It seem like most usage here would be short strings but may think about multiline text as a future enhancement.

PR #2013, opened 2023-04-04
fix(interactions): brushing over origin coordinates

### LINKED ISSUE #1991: Brushing disappears over origin point
**Describe the issue**

When brushing any chart, the user first clicks on the chart area, then starts to drag, this toggles the `interactions.pointer.dragging` state. But when the user hovers over or near the coordinates of the click that originated the dragging, the `dragging` is set to `false`. This causes a flashing to occur around this point.

This `interactions.pointer.dragging` state is used across many chart types, causing the issue to impact brushing interactions on many chart types.

**To Reproduce**
1. Go to 'https://elastic.github.io/elastic-charts/?path=/story/interactions--brush-tool&globals=theme:light
2. Click and start brushing
3. While continuing to hold the pointer down, return your cursor to the original click
4. Notice the brushing is disabled and the _hover_ interaction is shown instead
5. Moving the cursor away from the origin, the brushing UI reappears without ever lifting the pointer

Screen Recording 2023-03-07 at 12 19 37 PM

**Expected behaviour**
Once brushing is enabled, it is only disabled when the user unclicks (i.e. [`onmouseup`](https://www.w3schools.com/jsref/event_onmouseup.asp)) or presses the <kbd>esc</kbd>.

 - Elastic Charts: `latest`

nickofthyme:
## Summary

Fixes an issue with the brushing tool that would disable brushing when the pointer is returned to the coordinate where the pointer originated the drag.

Screen Recording 2023-04-04 at 03 33 04 PM

## Details

The issue was that enablement of `dragging` was always checked on the `onPointerMove` event without considering the present value of `dragging`, thereby thinking that there is no current dragging directly over the origin coordinate.

Also cleaned up the selector logic related to the `isBrushing` selector to centralize the shared logic in one place.

## Issues

fix #1991

### Checklist


nickofthyme:
buildkite update screenshots

nickofthyme:
buildkite update screenshots

nickofthyme:
buildkite update screenshots

e2e/tests/interactions.test.ts
```diff
@@ -427,4 +427,27 @@ test.describe('Interactions', () => {
       { left: 247, top: 76 }, // mouse over the second point
     );
   });
+
+  test('should not disable dragging when returning to origin', async ({ page }) => {
```
markov00, line 431: I'm considering rephrasing these names to something more specific to what should happen, like: "don't show a tooltip when dragging near the origin" or something along these lines because it specifies what the image should look like and not the configuration.
We can probably describe the test a bit more in a comment if we need

packages/charts/src/state/reducers/interactions.ts
```diff
@@ -68,7 +68,8 @@ export function interactionsReducer(
     case ON_POINTER_MOVE:
       // enable the dragging flag only if the pixel delta between down and move is greater then 4 pixel
       const dragging =
-        !!state.pointer.down && getDelta(state.pointer.down.position, action.position) > DRAG_DETECTION_PIXEL_DELTA;
+        state.pointer.dragging ||
+        (!!state.pointer.down && getDelta(state.pointer.down.position, action.position) > DRAG_DETECTION_PIXEL_DELTA);
```
markov00, line 72: Could you integrate the comment above with: "The dragging  is enabled when the delta between down and move positions is greater than a constant. After this initial threshold, the dragging still enabled until the mouse up event"

e2e/screenshots/interactions.test.ts-snapshots/interactions/should-not-disable-dragging-when-returning-to-origin-chrome-linux.png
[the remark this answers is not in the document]
nickofthyme: Yeah this is correct, the idea is to drag from an origin point and return to that point. It doesn't show the cursor in playwright but it correctly does **not** show the tooltip nor the drag area UI (zero height and width).

e2e/tests/interactions.test.ts
nickofthyme: Sure that's fine 8c1351a

packages/charts/src/state/reducers/interactions.ts
nickofthyme: Done bb1ae84

nickofthyme:
buildkite test this

PR #2054, opened 2023-06-01
build: change typescript target ES2019

nickofthyme:
## Summary

Changes typescript `target` to `ES2019`.

Tested locally against kibana with no noticeable issues.

BREAKING CHANGES: typescript `target` now set to `ES2019`


markov00 (approved):
LGTM

PR #2002, opened 2023-03-20
feat: chart title and desciption

markov00:
## Summary

This PR adds the ability to show a general title and description to a chart.

The title and description are properties from the `<Chart>` component, but we can probably find a different place for the future to expose add properties (settings or a custom meta info component could be good too).

To Do:
- enable/disable the title/description for certain chart types. For single metrics and, probably for the new redesigned gauge/goal charts, the title will be already available as part as the common chart rendering and we don't want to provide a way to render multiple duplicated titles. The chart that needs a title are:
    - [x] all cartesian charts
    - [x] all partition charts
    - [x] all heatmaps
    - [x] small multiple of metrics
    - [x] worldcloud
    - [ ] small multiple of gauge/goal/bullet (blocked by #1235)
- check the text for accessibility standards (use the right semantic tags) 

Screenshot 2023-03-20 at 11 23 25
Screenshot 2023-03-20 at 11 23 02

## Details

The biggest change is effectively moving the target of the chart canvas parent size dimension from the `.echChart` element to the `.echChartContent` element.

Added global toggles UI to storybook for controlling title/description on all stories.

Screen Recording 2023-05-31 at 11 49 19 AM

### Story type changes

Storybook provides a `Story` type that passes `any` args and the story `context` to each story. This type is not adaptive with generators so I created a new `ChartsStory` type that is identical to the `Story` type while adding types to define our customizations. No every story should be defined as such...

```tsx
export const Example: ChartsStory = (_, { title, description }) => {
  ...
}
```

> The `title` and `description` are the `kind` and `name` of the story and defined conditionally based on the value fo the global `toggles`. This logic is handled in `story_wrapper.tsx`.

This also injects the correct types for `Example.parameters` that are custom to our usage.

## Issues

fix #1862 
fix #1928 

### Checklist

- [x] The proper **chart type** label has been added (e.g. `:xy`, `:partition`)
- [x] The proper **feature** labels have been added (e.g. `:interactions`, `:axis`)
- [x] All related issues have been linked (i.e. `closes #123`, `fixes #123`)
- [x] Unit tests have been added or updated to match the most common scenarios
- [x] The proper documentation and/or storybook story has been added or updated
- [ ] The code has been checked for cross-browser compatibility (Chrome, Firefox, Safari, Edge)
- [ ] Visual changes have been tested with all available themes including `dark`, `light`, `eui-dark` & `eui-light`

PR #2055, opened 2023-06-01
docs: remove erroneous zero margins/paddings

nickofthyme:
## Summary

Back when we linked the partition chart to the global `Theme`, see #1402, the partition config had zeroed out `chartMargins` and `chartPaddings`. When linking to the `Theme` in an effort to avoid large VRT we added overrides to each story that was impacted, most all `Partition` stories.

In working on #2002 I noticed the Partition chart titles had different margins because of this zeroing. I no longer see a reason for this as our docs do not match the default look/behavior of the chart.

If different defaults are needed across different chart types we should build that functionality into the theme but for now removal is an easy correction.


nickofthyme:
buildkite update screenshots

markov00 (approved):
Even if I see the reason for removing the padding in the example, I feel this alignment is fine and cleans up a bit the code.

PR #2066, opened 2023-06-07
build: upgrade node to 18.x

markov00:
## Summary

Upgrade min node version to 18.x.
Removed unused ENV in release workflow


.nvmrc
```diff
@@ -1 +1 @@
-16.13.2
+18.16.0
```
nickofthyme, line 1: I knew this code would last!

https://github.com/elastic/elastic-charts/blob/011b56b3f51e2aa1fea8410b4f24e69403635933/.buildkite/utils/pipeline/plugins.ts#L18-L19

nickofthyme (approved):
LGTM if checks pass

.nvmrc
markov00: 😬

Issue #2067, opened 2023-06-12
[Flamegraph] Add properties to externally control the search text


cauemarcondes:
**Is your feature request related to a problem? Please describe.**
On Profiling, we'd like to filter the flamegraph when navigating from one page to another. That means I need to save in the URL the search value the user provides and pass it down to the flamegraph component. In case the user edits or clears out the search value I need to update its value in the URL too.

**Describe the solution you'd like**
The solution I propose is to expose two new properties: `searchText` and `onSearchTextChange`.

The first one `searchText`, is an optional string that will be used to pass down the searched value.
The second one `onSearchTextChange`, is an optional callback function with the new searched value, in case the user edited, or undefined, in case the user cleared the search. 

```js
<Flame
    id={id}
    columnarData={columnarData.viewModel}
    valueAccessor={(d: Datum) => d.value as number}
    valueFormatter={(value) => `${value}`}
    animation={{ duration: 100 }}
    searchText={searchText}
    onSearchTextChange={(newSearchText: string | undefined) => {
      //update url based on the new text
    }}
  />
```

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

**Kibana Cross Issues**
https://github.com/elastic/prodfiler/issues/3126

PR #2068, opened 2023-06-12
feat(flame): expose search field text and search text change listener

markov00:
## Summary

This PR exposes the search text as prop in the `Flame` component.
It also exposes a listener `onSearchTextChange` that is called whenever a change in the search text input provided by the chart happens. It always calls the function with a string, even if the `clean` button is clicked. There is no differences between passing an empty string or clearing the input, both of these actions results in cleaning the search and the API is anyway cleaner and doesn't need a union type.

The `search` prop is an object because in the future it can contains also the `useRegExp` and `caseSensitive` option that are available to the user.

## Issues

fix https://github.com/elastic/elastic-charts/issues/2067

### Checklist


markov00:
buildkite update screenshots

markov00:
buildkite update screenshots

packages/charts/src/chart_types/flame_chart/flame_api.ts
```diff
@@ -77,6 +77,8 @@ export interface FlameSpec<D extends BaseDatum = Datum> extends Spec, LegacyAnim
   valueAccessor: ValueAccessor<D>;
   valueFormatter: ValueFormatter;
   valueGetter: (datumIndex: number) => number;
+  search?: { text: string };
```
nickofthyme, line 80: What's the thought behind this being an object not a string?

nickofthyme (approved):
Code changes LGTM, tested on PR deployment 👍🏼

packages/charts/src/chart_types/flame_chart/flame_api.ts
markov00: I wrote that in the description actually: 
>The search prop is an object because in the future it can contains also the useRegExp and caseSensitive options that are available to the user.

It doesn't have a big effect right now and I know that planning for these extensions in the future is not very useful as a breaking change can be applied easily

PR #2071, opened 2023-06-15
fix: `Chart` component `children` type

nickofthyme:
## Summary

React removed the implicit typing of `children` in `react@17`, causing the `Chart` component to throw a type error. This PR adds the implicit type back.

## Details

I attempted to limit the allowable `children` that can be used in the `Chart` component in e14fe851c3ae39c82e5c22e2ca82f5a71c79f1b7. But react children types are not easy to restrict due to various cases.

When using a `boolean`, `null` or `undefined` value for gated rending.

```tsx
return (
  <Chart>
    { showLegend && <Settings showLegend /> }
  </Chart>
)
```

When mapping over an array of values to components. Children are typed as `(Element | Element[])[]`

```tsx
return (
  <Chart>
    { series.map(({ name, data }) => (
      <LineSeries
        name={name}
        data={data}
        ...
      />
    ) }
  </Chart>
)
```

Furthermore, the types would still allow prop-less components as children, see [SO post](https://stackoverflow.com/q/44475309/6943587).

### Checklist


packages/charts/src/components/chart.tsx
```diff
@@ -44,6 +46,7 @@ interface ChartProps {
   id?: string;
   title?: string;
   description?: string;
+  children?: ReactNode;
```
markov00, line 49: Why not using `PropsWithChildren` instead?

markov00: Sounds good, I was just wondering why

markov00 (approved):
LGTM

PR #2077, opened 2023-06-21
ci: fix execSync input parameter

### LINKED ISSUE #1823: Don't skip eslint step on VRT screenshot changes
If you push a commit with some linting errors and you immediately after pushing you also ask to update the VRTs, the eslint step is skipped (because it checks for changes in TS files) and the eslint error will never been catch and will propagate to master.

This situation could also happen every time I submit a "safe" commit (with changes to a file that is correctly linted) but I previously changed and committed a different file with some linting errors.

The CI eslint only checks file changes from the commit that runs the CI, but from the branch changes.

See for example this PR https://github.com/elastic/elastic-charts/pull/1819

Two way to solve:
- always run eslint across the repo
- run eslint to the changed file from the beginning of the branch

cc @nickofthyme

nickofthyme:
## Summary

Fixes issue with ci linting check where the files to lint were not properly being passed to the `yarn lint:it` command, resulting in a noop execution.

Added group naming to differentiate running linting checks on all vs some files, removed unneeded code and comments.

## Details

I don't understand the usage of the `input` parameter on the [`execSync`](https://nodejs.org/api/child_process.html#child_processexecsynccommand-options) util, it works to pass the buildkite `json` pipeline but apparently not not to pass arguments to the command as I had thought. Fixed by adding `args` option to `exec` util to append `command`.

## Issues

fixes #1823

### DIFF
--- a/.buildkite/scripts/steps/eslint.ts
+++ b/.buildkite/scripts/steps/eslint.ts
@@ -39,14 +39,14 @@ void (async () => {
     const filesToLint = changes.files.filter('**/*.ts?(x)').join(' ');
 
     if (filesToLint.length > 0) {
-      startGroup('Running eslint checks');
+      startGroup(`Running eslint checks - ${filesToLint.length} files`);
 
       await exec('yarn lint:it', {
         input: filesToLint,
       });
     }
   } else {
-    startGroup('Running eslint checks');
+    startGroup('Running eslint checks - all files');
     await exec('yarn lint');
   }
 })();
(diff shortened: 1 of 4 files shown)


nickofthyme:
buildkite update screenshots

nickofthyme:
Changes check out, now the linting rule runs and fails correctly, see b27fa7c9ea564b7d072f4d0e430e74db7c0c4b02 & [buildkite job](https://buildkite.com/elastic/datavis-ci/builds/2203#0188df96-76ef-4f94-9819-17ac39f7824c).

markov00 (approved):
Ok for me, let's check if this works in main!

PR #2080, opened 2023-06-23
fix(legend): use reading direction sorting for stacked bar charts

markov00:
## Summary

This commit revisits the sorting logic applied to the legend when rendering stacked bars.
We currently align visually what is rendered on the bottom of the chart to what is rendered at the bottom of the legend, and vice versa. This actually is not the right reading direction and can cause confusion, in particular when the legend needs to scroll due to too many series or to tiny available space: in this case, the prominent element, which usually stays at the baseline of a stacked bar chart, can be hidden in the legend due to the scroll behavior. 
This actually corrects the reading direction of the legend, which is always top to bottom, whereas on stacked bars is bottom to top.

## Issues

Two Kibana issues are fixed with this change
https://github.com/elastic/kibana/issues/159867
https://github.com/elastic/kibana/issues/154533

### Checklist


markov00:
buildkite update screenshots

MakoWish:
I don't see a version label on this. Will it be backported to 8.8.x?

markov00:
> I don't see a version label on this. Will it be backported to 8.8.x?

This is the PR related to the charting library (different repo than Kibana). It needs to be merged in Kibana and there I will add the required labels.
I don't think this will be backported to 8.8.x we are receiving some reconsideration about this change, so please hold on.

PR #2084, opened 2023-06-30
style(tooltip): update selected row styles

### LINKED ISSUE #2083: Charts tooltip - aesthetic details for selected series (checkmark and padding)
**Describe the issue**
Within the charts tooltip, the checkmark icon that appears once a series is selected has no margins and therefore looks cluttered. Due to the same issue, the series label has no additional padding while it should be slightly shifted to the right

**Current behaviour**
Screenshot 2023-06-29 at 17 09 04

**Expected behaviour**
Screenshot 2023-06-29 at 17 12 34

nickofthyme:
## Summary

Draft PR to demonstrate style changes.

Changes: 
- Add `5px` of padding around checkmark
- Reduced check `font-weight` from `700` to `200`
- Increased `font` size of check by `40%`.

See [sample story](https://ech-e2e-ci--pr-2084-rj20q1cf.web.app/?path=/story/components-tooltip--actions-with-custom-tooltip&globals=toggles.showHeader:true;toggles.showChartTitle:false;toggles.showChartDescription:false;theme:light&knob-attach%20click%20handler=true&knob-pinned=true&knob-show%20footer=true&knob-show%20header=true&knob-visible=true&knob-backgroundColor=#E8F9FD)

### Before

Screen Recording 2023-06-29 at 05 37 33 PM

### After

Screen Recording 2023-06-30 at 02 22 45 PM

## Issues

fixes #2083 

### Checklist


packages/charts/src/components/tooltip/components/_tooltip.scss
```diff
@@ -123,14 +130,19 @@ $tableRowHoverSelectedColor: tintOrShade($euiFocusBackgroundColor, 0, 10%);
 
   &__tableBody {
     .echTooltip__tableCell:not(.echTooltip__colorCell) {
-      padding: 2px 4px;
+      padding: 2px 8px;
```
nickofthyme, line 133: @gvnmagni per...

> the series label has no additional padding while it should be slightly shifted to the right

Do you want all cells in the table row to have extra left/right `padding`? Or just increase the `padding-left` between the color and the right-adjacent column?

This PR increase right/left padding for all cells in the row, excluding the color.

gvnmagni: I think I have explained this very poorly, I'm sorry. The padding that I was referring to was only the amount of pixels of shift to the right that we need once series are selected. I have assumed we could use a padding to solve the problem and I called it that way and this caused some confusion I guess.

If I get this correctly, we could probably go back to the original padding because this new values are causing the header label and the series label to be misaligned


Screenshot 2023-06-30 at 10 25 35

While the rest of things you did seems to be perfectly solve the issue! ❤️

gvnmagni:
A part from the the small misalignment that I pointed out in the previous comment (again, my bad, apologies), it seems that what you did Nick fixed the issue perfectly! This new behavior is beautiful 👌

packages/charts/src/components/tooltip/components/_tooltip.scss
nickofthyme: Ok thanks for clarifying. I reverted the padding change and added left-padding on the first columns cells to match that of the unselected rows. See 63000bb

nickofthyme:
buildkite update screenshots

markov00 (approved):
Tested, looks good to me!
We should probably change the tick from being an unicode to being a SVG to improve consistency across OS
I will open an issue to track this
Screenshot 2023-07-04 at 09 30 35

PR #2093, opened 2023-07-07
refactor: improve condition statements

wonseop:
## Summary
A valuable's property is accessed without null check. 
This patch fixes incorrect conditions and removes unnecessary code.

### DIFF
--- a/packages/charts/src/chart_types/heatmap/state/selectors/on_element_click_caller.ts
+++ b/packages/charts/src/chart_types/heatmap/state/selectors/on_element_click_caller.ts
@@ -38,14 +38,14 @@ export function createOnElementClickCaller(): (state: GlobalChartState) => void 
           if (!spec) {
             return;
           }
-          if (!settings.onElementClick) {
+          if (!settings || !settings.onElementClick) {
             return;
           }
           if (!isPickedCells(pickedShapes)) {
             return;
           }
           const nextPickedShapesLength = pickedShapes.length;
-          if (nextPickedShapesLength > 0 && isClicking(prevClick, lastClick) && settings && settings.onElementClick) {
+          if (nextPickedShapesLength > 0 && isClicking(prevClick, lastClick)) {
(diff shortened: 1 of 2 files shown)


packages/charts/src/utils/common.tsx
```diff
@@ -274,7 +274,7 @@ export function getAllKeys(object?: any, objects: any[] = []): Set<any> {
   return new Set(
     [object, ...objects].filter(Boolean).reduce((keys: any[], obj) => {
       if (obj && typeof obj === 'object') {
-        const newKeys = obj ? (obj instanceof Map ? obj.keys() : Object.keys(obj)) : [];
```
wonseop, line 277: "obj" is always true.

packages/charts/src/chart_types/heatmap/state/selectors/on_element_click_caller.ts
```diff
@@ -38,14 +38,14 @@ export function createOnElementClickCaller(): (state: GlobalChartState) => void
           if (!spec) {
             return;
           }
-          if (!settings.onElementClick) {
+          if (!settings || !settings.onElementClick) {
```
markov00, line 41: `settings` is always available because it comes from `getSettingsSpecSelector` that has already a fallback to a default object.
Do you see a situation where this could happen?
I feel that we should live this and accept the subsequent change in this file that removes a not necessary check

markov00:
Hey @wonseop thanks for your contribution, I just left a small comment but the rest works fine for me.

packages/charts/src/chart_types/heatmap/state/selectors/on_element_click_caller.ts
wonseop: Ok. I fixed it.

markov00 (approved):
Changes looks good to me, thanks for your contribution

nickofthyme (approved):
LGTM

PR #2110, opened 2023-07-18
test(vrt): verify ssim-cie94 comparator

markov00:
## Summary

wrt: https://github.com/microsoft/playwright/issues/13873

### Checklist

- [ ] The `:theme` label has been added and the `@elastic/eui-team` team has been pinged when there are `Theme` API changes

### DIFF
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -12,7 +12,7 @@
     "playwright": "playwright"
   },
   "dependencies": {
-    "@playwright/test": "^1.23.0",
+    "@playwright/test": "1.37.0-alpha-jul-18-2023",
     "change-case": "^4.1.2",
     "expect-playwright": "^0.8.0",
     "playwright-merge-html-reports": "^0.2.3",
--- a/e2e/playwright.config.ts
+++ b/e2e/playwright.config.ts
@@ -33,10 +33,14 @@ const config: PlaywrightTestConfig = {
   ],
   expect: {
     toMatchSnapshot: {
+      _comparator: 'ssim-cie94',
       threshold: 0,
       maxDiffPixelRatio: 0,
       maxDiffPixels: 0,
-    },
+    } as any,
+    toHaveScreenshot: {
+      _comparator: 'ssim-cie94',
+    } as any,
   },
   // TODO limit this to only flaky tests. Watch https://github.com/microsoft/playwright/issues/15657
   retries: 1,
--- a/e2e/tests/tooltip.test.ts
+++ b/e2e/tests/tooltip.test.ts
@@ -10,8 +10,6 @@ import { test } from '@playwright/test';
 
 import { common } from '../page_objects/common';
 
-process.env.ENV_URL = 'http://localhost:9002/';
-
 test.describe('Tooltip', () => {
   test.describe('Chart Types', () => {
     test.describe('Cartesian', () => {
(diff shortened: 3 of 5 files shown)


nickofthyme:
> @nickofthyme if you have time, do you mind check locally how many failures you see when running our tests against an M1 machine?
> 
> It is a bit strange but I can't make it run correctly locally if I just run `yarn test:e2e` but it runs successfully (and also manage to correctly compare most images) if I specify a test suite at time

@markov00 Ugh, it's cuz of this line ⬇️ 

https://github.com/elastic/elastic-charts/blob/ad3c5f6375abf330a50db28e937f6781458ca866/e2e/tests/tooltip.test.ts#L13

Whenever you specify a specific test suite, except `tooltip.test.ts`, it ignores this file. This line was overriding the `ENV_URL` from `http://host.docker.internal:9002` to `https://localhost:9002`. But this is only a problem running `yarn test:e2e` locally because we call `yarn playwright:test` from `e2e/` that's already running inside of the docker container on CI.

e2e/package.json
nickofthyme, line 1: I wonder if we make this shift if we should remove the `platform` flag here...

https://github.com/elastic/elastic-charts/blob/ad3c5f6375abf330a50db28e937f6781458ca866/e2e/scripts/start.sh#L22

markov00: Does that mean that we are already running on `arm64` CPU in our CI? So are we emulating arm64 or is just this command ignored by linux docker?

nickofthyme: No this command is only used locally. The Buildkite playwright jobs are configured in docker here...

https://github.com/elastic/elastic-charts/blob/f1b7d797e13cc5831e4ef2ea584dc7d3a9ac2759/.buildkite/utils/pipeline/plugins.ts#L51-L59

But to answer your question -- No, I don't think it is ignored locally, because it complains when you try to run locally with `amd64` on apple SOC, but it appears docker on apple SOC is running in `linux/arm64` already so no need for the flag.

nickofthyme:
After looking into the differences when using the experimental `ssim-cie94` comparator, we see a large improvement over the default going from `704` failures to `117` (`~83%` decrease ⬇️ ). Many of the failures with a pixel difference ratio of less than `0.01`.

We are currently running the playwright checks and updates in a buildkite CI agent, see specs below. We runs the tests inside a docker container using the `mcr.microsoft.com/playwright:<version>-focal` image.

### Buildkite agent
**User Agent:** buildkite-agent/3.49.0.6166 (linux; amd64)
**OS:** Ubuntu 20.04.6

The test runs below were run locally against the baseline images generate from the buildkite CI described above (i.e. `linux/amd64`.

## With `ssim-cie94` comparator

**playwright version:** `1.37.0-alpha-jul-18-2023`
**docker image**: `mcr.microsoft.com/playwright:v1.37.0-alpha-jul-18-2023-focal`
**arch**: Apple M1 max - `aarch64` - docker

image

See full html report [here](https://upload.elastic.co/d/a6d7fcc7e751f6597c3b2862a566560ffa35f8c0928215b220707d5e1b6376ec) (token: `35ed130e69993cbd`) ⚠️ **expires October 18, 2023**

## Without comparator

**playwright version:** `1.36.1`
**docker image**: `mcr.microsoft.com/playwright:v1.36.1-focal`
**arch**: Apple M1 max - `aarch64` - docker

image

See html report [here](https://upload.elastic.co/d/a6d7fcc7e751f6597c3b2862a566560ffa35f8c0928215b220707d5e1b6376ec) (token: `35ed130e69993cbd`) ⚠️ **expires October 18, 2023**

nickofthyme:
@markov00 I think we can close this for now as it doesn't seem to full solve our `arch` diff issue at the moment.

PR #2115, opened 2023-07-27
fix(legend): improve last value handling

markov00:
## Summary

This PR clean up the concept of what the value, shown after the legend item label, represents.

In cartesian charts, the value represents the last available data point laid on the X-axis.
When multiple series are displayed, the last value represents only the data points with the same X as the greatest value on the X axis.

This value, in the future, can also represent another aspect of the series, like the last non-null value, an average/max/min value, but this PR limits is scope to the last value.

We removed every concept of bucket/time interval from this calculation to clarify the legend item value concept.
This allows us to decouple our code from Elasticsearch or other data sources. 
A small, but fixable side-effect is:
- the last value depends entirely on the passed data, so if you pass just 10 values of a series that should cover 100 data points, we consider the last value the last of these 10 values: if the 10th value is not at the extreme edge of your data domain we still consider that 10th value as the last one. If you instead, are looking to depict the extreme edge of your data domain (the 100th element) that in the mentioned example is empty/null then you should pass all the 100 data points with null values where needed.

Related issues
https://github.com/elastic/kibana/issues/153079
https://github.com/elastic/kibana/issues/94280

### BREAKING CHANGE

In cartesian charts, the default legend value now represents the data points that coincide with the latest datum in the X domain. Please consider passing every data point, even the empty ones (like empty buckets/bins/etc) if your x data domain doesn't cover fully a custom x domain passed to the chart configuration.

### Checklist


markov00:
buildkite test this

markov00:
buildkite update screenshots

markov00:
@nickofthyme could you please review this?

storybook/stories/test_cases/21_domain_edges.story.tsx
```diff
@@ -0,0 +1,219 @@
...
+          timeAxisLayerCount={2}
+          style={{
+            tickLabel: {
+              alignment: { horizontal: 'left' },
+              padding: 0,
+            },
+            tickLine: {
+              padding: 0,
+              size: 5,
+            },
```
nickofthyme, line 97: With the new eui theme adoption, the `tickLine.visible` property is `false`. So with multilayer time axes I add it explicitly.

```suggestion
            tickLine: {
              visible: true,
              padding: 0,
              size: 5,
            },
```

storybook/stories/test_cases/21_domain_edges.story.tsx
```diff
@@ -0,0 +1,219 @@
...
```
nickofthyme, line 146: Same thing here

```suggestion
            tickLine: {
              visible: true,
              padding: 0,
              size: 5,
            },
```

storybook/stories/test_cases/21_domain_edges.story.tsx
```diff
@@ -0,0 +1,219 @@
...
+          timeAxisLayerCount={2}
+          style={{
+            tickLabel: {
+              alignment: { horizontal: 'center' },
+              padding: 0,
+            },
+            tickLine: {
+              padding: 0,
+              size: 5,
+            },
```
nickofthyme, line 193: Same thing here

```suggestion
            tickLine: {
              visible: true,
              padding: 0,
              size: 5,
            },
```

packages/charts/src/chart_types/xy_chart/state/utils/get_last_value.ts
```diff
@@ -6,58 +6,52 @@
...
-    if (series.stackMode === StackMode.Percentage) {
-      const y1InPercentage = y1 === null || y0 === null ? null : y1 - y0;
-      lastValues.set(seriesKey, { y0, y1: y1InPercentage });
-      return;
-    }
-    if (initialY0 !== null || initialY1 !== null) {
-      lastValues.set(seriesKey, { y0: initialY0, y1: initialY1 });
+  switch (type) {
+    case LegendValue.LastNonNullValue: {
+      const last = series.data.findLast((d) => d.x === xDomain.dataDomain[1] && valueAccessor(d) !== null);
```
nickofthyme, line 44: Nice use of `findLast`!

packages/charts/src/chart_types/xy_chart/state/utils/get_last_value.ts
```diff
@@ -6,58 +6,52 @@
...
 import { ScaleType } from '../../../../scales/constants';
 import { XDomain } from '../../domains/types';
 import { isDatumFilled } from '../../rendering/utils';
-import { DataSeries, getSeriesKey, XYChartSeriesIdentifier } from '../../utils/series';
-import { StackMode } from '../../utils/specs';
+import { DataSeries, DataSeriesDatum } from '../../utils/series';
+
+/** @internal */
+export const LegendValue = Object.freeze({
+  None: 'none' as const,
```
nickofthyme, line 18: What's the use of this `None` option, compared to setting `showLegendExtra` to `false`?

nickofthyme (approved):
Changes LGTM, just one suggestion on the story styles and a question.

packages/charts/src/chart_types/xy_chart/state/utils/get_last_value.ts
markov00: Right now this is only internal, but we will add more legend calculation in the future

storybook/stories/test_cases/21_domain_edges.story.tsx
markov00: fixed in https://github.com/elastic/elastic-charts/pull/2115/commits/6af36e5d71d0adc0fbbb643837888f58df5b35b2

storybook/stories/test_cases/21_domain_edges.story.tsx
markov00: fixed in https://github.com/elastic/elastic-charts/pull/2115/commits/6af36e5d71d0adc0fbbb643837888f58df5b35b2

storybook/stories/test_cases/21_domain_edges.story.tsx
markov00: fixed in https://github.com/elastic/elastic-charts/pull/2115/commits/6af36e5d71d0adc0fbbb643837888f58df5b35b2

PR #2141, opened 2023-08-13
chore: convert `substr` to `substring`

wonseop:
## Summary
'substr' is just as fine as it's string specific and well named. but it was deprecated. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

Therefore, I changed the code to use String.prototype.substring() instead of String.prototype.substr().

### Checklist

- [X] The code has been checked for cross-browser compatibility (Chrome, Firefox, Safari, Edge)


nickofthyme:
buildkite test this

PR #2167, opened 2023-09-12
ci: fix linting ci check for deleted files

nickofthyme:
## Summary

Fixes issues with the `eslint` ci step.

## Details

This step, like many, tries to only test what is necessary. It does so by checking which files have changed in the entire PR. Then we use these file paths to narrow the linting task.

The problem is with deleted files. When a files is deleted, we don't filter them out, thus we are telling eslint to lint a non-existent file.

To fix this we now request the `changeType` from github file diffs api and pass that along to the `FileContext` where we can filter `byType` and return a new `FileContext` without affecting the initial context.

Fix verified for `DELETED` and `RENAMED` files in [here](https://buildkite.com/elastic/datavis-ci/builds/2490#018a8a3b-ce64-4158-bba1-2a9bb46d8d90)


nickofthyme:
@markov00 See 28d2fa4 for confirmation of fix. Command runs

```
> Running eslint checks - 3 files | 19s
yarn run v1.22.19
$ NODE_ENV=production eslint --quiet --ext .tsx,.ts,.js .buildkite/scripts/steps/eslint.ts .buildkite/utils/github.ts packages/charts/src/specs/specs_parserrrrr.test.tsx
Done in 15.65s.
```

.buildkite/scripts/steps/eslint.ts
```diff
@@ -36,13 +36,13 @@ void (async () => {
   await exec('yarn build:ts');
 
   if (bkEnv.isPullRequest && !hasLintConfigChanges) {
-    const filesToLint = changes.files.filter('**/*.ts?(x)').join(' ');
+    const filesToLint = changes.files.byType('DELETED', true).filter('**/*.ts?(x)');
```
markov00, line 39: There was also an error in the previous code where we where showing `Running eslint checks - ${filesToLint.length} files` but `filesToLint` was already a concatenated string?

.buildkite/utils/github.ts
```diff
@@ -272,6 +303,11 @@ export const updateCheckStatus = async (
   }
 };
 
+interface FileDiff {
+  path: string;
+  changeType: 'ADDED' | 'DELETED' | 'RENAMED' | 'COPIED' | 'MODIFIED' | 'CHANGED';
```
markov00, line 308: I think these can be also picked from `@oktokit/graphql-schema`

markov00 (approved):
Changes looks good to me! thanks for the quick fix

.buildkite/utils/github.ts
nickofthyme: Oh good call, I can add that later.

PR #2169, opened 2023-09-13
fix(tooltip): rendering in react v18

### LINKED ISSUE #1910: Tooltip doesn't render when React 18 `createRoot` is used
**Describe the issue**

The tooltip does not show up/render when using the newly introduced React 18 function `createRoot` to initialise the root and  render the app.
[Codesandbox example for the above](https://codesandbox.io/s/elated-night-zy3mbh?file=/src/index.tsx:0-261)

If one reverts to the older `ReactDOM.render`, [then the tooltip shows up.](https://codesandbox.io/s/eloquent-fast-s4deou?file=/src/index.tsx:97-113)

**To Reproduce**
1. Create a React v18 app and initialise it like this:
  ```tsx
  import { StrictMode } from "react";
  import { createRoot } from "react-dom/client";
  import App from "./App";
  const rootElement = document.getElementById("root");
  const root = createRoot(rootElement!);
  root.render(
    <StrictMode>
      <App />
    </StrictMode>
  );
  ```
2. Hover the geometry
4. Notice the missing tooltip

**Expected behaviour**
The tooltip should render regardless of how the app is initialised.

 - OS: macOS 11
 - Browser: Firefox
 - Elastic Charts: latest, `v51.3.0`

**Errors in browser console**
The console shows neither errors nor warnings. The only noticeable thing is that the `#echTooltipPortalMainTooltip` node quickly has content. But one instant later it is gone again, nothing is ever painted.

**Checklist**

- [ ] Every related Kibana issue is listed under `Kibana Cross Issues` list and the `kibana cross issue` tag is applied

nickofthyme:
## Summary

Fixes issue where tooltip would not render when using `react@^18`.

## Details

Troubleshooting steps:

- So digging into this at first I noticed that popper was **not** applying the `styles` to the `portalNode` element.
- I then realized that the `styles` were mostly zeros which must be why they didn't apply the styles to the element.
- Next I noticed the popper's computed size of the `portalNode` element was all zeros including the `height` and `width`.
- Considering the zero height I looked into the popper code and compared between the current `main` playground and the playground using `react@^v18`. I noticed that when the size computation was performed, the current code rendered the tooltip size accurately but the `react@^v18` did not show a tooltip at all.
- Capturing the `portalNode` element ref from the code and searching for it in the dom, nothing! It was missing.
- Thus the reference to the `portalNode` was broken.

Why??? I have no clue...

I searched the react changelog for changes to `createPortal` and saw nothing. But for some reason the old code with `react@^16` was able to maintain an active reference to the `portal` dom node and `react@^18` was not.

The solution here is to ensure the portal node reference is maintained across re-renders. Before we did so by creating a new portalNode each time but here I simply append the saved reference node to the dom each render.

## Issues

fixes #1910

### Checklist


nickofthyme:
@markov00 can you take a look at this? Happy to take ideas. You can test it by pulling down this branch and running `yarn && yarn playground`. Make sure you are on `node16` or it will complain. Happy to chat tomorrow.

nickofthyme:
buildkite update screenshots

nickofthyme:
Slight pixel shift in placement from updated `@popperjs/core` version

e2e/screenshots/all.test.ts-snapshots/baselines/annotations-lines/advanced-markers-chrome-linux.png
markov00, line 1: I don't think we are using popperjs to place the text markers here in annotations. Is there a different reason? Do we use a portal for that?

markov00 (approved):
It is fine for me, but we probably should consider a way to test the library with react 18

e2e/screenshots/all.test.ts-snapshots/baselines/annotations-lines/advanced-markers-chrome-linux.png
nickofthyme: Yeah we are using popper for the line annotation `markerBody` since #1116

https://github.com/elastic/elastic-charts/blob/0bb3ab1f3afc0a04c7e1af2418c6fa0a3bdd1af4/packages/charts/src/chart_types/xy_chart/renderer/dom/annotations/line_marker.tsx#L165-L169

PR #2174, opened 2023-09-20
fix: add treeshaking

markov00:
## Summary

Add `sideEffects` in package.json to provide bundle treeshaking


markov00:
buildkite test this

PR #2184, opened 2023-10-02
feat: allow alpha colors and improve contrast logic 

### LINKED ISSUE #2183: [Metric] Opacity is ignored when passed as RGBA or HEX string
**Describe the issue**

Screenshot 2023-09-29 at 14 55 01

**Kibana Cross Issues**
https://github.com/elastic/kibana/issues/167636

nickofthyme:
## Summary

Adds support for colors with alpha channel on `Metric` charts including sparkline and progress bar colors.

https://github.com/elastic/elastic-charts/assets/19007109/d6ade9ad-0ccc-4598-8eca-b8231aae63c1

> It's hard to see the changes from a simple demo so please play around with the story [here](https://ech-e2e-ci--pr-2184-mijxezir.web.app/?path=/story/metric-alpha--basic&globals=toggles.showHeader:true;toggles.showChartTitle:false;toggles.showChartDescription:false;theme:light&knob-EUI%20icon%20glyph%20name=warning&knob-EUI%20value%20icon%20glyph%20name=sortUp&knob-color=rgba(71,%2031,%2031,%200.46)&knob-extra=last%20%3Cb%3E5m%3C/b%3E&knob-is%20numeric%20metric=true&knob-progress%20bar%20direction=vertical&knob-progress%20max=100&knob-progress%20or%20trend=trend&knob-subtitle=Cluster%20CPU%20usage&knob-title=21d7f8b7-92ea-41a0-8c03-0db0ec7e11b9&knob-trend%20a11y%20description=The%20trend%20shows%20a%20peak%20of%20CPU%20usage%20in%20the%20last%205%20minutes&knob-trend%20a11y%20title=The%20Cluster%20CPU%20Usage%20trend&knob-trend%20data%20points=30&knob-trend%20shape=area&knob-value=55.23&knob-value%20color=#3c3c3c&knob-value%20prefix=&knob-value%20postfix=%20%&knob-use%20value%20color=&knob-show%20icon=&knob-show%20value%20icon=).

## Details

These changes clean up a bit of the background color and contrast color for the `Metric` chart.

Notable changes:
- Removes the `backgroundColor` option on the `Theme.metric` (aka `MetricStyle`) in favor of the top-level `Theme.background`
- Fixes color logic converting `hsl` to `rgba` while persisting the `alpha` value, see 48f20769fc9f990f958f6742b517fe33e0f92cef.
- Enables setting transparent colors for progress bar, overall metric color and implicitly the sparkline color.

## Issues

Fixes #2183

### Checklist


nickofthyme:
buildkite update screenshots

packages/charts/src/utils/themes/dark_theme.ts
```diff
@@ -410,7 +410,6 @@ export const DARK_THEME: Theme = {
     },
     border: '#343741',
     barBackground: '#343741',
-    background: '#1D1E23',
     nonFiniteText: 'N/A',
```
nickofthyme, line 413: @markov00 was there a reason for adding this at this level?

e2e/screenshots/all.test.ts-snapshots/baselines/metric-alpha/grid-chrome-linux.png
nickofthyme, line 1: @gvnmagni & @markov00 - This change made the text contrast much more pronounced than it was before. Are we ok with this or should I revert the color to the old text color?

gvnmagni: I am looking at the storybook and this might be a stupid question but, why did the text color became full back or full white? instead of classic text.color

nickofthyme: Yeah you are right, Ii didn't see those colors defined in the `Theme.metric.text`. I updated the contrast logic to respect those values. See 29f198d

nickofthyme:
buildkite update screenshots

nickofthyme:
buildkite update screenshots

e2e/screenshots/all.test.ts-snapshots/baselines/metric-alpha/grid-chrome-linux.png
gvnmagni: thank you Nick!

packages/charts/src/common/color_library_wrappers.ts
```diff
@@ -83,18 +83,20 @@ export function colorToRgba(color: Color): RgbaTuple {
 }
 
 /** @internal */
-export function colorToHsl(color: Color) {
-  const [r, g, b] = colorToRgba(color);
-  return chroma.rgb(r, g, b).hsl();
+export function colorToHsl(color: Color): [h: number, s: number, l: number, a: number] {
+  const [r, g, b, a] = colorToRgba(color);
+  const [h, s, l] = chroma.rgb(r, g, b).hsl(); // alpha not preserved
+  return [h, s, l, a];
```
markov00, line 89:
```suggestion
  return chroma.rgb(r, g, b, a).hsl();
```

packages/charts/src/common/color_library_wrappers.ts
```diff
@@ -83,18 +83,20 @@ export function colorToRgba(color: Color): RgbaTuple {
...
 }
+
 /** @internal */
-export function hslToColor(h: number, s: number, l: number): Color {
-  const rgba = chroma.hsl(h, s, l).rgba();
+export function hslToColor(h: number, s: number, l: number, a = 1): Color {
+  const rgba = chroma.hsl(h, s, l).alpha(a).rgba();
```
markov00, line 94:
```suggestion
  const rgba = chroma.hsl(h, s, l, a).rgba();
```

packages/charts/src/chart_types/metric/renderer/dom/metric.tsx
```diff
@@ -75,34 +77,34 @@ export const Metric: React.FunctionComponent<{
...
-
+  const highContrastTextColor = fillTextColor(
+    backgroundColor,
+    isMetricWProgress(datum) ? backgroundColor : datum.color,
+    undefined,
+    {
+      lightColor: colorToRgba(style.text.lightColor),
+      darkColor: colorToRgba(style.text.darkColor),
+    },
+  );
```
markov00, line 105: I believe we should improve this a bit because the text could be above the background but also above the area background that has a different luminosity and that could change the contrast.
I think we can do the following:
- find the color contrast ratios of each text color for both background colors (metric and area)
- pick the text color that is more close to pass the WCAG threshold in both situations

packages/charts/src/chart_types/metric/renderer/dom/index.tsx
```diff
@@ -94,8 +98,18 @@ class Component extends React.Component<StateProps & DispatchProps> {
...
     const panel = { width: width / maxColumns, height: height / totalRows };
 
+    if (colorToRgba(background.color)[3] < 1) {
+      // TODO: centralize this check and bg color fallback across all chart types
+      Logger.expected(
+        'Text contrast requires an opaque background color, using fallbackColor blend',
+        'an opaque color',
+        background.color,
+      );
+    }
```
markov00, line 108: I know we are already logging this in other charts, but we can probably get rid of this, as we have a fallback and I haven't see a reaction from developers to fix their choices we can remove this warning.

packages/charts/src/common/color_library_wrappers.ts
nickofthyme: Yeah I tried this but it doesn't work, `hsl` only return 3 values.

```ts
chroma.Color.hsl: () => [number, number, number]
```

Also `rgb` does not allow alpha values so you'd need to do `.alpha(a)`.

packages/charts/src/chart_types/metric/renderer/dom/metric.tsx
nickofthyme: Are you talking about the checking between the background and the sparkline area?

packages/charts/src/chart_types/metric/renderer/dom/sparkline.tsx
```diff
@@ -36,8 +36,8 @@ export const SparkLine: FunctionComponent<{
     trendShape === MetricTrendShape.Bars ? CurveType.CURVE_STEP_AFTER : CurveType.LINEAR,
   );
 
-  const [h, s, l] = colorToHsl(color);
-  const pathColor = hslToColor(h, s, l >= 0.8 ? l - 0.1 : l + 0.1);
+  const [h, s, l, a] = colorToHsl(color);
```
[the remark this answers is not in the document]
nickofthyme: I'm not quite sure what you're saying here, I need to blend the sparkline color with an opaque background?

packages/charts/src/chart_types/metric/renderer/dom/index.tsx
nickofthyme: That's fine. I do think we should centralize this logic to get the blended bg color though, because in several places we use `highContrastColor` instead of the more thorough `fillTextColor`. See e7c9b93bd9444a433ffa8125dffa68a08bb7554b

packages/charts/src/common/color_library_wrappers.ts
nickofthyme: The `chroma.hsl` method does not support passing alpha value.

```ts
(method) chroma.ChromaStatic.hsl(h: number, s: number, l: number): chroma.Color
```

packages/charts/src/common/color_library_wrappers.ts
nickofthyme: Yeah I tried this but it doesn't work, `hsl` only return 3 values.

```ts
chroma.Color.hsl: () => [number, number, number]
```

Also `rgb` does not allow alpha values so you'd need to do `.alpha(a)`.

packages/charts/src/chart_types/metric/renderer/dom/sparkline.tsx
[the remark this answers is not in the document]
nickofthyme: I think you are right here since the transparent sparkline would be rendered **on top** of the transparent back area. I think a good fix of this is to just mask the filled sparkline from the background. See 6bd75a4590c1f818482d08b867a41530bdbaee5d

Screen Recording 2023-10-06 at 02 14 49 PM

packages/charts/src/chart_types/metric/renderer/dom/metric.tsx
nickofthyme: Ok I took a stab at this and indeed there is a small range where the text color contrast selection is opposing.

So I'm checking when there is a trend line for the contrast over the sparkline, then if different I take the color with the _higher_ ratio. This is the part I'm not sure about, if it should be the higher one but also if we are even comaring the right ration since one is based on the `light` color and the other on the `dark`.

The main logic is [here](https://github.com/elastic/elastic-charts/pull/2184/commits/e254ac88d360363009a2c8fefadf2a63ddb97563#diff-e03d991817e697e8dbd379ce20ed45a05c0cbd4be41833633d6d67ee7d97130c)...

https://github.com/elastic/elastic-charts/blob/e254ac88d360363009a2c8fefadf2a63ddb97563/packages/charts/src/chart_types/metric/renderer/dom/metric.tsx#L100-L120

nickofthyme:
buildkite update screenshots

Issue #2148, opened 2023-08-24
[partition] issue with sparse date histogram result in treemap


dej611:
**Describe the issue**

In some specific scenarios the Treemap throws a Maximum stack error.
Note that Treemap is the only partition visualization type affected by this issue.

**To Reproduce**
Steps to reproduce the behavior for Lens in the linked issue.

For local reproduction, this is the dataset used:
```
"timestamp per 30 seconds","Count of records"
1692887640000,
1692887670000,
1692887700000,1
1692887730000,
1692887760000,
1692887790000,
1692887820000,
1692887850000,
1692887880000,
1692887910000,
1692887940000,
1692887970000,
1692888000000,
1692888030000,
1692888060000,
1692888090000,
1692888120000,
1692888150000,3
1692888180000,
1692888210000,
1692888240000,
1692888270000,
1692888300000,
1692888330000,
1692888360000,
1692888390000,
1692888420000,1
1692888450000,1
1692888480000,
1692888510000,
1692888540000,
```

**Expected behaviour**

**Screenshots**

![treemap_forever](https://github.com/elastic/elastic-charts/assets/924948/4db9cf40-a549-438a-826e-da8b522fcb80)

**Errors in browser console**
```
Uncaught RangeError: Maximum call stack size exceeded
    at bestVector (treemap.js:39:1)
    at treemap (treemap.js:76:1)
    at treemap (treemap.js:98:1)
    at treemap (treemap.js:98:1)
    at treemap (treemap.js:98:1)
    at treemap (treemap.js:98:1)
    at treemap (treemap.js:98:1)
    at treemap (treemap.js:98:1)
    at treemap (treemap.js:98:1)
    at treemap (treemap.js:98:1)
```

**Kibana Cross Issues**
https://github.com/elastic/kibana/issues/164730

**Checklist**

- [x] The proper **chart type** label has been added (e.g. `:xy`, `:partition`)
- [x] The proper **feature** labels have been added (e.g. `:interactions`, `:axis`)
- [x] Every related Kibana issue is listed under `Kibana Cross Issues` list and the `kibana cross issue` tag is applied

markov00:
closed in https://github.com/elastic/elastic-charts/pull/2214

nickofthyme:
@markov00 #2214 was never merged 🤔

markov00:
@nickofthyme correct, sorry about that, but I remember we fixed that somehow. Let me find how and in case I will close it again

[...]

PR #2214, opened 2023-10-23
fix(partition): reduce recursive call stack

dej611:
## Summary

The treemap no longer hangs with specific datasets and completes the rendering without the `Maximum call stack` error.

## Details

The issue was generated by the recursive nested functions of the `areaAccessor` during the treemap computation.
This PR passes always the same function reference now and decouple the rescaling factor as an additional parameter to avoid nested recursion.

Unfortunately I haven't managed to reproduce here the stack error as I suspect it happens only when an existing long stack is already present (as in Kibana). I'll try to test-case this with à-la carte instance once created the PR.

## Issues

#2148

### Checklist


packages/charts/src/chart_types/partition_chart/layout/utils/treemap.ts
```diff
@@ -124,14 +126,15 @@ export function treemap(
     height: outerHeight,
   }: { x0: number; y0: number; width: number; height: number },
   layouts: LayerLayout[],
+  rescaleAreaFactor: number = 1,
 ): Array<Part> {
   if (nodes.length === 0) return [];
```
dej611, line 131:
```suggestion
  if (nodes.length === 0 || Number.isNaN(outerX0) || Number.isNaN(outerY0)) return [];
```

PR #2218, opened 2023-10-25
fix(metric): fix rendering artifact

nickofthyme:
## Summary

Fixes minor rendering artifact in Chrome introduced in #2184

image

## Details

With the changes from #2184 we introduced support for transparent colors. This change would have overlapping colors on the trend line metric causing blended colors. To prevent this we split the top and bottom half of the sparkline in two so the shapes rendered independently directly on the metric background.

Doing this caused a slight white sliver to be rendered between the two halves for colors with hight constant to the background, even with `strokeWidth` set to `0`.

To fix this we render a thin line along the spark line to fill this render defect without making the line too thick where a transparent color overlap would create a noticeable line.

Screen Recording 2023-10-25 at 12 47 53 PM

### Checklist


nickofthyme:
buildkite update screenshots

nickofthyme:
Closing this in favor of #2255

PR #2225, opened 2023-11-01
fix(BarSeries): ignore histogram mode in determining stacked series

nickofthyme:
## Summary

This PR removes the forced stacking of bar charts with `enableHistogramMode` set to `true`.

### Before

Screen Recording 2023-11-01 at 09 51 39 AM

> Notice the domain fit appears to be disabled when `enableHistogramMode` is used, due to the forced stacking.

### After

Screen Recording 2023-11-01 at 04 20 31 PM

> Notice the domain is able to fit the data with `enableHistogramMode` enabled even for many bar series. Now only stacking the series disables the fit behavior as required by the current _filling_ logic used to stacking bars.

## Details

This was added in https://github.com/elastic/elastic-charts/pull/751 to prevent using clustered bars in a histogram bar chart.

We also implemented this logic in #2226 to avoid the forced stacking of histogram bars when only one series was present. However, this implementation would still impact previous fitting behavior allow in kibana related to https://github.com/elastic/kibana/issues/170316, specifically fitting multiple unstacked histogram bars to data domain.

## Issues

fix #2223

### Checklist

- [x] The proper **chart type** label has been added (e.g. `:xy`, `:partition`)
- [x] The proper **feature** labels have been added (e.g. `:interactions`, `:axis`)
- [X] All related issues have been linked (i.e. `closes #123`, `fixes #123`)
- [x] Unit tests have been added or updated to match the most common scenarios
- [x] The proper documentation and/or storybook story has been added or updated

PR #2226, opened 2023-11-01
fix(BarSeries): forced stacking for single series histogram

### LINKED ISSUE #2223: Do not force stacking on single histogram series
**Describe the issue**

Whenever a user creates a bar series and enables the `enableHistogramMode` flag, the series is treated as stacked.

This unexpectedly disabled the domain fitting due to the filling of dataseries values for stacked series.

**To Reproduce**
1. Go to https://codesandbox.io/s/empty-pond-lfn4t2
2. Toggle `enableHistogramMode` option
3. Notice how the domain `fit` is applied in each case

**Expected behaviour**
Domain fit should not be impacted by whether the `enableHistogramMode` flag is enabled or not.

 - Elastic Charts: `v21.0.0`

**Additional context**

This issues was introduced in https://github.com/elastic/elastic-charts/pull/751 by treating any bar series with `enableHistogramMode` set to `true` as _stacked_. This was done to prevent clustered bar charts in a histogram.

The solution here would be to only force the stacking behavior for histogram bar charts when there is more than one bar series.

**Kibana Cross Issues**
- https://github.com/elastic/kibana/issues/170316

**Checklist**

nickofthyme:
## Summary

This PR was the first solution designed to fix #2223 however after more thought, this would solve one breaking behavior but not all. Namely, this would not solve the case where the user has multiple histogram bar series in a chart and wants to fit the domain to the data. This would force the bars to be stacked, against the chart config, and make domain fitting impossible.

### Before

Screen Recording 2023-11-01 at 09 51 39 AM

### After

Screen Recording 2023-11-01 at 02 39 54 PM

> Notice, when there are 2 bar series in the chart with `enableHistogramMode` set to `true` they are forced to stack and the domain fitting is set to start at `0`.

### DIFF
--- a/packages/charts/src/chart_types/xy_chart/domains/y_domain.ts
+++ b/packages/charts/src/chart_types/xy_chart/domains/y_domain.ts
@@ -141,7 +141,8 @@ export function groupSeriesByYGroup(specs: YBasicSeriesSpec[]) {
     { stackMode: StackMode | undefined; stacked: YBasicSeriesSpec[]; nonStacked: YBasicSeriesSpec[] }
   >();
 
-  const histogramEnabled = isHistogramEnabled(specs);
+  const isStackedSpec = isStackedSpecFn(specs);
+
   // split each specs by groupId and by stacked or not
   specs.forEach((spec) => {
     const group = specsByGroupIds.get(spec.groupId) || {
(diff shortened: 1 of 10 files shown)


nickofthyme:
Closed in favor of #2225

PR #2228, opened 2023-11-02
fix: `onRenderChange` callback trigger on resize

### LINKED ISSUE #2221: `onRenderChange` not called on chart resize
**Describe the issue**
When a chart is resized with no configuration changes, `onRenderChange` fails to fire leaving no way for a consumer to detect the rerender.

**To Reproduce**
1. Create a chart and instrument `onRenderChange` with a console statement
2. Resize the chart canvas (resizing the browser works)

See https://codesandbox.io/s/focused-rubin-n6s97f?file=/src/App.tsx

The chart resizes, but the `onRenderChange` isn't called.

https://github.com/elastic/elastic-charts/assets/19007109/56a2f383-5cb6-410c-8fe4-135e18eaedbd

nickofthyme:
## Summary

The `onRenderChange` was not being fired on resize of the chart even though the resize caused the chart to re-render. Adds an `onWillRender` and `onResize` events to the `Settings` spec.

This also reworks the `resizeDebounce` logic to update the debounce time when changed and remove the debouncer completely when `resizeDebounce` is set to `0`.

## Details

The `ChartResizer` triggers the update of the parent dimensions in the redux state but for some reason this does not affect the `state.chartId`, `state.chartRendered`, `state.chartRenderedCount`, nor `getDebugStateSelector(state)`, the `ChartStatus` component never updates, and thus never triggers the `onRenderChange` callback event.

## Issues

fixes #2221

### Checklist


nickofthyme:
buildkite update screenshots

packages/charts/src/components/chart_resizer.tsx
```diff
@@ -11,24 +11,25 @@ import { connect } from 'react-redux';
...
 }
 
 interface ResizerDispatchProps {
-  updateParentDimensions(dimension: Dimensions): void;
+  updateParentDimensions: typeof updateParentDimensions;
 }
 
 type ResizerProps = ResizerStateProps & ResizerDispatchProps;
-
-const DEFAULT_RESIZE_DEBOUNCE = 200;
```
nickofthyme, line 31: This was never used cuz setttings has it's own default of `10`.

packages/charts/src/components/chart_resizer.tsx
```diff
@@ -52,18 +50,27 @@ class Resizer extends React.Component<ResizerProps> {
...
-    this.onResizeDebounced = debounce(this.onResize, this.props.resizeDebounce);
+    this.setupResizeDebounce();
     if (this.containerRef.current) {
       this.ro.observe(this.containerRef.current as Element);
     }
   }
 
+  componentDidUpdate({ resizeDebounce }: Readonly<ResizerProps>): void {
+    if (resizeDebounce !== this.props.resizeDebounce) this.setupResizeDebounce();
+  }
```
nickofthyme, line 61: If `resizeDebounce` value was changed after the initial chart render, the value and thus `this.onResizeDebounce` was never updated. This adds an update to `this.onResizeDebounce` on `componentDidUpdate`.

Even setting the value directly on the settings with a storybook knob was too async to update the value and thus never using the initial knob value.

packages/charts/src/components/chart_resizer.tsx
nickofthyme, line 1: We could consider removing this deboucer altogether as it was an original implementation that may not be needed or at the very least default to `0` (aka disabled).

PR #2234, opened 2023-11-08
test: add resize config for easy story and e2e resizing

nickofthyme:
## Summary

We've started adding `resize` wrapper elements around certain stories to allow for a better size per chart while also allowing the docs consumer to easily resize the chart to see how it changes.

This caused a bit of an issue in #2156 with clipping of the vrt screenshots.

This PR adds a `resize` config to the `Story.parameters`. This allows us to pass a `boolean` where if true we wrap the story in resizable `div`, matching the default size of the `.echChart`. You may also pass this parameter any css styles to be applied to this resize wrapper. For example the `Metric - Basic` story just adds the following to override the `height` and `width` of the resize wrapper to `200px`.

```tsx
Example.parameters = {
  resize: {
    height: '200px',
    width: '200px',
  },
};
```

Any provided styles override the default styles here...

https://github.com/elastic/elastic-charts/blob/3b83f53e31534d067df543d0c25889e87ee40f5e/storybook/style.scss#L144-L151

As for the the e2e tests, I added the `setResizeDimensions` method to the `common` page object for easily updating the height and width of this resize wrapper. This would look like...

```ts
test('should resize metric', async ({ page }) => {
    await common.expectChartAtUrlToMatchScreenshot(page)(
      'http://localhost:9001/xxxxx,
      {
        action: async () => {
          await common.setResizeDimensions(page)({ width: '400px' });
        },
      },
    );
  });
```

If no `resize` parameter was provided this method will throw and error!

> ⚠️ This feature only impacts the storybook docs and e2e testing.

Changes added to facilitate testing of Bullet as Metric in #2156.

### DIFF
--- a/e2e/tests/tooltip.test.ts
+++ b/e2e/tests/tooltip.test.ts
@@ -10,8 +10,6 @@ import { test } from '@playwright/test';
 
 import { common } from '../page_objects/common';
 
-process.env.ENV_URL = 'http://localhost:9002/';
-
 test.describe('Tooltip', () => {
   test.describe('Chart Types', () => {
     test.describe('Cartesian', () => {
--- a/e2e/page_objects/common.ts
+++ b/e2e/page_objects/common.ts
@@ -556,7 +556,6 @@ export class CommonPage {
     }
 
     await el.evaluate((element, { height, width }) => {
-      console.log({ height, width });
       if (height !== undefined) element.style.height = height;
       if (width !== undefined) element.style.width = width;
     }, dimensions);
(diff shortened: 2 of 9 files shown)


e2e/tests/tooltip.test.ts
```diff
@@ -10,8 +10,6 @@ import { test } from '@playwright/test';
 
 import { common } from '../page_objects/common';
 
-process.env.ENV_URL = 'http://localhost:9002/';
```
nickofthyme, line 13: This straggler was causing issues pointing to the wrong port.

Issue #797, opened 2020-08-27
Value labels for line and area charts


wylieconlon:
I couldn't figure out what value labels were supposed to look like for line and area charts, so I've pulled some examples from popular chart-building tools:

**Line charts**:

<img width="400" alt="Screen Shot 2020-08-27 at 2 13 29 PM" src="https://user-images.githubusercontent.com/666475/91479946-3ddb8800-e870-11ea-970c-39a3a3f1d287.png"> <img width="200" alt="Screen Shot 2020-08-27 at 2 17 59 PM" src="https://user-images.githubusercontent.com/666475/91480023-564ba280-e870-11ea-891f-4854da6a85b2.png">

**Area charts**:

<img width="443" alt="Screen Shot 2020-08-27 at 2 13 40 PM" src="https://user-images.githubusercontent.com/666475/91479970-43d16900-e870-11ea-945a-bd2bef338c5a.png">

**Stacked area charts**:

<img width="300" alt="Screen Shot 2020-08-27 at 2 13 49 PM" src="https://user-images.githubusercontent.com/666475/91480002-4fbd2b00-e870-11ea-8ea3-fbc2b2b2a125.png"> <img width="300" alt="Screen Shot 2020-08-27 at 2 18 15 PM" src="https://user-images.githubusercontent.com/666475/91480168-91e66c80-e870-11ea-864e-c488bdab61ab.png">

dej611:
Actually this ![colored-labels](https://user-images.githubusercontent.com/666475/91480002-4fbd2b00-e870-11ea-8ea3-fbc2b2b2a125.png) label colored per category trick may be good also for bar charts

qbob1:
To anyone looking for a workaround on this issue:
You can layer a line chart on top of some transparent bar charts with a small padding value:
```{js}
theme: { ...defaultChartTheme, scales: { barsPadding: 1 } },
```
```{js}
{yAccessors?.map((yAccessor, index) => (
            <LineSeries
              id={toTitleCase(yAccessor)}
              key={yAccessor}
              name={yLegendOverrides?.[index] ?? toTitleCase(yAccessor)}
              xScaleType={xAxisProps?.scaleType}
              yScaleType={yAxisProps?.scaleType}
              xAccessor={xAccessor}
              yAccessors={[yAccessor]}
              data={baseData ?? []}
            />
          ))}
          {yAccessors?.map((yAcc, i) => (
            <BarSeries
              id={yAcc}
[...]

PR #2245, opened 2023-11-15
feat: Custom point renderer

qbob1:
## Summary
This pr allows the user to specify a custom point renderer for line charts. Use cases include custom shapes displayed in line charts.

## Details
This is done by exposing the canvas context during the point rendering. Specifically, by extending the point geometry style type to include a custom rendering function.

### Checklist
- [] The proper **chart type** label has been added (e.g. `:xy`, `:partition`)
- [] The proper **feature** labels have been added (e.g. `:interactions`, `:axis`)

### DIFF
--- a/packages/charts/src/chart_types/xy_chart/renderer/canvas/points.ts
+++ b/packages/charts/src/chart_types/xy_chart/renderer/canvas/points.ts
@@ -15,7 +15,7 @@ import { Circle, Fill, Stroke } from '../../../../geoms/types';
 import { Rotation } from '../../../../utils/common';
 import { Dimensions } from '../../../../utils/dimensions';
 import { PointGeometry } from '../../../../utils/geometry';
-import { GeometryStateStyle } from '../../../../utils/themes/theme';
+import { GeometryStateStyle, PointShape } from '../../../../utils/themes/theme';
 
 /**
  * Renders points from single series
@@ -33,6 +33,9 @@ export function renderPoints(ctx: CanvasRenderingContext2D, points: PointGeometr
         ...style.stroke,
         color: overrideOpacity(style.stroke.color, (fillOpacity) => fillOpacity * opacity),
       };
+      if (typeof style.shape === 'function') {
+        return style.shape(ctx, coordinates);
+      }
       renderShape(ctx, style.shape, coordinates, fill, stroke);
     });
 }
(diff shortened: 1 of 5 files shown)


nickofthyme:
buildkite test this

nickofthyme:
Hey @qbob1 Thanks for these changes, I'll take a look at them soon.

packages/charts/src/chart_types/xy_chart/renderer/canvas/points.ts
```diff
@@ -33,6 +33,9 @@ export function renderPoints(ctx: CanvasRenderingContext2D, points: PointGeometr
         ...style.stroke,
         color: overrideOpacity(style.stroke.color, (fillOpacity) => fillOpacity * opacity),
       };
+      if (typeof style.shape === 'function') {
+        return style.shape(ctx, coordinates);
```
nickofthyme, line 37: Hey @qbob1 thanks again for contributing to elastic charts!

Looking over these changes I wonder what is the primary goal or feature that you are looking for from these changes?

Are you looking for a way to define unique shape for the points aside from the limited `PointShape`s? In that case we could add an option to provide a [`Path2D`](https://developer.mozilla.org/en-US/docs/Web/API/Path2D/Path2D), to support any abstract path/symbol.

Or are you looking for support of value labels on line/area charts (i.e. https://github.com/elastic/elastic-charts/issues/797)?

The reason I ask is because we are not big fans of exposing the `ctx` directly to the user to do with as they please. With that said I'm happy to hear your thoughts and get the changes you are looking for added to the library.

qbob1: Totally understandable!
Fortunately, I was able to find a work around for #797 and posted a comment to help anyone else that may need to display the labels.
I'd say a Path2D would be excellent and a much cleaner solution.

nickofthyme: Thanks for adding that comment to #797.

Perfect, if you'd like to refactor this PR or create a fresh PR to add the option to assign a `Path2D` to `PointStyle.shape` I'll take another look. Otherwise it'll probably be a few weeks before we get around to adding it ourselves.

markov00:
Closing for now as I don't see updates since January

PR #2247, opened 2023-11-18
fix(metric): add option to set empty cell background color

nickofthyme:
## Summary

Adds option to `Theme.metric.emptyBackground` to set the color of empty backgrounds.

Screen Recording 2023-11-18 at 09 09 42 AM

## Details

In updating chart in kibana following changes in #2184 I noticed a difference in the empty/missing cells in the grid form. This was not presented in our VRT screenshots because we use the same background color for the chart as the page which Lens does not. Also we pass the real background color where Lens was passing `transparent` which would not enable color blending as desired in #2184. Thus in order to provide a way to color these empty/missing cells I added a theme option to set the background color.

### Checklist


nickofthyme:
buildkite update screenshots

PR #2250, opened 2023-11-19
fix(metric): add option to set empty cell background color (#2247) [61.0.x]

nickofthyme:
Backports the following commits to 61.0.x:
 - fix(metric): add option to set empty cell background color (#2247)


nickofthyme:
buildkite update screenshots

PR #2248, opened 2023-11-19
fix(metric): background color for bar with interactions

nickofthyme:
## Summary

Fixes changes made in #2184 with the interaction hover coloring for metrics with progress bar.

### Before
Screen Recording 2023-11-19 at 10 20 57 AM

### After

Screen Recording 2023-11-19 at 10 18 21 AM

## Details

When adding support for transparent colors in #2184 we used the default `chartBackground` component to supply the background and have the browser blend the colors. However, these changes removed the interaction lightening affect on metrics with bar charts because this would simply set no background on the `.echMetric` so the background would show through. But without any color we cannot lighten it to render the interaction affect.

### Checklist

- [x] The proper **chart type** label has been added (e.g. `:xy`, `:partition`)
- [x] The proper **feature** labels have been added (e.g. `:interactions`, `:axis`)
- [x] Unit tests have been added or updated to match the most common scenarios
- [x] Visual changes have been tested with `light` and `dark` themes

PR #2255, opened 2023-11-20
fix(metric): background colors and sparkline rendering

nickofthyme:
## Summary

This PR fixes issues created in #2184 related to rendering of the sparkline with transparent backgrounds. Additions include:
- Added optional `Theme.metric.blendingBackground` option used to blend transparent colors when used with a non-default background.
- Added optional `background` option `MetricBase` to override the background of individual metric cells.

## Details

In #2184 we added support for transparent colors. For this change we rendered the sparkline using an `svg` `mask` to separate the fill from the sparkline (see https://github.com/elastic/elastic-charts/pull/2184#discussion_r1349328772) to allow both to have a transparent background that would then blend with the `echBackground` color. But this approach created a small rendering artifact (see #2218) and also created a rendering issue with duplicated masking ids...

Screenshot 2023-11-20 at 9 26 42 AM

> Note all successive cells use the mask from the first cell. The third is fixed with a unique mask id.

The changes in #2184 created other issues (see #2248) causing a revert back to filling the metric elements with blend opaque colors. And thus the need to separate the sparkline svg elements is no longer valid. We now fill the svg sparkline with the respective blended background color.

Additionally we added the `Theme.metric.blendingBackground` option to specify the background color to blend transparent colors. This is useful in cases where the metric has a set background color like gray but has a transparent color fill. This option allows setting the blend color the the empty color of the theme rather than the gray background for more theme-specific coloring.

Screen Recording 2023-11-20 at 11 23 35 AM

> In the demo above we set the color fill to very transparent, change the background color to pink, then set the color transparency to ~50%. At this point you can see the pink blending with the metric color. But this new option allows us to use the theme background, in this case white, to blend the color. The background color is used elsewhere as needed (i.e. to fill text area of metrics with progress bar, empty/missing cells, etc.

Also fixes error in lightness logic with alpha values.

image

### Checklist


nickofthyme:
buildkite update screenshots

packages/charts/src/common/fill_text_color.ts
```diff
@@ -40,5 +40,5 @@ export function fillTextColor(
     return highContrastColor(blendedFgBg, 'WCAG2', options);
   }
 
-  return highContrastColor(backgroundRGBA);
+  return highContrastColor(backgroundRGBA, 'WCAG2', options);
```
nickofthyme, line 43: Missed this last time when passing text contrast options

nickofthyme:
buildkite test this

Issue #2048, opened 2023-05-23
[tooltip] Reconsider tooltip min-width and label truncations


markov00:
**Is your feature request related to a problem? Please describe.**
There are valid cases where longer labels are required to be displayed in their entirety.
Some of these cases are:
- You can quickly understand from the structure of the label if the value refers to some sort of subcategory or not, for example when showing hierarchical path names like `root.package.app.service.version` 
- The right-hand side of the label is more important than the left-hand side

![image](https://github.com/elastic/elastic-charts/assets/1421091/79aca4b6-e311-4b05-bb82-463e88027baf)

**Describe the solution you'd like**
The ideal path is to be always able to visualize the entire label on the tooltip but, as always, it depends on the length of it.
We can't represent a label that is as long as the current screen width (it will occlude too much in the current view), but is also true that we cannot limit too much the width and always truncate the label.
A trade-off should be considered.
We should also consider its effects on the legend (longer label on legend will shrink the chart)

**Describe alternatives you've considered**
The multi-line wrapping was already considered in the design but could increase the height of the tooltip too much depending on the number of lines to wrap every single text.  The design idea was to wrap the text at a max 300px width.
<img width="429" alt="Screenshot 2023-05-23 at 11 14 21" src="https://github.com/elastic/elastic-charts/assets/1421091/c9cfd500-70ec-42f4-860d-c5eff39f5d08">

The beginning/middle truncation of text is an alternative option, but doesn't fully cover the extent of the problem and is a custom solution (we need to compute that manually because this type of text truncation is not available in HTML out of the box)

**Additional context**
This and other similar situations are examples of the need for a full extent tooltip width or a tooltip with a different truncation.
![](https://files.slack.com/files-pri/T0CUZ52US-F0593CKM57W/image.png)

cc @gvnmagni

timductive:
Some more examples of where this is a problem.
![image](https://github.com/elastic/elastic-charts/assets/1280964/9a75bd93-6bcc-43ae-8897-4a46104a36d4)
![image_720](https://github.com/elastic/elastic-charts/assets/1280964/5185e60e-c084-4ff1-bf65-569760175a55)

gvnmagni:
After a few conversations we decided that the optimal solution would be to provide these series of solution to our users.

1. We should offer different ways of truncation that could be customized according to need. Users would be able to pick which option they prefer between `begin`, `middle` and `end`
[...]

PR #2261, opened 2023-11-28
feat: increase tooltip width to 500px and truncate items to 2 lines

mbondyra:
## Summary

Fixes https://github.com/elastic/elastic-charts/issues/2048 (I am not sure if we consider this issue fixed after merging this one)

Auto-expands the tooltip containter to 500px for wider content. The data items in the tooltip now truncate to 2 lines instead of just one. 

Screenshot 2023-11-28 at 15 06 15

Screenshot 2023-11-28 at 15 05 32


packages/charts/src/components/tooltip/components/_tooltip.scss
```diff
@@ -56,7 +56,11 @@ $transitionTime: 200ms;
   }
 
   .echTooltip__tableCell--truncate {
-    @include euiTextTruncate;
+    display: -webkit-box;
+    -webkit-box-orient: vertical;
+    line-clamp: 2;
+    -webkit-line-clamp: 2;
+    overflow: hidden;
```
markov00, line 63: you can probably use also the `lineClamp` sass mixin, it has a few more properties but it should work

packages/charts/src/chart_types/metric/renderer/dom/_text.scss
```diff
@@ -1,12 +1,3 @@
-@mixin lineClamp($maxLines) {
```
mbondyra, line 1: this was not used in any place so I moved it to generic mixins

markov00 (changes requested):

mbondyra:
Thank you @markov00 great tips! I addressed all your comments! 🙇🏼‍♀️

markov00 (approved):
Changes looks good to me, tested on deployed storybook.
Just re-generate the changed screenshots from the CI rather than from the local machine to correctly match with what the CI is going to match.

mbondyra:
buildkite update vrt

Issue #909, opened 2020-11-19
Improve popper.js performance with removing debounce.


nickofthyme:
**Is your feature request related to a problem? Please describe.**
Popper.js has a debounce on the `createPopper` method that may limit the performance of the tooltip with composite and other chart types. This debounce limits this method to _run at most once-per-tick_ which may be unacceptable performance.

**Describe the solution you'd like**
Investigate the removal of the debounce to replace with our own debounce flag to improve performance when not needed.

See https://github.com/popperjs/popper-core/issues/1194

monfera:
Thanks @nickofthyme. The goal is, consistent, synchronous repositioning of the `popper.js`-positioned tooltip AND whatever else moves together with the mouse, eg. the bin highlight (grey rectangle). Right now, if I swipe from left to right, the bin highlight is "glued" to the mouse X, while the tooltip is less in sync with the mouse. So the tooltip and the bin highlight can disturbingly separate from one another, esp. if there's some lag. And it's different when sweeping from left to right and right to left. Happens on current `elastic-charts` [too](https://elastic.github.io/elastic-charts/?path=/story/bar-chart--time-clustered-using-various-specs), more visible on less powerful hardware and/or more contrasty bin highlight, as the current one is not too different from the background.

It might even be the case that we want to debounce, but in a more controllable way, eg. the bin highlight and the tooltip together, ie. outside `popper.js`

nickofthyme:
Relates to https://github.com/elastic/elastic-charts/issues/615

markov00:
@nickofthyme can we close this after #2309 and #2270?

nickofthyme:
Yeah I think so, didn't even remember creating this issue.

PR #2270, opened 2023-12-06
fix: use passed size in pixel without waiting for resizeObserver

### LINKED ISSUE #2238: Metric text bounces when container size changed
**Describe the issue**
When the container size is updated along with the layout of the tiles, the text of the metric briefly changes size before adjusting itself.

**To Reproduce**
1. Visit the metric grid story [here](https://elastic.github.io/elastic-charts/?path=/story/metric-alpha--grid&globals=toggles.showHeader:true;toggles.showChartTitle:false;toggles.showChartDescription:false;toggles.showChartBoundary:false;theme:light&knob-show%20grid%20border=&knob-attach%20click%20handler=true&knob-use%20progress%20bar=true&knob-progress%20bar%20direction=vertical&knob-max%20trend%20data%20points=30&knob-number%20of%20columns=4&knob-debug%20randomized%20data=)
2. Change the column count

**Expected behaviour**
The text shouldn't resize.

**Screenshots**

https://github.com/elastic/elastic-charts/assets/315764/e499b29c-91ba-4ef5-80aa-de9fb59ac355

**Kibana Cross Issues**
Blocking https://github.com/elastic/kibana/issues/136993

**Checklist**

markov00:
## Summary

This change bypasses the need for one ResizeObserver event when passing chart size in pixel.
This avoid double re-renders with different sizes in the case of a datum + size change from the caller component.

## Issues

fix #2238
fix https://github.com/elastic/elastic-charts/issues/909

### Checklist


markov00:
buildkite test this

drewdaemon (commented):
Hmm, I still see the text bounce... did I miss something?


https://github.com/elastic/elastic-charts/assets/315764/76adb27f-0cbf-41cb-8c84-239c6f85590a

markov00:
> Hmm, I still see the text bounce... did I miss something?

Correct, I may have messed up with something here, let me check again thanks for the heads up

markov00:
buildkite update screenshots

markov00:
@drewdaemon ok so I've adjusted the example in Storybook, it works fine now:
- if the chart is resizable (like a panel in a dashboard) external container and the `Chart` size should be configured at `100%` width/height
- if the sizing is constrained then both the container and the `Chart` size need to be configured in pixels

markov00:
> The changes to the storybook may mark a departure from how this will be used in Kibana.
> 
> The way we have things set up creates a delay between when we apply the new CSS to the parent container and when we find out the true container dimensions the browser assigned.
> 
> For example, I could set the CSS on the parent container to the following
> 
> ```css
>     height: 100%;
>     width: 100%;
>     max-width: 200px;
>     max-height: 1200px 
> ```
> 
> Here I'm requesting a width of 200px and a height of 1200px. But, the parent of the parent container may only be 600px tall.
> 
> So, the application code would be telling Elastic charts that the text size should be based on a 200x1200 container when it's really in a 200x600 container.
> 
> If the client detects that the true size is 200x600 and tells Elastic charts, the text size will be corrected. But this just leads to the same bouncing behavior as we see today.
> 
> This is the reason I wondered about using [CSS container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_container_queries) to decouple the text size from the React render cycle.
> 
> As far as I can see, forcing the client to choose between "resizable" mode and "static" mode doesn't fit the experience we're trying to deliver.
> 
> Might be easiest to sync on this problem. Maybe there's some change I can make on my end as well.

@drewdaemon let's sync on this together, I'm not fully sure I'm aligned with how you are currently handing this in Kibana

markov00:
just waiting on https://github.com/elastic/kibana/pull/173772 to be merged

markov00:
just waiting on https://github.com/elastic/kibana/pull/173772 to be merged

markov00:
@nickofthyme can you review this please?

packages/charts/src/components/chart.tsx
```diff
@@ -121,10 +121,17 @@ export class Chart extends React.Component<ChartProps, ChartState> {
     this.unsubscribeToStore();
   }
 
-  componentDidUpdate({ title, description }: Readonly<ChartProps>) {
+  componentDidUpdate({ title, description, size }: Readonly<ChartProps>) {
     if (title !== this.props.title || description !== this.props.description) {
       this.chartStore.dispatch(updateChartTitles(this.props.title, this.props.description));
     }
+    if (size !== this.props.size) {
```
nickofthyme, line 128: Does this only need to check the reference not that actual values?

packages/charts/api/charts.api.md
nickofthyme, line 1: Did you confirm with Drew that this fixes his issue? https://github.com/elastic/elastic-charts/pull/2270#issuecomment-1849702312

nickofthyme, line 1: I think you can remove the `Test Cases > Resize Debounce` [story](https://ech-e2e-ci--pr-2270-lggsv47e.web.app/?path=/story/test-cases--resize-debounce&globals=toggles.showHeader:true;toggles.showChartTitle:false;toggles.showChartDescription:false;toggles.showChartBoundary:false;theme:light).

nickofthyme (approved):
Looks good to me, I don't see any issues with these change in code or in storybook.

packages/charts/src/components/chart.tsx
markov00: good one, I've fixed that in https://github.com/elastic/elastic-charts/pull/2270/commits/d17a87e86df639ca7ea810aa294bfde2c2324082

nickofthyme (approved):

PR #2275, opened 2023-12-07
style(tooltip): increase width to 500px and 2 lines items (#2261)

markov00:
Backport of #2261


markov00:
buildkite update screenshots

PR #2303, opened 2024-01-16
feat(metric): custom slot to render contents in gap

### LINKED ISSUE #2182: Add render slot for gap area in Metric
Solutions teams have requested a way to consistently place tags below the title on the Metric component. This is not easy to force externally with consistency.

image

Ideally we allow a render prop on the `Metric` chart that we pass each datum to in order to render the contents into the gap area.

image

The user would be responsible for padding and sizing of the gap content as applicable to their use case.

Rendering custom elements absolutely relative to the `Metric` chart. Causes inconsistent placement of contents.

**Additional context**
Since the `Metric` uses dom elements this approach is preferred over dimensional placement of the render contents.

nickofthyme:
## Summary

This PR adds the ability to render arbitrary contents into the gap space of the metric. This adds the optional `slot` prop to the `Metric` spec.

Screen Recording 2024-01-16 at 01 06 53 PM

## Details

The contents of the slot are contained inside of an absolutely positioned container to allow rendering contents without contents size impacting the layout of the Metric elements. Alternatively, we could add an option to disable this behavior ad allow the Metric elements to be pushed out of the render area or we could account for the slot size at the time of rendering.

## Issues

closes #2182

### Checklist


nickofthyme:
buildkite update screenshots

packages/charts/api/charts.api.md
```diff
@@ -1987,6 +1987,8 @@ export interface MetricSpec extends Spec {
     // (undocumented)
     data: (MetricDatum | undefined)[][];
     // (undocumented)
+    slot?: ReactNode;
```
markov00, line 1990: 🦥 do we have a better name for that? isn't too generic?

markov00 (approved):
I'm fine with the code changes, just the naming is a bit too vague

packages/charts/api/charts.api.md
nickofthyme: Yeah what are you thinking?

markov00: I don't know, could be something like `content` or `body`

nickofthyme: Renamed to `body` in 5156195

nickofthyme:
buildkite update screenshots

PR #2305, opened 2024-01-17
refactor(axes): replace `integersOnly` with `maximumFractionDigits`

### LINKED ISSUE #2220: Replace `Axis.integersOnly` option with `maximumFractionDigits `
With formatting the `Axis` tick labels, it is sometimes desirable to alter the values prior to their formatting. This is because the formatting can alter the true numerical value of label to the point where the label is no longer accurate with respect to the axis scale.

We added the option to limit axis values to integers in https://github.com/elastic/elastic-charts/issues/387, but I think making this more flexible as to define a `maximumFractionDigits` would provide greater utility to the chart consumer.

This option's naming comes from the Swift documentation for the `NumberFormatter`, see [docs](https://developer.apple.com/documentation/foundation/numberformatter/1415364-maximumfractiondigits).

There is no true alternate to this other than a way for the chart consumer to manually dictate the ticks to be shown which would be too flexible from a configuration level.

**Kibana Cross Issues**

- https://github.com/elastic/kibana/issues/170151

nickofthyme:
## Summary

Adds a way to round axes tick values prior to formatting the numerical values into a string.

⚠️  The `Axis.integersOnly` prop is now deprecated and replace with `maximumFractionDigits` ⚠️ 

### Usages

Below shows, toggling `maximumFractionDigits` between `0` (integers only) and `1` (single decimal or tenths only)

Screen Recording 2024-01-16 at 07 44 46 PM

Below shows, toggling `maximumFractionDigits` between `0` (integers only), `1` (single decimal or tenths only) and  `2` (double decimal or hundredths only)

Screen Recording 2024-01-16 at 08 09 41 PM

## Details

The original logic to add the `integersOnly` only allowed rounding of the numerical tick value in the case of an integer. This PR expands on that by allowing the chart consumer to set the max number of fractional digits (aka decimals) in a numerical value. This is different than **_formatting_** in order to show a value to a max fractional digits, as this can render deceptive ticks and their respective placements, see linked issue.

### Drawbacks

This is not an active assignment of ticks given the rounded constraint but rather a passive removal of ticks that do not abide by the constraint. Currently `d3-scale` does not support asking for ticks in this manner. Thus there are drawbacks to this approach, namely those listed below:

- Nicing end ticks can be removed - Given the range extents, there are some cases where nicing a domain with a restrictive `maximumFractionDigits` value with result in one or both of the end ticks being removed. See the second gif above as an example of this.
- Tick count is more variable - Say we set `maximumFractionDigits` to `1` and out data domain is between `0` and `1`. Then let's say we desire `~10` ticks, that would be a tick every `0.1` unit. Here we would keep all the ticks as it would maintain all `0.X` values. But what if we specify `~4` ticks, d3 would likely give us ticks at every `0.25` unit, in this case all ticks but the `0` and `1` would be removed as then are `0.XX` not `0.X`. I don't see a way to fix this without iterating a few times, or with custom tick logic.
- Less density of ticks - This is hard to avoid as the very nature of rounding would limit the possible ticks available to show.

## Issues

Closes #2220

### Checklist


nickofthyme:
buildkite update screenshots

nickofthyme:
buildkite update screenshots

storybook/stories/axes/15_integers_only.story.tsx
markov00, line 1: We should probably remove the screenshot related to this story

markov00 (commented):
Screenshot 2024-01-24 at 15 38 57

This doesn't look correct to me, with 0 digit the `1` should not be there, the nice scale shows `0.7` but when changing to 0 fractional digits the `0.7` transforms to `1`

storybook/stories/axes/15_integers_only.story.tsx
nickofthyme: Yeah this is very strange. This seems to be a bug in the `visible_ticks.ts` logic. When I disable the `adaptiveTickCount` it works as expected. Do you know what this adaptive logic is doing and why it's computed independently from the actual scale?

https://github.com/elastic/elastic-charts/blob/3b3aa03456be76f2ef1e8398e9bf140bb14bd54a/packages/charts/src/chart_types/xy_chart/state/selectors/visible_ticks.ts#L283-L327

nickofthyme: So taking your example, like below, you can see the issue. But if I log the domain, range and ticks for all iterations of the y axis. You can see the first 3 are determined using the correct scale, the last 3 are from the code above, which appears to be walking the domain out from `0.7` to `0.8` to `1`. This returns the ticks but these ticks are used with the scale computed originally with a different domain. 🤔 What the??

image


```
data: (5) [
  ['Sensor 1', 0.68574]
  ['Sensor 2', 0.5857399999999999]
  ['Sensor 3', 0.48573999999999995]
  ['Sensor 4', 0.38573999999999997]
  ['Sensor 5', 0.28574]
]

Tick iterations of y axis continuous scale
-------------------------------------------
 domain: (2) [0, 0.7]
 range: (2) [0, 1]
 ticks: (8) [0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7]
-------------------------------------------
 domain: (2) [0, 0.7]
 range: (2) [307, 0]
 ticks: (8) [0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7]
-------------------------------------------
 domain: (2) [0, 0.7]
 range: (2) [307, 0]
 ticks: (8) [0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7]
-------------------------------------------
 domain: (2) [0, 0.8]
 range: (2) [307, 0]
 ticks: (5) [0, 0.2, 0.4, 0.6, 0.8]
-------------------------------------------
 domain: (2) [0, 0.8]
 range: (2) [307, 0]
 ticks: (5) [0, 0.2, 0.4, 0.6, 0.8]
-------------------------------------------
 domain: (2) [0, 1]
 range: (2) [307, 0]
 ticks: (3) [0, 0.5, 1]
```

nickofthyme: Created #2311

markov00: I will take a look at this

markov00 (approved):
Let's merge this and fix the other issue in a subsequent PR, thanks

PR #2310, opened 2024-01-23
perf(tooltip): improve placement logic

### LINKED ISSUE #2309: Popperjs is excessively creating new instances for external tooltips
**Describe the issue**
When using a tooltip synced across multiple charts, the performance can quickly be diminished. The main culprit in that the `createPopper` method is called on every mouse move, due to the settings changes on each render.

**To Reproduce**
1. Go to this [story](https://elastic.github.io/elastic-charts/?path=/story/interactions--multi-chart-cursor-sync&globals=toggles.showHeader:true;toggles.showChartTitle:false;toggles.showChartDescription:false;toggles.showChartBoundary:false;theme:light)
2. Place a log within this [hook](https://github.com/elastic/elastic-charts/blob/48cade4b42e873ab9dcac88b958f29ef35298a0b/packages/charts/src/components/tooltip/tooltip.tsx#L150-L167) via chrome source lookup.
3. Notice the log is called to update the `popperSettings` on every mouse change for each external chart.

**Expected behaviour**
`popperSettings` are not excessively updated unnecessarily. 

 - Elastic Charts: `latest`

nickofthyme:
## Summary

Fixes performance issue notice with synced tooltips.

## Details

The `popperSettings` value here...

https://github.com/elastic/elastic-charts/blob/48cade4b42e873ab9dcac88b958f29ef35298a0b/packages/charts/src/components/tooltip/tooltip.tsx#L150-L167

...was triggering an update for every new render even though the settings values had not changed.

This was caused by the `getTooltipSettings` helper method here...

https://github.com/elastic/elastic-charts/blob/48cade4b42e873ab9dcac88b958f29ef35298a0b/packages/charts/src/components/tooltip/tooltip.tsx#L332-L343

...which was returning a new merged instance of the settings for every external event.

## Issues

fixes #2309

### Checklist

### DIFF
--- a/packages/charts/src/components/portal/tooltip_portal.tsx
+++ b/packages/charts/src/components/portal/tooltip_portal.tsx
@@ -109,11 +109,11 @@ const TooltipPortalComponent = ({
    * Popper instance used to manage position of tooltip.
    */
   const popper = useRef<Instance | null>(null);
-  const popperSettings = useMemo(
+  const popperSettings = useMemo(() => {
     // @ts-ignore - nesting limitation
-    () => mergePartial(DEFAULT_POPPER_SETTINGS, settings),
-    [settings],
-  );
+    return mergePartial(DEFAULT_POPPER_SETTINGS, settings);
+  }, [settings]);
+
   const destroyPopper = useCallback(() => {
     if (popper.current) {
       popper.current.destroy();
--- a/packages/charts/src/components/portal/tooltip_portal.tsx
+++ b/packages/charts/src/components/portal/tooltip_portal.tsx
@@ -103,7 +103,7 @@ const TooltipPortalComponent = ({
    */
   useEffect(() => {
     document.body.appendChild(portalNode);
-  });
+  }, [portalNode]);
 
   /**
    * Popper instance used to manage position of tooltip.
--- a/packages/charts/src/components/tooltip/tooltip.tsx
+++ b/packages/charts/src/components/tooltip/tooltip.tsx
@@ -30,7 +30,7 @@ import {
 } from './types';
 import { Colors } from '../../common/colors';
 import { SeriesIdentifier } from '../../common/series_id';
-import { BaseDatum, DEFAULT_TOOLTIP_SPEC, SettingsSpec, TooltipProps, TooltipSpec, TooltipValue } from '../../specs';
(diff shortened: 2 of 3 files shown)


packages/charts/src/components/portal/tooltip_portal.tsx
```diff
@@ -109,11 +109,11 @@ const TooltipPortalComponent = ({
...
   const popper = useRef<Instance | null>(null);
-  const popperSettings = useMemo(
+  const popperSettings = useMemo(() => {
     // @ts-ignore - nesting limitation
-    () => mergePartial(DEFAULT_POPPER_SETTINGS, settings),
-    [settings],
-  );
+    return mergePartial(DEFAULT_POPPER_SETTINGS, settings);
+  }, [settings]);
+
```
nickofthyme, line 116: Got tired of changing this every time to add a log 🤦‍♂️

packages/charts/src/components/tooltip/tooltip.tsx
```diff
@@ -329,17 +329,19 @@ export const TooltipComponent = <D extends BaseDatum = Datum, SI extends SeriesI
...
-
-  return {
-    ...tooltip,
-    ...externalPointerEvents.tooltip,
-  };
+  return Object.assign(
+    tooltipSettings,
+    // @ts-ignore - nesting limitation
+    mergePartial(tooltip, externalPointerEvents.tooltip),
+  );
```
nickofthyme, line 344: This is the main issue, whenever `isExternalTooltipVisible` is `true` this issue occurs.

I think this is the best approach for the time being to always return the same referenced object. I thought about adding logic to the `mergePartial` function to allow passing a referenced object target to add the values to like the `target` of the `Object.assign` method but this would be a pretty big change for a small use case.

LMK what you think.

markov00: I'm not super happy about this solution, the `tooltipSettings` is not a module instance and is reused across multiple charts, this could possibly create some issue

markov00: @nickofthyme

nickofthyme: Yeah good point. I'm struggling to find the best solution here. The best I found is to extract this into a selector, which in itself does not solve the problem but if we create a cached selector to return a new tooltip setting singleton cached by `chartId`, this only updates `popperSettings` in `tooltip.tsx` upon entering the chart.

What do you think? See 906a9c1

PR #2350, opened 2024-03-08
fix(bullet): full circle overlapping tick labels

nickofthyme:
## Summary

Simple fix for overlapping ticks on `Circle` Bullet chart. Also now allows passing `ticks` as `number[]` instead of `(domain: [number, number]) => number[]`.

Zight Recording 2024-03-08 at 08 52 31 AM

## Details

This is a basic fix to remove the ending tick on the circular bullet whenever the start and end tick values are the same. This is to prevent overlapping.

This is not a perfect fix as it does not address the issue for when the label position are not equal but still overlap due to the length of the labels.

### Checklist


nickofthyme:
buildkite update screenshots

nickofthyme:
@markov00 could you take a look and this quick PR. As a first pass I am just adopting the approach from kibana to remove the max value of the 2 overlapping ticks.

markov00 (approved):
Looks good to me, I'm fine with that fix even if this partially cover the problem

Issue #1053, opened 2021-03-04
Percentage area charts can have gaps


dej611:
**Describe the bug**

Chromium bug: https://bugs.chromium.org/p/chromium/issues/detail?id=1163912

When series have `null` or missing values for a date histogram with area percentage, a gap is rendered within the chart:

<img width="560" alt="Screenshot 2021-03-04 at 18 34 41" src="https://user-images.githubusercontent.com/924948/110008935-9d818700-7d1c-11eb-8553-0e7f78ce4251.png">

<img width="173" alt="Screenshot 2021-03-04 at 18 45 37" src="https://user-images.githubusercontent.com/924948/110008954-a1ada480-7d1c-11eb-8b04-525a3bd36630.png">

**To Reproduce**

<details>
<summary>CSV data dump:</summary>

```
"Top values of FlightDelayType","timestamp per 30 minutes","Count of records"
"No Delay","08:30",1
"No Delay","09:00",2
"No Delay","09:30",6
"No Delay","10:00",6
"No Delay","10:30",5
"No Delay","11:00",3
"No Delay","11:30",6
"No Delay","12:00",11
"No Delay","12:30",8
"No Delay","13:00",3
"No Delay","13:30",8
"No Delay","14:00",9
"No Delay","14:30",4
"No Delay","15:00",4
"No Delay","15:30",8
"No Delay","16:00",5
"No Delay","16:30",4
"No Delay","17:00",6
"No Delay","17:30",2
"No Delay","18:00",6
"No Delay","18:30",4
"No Delay","19:00",3
"No Delay","19:30",3
"No Delay","20:00",5
"No Delay","20:30",1
"No Delay","21:00",1
"No Delay","21:30",7
"No Delay","22:00",6
"No Delay","22:30",9
"No Delay","23:00",3
"No Delay","23:30",9
"No Delay","00:00",3
"No Delay","00:30",6
"No Delay","01:00",4
"No Delay","01:30",5
"No Delay","02:00",2
"No Delay","02:30",4
"No Delay","03:00",4
"No Delay","03:30",7
"No Delay","04:00",4
"No Delay","04:30",6
"No Delay","05:00",7
"No Delay","05:30",4
"No Delay","06:00",5
"No Delay","06:30",4
"No Delay","07:00",9
"No Delay","07:30",3
"No Delay","08:00",6
"No Delay","08:30",5
"No Delay","09:00",10
"No Delay","09:30",6
"No Delay","10:00",4
[...]

PR #2407, opened 2024-04-11
fix(barSeries): error rendering bars with negative log scale

### LINKED ISSUE #607: Band charts should show y0 value in tooltip, be treated as a single series
I'm exploring how we will visualize bounded data in Lens, and found this issue. As seen in [this band chart](https://codesandbox.io/s/focused-field-tpjm0), the lower value is not shown in the tooltip.

Screenshot 2020-03-25 17 46 28

Also, I would like to treat a band as a single series, not as a two series. It doesn't seem like it's possible to get the series name to only show up once in the legend, even using code like:

```
          y0AccessorFormat={''}
          y1AccessorFormat={''}
```

### LINKED ISSUE #2406: Log scale uses wrong baseline for negative domains
When using a logarithmic scale with all negative values, the areas or bars are rendered from the bottom up (towards from `0`) instead of the top down (aways from `0`)

image

The expected behavior should match that of the linear scales, as shown below.

image

### LINKED ISSUE #2408: Banded `BarSeries` not showing lower value in tooltip
When using `y1Accessors` and `y0Accessors`, the tooltip values exclude the value from the `y0Accessors`.

image

nickofthyme:
## Summary

This PR mainly fixes the rendering of bars on a negative log scale. Also fixes the tooltip for banded bar charts.

Zight Recording 2024-04-11 at 10 27 05 AM

In addition this PR fixes some other issues including
- Banded bar charts missing lower bound series in tooltip and legend
- Wrong highlighting of tooltips value and highlighted bar geometries, particularly for banded bar charts.
- Banded series items in tooltip were being shuffled according to the distance to the cursor position. Now I pre-sort them to not abruptly switch.
- The Chrome bug from #1053 associated with filling areas containing sharp corners, is now fixed and since using the shared `getY1ScaledValueFn` and `getY0ScaledValueFn` functions I just removed this logic to avoid checking.

## Details

The bar geometry logic was using custom logic to handle the scaled `y1` and `y0` values, this was imperfect. This same issue was not present with `AreaSeries` as we had already addressed it. Using the same logic from area geometries fixed the issue.

Secondly, while testing the above fix for banded `BarSeries`, I noticed that the tooltip never showed the `lower` bound item on the tooltip. This again only an issue for bar charts. The issue is that unlike the `AreaSeries` the `BarSeries` geometry is a single geometry for the upper and lower whereas the area is 2 points. So simply adding the `y0` geometry to the `indexedGeometryMap` solved the issue, see bed2b45.

## Issues

fix #2406, fix #2408, fix #607

### Checklist


nickofthyme:
buildkite update screenshots

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -143,6 +148,24 @@ export function renderBars(
...
+          y: getDatumYValue(datum, true, isBandedSpec, stackMode),
+          mark: null,
+          accessor: BandedAccessorType.Y0,
+          datum: datum.datum,
+        },
+        bandedY: finalY1,
+      });
+
+      barGeometry.bandedY = finalY0;
+    }
```
nickofthyme, line 167: Adds lower (`y0`) tooltip and legend value.

packages/charts/src/chart_types/xy_chart/rendering/utils.ts
```diff
@@ -136,8 +137,22 @@ export function isPointOnGeometry(
...
+  if (!isWithinRange([x, x + width])(xCoordinate)) return false;
+
+  if (isBandedSpec) {
+    if (bandedY === undefined) return false;
+
+    const rangeUtils = inRange(y, bandedY);
+    return accessor === 'y0' ? rangeUtils.firstHalf(yCoordinate) : rangeUtils.firstHalf(yCoordinate);
+  }
+
+  return isWithinRange([y, y + height])(yCoordinate);
```
nickofthyme, line 155: Only highlights the closest half of banded bars in tooltip.

Zight Recording 2024-04-26 at 03 03 07 PM

packages/charts/src/chart_types/xy_chart/state/selectors/get_tooltip_values_highlighted_geoms.ts
```diff
@@ -129,58 +129,63 @@ function getTooltipAndHighlightFromValue(
...
-
-    xValues.add(indexedGeometry.value.x);
-
-    return [...acc, formattedTooltip];
-  }, []);
+  const values = matchingGeoms
+    .toSorted((a, b) => {
+      // presort matchingGeoms to group by series then y value to prevent flipping
+      return b.seriesIdentifier.key.localeCompare(a.seriesIdentifier.key) || b.value.y - a.value.y;
+    })
```
nickofthyme, line 136: Prevents `lower` and `upper` tooltip items from swapping based on cursor position. Code in `reduce` below is unchanged.

Zight Recording 2024-04-26 at 03 04 49 PM

packages/charts/src/utils/common.tsx
```diff
@@ -732,6 +732,32 @@ export const isWithinRange = (range: [number, number], exclusive = false) => {
...
+      return start === min ? isHalfFromMin(n) : isHalfFromMax(n);
+    },
+    /**
+     * Returns true if values are within the last half of range, from end halfway to start
+     */
+    lastHalf: (n: number) => {
+      return end === max ? isHalfFromMax(n) : isHalfFromMin(n);
+    },
+  };
+};
```
nickofthyme, line 759: Utility to simplify checking if a value is in the first or last half of a range relative to the start/end regardless of polarity.

packages/charts/src/utils/geometry.ts
```diff
@@ -103,6 +103,7 @@ export interface BarGeometry {
   value: GeometryValue;
   seriesStyle: BarSeriesStyle;
   panel: Dimensions;
+  bandedY?: number;
```
nickofthyme, line 106: We need to know what the other part of the range, as these are independently checked if cursor is over geometry.

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -43,40 +43,45 @@ export function renderBars(
...
+    const finiteHeight = y0Scaled - y1Scaled || 0;
     const absHeight = Math.abs(finiteHeight);
-    const height = absHeight === 0 ? absHeight : Math.max(minBarHeight, absHeight); // extend nonzero bars
-    const heightExtension = height - absHeight;
+    const crossesZero = y0 && y1 && isWithinRange([y0, y1])(0);
+    const adjustedHeight =
+      absHeight === 0
+        ? absHeight
+        : // extends nonzero bar height - doubled for bars crossing 0
+          Math.max((crossesZero ? 2 : 1) * minBarHeight, absHeight);
```
nickofthyme, line 75: Using `minBarHeight` on banded bars is a little strange, but for small value of say `~2px` it's ok.

The issue is that the `minBarHeight` can be applied away from the baseline, altering the accuracy of the scale.

Additionally, when a banded bar crosses the `0` baseline, we only applied the `minBarHeight` to one side, which looks very strange. This code fixes this issue but simply doubling the `minBarHeight` for cases where the `minBarHeight` is applied and where the bar crosses `0`.

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -43,40 +43,45 @@ export function renderBars(
...
+    const y1Extension = (adjustedHeight - absHeight) / (crossesZero ? 2 : 1);
+    const y0Extension = crossesZero ? y1Extension : 0;
     const isUpsideDown = finiteHeight < 0;
-    const finiteY = Number.isNaN(y0Scaled + rawY) ? 0 : rawY;
-    const y = isUpsideDown ? finiteY - height + heightExtension : finiteY - heightExtension;
+    const height = isUpsideDown ? -adjustedHeight : adjustedHeight;
```
nickofthyme, line 79: Previously we were basically swapping `y1` and `y0` to render an adjusted bar to always render top down. This was very confusing but makes sense, however attempting to alter the highlighter logic for this case was extremely difficult and after many tries I gave up.

Now, the `y0` and `y1` values are what they are, we compute the absolute `height` and apply the original sign to the height to indicate that the bar is upsidedown. This allow us to check in the highlighter for this case and handle it accordingly. The canas renderer correctly renders the bar without translating this negative height.

This also helps to check for cursor over geometry.

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -143,6 +148,24 @@ export function renderBars(
...
+          x: datum.x,
+          y: getDatumYValue(datum, true, isBandedSpec, stackMode),
+          mark: null,
+          accessor: BandedAccessorType.Y0,
+          datum: datum.datum,
+        },
+        bandedY: finalY1,
+      });
+
+      barGeometry.bandedY = finalY0;
```
nickofthyme, line 166: Only need this value if it is a banded spec

packages/charts/src/chart_types/xy_chart/renderer/dom/highlighter.tsx
```diff
@@ -100,13 +103,14 @@ class HighlighterComponent extends React.Component<HighlighterProps> {
...
+
           return (
             <rect
               key={i}
               x={x}
-              y={y}
+              y={geom.height < 0 ? (accessor === 'y1' ? y + geom.height : y) : accessor === 'y1' ? y : y - geom.height}
               width={geom.width}
-              height={geom.height}
+              height={Math.abs(geom.height)}
```
nickofthyme, line 113: Transforms geoms to correctly render negative height svg for highlighting either `y0` or `y1` bars.

nickofthyme (commented):
Just leaving comments to explain specific code changes.

nickofthyme:
buildkite update screenshots

packages/charts/src/chart_types/xy_chart/rendering/utils.ts
nickofthyme: Removed. No longer applies

nickofthyme:
buildkite update screenshots

PR #2411, opened 2024-04-18
fix:(metric) trend issue with multiple date histograms

### LINKED ISSUE #2410: [Metric] When trend is enabled the date histogram doesn't get sorted when ingested
In a CCS context Kibana can get a table with multiple stacked date histograms not correctly sorted (for more info see here).
If the background trend is enabled in the Metric chart type it can get into a broken state like the following:

Screenshot 2024-04-17 at 09 43 34

Kibana related issue: https://github.com/elastic/kibana/issues/181012

dej611:
## Summary

The `trend` prop now handles correctly data coming from multiple date histograms.

Before:
Screenshot 2024-04-18 at 15 01 54
Screenshot 2024-04-18 at 15 01 48

After:
Screenshot 2024-04-18 at 15 00 49
Screenshot 2024-04-18 at 15 00 42

## Details

When multiple date histograms with zero or null values are handled to the `trend` prop, values were not stacked correctly for the same `x` bucket.

## Issues

Fix #2410

### Checklist


packages/charts/src/chart_types/metric/renderer/dom/sparkline.tsx
```diff
@@ -21,6 +21,37 @@ export const getSparkLineColor = (color: MetricWTrend['color']) => {
...
+        return true;
+      }
+      lastX = x;
+    });
+  if (!shouldBeSorted) {
+    return trend;
+  }
+  return trend.toSorted((a, b) => {
+    // make null behave like 0
+    return a.x - b.x || a.y - b.y;
```
dej611, line 51:
```suggestion
    return a.x - b.x || +a.y - b.y;
```

nickofthyme:
buildkite update screenshots

packages/charts/src/chart_types/metric/renderer/dom/sparkline.test.ts
```diff
@@ -20,9 +20,10 @@ const sampleDataWithNull = [
   { x: 1, y: null },
   { x: 2, y: 3 },
 ];
+
 describe('getSafeTrendData', () => {
   it('should avoid to process the data if already sorted', () => {
-    expect(getSafeTrendData(sampleData)).toEqual(sampleData);
+    expect(getSafeTrendData(sampleData)).toBe(sampleData);
```
nickofthyme, line 26: Check strict equality

packages/charts/src/chart_types/metric/renderer/dom/sparkline.tsx
```diff
@@ -21,21 +21,20 @@ export const getSparkLineColor = (color: MetricWTrend['color']) => {
...
   const shouldBeSorted =
-    // eslint-disable-next-line eqeqeq
-    lastX != null &&
+    lastX !== undefined &&
     trend.some(({ x }) => {
-      // eslint-disable-next-line eqeqeq
-      if (lastX == null) {
+      if (lastX === undefined) {
         return false;
       }
```
nickofthyme, line 39: We avoid the loose equality, especially in this case where the value cannot be `null`.

image

nickofthyme (approved):
Changes look good to me, I pushed a couple changes.

packages/charts/src/chart_types/metric/renderer/dom/sparkline.test.ts
dej611: Just for curiosity what's the difference in this context?
As long as the content is the same, having the same reference or not to the same array object doesn't add particular value, to me, to this test.
To me the particular reference equality is just an implementation detail, but maybe I'm missing some context.

packages/charts/src/chart_types/metric/renderer/dom/sparkline.tsx
dej611: But values can be `null` in real life.
Based on types, values cannot be `undefined` either.
Maybe the change doesn't have any particular impact, but it would be useful to understand some context here as well.

packages/charts/src/chart_types/metric/renderer/dom/sparkline.test.ts
nickofthyme: The test claims to assert that no sort was performed, but `.toEqual` does not actually assert that, it just says you started with a sorted and ended with the same sorted array. The `.toBe` asserts that the array reference is the same and thus no sorting occurred.

Not sure if that what you indented, with this test or not. I'm fine with it either way.

packages/charts/src/chart_types/metric/renderer/dom/sparkline.tsx
nickofthyme: Yeah we simply go by the typescript types. In this case that value of `trend.x` is `number`.

https://github.com/elastic/elastic-charts/blob/785f6351737e1df6798d93200584646d49b83bae/packages/charts/src/chart_types/metric/specs/index.ts#L81-L82

Additionally in this case you are pulling off the first value from the array which technically can be `undefined`. Unlike kibana, we enforce this check by enabling the [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) option which says the value of any array access, except for explicit tuple type arrays, can be `undefined`.

So the resulting type for `lastX` is `number | undefined`, which does not include `null`. If this were used in javascript you could pass `null` but we assume everyone used TS for the purpose of type safety.

If the case warranted you to handle `null` values, then we would change the type of `trend.x` to include `null`. Then in this case we would check for `null` and `undefined` individually an not the shorthand `lastX != null`.

dej611: I think the index access type check can be enabled without the `===` eslint rule.
Anyway, thanks for the context explanation.

packages/charts/src/chart_types/metric/renderer/dom/sparkline.tsx
```diff
@@ -21,6 +21,35 @@ export const getSparkLineColor = (color: MetricWTrend['color']) => {
...
+        return false;
+      }
+      if (lastX > x) {
+        return true;
+      }
+      lastX = x;
+    });
+  if (!shouldBeSorted) {
+    return trend;
+  }
```
markov00, line 47:
```suggestion
  const shouldBeSorted = trend.some(({x}, index, array) => {
    if(index===0) return false;
    if(x < array[index-1]?.x){
        return true;
    }
    return false;
});
```
Can we use something like that (just pseudocode not checked)? To simplify a bit the code and also to avoid mutating the lastX variable from within the .some loop?

packages/charts/src/chart_types/metric/renderer/dom/sparkline.tsx
```diff
@@ -21,6 +21,35 @@ export const getSparkLineColor = (color: MetricWTrend['color']) => {
...
 
+/**
+ * Aligns and implicitly stacks all histogram trend data.
+ *
+ * Sometimes the trend data arrives with multiple series
+ * appended one on top of each other.
+ *
+ * @internal
+ */
+export const getSafeTrendData = (trend: MetricWTrend['trend']) => {
```
markov00, line 32: more than safeTrendData, I think we can call this function: `getSortedTrendData`, the word safe is not very clear to me. We are also not stacking data (so also the comment is wrong)

packages/charts/src/chart_types/metric/renderer/dom/sparkline.tsx
```diff
@@ -29,8 +58,9 @@ export const SparkLine: FunctionComponent<{
   if (!trend) {
     return null;
   }
-  const [xMin, xMax] = extent(trend.map((d) => d.x));
-  const [, yMax] = extent(trend.map((d) => d.y));
+  const groupedTrendData = getSafeTrendData(trend);
```
markov00, line 61: and in the end this is not a groupedTrendData but just the sortedTrend array

packages/charts/src/chart_types/metric/renderer/dom/sparkline.tsx
```diff
@@ -29,8 +58,9 @@ export const SparkLine: FunctionComponent<{
...
+  const [xMin, xMax] = extent(groupedTrendData.map((d) => d.x));
```
markov00, line 62: At this point we can also consider to specify `xMin = groupedTrendData.at(0).x` and `xMax = groupedTrendData.at(-1).x`;
Because we are ensuring the order of the elements, the extremes are the min and max and we save another loop.

Issue #2096, opened 2023-07-10
Legend design v2


gvnmagni:
A new legend design is proposed here.

Instead of the existing legend behavior, we will have two legend formats: `table` and `inline`. Each version has its own pros and cons and its use depends case by case.

Each legend format can be used with each 4 possible positions around the chart (top, left, bottom, right), each format will behave slightly differently on each position.

Both format can support series statistics (or multiple values) depending on the chart type as highlighted [here](https://github.com/elastic/kibana/issues/176583)

## Table format

Series names are aligned to the left, and values are aligned to the right (such as their header) in order to simplify comparison between rows.

![Frame 150](https://github.com/elastic/elastic-charts/assets/18484576/71882143-6b5f-4ddf-a56f-1e90cfdf72b5)

Table header:
- can be hidden by default when only the series/categories name are displayed and no statistic/value is visible
- the series header label should be passed from the API
- the other statistics header labels are in English by default, but we should provide a way to internationalize them (also on a subsequent PR)

Table content:
- using what was implemented [here](https://github.com/elastic/elastic-charts/pull/2355)  is possible to ask for one or many statistics to be shown in the table/inline version.
- the same statistics order should apply

**Inline format**

Values follow the data series name, for each data series columns header are repeated. We use a `:` char to separate them

![Frame 149](https://github.com/elastic/elastic-charts/assets/18484576/e74b83d7-feef-409b-95b5-4e41a10f0b6a)

## Legend sizing, truncation, line wrapping

The legend can be sized automatically or statically with a fixed width/height.

Currently, when auto-sized legend and is rendered at the left/right side of the chart, the legend computes its maximum size automatically before rendering (via canvas measureText and not via DOM requests to avoid multiple rendering passes before rendering the chart). The current size is computed based on the label length and the legend value that is currently displayed without interactions.

With these new formats, we should improve this calculation and find a good way to measure the max legend size based on something similar `clamp(max(table headers, formatted cell values), 1/2 of the chart parent size)`.

| format | position | size | scroll | wrap/truncate | 
| -- | -- | -- | -- | -- |
| table | left/right | clamp(max(header/values), 1/2parent)) | horizontal + vertical |  values and their header are fixed to their max size, label/series name title instead will take at maximum the 3/4 of the available space, then it goes to 2 lines max and then truncates | 
[...]

PR #2426, opened 2024-05-08
feat(legend): add legend stats (table view)

mbondyra:
## Summary

This PR enables users to add some statistics (e.g., avg, min, max, etc.) for their metrics in the legend. This way they can get their insights faster by easily accessing more information within the chart. All statistics are computed client-side considering the entire dataset range selected.

Screenshot 2024-05-28 at 10 29 56

Screenshot 2024-05-28 at 10 31 13

## Details

Story to test it: Legend/Tabular data

Implements table view for legend statistics. The table shows up automatically when user chooses any statistic other than `CurrentAndLastValue` for xy charts only. 

For the charts that don't have the new legend values added, the legacy view is displayed.

## Issues

part of https://github.com/elastic/elastic-charts/issues/2096

### Checklist


storybook/stories/legend/17_tabular_data.story.tsx
dej611, line 1: On the UX/UI side I've just noticed an edge case with a long list of items, where the alignment between the chart and the legend seems a bit off.

Take this example from storybook:
Screenshot 2024-05-28 at 11 04 55

In particular the bottom side of the legend goes beyond the lower limit of the chart axis. I think some padding there as with the chart would be nice.
Note that on the top side the padding applies for both chart and legend, so that feels inconsistent.
Not a major issue, as legend always worked like that, but with the new table layout this issue became more prominent.

dej611 (commented):
Left some code reviews on a first pass.

packages/charts/src/state/selectors/get_legend_table_size.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  const widestValuesWidths = items.reduce((acc, { values }) => {
+    const valuesWidths = values.map(calcApprValuesWidth);
+    return acc.map((w, i) => Math.ceil(Math.max(w, valuesWidths[i] || 0)));
+  }, valuesTitlesWidth);
+
+  const seriesWidth = Math.ceil(widestLabelWidth + GRID_CELL_PADDING.width * 2);
+
+  const legendItemWidth =
+    seriesWidth + widestValuesWidths.reduce((acc, w) => acc + w + GRID_CELL_PADDING.width * 2, 0) + 1;
```
markov00, line 96: calculation here are a bit convoluted but ok for now.
I can suggest to remove the `withTextMeasure` calls inside the function, pass a `textMeasure` from the caller of `getLegendTableSize`.
In its caller, `getLegendSizeSelector` we already have a function wrapped into a `withTextMeasure` so we can just adjust that function and pass the textMeasure to `getLegendTableSize`.

Something like that is fine:
```
if (shouldDisplayTable(legendValues)) {
      return withTextMeasure((textMeasure) =>getLegendTableSize(config, theme, parentDimensions, items, textMeasure));
    }

File: storybook/stories/legend/17_tabular_data.story.tsx
Comment (markov00) [reply storybook/stories/legend/17_tabular_data.story.tsx:1]: cc @gvnmagni

**Review by markov00 — COMMENTED** [review]
Just minor comments, in general it looks great and I haven't found any bug right now. I will keep testing it

File: storybook/stories/legend/17_tabular_data.story.tsx [hunk 2]
```diff
@@ -0,0 +1,118 @@
...
+
+      <BarSeries id="bars" xScaleType={ScaleType.Linear} yScaleType={ScaleType.Linear} {...datasets[datasetSelect]} />
+    </Chart>
+  );
+};
+
+Example.parameters = {
+  markdown:
+    'The `legendAction` action prop allows you to pass a render function/component that will render next to the legend item.\n\n __Note:__ the context menu, color picker and popover are supplied by [eui](https://elastic.github.io/eui/#).',
+};
```
markov00, line 118: I believe this is just a copy coming from a different storybook story, please update it or just remove it

markov00 (approved):

packages/charts/src/common/legend.ts
```diff
@@ -91,3 +90,33 @@ export type LegendItem = {
 
 /** @internal */
 export type LegendItemExtraValues = Map<LegendItemChildId, LegendItemValue>;
+
+/** @internal */
+export const shouldDisplayTable = (legendValues: LegendValue[]) =>
+  !!legendValues.filter((v) => v !== LegendValue.CurrentAndLastValue && v !== LegendValue.Value).length;
```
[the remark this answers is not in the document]
mbondyra: I'd rather stick to mine because this suggestion doesn't include all the cases (empty array case). What we would have to check with the usage of every would be not only if there is a statistic other than CurrentAndLastValue, but also if after excluding the CurrentAndLastValue the length is 1 or more.

packages/charts/src/components/legend/label.tsx
```diff
@@ -106,8 +99,34 @@ export function Label({
...
-    <div dir={dir} className={labelClassNames} title={label} style={clampStyles}>
+    <NonInteractiveLabel label={label} options={options} />
+  );
+}
+
+/** @internal */
+export function NonInteractiveLabel({ label, options }: { label: string; options: LegendLabelOptions }) {
+  const { className, dir, clampStyles } = getSharedProps(label, options);
+  return (
+    <div dir={dir} className={className} title={label} style={clampStyles}>
```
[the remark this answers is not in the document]
mbondyra: I am using this component for a <LegendTableHeader/>  where I don't want to pass all the props for <Label/> (  label,  onToggle,  isToggleable,  isSeriesHidden,  options,  hiddenSeriesCount,  totalSeriesCount). I'll remove it from the usage from 'Label' but will keep it for <LegendTableHeader/>

storybook/stories/legend/17_tabular_data.story.tsx
mbondyra: Here's the option with ditching the paddings we add:

Screenshot 2024-05-28 at 13 50 19

packages/charts/src/common/legend.ts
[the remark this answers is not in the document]
mbondyra: Just saw @markov00's suggestion, thanks!

storybook/stories/legend/17_tabular_data.story.tsx
gvnmagni: what would you think about using the zero axix as a lower limit? the problem here is that chart and legend don't have a shared visual line to lay on, I know that we would take a few pixels because of this but we will probably gain some elegance

mbondyra: I like the idea and tried to add it quickly, but unfortunately it's not so simple - we have a lot of variables here and I have to be more cautious doing so. I've added it as an [issue](https://github.com/elastic/elastic-charts/issues/2442) here, because it's not directly related to stats, but also applies to the old view.

e2e/tests/legend_stories.test.ts
```diff
@@ -248,7 +248,7 @@ test.describe('Legend stories', () => {
       (l) => `should display nested legend extra values on ${l}`,
       async (page, layout) => {
         await common.expectChartAtUrlToMatchScreenshot(page)(
-          `http://localhost:9001/?path=/story/legend--piechart&knob-Partition Layout=${layout}&knob-flatLegend=false&knob-showLegendExtra=true&knob-legendMaxDepth=2`,
+          `http://localhost:9001/?path=/story/legend--piechart&knob-flatLegend=false&knob-showLegendExtra=true&knob-legendMaxDepth=2`,
```
mbondyra, line 251:
```suggestion
          `http://localhost:9001/?path=/story/legend--piechart&knob-Partition Layout=${layout}&knob-flatLegend=false&knob-showLegendExtra=true&knob-legendMaxDepth=2`,
```

packages/charts/src/components/legend/legend_table/legend_table_item.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { LegendTableRow } from './legend_table_row';
+import { LegendValueComponent } from './legend_value';
+import { SeriesIdentifier } from '../../../common/series_id';
+import { LayoutDirection } from '../../../utils/common';
+import { Label as ItemLabel } from '../label';
+import { useLegendColorPicker } from '../legend_color_picker';
+import { LegendItemProps, prepareLegendValues } from '../legend_item';
+
+/** @internal */
+export const LegendListItem: React.FC<LegendItemProps> = (props) => {
```
nickofthyme, line 22: nit: Component name should match file or vis versa

packages/charts/src/components/legend/legend_table/legend_table.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { SharedLegendItemProps } from '../types';
+
+/** @internal */
+export interface LegendTableProps extends SharedLegendItemProps {
+  items: LegendItem[];
+  seriesWidth?: number;
+}
+
+/** @internal */
+export const GRID_COLOR_WIDTH = 10;
```
nickofthyme, line 24: This value appears to be slightly too small to fit the color icon with its focus state. Looks like `20` would suffice.

nickofthyme (commented):
Excellent job 🎉 🚀 

The code looks great and playing around with it works perfectly! I think the way you built it make it easily to merge this with the tooltip table if we decide to go that route.

I just left 2 follow-up things below...

packages/charts/src/components/legend/legend_table/legend_table.tsx
mbondyra: Yeah, I noticed this one but I prioritized the fact that we don't want to occupy too much precious space here. Thank you for both comments, I'll followup in the next PR! <3

gvnmagni:
Adding a little comment here for future considerations, but not a blocker for this specific PR.

- The horizontal scroll bar should be superimposed on a sort of legend container instead of being blocked at the bottom of the table, users will find a little annoying the need to scroll down in order to scroll right. We already mentioned this in our conversations so nothing new, just wanted to write it down

Issue #2306, opened 2024-01-18
[Metric] Enhance typography positioning and proportions


gvnmagni:
Due to a relevant amount of feedback, I would propose to improve the positioning and alignment of the primary and secondary value within Metric charts. 

The current status of this component force these block of text to be aligned to the right side in order to facilitate vertical comparison of numbers:

![Metric Chartaaaa](https://github.com/elastic/elastic-charts/assets/18484576/707af090-97d2-4684-8bcd-01b92cb81b5e)

I would propose to:
1. include the possibility for the user to decide where to align these elements (left or right)
2. re-evaluate the current handling of font-size to see if here is room enough to make improvements or to adjust our current breakpoints
3. once implemented the new alignment, having it by default to the left and having the right one as an alternative

So that we could get this:

![Metric Chartbbb](https://github.com/elastic/elastic-charts/assets/18484576/88d47666-c302-4876-b70e-f10b89350984)

gvnmagni:
Hi all, I wanted to share a proposal for a Metric Chart update that takes in consideration the list of feedback received in these last month. I will list here the main concepts and then we will have the chance to discuss in detail...

**1. Font sizes: breakpoints**
Breakpoints would be defined by panel height instead of width and vertical responsiveness of elements would then be required only in the smallest breakpoints (hide and show elements according to heights). Font sizes have also been enlarged.

![Frame 1metric](https://github.com/elastic/elastic-charts/assets/18484576/212db853-9a68-4933-b9cc-6fafe7d6e406)

**2. Font sizes: new functionalities**
It would be great to introduce two new options that would NOT be activated by default but that would be available to users.
`Full-width text` that automatically resize it to make it as large as possible, and `Custom font size` that would allow users to specify the desired font size. This second possibility would be really helpful in case of monitoring dashboards in control rooms where customers might have a better understanding of the context that we can't have on our side.

![Frame 2metric](https://github.com/elastic/elastic-charts/assets/18484576/28df8f48-f636-45b4-bf39-a32a19c58cfb)


**3. Alignments and positioning of elements**
Possibility to choose the desired alignment, while having as a default an overall alignment to the left.

![Frame 4metric](https://github.com/elastic/elastic-charts/assets/18484576/72a3a860-a542-4f90-bec1-1f78b81e30b1)

**4. Padding**
A little aesthetic detail, a bigger inner padding to make the whole component a little more elegant and clean.

[...]

PR #2428, opened 2024-05-09
feat(metric): support array of values

### LINKED ISSUE #2111: [Metric] allow array of values
Add the possibility to render an array of values in a concatenated way.

When passing an array of values a bar can't be rendered and the background color could only have a categorical meaning because is not possible to link a single color to a set of values.

**Screenshots (if relevant):**

This is how the legacy metric will render concatenated strings:

Screenshot 2023-06-28 at 2 44 16 PM

**Kibana cross issue**
https://github.com/elastic/kibana/issues/160815

nickofthyme:
## Summary

The `Metric` now supports array of number or array of string values.

- Array values must be homogeneous.
- Numerical `valueFormatter` applies to each number in array.
- Current display does not nicely support the longer values, this will be addressed in #2306

## Issues

closes #2111

### Checklist


nickofthyme:
buildkite update screenshots

packages/charts/src/chart_types/metric/renderer/dom/text.tsx
```diff
@@ -305,7 +313,33 @@ export const MetricText: React.FunctionComponent<{
...
+    );
+    return [...numericValueParts, { emphasis: 'normal', text: ']' }];
+  }
+
+  return isMetricWNumber(datum)
+    ? isFiniteNumber(datum.value)
+      ? splitNumericSuffixPrefix(datum.valueFormatter(datum.value))
+      : [{ emphasis: 'normal', text: style.nonFiniteText }]
+    : [{ emphasis: 'normal', text: datum.value }];
+}
```
markov00, line 340: You can avoid writing the same operation if you transform your `isMetricWNumber` into a single array element.

```suggestion
 
    const numericValueParts = (isMetricWNumberArrayValues(datum) ? datum.value : [datum.value]).reduce<TextParts[]>(
      (acc, value, i, { length }) => {
        const parts: TextParts[] = isFiniteNumber(value)
          ? splitNumericSuffixPrefix(datum.valueFormatter(value))
          : [{ emphasis: 'normal', text: style.nonFiniteText }];
        if (i < length - 1) {
          parts.push({ emphasis: 'normal', text: ', ' });
        }
        return [...acc, ...parts];
      },
      [{ emphasis: 'normal', text: '[' }],
    );
    return [...numericValueParts, { emphasis: 'normal', text: ']' }];
}
```

packages/charts/src/chart_types/metric/specs/index.ts
```diff
@@ -37,6 +37,19 @@ export type MetricBase = {
 /** @alpha */
 export type MetricWText = MetricBase & {
   value: string;
+  valueFormatter?: never;
```
markov00, line 40: I know that a string can be preformatted, but do we need to enforce that we don't want a valueFormatter?
Can we just remove that prop all together in this case?

markov00 (approved):

packages/charts/src/chart_types/metric/specs/index.ts
nickofthyme: Yes we can but if we have an array of `MetricDatum` values, if this is not present, typescript will not error when a `valueFormatter` is added to one of the union types that does not support it.

With this `never` is added it will try to match it to the `MetricWNumberArrayValues` type.

image

Not exactly necessary but better type errors.

packages/charts/src/chart_types/metric/renderer/dom/text.tsx
nickofthyme: No this code would treat everything like an array.

I was trying to keep it simple and easier to read but I combined them all in a66c6b175f6a01069c58ad52980d836cede51d42. There's more type checking here with all unions also checking for `NaN`. But this is as compact as I can get it. WDYT?

markov00: >this code would treat everything like an array.

correct, this to limit the code duplication, because in both cases you are checking for finiteNumber, splitting, etc.
All fine for me, except that you can move the `valueFormatter` call outside the reduce to avoid calling it everytime.

nickofthyme: Done f65840a

PR #2431, opened 2024-05-16
feat(legend): change click on item behaviour

dej611:
Re-apply the same commit from #2427.

## Summary

Clicking on a legend item has been improved to isolate a series on first click, show all series on second click.
Also, when clicking on a hidden series now will add the selected series into the visible set.

### BREAKING CHANGE
The modifier key has been changed to `CTRL` (or `CMD` on Mac) from `SHIFT`. The `SHIFT` key will no longer have any effect on click.

## Details

A full description of the new behaviour:
> First click isolates selected element, then behave according to context (if single element is selected, restore all, otherwise toggle on off).

### Checklist


markov00:
buildkite test this

markov00 (approved):

PR #2440, opened 2024-05-21
fix: react component type errors

### LINKED ISSUE #2398: React type errors rendering `ComponentType`
In a few places throughout charts we render custom components on the type `ComponentType<Props>`. Like the `LegendColorPicker` to define a custom color picker.

https://github.com/elastic/elastic-charts/blob/a602838462721b6de3643da8263beab46508ac0d/packages/charts/src/specs/settings.tsx#L340-L341

Lately I've started seeing error where typescript is complaining about these components with the following error...

image

```
'ColorPicker' cannot be used as a JSX component.
  Its element type 'ReactElement<any, any> | Component<LegendColorPickerProps, any, any> | null' is not a valid JSX element.
    Type 'Component<LegendColorPickerProps, any, any>' is not assignable to type 'Element | ElementClass | null'.
      Type 'Component<LegendColorPickerProps, any, any>' is not assignable to type 'ElementClass'.
        The types returned by 'render()' are incompatible between these types.
          Type 'React.ReactNode' is not assignable to type 'import("/Users/nickpartridge/Documents/repos/elastic-charts/docs/node_modules/@types/react/index").ReactNode'.ts(2786)
```

If I were to use the `renderWithProps` utility like below, the error goes away.

```
return renderWithProps(ColorPicker, {
  anchor: this.colorRef.current,
  color,
  onClose: handleClose,
  onChange: handleChange,
  seriesIdentifiers,
});
```

I'm not sure if this is a typing issue or what but the functionality is unaffected. But we should look into this to resolve the type errors.

These errors do not appear as type errors in `main` but do appear when you add the exact `@types/react` to the top-level `package.json`. I think they always show in the IDE.

See example build errors at https://buildkite.com/elastic/datavis-ci/builds/3508#018ec3c1-2ccd-405a-834b-3a7ce5879bbb

nickofthyme:
## Summary

Fixes type errors related to react types.

## Details

I think the root issue here is that `react` now has a `peerDependency` on `@types/react` as `*` which would be fine if only one react was used in the project. But we are using `react@18` in `docs/` and `react@16` everywhere else and typescript does not like this dual usage.

Since we use yarn workspaces, the react types are hoisted to the top. This fix adds the explicit dependencies of `@types/react` and `@types/react-dom` in `packages/charts/package.json`. Additionally, VSCode will still not like this unless there is a local `tsconfig.json` file at the base of the `packages/charts/` directory, which affectively points to the local `packages/charts/node_modules/` before the top `node_modules/`.

This is a temporary fix, I will clean up the react types when upgrading to `react@18`.

## Issues

fixes #2398


package.json
```diff
@@ -150,11 +150,8 @@
     "postcss-loader": "^4.2.0",
     "prettier": "^3.2.5",
     "pretty-quick": "^4.0.0",
-    "react": "^16.13.0",
```
markov00, line 153: last time with @mbondyra adding just `"@types/react": "^16",` in the root package.json did the trick. Can we just do that?

nickofthyme: Doing that does not fix the problem. I still end up with 33 errors both in the IDE and `tsc`.

```
Found 33 errors in 15 files.

Errors  Files
     1  packages/charts/src/chart_types/bullet_graph/chart_state.tsx:35
     1  packages/charts/src/chart_types/flame_chart/flame_chart.tsx:1401
     1  packages/charts/src/chart_types/goal_chart/state/chart_state.tsx:82
     4  packages/charts/src/chart_types/heatmap/state/chart_state.tsx:89
     1  packages/charts/src/chart_types/metric/state/chart_state.tsx:28
     3  packages/charts/src/chart_types/partition_chart/renderer/dom/layered_partition_chart.tsx:22
     1  packages/charts/src/chart_types/timeslip/timeslip_chart.tsx:160
     1  packages/charts/src/chart_types/wordcloud/state/chart_state.tsx:60
     1  packages/charts/src/chart_types/xy_chart/renderer/dom/annotations/annotation_tooltip.tsx:81
     7  packages/charts/src/chart_types/xy_chart/state/chart_state.tsx:121
     1  packages/charts/src/components/accessibility/partitions_data_table.tsx:141
     1  packages/charts/src/components/accessibility/screen_reader_summary.tsx:71
     6  packages/charts/src/components/chart.tsx:190
     1  packages/charts/src/components/legend/legend.tsx:121
     3  packages/charts/src/components/tooltip/tooltip.tsx:239
```

mbondyra: Yeah, adding it didn't fix everything for me either :(

markov00 (approved):
Let's merge it like that but I don't like this solution so much because now the API docs is a mix of 2 defined React versions.
We can definitely find a solution when we are going to move to react 18 everywhere

PR #2441, opened 2024-05-23
fix(Metric): Cards should only show one focus halo on `::focus` event

alexwizp:
Closes: https://github.com/elastic/observability-dev/issues/3430
Closes: https://github.com/elastic/observability-dev/issues/3391

When we use the `Metric Chart` in `Kibana`, we have a problem where the focus halo is shown twice. I think we can explicitly add the property `outline: none`; to the `echMetricText__title button` to resolve this issue

image

### DIFF
--- a/packages/charts/src/chart_types/metric/renderer/dom/_text.scss
+++ b/packages/charts/src/chart_types/metric/renderer/dom/_text.scss
@@ -15,6 +15,7 @@
       font-weight: bold;
       width: 100%;
       text-align: left;
+      outline: none;
     }
   }
   &__icon {
--- a/packages/charts/src/chart_types/metric/renderer/dom/_text.scss
+++ b/packages/charts/src/chart_types/metric/renderer/dom/_text.scss
@@ -15,7 +15,6 @@
       font-weight: bold;
       width: 100%;
       text-align: left;
-      outline: none;
     }
   }
   &__icon {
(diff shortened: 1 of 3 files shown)


packages/charts/src/chart_types/metric/renderer/dom/_text.scss
```diff
@@ -15,6 +15,7 @@
       font-weight: bold;
       width: 100%;
       text-align: left;
+      outline: none;
```
markov00, line 18: We already have that `outline:none` applied but this is anyway overridden by the EUI or Kibana style:
```
:focus:not([class^=eui]):not(.kbn-resetFocusState):not(.euiSelectableList__list>ul) {
    outline: 2px solid currentColor;
}
```
that has a higher specificity. I've tested your change directly in the observability page and this doesn't fix the issue.

possibilities are: add an "hard" `important!` or fix that in Kibana

markov00: if you are looking where we are already applying `outline: none` is here packages/charts/src/chart_types/metric/renderer/_index.scss:
```
&:focus-within button {
    outline: none;
  }
```

alexwizp: Thank you, @markov00. The PR has been updated by using `inline` styles, which avoids the need to use `!important` + don't require to override styles on Kibana side.

image

markov00 (approved):

Issue #2445, opened 2024-05-30
[Metric] SparkLine - chart not rendered when the trend is empty


alexwizp:
Regression of #2411

When the `metric` chart is visualized with an empty `trend`, the chart component fails to render, and the following error message appears in the console:
<img width="1152" alt="image" src="https://github.com/elastic/elastic-charts/assets/20072247/4517dcd9-d186-4f85-b55a-6b681d9d4a10">

**How to reproduce:**
1. Configure any `Metric` chart 
2. Pass `[]` as value for trend attribute.

**Possible solution:**
1. Apply the following fix. This diff represents the change in line 56 of the `sparkline.tsx` file.

```diff
- if (!trend) {
+ if (!trend?.length) {
```
**Screen:**

> Trend was set to []

<img width="1669" alt="image" src="https://github.com/elastic/elastic-charts/assets/20072247/5e9de40b-99b7-4f01-9367-9cd04986d1a5">

PR #2447, opened 2024-05-30
fix(metric): chart not rendered when the trend is empty

alexwizp:
Closes: https://github.com/elastic/elastic-charts/issues/2445

## Summary

This change ensures that the chart component can handle `empty` trend data appropriately, preventing rendering failures and console errors.

## Screen 

Screen:

`Trend was set to []`

image

No `console` issues


packages/charts/src/chart_types/metric/renderer/dom/sparkline.tsx
```diff
@@ -50,7 +50,7 @@ export const SparkLine: FunctionComponent<{
   id: string;
   datum: MetricWTrend;
 }> = ({ id, datum: { color, trend, trendA11yTitle, trendA11yDescription, trendShape } }) => {
-  if (!trend) {
+  if (!trend?.length) {
```
markov00, line 53: there is no need for the optional chaining. `trend` is always an array in typescript

markov00 (changes requested):
This doesn't solve the issue completely.

The root cause is a couple of lines below that changed line where:
```
  const [xMin, xMax] = [sortedTrendData.at(0)!.x, sortedTrendData.at(-1)!.x];
```
We need to avoid these type of assertions because they can cause all sort of issues around them (cc @dej611)

Even if with a single element the chart renders, the SVG is malformed (will contain NaN numbers that throws in the console).

The correct fix should check if the trend contains at least 2 elements. If not you should assure a correct background/foreground color selection (changing it to `if(trend.length < 2) return null` doesn't solves it quickly unfortunately)

packages/charts/src/chart_types/metric/renderer/dom/sparkline.tsx
alexwizp: @markov00, the optional chaining is used here because we already have this condition checked before. However, based on your comment, I see that the correct fix should be more complex. Could someone more involved in the `Chart` project prepare a fix for this?

I believe this issue is the reason for the flaky `CI` in `Kibana`.

nickofthyme: Yeah I can take a look. Thanks @alexwizp

nickofthyme: @markov00 Check new changes.

packages/charts/src/chart_types/metric/renderer/dom/sparkline.tsx
```diff
@@ -50,11 +50,9 @@ export const SparkLine: FunctionComponent<{
   id: string;
   datum: MetricWTrend;
 }> = ({ id, datum: { color, trend, trendA11yTitle, trendA11yDescription, trendShape } }) => {
-  if (!trend) {
-    return null;
-  }
```
nickofthyme, line 55: I removed this `null` return completely and let it just render what it can given the computed values. This only renders sparklines for trend with 2 or more values.

If we block this it does not render the require `Sparkline` elements to property fill the metric, see first (white) metric in the demo below...

image

When it should look like...

image

nickofthyme:
buildkite update screenshots

packages/charts/src/chart_types/metric/renderer/dom/sparkline.tsx
```diff
@@ -50,11 +50,9 @@ export const SparkLine: FunctionComponent<{
...
   const sortedTrendData = getSortedData(trend);
-  const [xMin, xMax] = [sortedTrendData.at(0)!.x, sortedTrendData.at(-1)!.x];
+  const xMin = sortedTrendData.at(0)?.x ?? NaN;
+  const xMax = sortedTrendData.at(-1)?.x ?? NaN;
```
markov00, line 55: this doesn't solves the problem either.
if the trend has 1 only value then the x scale will first do a division by zero then it returns NaN on the path generation causing the SVG to errors out with something like `MNaN,0LNaN,1Z`

markov00 (changes requested):

alexwizp:
@markov00  what if we base our logic for showing the trend on meeting two conditions: the difference between `xMin` and `xMax` should not be 0, and `yMax` should not be zero either? I think it will be more correct instead of  checking the length of `trend`/`sortedTrendData`.

nickofthyme (approved):
New changes LGTM, let's see what @markov00 thinks.

packages/charts/src/chart_types/metric/renderer/dom/sparkline.tsx
nickofthyme: Good point, I didn't check the console errors. Looks like the latests changes fix this.

markov00 (approved):
LGTM, thanks for taking care of it

PR #2448, opened 2024-05-31
chore: reduce non-numeric number noise in logs

markov00:
## Summary
This PR reduces the noise generated in the console logs by passing non-numerical values as Y values in cartesian charts when in development mode.


nickofthyme (approved):
Nice!

PR #2471, opened 2024-06-23
fix: outside rect annotation placement and group relations

### LINKED ISSUE #2470: Defining a `RectAnnotation` as `outside` has not affect
**Describe the issue**
If a `RectAnnotation` is defined without using a `groupId` the `outside` prop has no affect.

**To Reproduce**
1. Go to https://codesandbox.io/p/sandbox/quizzical-saha-s8k5p2?file=%2Fsrc%2FApp.tsx%3A38%2C28
2. Comment/uncomment the `groupId` on the `RectAnnotation` spec
3. Notice the annotation is not rendered as `outside` annotation when `groupId` is missing

**Expected behaviour**
The `RectAnnotation` is always rendered on the outside of the chart when the `outside` prop is set to `true`.

**Screenshots**
image

I think this issue is original to the `outside` feature added in #1207

**Additional context**
The code has a break branch from #842 to render annotations when we have no `yScale`.

https://github.com/elastic/elastic-charts/blob/ae2af1ccff3124ef2a9254c9163ae511c857354d/packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts#L75

This is the path taken when a `groupId` is applied. But when a `yScale` **is** defined, it renders the outside only if `x0` and `x1` where initially `undefined`, this is the problem!

https://github.com/elastic/elastic-charts/blob/ae2af1ccff3124ef2a9254c9163ae511c857354d/packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts#L129-L134

nickofthyme:
## Summary

This PR fixes an issue with `RectAnnotation` when defined as `outside`. Now all `outside` `RectAnnotation`s render correctly within their respective `Axis`.

Zight Recording 2024-06-23 at 03 48 12 PM

## Details

This prop is meant to render the annotation outside of the chart in the axis area, however this only works when there is an arbitrary `groupId` applied, because this would cause the `yScale` to be undefined as if it's the same case as fixed in #842.

This was due to a combination of errors in the logic. First was the missing rotation when looking up the axis specs here...

https://github.com/elastic/elastic-charts/blob/ae2af1ccff3124ef2a9254c9163ae511c857354d/packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts#L43

This would provide wrong axis for the given annotation. I have now made the `rotation` argument required to avoid this error in the future.

Secondly, the logic to determine the placement dimensions of the `outside` annotation was extremely limiting for all cases particularly related to chart rotations.

https://github.com/elastic/elastic-charts/blob/ae2af1ccff3124ef2a9254c9163ae511c857354d/packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts#L77-L81

This was fixed by factoring in each case for rotation and domain, as is done for the `LineAnnotation` markers.

## Issues

fixes #2470

### Checklist


packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts
```diff
@@ -40,7 +43,7 @@ export function computeRectAnnotationDimensions(
   isHistogram: boolean = false,
 ): AnnotationRectProps[] | null {
   const { dataValues, groupId, outside, id: annotationSpecId } = annotationSpec;
-  const { xAxis, yAxis } = getAxesSpecForSpecId(axesSpecs, groupId);
+  const { xAxis, yAxis } = getAxesSpecForSpecId(axesSpecs, groupId, chartRotation);
```
nickofthyme, line 46: Needed to match to correct axis

packages/charts/src/chart_types/xy_chart/annotations/rect/dimensions.ts
```diff
@@ -99,10 +96,33 @@ export function computeRectAnnotationDimensions(
...
+      return; // cannot scale y values without a scale
+    }
+
+    const hasXValues = isDefined(initialX0) || isDefined(initialX1);
+
+    if (outside) {
+      if (hasXValues && hasYValues) {
+        Logger.warn(
+          `The RectAnnotation (${annotationSpecId}) was defined as outside but has both x and y values defined.`,
+        );
```
nickofthyme, line 108: Warns when attempting to render annotation as `outside` but both x and y coordinates are defined. The `outside` annotation should be limited to a single axis. Renders inside as fallback to avoid not rendering at all.

nickofthyme:
buildkite update screenshots

nickofthyme:
buildkite update screenshots

PR #2474, opened 2024-06-25
fix: (Legend values) percentage formatter should be distict from tick formatter

mbondyra:
## Summary

Fixes a few small issues for legend values:
- ✔️ Percent and difference percent formatting should not take the format from tick formatter. Instead, we override it with a percent formatter with a precision 0.1. Fixed, visual test added
- ✔️  for formatting with `MB` as a suffix, unnecessary scroll appears. That's because we calculate the width of each cell approximately and we underestimated. I corrected it by changing the width to 8.5 for each letter.
- ✔️ for the value headers that are longer than the value (for example last non null), unnecessary scroll appear because we didn't apply the proper font width that is bold. Fixed, visual test added

I changed one of the tests to make sure we test at least one stacked chart.


mbondyra:
buildkite update screenshots

mbondyra:
buildkite update screenshots

mbondyra:
buildkite update screenshots

mbondyra:
buildkite update screenshots

mbondyra:
buildkite update screenshots

mbondyra:
buildkite update screenshots

mbondyra:
buildkite update screenshots

mbondyra:
buildkite update screenshots

mbondyra:
buildkite update screenshots

mbondyra:
buildkite update screenshots

mbondyra:
buildkite update screenshots

e2e/screenshots/legend_stories.test.ts-snapshots/legend-stories/legend-tabular-data/should-correctly-display-current-and-last-value-median-values-chrome-linux.png
mbondyra, line 1: the difference is just a bit more space reserved for the legend

e2e/screenshots/legend_stories.test.ts-snapshots/legend-stories/legend-tabular-data/should-correctly-display-long-copy-dataset-dataset-chrome-linux.png
mbondyra, line 1: changing the chart to stacked

e2e/screenshots/legend_stories.test.ts-snapshots/legend-stories/legend-tabular-data/should-correctly-display-median-values-chrome-linux.png
mbondyra, line 1: the difference is just a bit more space reserved for the legend

packages/charts/src/chart_types/xy_chart/state/utils/get_legend_values.ts
```diff
@@ -38,18 +38,26 @@ export function getLegendValues(
...
-      label: typeof value === 'number' ? formatter(value) : '',
+      label,
       value,
     };
   });
 }
 
+function percentFormatter(value: number) {
+  return `${(value * 100).toFixed(1)}%`;
+}
```
nickofthyme, line 59: At some later point it might be nice to expose this to the chart consumer, with this as the fallback formatter.

packages/charts/src/chart_types/xy_chart/state/utils/get_legend_values.ts
```diff
@@ -38,18 +38,26 @@ export function getLegendValues(
...
   types: LegendValue[],
   valueAccessor: (d: DataSeriesDatum) => number | null,
-  formatter: TickFormatter<any> | ((tick: unknown) => string),
+  tickFormatter: TickFormatter<any> | ((tick: unknown) => string),
 ) {
   return types.map((type) => {
     const value = getLegendValue(series, xDomain, type, valueAccessor);
+    const formatter =
+      type === LegendValue.Percent || type === LegendValue.DifferencePercent ? percentFormatter : tickFormatter;
+    const label = typeof value === 'number' && isFinite(value) ? formatter(value) : '';
```
nickofthyme, line 47: For future consideration...

In the `Tooltip` we have an option called `showNullValues` to prevent filtering out `null` values. This then allows the user to format these `null` values as they wish with the `tickFormat` (e.g. `N/A`), see demo [here](https://elastic.github.io/elastic-charts/storybook/?path=/story/interactions--interaction-with-null-values&globals=toggles.showHeader:true;toggles.showChartTitle:false;toggles.showChartDescription:false;toggles.showChartBoundary:false;theme:light).

image

I think we might want to match that here. It's a little different in that the values are indirectly formatted with the `tickFormatter` but mostly the same idea. I think it would be better if we just let the user handle both cases if for some odd reason the legend value is `NaN` or `null`.

nickofthyme (commented):
Change look good!

It kinda feels to me that we are in need of a some type of separate `legendValueFormatter` instead of just relying on the `tickFormat` for everything. This way we could let the user handle the percent and null cases directly.

Similar to how we have a `labelFormat` option just to format the axis label values apart from the tooltip values, see demo [here](https://elastic.github.io/elastic-charts/storybook/?path=/story/axes--label-formatting&globals=toggles.showHeader:true;toggles.showChartTitle:false;toggles.showChartDescription:false;toggles.showChartBoundary:false;theme:light).

No changes required, just commenting for future discussion.

e2e/screenshots/legend_stories.test.ts-snapshots/legend-stories/legend-tabular-data/should-correctly-display-null-values-dataset-dataset-chrome-linux.png
nickofthyme, line 1: Does this test case still exist?

Playwright does not automatically cull/prune unused images.

mbondyra: good call, removed!

mbondyra:
I am gonna go ahead and merge it so I can start working on upgrading it in Kibana

PR #2502, opened 2024-08-07
fix(flame): failure when accessing `visualViewport.visualViewport` from within `iFrame`

nickofthyme:
## Summary

Fixes issue where the `Flame` chart was attempting to access [`window.visualViewport`](https://developer.mozilla.org/en-US/docs/Web/API/Window/visualViewport) from within an iframe.

## Details

The `Flame` chart was attempting to access the [`window.visualViewport`](https://developer.mozilla.org/en-US/docs/Web/API/Window/visualViewport) of the top most `window` from an iframe. However this triggered a `DOMException` error as the `visualViewport` property is forbidden expect when accessed by the top `window` context, for cross-origins.

```
DOMException: Failed to read a named property 'visualViewport' from 'Window': Blocked a frame with origin "https://my.kibana.com" from accessing a cross-origin frame.
```

We were only updating this value every `100ms` to capture the latest `pinchZoomScale`. Since this is not accessible from an iframe I elected to simply disable this update loop when in the iframe and access the value in place directly from `window.visualViewport?.scale`;

Now renders fine within iframe 👇🏼 

Zight Recording 2024-08-07 at 01 21 10 PM

## Issues

Related to https://github.com/elastic/kibana/issues/188288

### Checklist

### DIFF
--- a/packages/charts/src/chart_types/flame_chart/flame_chart.tsx
+++ b/packages/charts/src/chart_types/flame_chart/flame_chart.tsx
@@ -63,10 +63,14 @@ const NODE_TWEEN_DURATION_MS = 500;
 const unitRowPitch = (position: Float32Array) => (position.length >= 4 ? (position[1] ?? 0) - (position[3] ?? 0) : 1);
 const initialPixelRowPitch = () => 16;
 const specValueFormatter = (d: number) => d; // fixme use the formatter from the spec
-const browserRootWindow = () => {
-  let rootWindow = window; // we might be in an iframe, and visualViewport.scale is toplevel only
+
+// Not cached to avoid holding a reference to the `window` object
+const getVisualViewportScale = (): number => {
+  let rootWindow = window;
   while (window.parent && window.parent.window !== rootWindow) rootWindow = rootWindow.parent.window;
-  return rootWindow;
+
+  // we might be in an iframe, and `visualViewport` is toplevel only
+  return window !== rootWindow ? 1 : rootWindow.visualViewport?.scale ?? 1;
 };
 
 const columnToRowPositions = ({ position1, size1 }: FlameSpec['columnarData'], i: number) => ({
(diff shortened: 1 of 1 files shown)


packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -259,7 +254,7 @@ class FlameComponent extends React.Component<FlameProps> {
 
     // browser pinch zoom handling
     this.pinchZoomSetInterval = NaN;
-    this.pinchZoomScale = browserRootWindow().visualViewport?.scale ?? 1;
+    this.pinchZoomScale = 1; // we set this to the accurate value next
```
dej611, line 257: If I get this right the problem arise only for cross-origin `iframe`, not all `iframe` scenarios.
What about wrapping the `browserRootWindow().visualViewport` call into a `try/catch` function wrapper and return `1` as fallback?

nickofthyme: OK I changed it to test the access and apply the interval if not forbidden. See 707a407

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -370,10 +382,15 @@ class FlameComponent extends React.Component<FlameProps> {
...
+  /**
+   * Setup interval to update pinch zoom scale factor
+   *
+   * Note: method accesses `window.visualViewport`
+   */
   private setupViewportScaleChangeListener = () => {
     window.clearInterval(this.pinchZoomSetInterval);
     this.pinchZoomSetInterval = window.setInterval(() => {
-      const pinchZoomScale = browserRootWindow().visualViewport?.scale ?? 1; // not cached, to avoid holding a reference to a `window` object
+      const pinchZoomScale = browserRootWindow().visualViewport?.scale ?? 1;
```
dej611, line 393: shouldn't this be wrapper as well?

nickofthyme: Sure, I put the check at the call site but I guess it makes more sense to add the check in the setup method.

See bf0bd1a

dej611 (approved):

PR #924, opened 2020-11-30
refactor: add manual sorting for tooltip, legend and rendering

markov00:
## Summary

fixes #303

Adds a way to sort tooltip, legend and rendered series via chart specification:

The API is currently hidden and not exposed to the final consumer to allow us cleaning this feature once we have found a better type for the SeriesIdentifier.
Although the API is hidden, you can ignore ts errors and using it at your own risk using the following hidden API and types:

```ts
export interface SettingsSpec extends Spec {
  /**
   * A compare function or an object of compare functions to sort
   * series in a different part of the chart like tooltip, legend and
   * the rendering order on the screen. To assign the same compare function.
   *  @defaultValue the series are sorted in order of appearance in the chart configuration
   */
  sortSeriesBy?: SeriesCompareFn | SortSeriesByConfig;
}


/**
 * An object of compare functions to sort
 * series in different part of the chart like tooltip, legend and rendering order.
 */
export interface SortSeriesByConfig {
  /**
   * A SeriesSortFn to sort the legend values (top-bottom)
   * It has precedence over the general one
   */
  legend?: SeriesCompareFn;
  /**
   * A SeriesSortFn to sort tooltip values (top-bottom)
   * It has precedence over the general one
   */
  tooltip?: SeriesCompareFn;
  /**
   * A SeriesSortFn to sort the rendering order of series.
   * Left/right for cluster, bottom-up for stacked.
   * It has precedence over the general one
   * Currently available only on XY charts
   */
  rendering?: SeriesCompareFn;
  /**
   * The default SeriesSortFn in case no other specific sorting fn are used.
   * The rendering sorting is applied only to XY charts at the moment
   */
  default?: SeriesCompareFn;
}

export type SeriesCompareFn = (siA: SeriesIdentifier, siB: SeriesIdentifier) => number;

```

An example of usage is:

```tsx
...
// sort ascending by group and yAccessor (a,y1)(a,y2)(b,y1)(b,y2)
function sortSeriesFn(a: SeriesIdentifier, b: SeriesIdentifier) {
  const { splitAccessors: aSplit, yAccessor: aYAccessor, specId: specIdA } = a as XYChartSeriesIdentifier;
  const { splitAccessors: bSplit, yAccessor: bYAccessor, specId: specIdB } = b as XYChartSeriesIdentifier;
  const aGroup = `${aSplit.get('g') ?? ''}`;
  const bGroup = `${bSplit.get('g') ?? ''}`;

  // sort by group descending within stacked bars
[...]

PR #2524, opened 2024-09-13
feat(xy): render sorting

nickofthyme:
## Summary

Adds render sorting for xy charts. See demo story [here](https://ech-e2e-ci--pr-2524-xnw4abb2.web.app/storybook/?path=/story/bar-chart--rendering-sort).

Zight Recording 2024-09-12 at 07 02 47 PM

## Details

Render order
- Default sort based on data order
- Override default sort with `Settings.renderSort` function

Legend items order
- Default to opposite of render order
- Override default using `Settings.legendSort`, original array is based on render order.

Tooltip items order
- Default to match legend order
- Override default using `Tooltip.sort`, original array is based on render order.

## Breaking changes

The way mixed stacked/nonstacked series are colored now is different from the previous behaviour. Now we color them not by their insert index but by the way we display them in the rendering: from the left to right, bottom top, stacked, nonstacked. This align correctly also the legend colors by default. This does **not** affect colors assigned via a `SeriesColorAccessor`.

Screenshot 2024-10-08 at 09 48 24

## Issues

Related to https://github.com/elastic/kibana/issues/191988

### Checklist


nickofthyme:
buildkite update screenshots

nickofthyme:
buildkite update screenshots

nickofthyme:
buildkite test this

nickofthyme:
buildkite update screenshots

nickofthyme:
buildkite test this

packages/charts/src/chart_types/xy_chart/legend/legend.ts
```diff
@@ -183,16 +183,24 @@ export function computeLegend(
...
+      .sort((a, b) =>
+        a.seriesIdentifiers[0] && b.seriesIdentifiers[0]
+          ? baseLegendSortFn(a.seriesIdentifiers[0], b.seriesIdentifiers[0])
+          : 0,
+      )
+      .sort((a, b) =>
+        settingsSpec.legendSort && a.seriesIdentifiers[0] && b.seriesIdentifiers[0]
+          ? settingsSpec.legendSort(a.seriesIdentifiers[0], b.seriesIdentifiers[0])
+          : 0,
+      ),
```
markov00, line 203: Why sorting two times one after the other? 
Why not  doing that in a single sort function as before?

packages/charts/src/chart_types/xy_chart/state/selectors/get_tooltip_values_highlighted_geoms.ts
```diff
@@ -192,20 +192,28 @@ function getTooltipAndHighlightFromValue(
...
+    .sort((a, b) => {
+      return baseTooltipSortFn(a.seriesIdentifier, b.seriesIdentifier);
+    })
+    .sort(
+      tooltipSortFn
+        ? (a, b) => {
+            return tooltipSortFn(a.seriesIdentifier, b.seriesIdentifier);
+          }
+        : defaultSeriesSort,
+    );
```
markov00, line 212: Same as above, why sorting two times?

packages/charts/src/chart_types/xy_chart/state/utils/utils.test.ts
```diff
@@ -136,15 +136,15 @@ describe('Chart State utils', () => {
     );
     expect(domains.yDomains).toEqual([
       MockYDomain.fromScaleType(ScaleType.Log, {
-        domain: [0, 5],
-        groupId: 'group1',
+        domain: [0, 9],
+        groupId: 'group2',
```
markov00, line 140: Do you think this order change has any implications?

packages/charts/src/chart_types/xy_chart/utils/series.ts
```diff
@@ -431,10 +435,16 @@ export function getDataSeriesFromSpecs(
...
+  const dataSeries = globalDataSeries
+    .map((d, i) => ({
+      ...d,
+      insertIndex: i, // capture original insert order for coloring
+    }))
+    .toSorted(seriesSort)
+    .map((d, i) => ({
+      ...d,
+      sortOrder: i,
+    }));
```
markov00, line 447: 🙈

packages/charts/src/chart_types/xy_chart/utils/series.ts
```diff
@@ -587,9 +597,8 @@ export function getSeriesColors(
 ): Map<SeriesKey, Color> {
   const seriesColorMap = new Map<SeriesKey, Color>();
   let counter = 0;
-  const sortedDataSeries = [...dataSeries].sort((a, b) => a.insertIndex - b.insertIndex);
   groupBy(
-    sortedDataSeries,
+    dataSeries.toSorted((a, b) => a.insertIndex - b.insertIndex),
```
markov00, line 601: So in this case why the color are now applied by insertIndex instead of the sortIndex?

storybook/stories/bar/10a_rendering_sort.story.tsx
```diff
@@ -0,0 +1,118 @@
...
+        tickFormat={(d) => moment(d).format('lll')}
+        style={{
+          tickLine: { size: 4, padding: 4, visible: true },
+          tickLabel: {
+            alignment: { horizontal: Position.Left, vertical: Position.Bottom },
+            padding: 0,
+            offset: { x: 0, y: 0 },
+          },
+        }}
+        timeAxisLayerCount={3}
```
markov00, line 93:
```suggestion
        timeAxisLayerCount={2}
```

packages/charts/src/chart_types/xy_chart/utils/series.ts
```diff
@@ -431,10 +435,16 @@ export function getDataSeriesFromSpecs(
...
         );
 
-  const dataSeries = globalDataSeries.map((d, i) => ({
-    ...d,
-    insertIndex: i,
-  }));
+  const dataSeries = globalDataSeries
+    .map((d, i) => ({
+      ...d,
+      insertIndex: i, // capture original insert order for coloring
```
markov00, line 441: Are we sure we want that?

packages/charts/src/chart_types/xy_chart/utils/series.ts
nickofthyme: This was applied previously [here](https://github.com/elastic/elastic-charts/pull/924/files?file-filters%5B%5D=.md&file-filters%5B%5D=.snap&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx&show-viewed-files=true#diff-2a2af4b892a573eb602ef0e136a23d3bb8396d6ef7cd92ea3938ebe02aa2da8dR584-R595) when the render sorting was first added to maintain the assigned colors when the render order changes.

I don't know that this is really necessary as the colors can be manually assigned.

packages/charts/src/chart_types/xy_chart/legend/legend.ts
nickofthyme: I was thinking it would be best if the original order starts with the default in case a rendering order was provided but I think now just using one or the other is fine. I'll fix it.

packages/charts/src/chart_types/xy_chart/state/utils/utils.test.ts
nickofthyme: No I don't think so, even if it did there would surly be a regression somewhere in the e2e tests.

packages/charts/src/chart_types/xy_chart/legend/legend.ts
markov00: I believe we should just sorting once for all. If you have a specific legendSort then use that, if not use the default.
Your applied sort strategy should be a "complete" strategy, and should not rely on the original sorting of the objects.

packages/charts/src/chart_types/xy_chart/utils/series.ts
markov00: Let's discuss offline, but we should consider this:
- we need to use an index to correctly sort the dataseries
- if no sorting is applied, that index refers to the 'insertion index'
- if sorting is applied, that index refers to the 'after sort index'
IMHO we can get rid of one of them, the index should always reflect the sorting

nickofthyme: Removed in fdce037

packages/charts/src/chart_types/xy_chart/utils/series.ts
nickofthyme: Removed in fdce037. Now we just color series by sort order not insert order.

packages/charts/src/chart_types/xy_chart/state/selectors/get_tooltip_values_highlighted_geoms.ts
nickofthyme: Removed in 86a3836, now only sorts once.

packages/charts/src/chart_types/xy_chart/legend/legend.ts
nickofthyme: Removed in 86a3836, now only sorts once.

nickofthyme:
buildkite update screenshots

nickofthyme:
buildkite update screenshots

nickofthyme:
buildkite update screenshots

nickofthyme:
buildkite test this

PR #2534, opened 2024-09-27
fix(ChartStatus): render complete if same parent size is dispatched

markov00:
## Summary

A [PR in Kibana](https://github.com/elastic/kibana/pull/193798) identified a bug in our current code (and probably also in the ResizeObserver API). This PR showed that the ResizeObserver can dispatch the same witdh/height/left/top values under certain circumstances (not really clear from the PR).
In code we don't have a check for this case thus a duplicated `updateParentDimension` action can be dispatched causing the chart to remain in a wrong state: the `renderedComplete:false` but the chart still rendered on the screen. The configured chart renderer doesn't get triggered because the parentDimension in effect is not changed, thus the render count doesn't increase and no dispatch to `onRenderComplete`is issued, causing a wrong ChartStatus to be reported.

The fix just trigger a `UPDATE_PARENT_DIMENSION` state change only if the dimensions are different.


nickofthyme (approved):
Nice! Code changes LGTM, tested on Storybook with no issues.

Issue #2543, opened 2024-10-07
[Metric Chart] New font-sizes handling might cause existing Metric charts to look off


gvnmagni:
In `8.15` a new handling of font-sizes has been introduced, moving it from being `width-related` to `height-related`. This was motivated by a series of considerations that are still valid but it could result in weird appearances for panels that are taller than large, or with very large numbers (lot of digits).

Example of problematic Metrics
![Image](https://github.com/user-attachments/assets/6da0217e-49cf-497f-9022-7626087bb74d)

In order to solve this, I would propose to introduce a way to take in consideration also the width of panel when defining the font-size of values so that value don't get cut off.

I would see two approaches to implement this solution:

1. Font-sizes are still determined according to break-points that take in consideration height of the panel, if the numbers are cut off their font-size automatically jumps to a smaller breakpoints, until it fit.

2. Same but instead of jumping to a smaller breakpoint, we automatically swtich to `FIT` function and get back to `DEFAULT` function as soon as the panel gets back to be big enough

Both options have pros and cons in terms of final look & feel but both will solve the original issue.

Thanks to @MichaelMarcialis and @markov00 and @nickofthyme for the conversation around this topic.

gvnmagni:
Expected result of two proposed options:

![Image](https://github.com/user-attachments/assets/dbb9f1a8-3627-49a7-8adb-1d85a569b9cc)

muthu-mps:
@gvnmagni, @markov00  - The integrations upgraded to 8.15.0 needs to get upgraded to the latest version to get this issue resolved without any changes to the metric visualisation?

nickofthyme:
Correct, changes from #2548 will be in [`@elastic/charts@v68.0.1`](https://github.com/elastic/elastic-charts/releases/tag/v68.0.1) and targeted to be in `8.16.1`/`8.17.0` (https://github.com/elastic/kibana/pull/197473).

markov00:
probably also in `8.16.0`

PR #2548, opened 2024-10-14
fix(Metric): improve default font-sizing

markov00:
## Summary

This feature adds the snapping option for metric value font size. This behaviour will apply the breaking point font size that is big enough to completely render the `value` without any truncation/overflow.

fix https://github.com/elastic/elastic-charts/issues/2543

https://github.com/user-attachments/assets/366e488d-2a68-4bbd-b9f8-12004b3d6314

## Details

I've refactored a bit the metric text code, extracting the various functions and moving the processing/measurement logic in separate files.

### Checklist


markov00:
buildkite update screenshots

markov00:
buildkite update screenshots

packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx
```diff
@@ -0,0 +1,304 @@
...
+  titleMaxLines: number;
+  subtitleMaxLines: number;
+  title: boolean;
+  subtitle: boolean;
+  extra: boolean;
+};
+/**
+ * synced with scss variables
+ */
+const PROGRESS_BAR_WIDTH = 10;
```
nickofthyme, line 40:
```suggestion
// synced with scss variables
const PROGRESS_BAR_WIDTH = 10;
```

This was a general comment not a JSdocs for this value

packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx
nickofthyme, line 1: The `Custom` mode appears broken, does not respect the defined font size, no different than the `Fit` mode.

Zight Recording 2024-10-21 at 08 52 47 PM

nickofthyme, line 1: Should there be an option to sync all font sizes across the panels? Right now it's just assumed that they are homogenous but this may not always be desirable. I think in the Lens application this case would be rare to have a large different in value lengths so more of a general/future question.

packages/charts/src/chart_types/bullet_graph/renderer/canvas/index.tsx
nickofthyme, line 1: I noticed the visibility logic for `Default` and `Custom` seems to still use the original font sizing not the limiting one, causing overly eager hiding of element. Mainly noticeable when the limiting font is much smaller than the original height-based font size.

#### This PR storybook

Zight Recording 2024-10-21 at 08 36 55 PM

> Notice the inconsistent hiding of the subtitles and secondary values.

#### Current storybook

Zight Recording 2024-10-21 at 08 38 55 PM

nickofthyme (commented):
Really nice code structure improvements!

Overall, I like the new behaviors but I had a few questions below. I think we could merge this with the `Custom` fixes and save the others for later improvements.

packages/charts/src/chart_types/bullet_graph/renderer/canvas/index.tsx
markov00: Not sure about the custom (I need to check it), but the default method is currently reducing the fontsize to fit as much as possible the long array list in the first metric. This reduces the fontsize to the minimum for all the panels at once. The hiding logic instead follows a different pattern, that we can probably adjust to handle that font size better.

packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx
markov00: definitely a future step. We first need to expose the possibility to select a custom font size that is based on our breakpoints, and then probably we need to find a good way to expose the actual minimum font size of each panel.
Definitely not perfect and easy to achieve (as each panel in dashboard is independent, and we will have something like a callback on each fontsize change, or a preemptive calculation that needs to be confirmed or adjusted before rendering.

markov00: done in [c85c304](https://github.com/elastic/elastic-charts/pull/2548/commits/c85c30485ea6d246dafe418188d20f6cc3f50bd5)

markov00:
buildkite update screenshots

PR #2571, opened 2024-12-02
fix(xy): compute per series and global minPointsDistance

markov00:
## Summary

This PR improves the logic behind the `auto` visibility of points in Lines and Area charts.
Instead of using the bandwidth or the minInterval, it uses the actual minimum distance between points on each series.
If there are multiple series it uses the minimum of all the series to avoid having points on a line and not on the other.

## Details

A storybook example was changed to introduce a different data cadence.


markov00:
buildkite update screenshots

nickofthyme (approved):
These changes look to be a good improvement. Hopefully we are done with changing the iso point behavior for a while 😅.

PR #2573, opened 2024-12-04
perf: fix unnecessary re-render

markov00:
## Summary

The `Chart` component `render` function is called everytime something change in the redux store (for example every time you move the mouse over the chart).
This is well visible if you add a  console.log in the render method of the `Chart` component and move the mouse over the chart.

This PR  avoids that rendering by moving some store-subscribed logic into re-selectors.


nickofthyme (approved):
Excellent find! :shipit:

PR #2580, opened 2025-01-11
feat: remove error boundary in favor of external boundary

### LINKED ISSUE #2579: Remove error boundary

Currently it has come to our attention that the current error boundary in charts is not really used or needed.

Chart consumers would like a way to catch errors that are thrown inside the chart and it's child components.

Currently we only catch so-called **_graceful errors_** and render the chart, or catch other errors and render a blank canvas.

**Additional context**

We should verify this works to render the outer error boundary in place of the chart, meaning the chart should not be rendered at all. Currently, it renders side-by-side...

Image

nickofthyme:
## Summary

Removed internal chart error boundary. Now users should wrap the `<Chart>` component if they expect it may throw. See example below using `EuiErrorBoundary`.

Zight Recording 2025-01-10 at 06 48 54 PM

## Details

This PR removes the error boundary that wrapped internal chart specs. This boundary was not blocking errors thrown in charts from bubbling up, even for non-graceful errors. In this case it would leave a black white screen.

> [!CAUTION]
> The same limitations apply to all error boundaries as mention in the react docs [here](https://legacy.reactjs.org/docs/error-boundaries.html#introducing-error-boundaries). Error boundaries do not catch errors if any of the follow apply:
> 
> - Event handlers ([learn more](https://legacy.reactjs.org/docs/error-boundaries.html#how-about-event-handlers))
> - Asynchronous code (e.g. `setTimeout` or `requestAnimationFrame` callbacks)
> - Server side rendering
> - Errors thrown in the error boundary itself (rather than its `children`)

## Issues

Closes #2579

### Checklist


nickofthyme:
buildkite update screenshots

storybook/stories/test_cases/14_error_boundary.story.tsx
markov00, line 1: can you add a test with a thrown error? thanks

nickofthyme: Added test in dac798fd84e18503a5529e9f72745852cef2d526

markov00 (approved):

nickofthyme:
buildkite update screenshots

PR #2582, opened 2025-01-13
feat(xy)!: change `timeAxisLayerCount` the default from `0` to `2`

walterra:
## Summary

`timeAxisLayerCount: 2` is now the default to render multiple layers for time based axes. This feature [was originally added in 2021 ](https://github.com/elastic/elastic-charts/issues/1310) and now just a small set of consumers are using the legacy time axis.

### BREAKING CHANGE
The `timeAxisLayerCount` options now defaults to `2` instead of `0`.

## Details

- This includes some tricky changes: Because `timeAxisLayerCount` defaulted to `0`, code in some places assumed that a change to `2` was only done by users when the x axis was then time based. Thus changing the default to `2` changed the calculated padding for all axis types.
- On top of that this surfaced a limitation of this setting: When `timeAxisLayerCount` is set to `2`, there was no check for axes if they were actually time based. This means this also affected vertical axes so the left and right margin would change. I think that's the reason why we see so many baseline changes in the playwright e2e image assertions.

## Issues

This PR is part of https://github.com/elastic/kibana/issues/193683

### Checklist


storybook/stories/components/tooltip/10_cartesian_chart.story.tsx
```diff
@@ -95,16 +95,7 @@ export const Example: ChartsStory = (_, { title, description }) => {
         gridLine={{
           visible: true,
         }}
-        style={{
-          tickLine: { size: 0.0001, padding: 4 },
```
walterra, line 99: In `MULTILAYER_TIME_AXIS_STYLE` which I copied from Kibana `size` is just `0`. Should we reuse the `0.0001` used here instead?

walterra: Discussed with @markov00 and checked baseline change, it can stay `0.`

storybook/stories/test_cases/21_domain_edges.story.tsx
```diff
@@ -81,24 +81,7 @@ export const Example: ChartsStory = (_, { title, description }) => {
             },
```
walterra, line 81: This still causes a baseline change, right padding in the lowest chart:

CleanShot 2025-01-20 at 15 24 09@2x

walterra: This reason for this regression: We now have to check for both `timeAxisLayerCount > 0` and if the scale type is `time`, however, the lowest chart uses `ordinal`.

walterra: If I change the `xScaleType` of the `BarSeries` of the last chart to `ScaleType.Time` it looks correct:

CleanShot 2025-01-20 at 16 01 07@2x

docs/docs/api/specs/xy/axis.mdx
```diff
@@ -132,7 +132,7 @@ Show duplicated ticks
 
 ### `timeAxisLayerCount`
 - Type: <code>timeAxisLayerCount: number</code>
-- Default: `0`
+- Default: `2`
```
walterra, line 135: Posting this here for reference to start a thread, because the affected example file has not been touched.

There's a baseline change for stories with custom markers. Previously the axis ticks and markers would render overlapping, now they render with additional offset. Is the new version ok? I'm not sure the original overlapping was intentional or a bug?

CleanShot 2025-01-20 at 16 44 45@2x

storybook/stories/test_cases/21_domain_edges.story.tsx
walterra: For `test-cases-stories/legend-last-value-should-be-aligned-across-areas-and-bars/custom-1ms`, the lowest char is missing the last empty bucket:

CleanShot 2025-01-24 at 08 56 13@2x

packages/charts/src/chart_types/xy_chart/axes/axes_sizes.ts
```diff
@@ -23,11 +31,14 @@ const getAxisSizeForLabel = (
...
   { maxLabelBboxWidth = 0, maxLabelBboxHeight = 0 }: TickLabelBounds,
   smSpec: SmallMultiplesSpec | null,
+  scaleConfigs: ScaleConfigs,
+  { rotation }: SettingsSpec,
 ) => {
   const { tickLine, axisTitle, axisPanelTitle, tickLabel } = axesStyles.get(axisSpec.id) ?? sharedAxesStyles;
   const horizontal = isHorizontalAxis(axisSpec.position);
   const maxLabelBoxGirth = horizontal ? maxLabelBboxHeight : maxLabelBboxWidth;
-  const allLayersGirth = getAllAxisLayersGirth(axisSpec.timeAxisLayerCount, maxLabelBoxGirth, horizontal);
+  const isMultilayerTimeAxis = isMultilayerTimeAxisFn({ axisSpec, scaleConfigs, rotation });
```
walterra, line 40: This could be named `const multilayerTimeAxis = isMultilayerTimeAxis(...)` instead, wdyt?

markov00: Yes, agree

packages/charts/src/chart_types/xy_chart/state/selectors/get_axis_styles.ts
```diff
@@ -6,20 +6,54 @@
...
+
+const MULTILAYER_TIME_AXIS_STYLE: RecursivePartial<AxisStyle> = {
+  tickLabel: {
+    visible: true,
+    padding: 0,
+    rotation: 0,
+    alignment: {
+      vertical: Position.Bottom,
+      horizontal: Position.Left,
+    },
```
markov00, line 28: For some reason multiple tests shows centered aligned bars instead of left aligned bars. The first example is here https://ech-e2e-ci--pr-2582-n5t2yfqp.web.app/e2e-report/#?testId=4a59eca736035d93205f-0a486745228f4b4ff460

Screenshot 2025-01-27 at 12 13 34

Do you mind checking and see why the bars are centered?

packages/charts/src/chart_types/xy_chart/axes/axes_sizes.ts
```diff
@@ -23,11 +31,14 @@ const getAxisSizeForLabel = (
   axesStyles: Map<AxisId, AxisStyle | null>,
   { maxLabelBboxWidth = 0, maxLabelBboxHeight = 0 }: TickLabelBounds,
   smSpec: SmallMultiplesSpec | null,
+  scaleConfigs: ScaleConfigs,
+  { rotation }: SettingsSpec,
```
markov00, line 35: Instead of passind down these two complex properties, we can move the `isMultilayerTimeAxisFn` call outside this method and pass just the `isMultilayerTimeAxis` boolean to this function

packages/charts/src/chart_types/xy_chart/renderer/canvas/axes/tick.ts
```diff
@@ -21,14 +21,18 @@ export function renderTick(
   ctx: CanvasRenderingContext2D,
   { position, domainClampedPosition: tickPosition, layer, detailedLayer }: AxisTick,
   {
-    axisSpec: { position: axisPosition, timeAxisLayerCount },
+    axisSpec,
     size: { width, height },
     axisStyle: { tickLine, gridLine },
     layerGirth,
+    scaleConfigs,
+    settingsSpec: { rotation },
```
markov00, line 29: same comment as in the `axes_sizes.ts` file. we can directly pass the `isMultilayerTimeAxis` boolean and compute it before calling the renderTick method

markov00: we can also consider adding this information to the `AxisTick` property

packages/charts/src/chart_types/xy_chart/renderer/canvas/axes/tick_label.ts
```diff
@@ -47,7 +48,8 @@ export function renderTickLabel(
     }
   }
 
-  const tickOnTheSide = timeAxisLayerCount > 0 && Number.isFinite(tick.layer);
+  const isMultilayerTimeAxis = isMultilayerTimeAxisFn({ axisSpec, scaleConfigs, rotation });
```
markov00, line 51: same comment, let's find a place where we can call this and store it without the need to pass down these info.

packages/charts/src/chart_types/xy_chart/renderer/canvas/panels/title.ts
```diff
@@ -54,7 +65,8 @@ export function renderTitle(
   const font: TextFont = { ...titleFontDefaults, ...axisTitleToUse, textColor: axisTitleToUse.fill };
   const tickDimension = shouldShowTicks(tickLine, hideAxis) ? tickLine.size + tickLine.padding : 0;
   const maxLabelBoxGirth = horizontal ? maxLabelBboxHeight : maxLabelBboxWidth;
-  const allLayersGirth = getAllAxisLayersGirth(timeAxisLayerCount, maxLabelBoxGirth, horizontal);
+  const isMultilayerTimeAxis = isMultilayerTimeAxisFn({ axisSpec, scaleConfigs, rotation: chartRotation });
```
markov00, line 68: same

packages/charts/src/chart_types/xy_chart/axes/axes_sizes.ts
walterra: updated in b07adcf8ea.

packages/charts/src/chart_types/xy_chart/axes/axes_sizes.ts
walterra: Updated this one to have `multilayerTimeAxis: boolean` as an arg and calculate it outside (fc7edea906).

packages/charts/src/chart_types/xy_chart/renderer/canvas/axes/tick.ts
walterra: Added `multilayerTimeAxis` to `AxisTick` in fc7edea906.

packages/charts/src/chart_types/xy_chart/renderer/canvas/panels/title.ts
walterra: I added it to `AxisProps` (which `TitleProps` extends) so it doesn't need to be recalculated here. (fc7edea906)

walterra:
@markov00 I updated code related to `multilayerTimeAxis` so it's not recalculated that often. I pushed the calculation as far up as possible to the level of selectors, or where we iterate over all axes and need to calculate it for each axis.

storybook/stories/test_cases/21_domain_edges.story.tsx
walterra: With the update (c6fdc74a69) to use `enableHistogramMode:true` for time based x-axis the axis now render the same across all three charts:

CleanShot 2025-02-20 at 09 16 04@2x

packages/charts/src/chart_types/xy_chart/specs/bar_series.ts
```diff
@@ -44,7 +44,14 @@ export const BarSeries = function <D extends BaseDatum = Datum>(
...
   const { defaults, overrides } = buildProps;
-  useSpecFactory<BarSeriesSpec<D>>({ ...defaults, ...stripUndefined(props), ...overrides });
+  const constraints: Pick<typeof props, 'enableHistogramMode'> = {};
+
+  // enable histogram mode for time scales to align with the multi-layer time axis.
+  if (props.xScaleType === ScaleType.Time) {
+    constraints.enableHistogramMode = true;
+  }
+
+  useSpecFactory<BarSeriesSpec<D>>({ ...defaults, ...stripUndefined(props), ...overrides, ...constraints });
```
walterra, line 54: @markov00 Is this the right place to update `enableHistogramMode` internally? I saw the goal chart does something similar. Is this sufficient or do we also need to check for rotation (e.g. to be 0 or 180) here?

markov00: I think is fine checking just this, the rotation is not necessary I believe

walterra:
buildkite update screenshots

docs/docs/api/specs/xy/axis.mdx
nickofthyme: I think this change is good, not sure why it previously was not adding the padding. But if a user want to make space for the annotation, either the icon or `markerBody`, they need to add the required padding manually.

packages/charts/src/chart_types/xy_chart/specs/bar_series.ts
nickofthyme: @markov00 do you think it's needed to add a `Logger.warn` like we do in the `Goal` spec below?

https://github.com/elastic/elastic-charts/blob/ac16be0a5d3e12d35562bcdc85529a547e93385d/packages/charts/src/chart_types/goal_chart/specs/index.ts#L115-L124

I'd say probably not but we are unknowingly overriding the spec when someone has `enableHistogramMode` set to `false`, even if it's not a major impact to their chart.

nickofthyme (approved):
Overall LGTM. I would just ask that we swap out the manual object creation with the equivalent mocks in the unit tests.

docs/docs/api/specs/xy/axis.mdx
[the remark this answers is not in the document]
markov00: I believe we can fix this in a follow up PR, please open an issue for that,thanks

packages/charts/src/chart_types/xy_chart/specs/bar_series.ts
markov00: I don't think so, I believe we can also remove the "histogramMode" in the near future since this is now the default behavhiour when using a specific configuration.
The warning is also just used in dev mode so not much you can do most of the time probably

markov00 (approved):

packages/charts/src/chart_types/xy_chart/utils/dimensions.test.ts
```diff
@@ -82,7 +97,16 @@ describe('Computed chart dimensions', () => {
...
-    const { chartDimensions } = computeChartDimensions(parentDim, chartTheme, axisDims, axisStyles, axisSpecs, null);
+    const { chartDimensions } = computeChartDimensions(
+      parentDim,
+      chartTheme,
+      axisDims,
+      axisStyles,
+      axisSpecs,
+      null,
+      getScaleConfigs(),
+      getSettingsSpec(),
```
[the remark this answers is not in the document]
walterra: Updated in 4f9927da1e.

docs/docs/api/specs/xy/axis.mdx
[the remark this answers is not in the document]
walterra: Created an issue here: https://github.com/elastic/elastic-charts/issues/2625

packages/charts/src/chart_types/xy_chart/state/selectors/visible_ticks.ts
```diff
@@ -266,7 +285,7 @@ function getVisibleTickSets(
               detailedLayer,
               ticks,
               labelFormatter,
-              isMultilayerTimeAxis,
+              isXAxis && multilayerTimeAxis,
```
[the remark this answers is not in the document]
walterra: Thanks for catching this one! In earlier commits `multilayerTimeAxis` didn't include that check so this was a leftover. Removed `isXAxis` in f161d881df.

nickofthyme (approved):
Thanks for the changes! LGTM

packages/charts/src/chart_types/xy_chart/utils/axis_utils.test.ts
```diff
@@ -1087,6 +1092,53 @@ describe('Axis computational utils', () => {
     expect(horizontalY).toBe(true);
   });
 
+  describe('isMultilayerTimeAxis', () => {
+    test('should return true if chartType is xy_axis, timeAxisLayerCount = 2, position is bottom, x axis type is time, rotation is 0', () => {
+      const multilayerTimeAxis = isMultilayerTimeAxis(
+        { chartType: 'xy_axis', timeAxisLayerCount: 2, position: 'bottom' } as unknown as AxisSpec,
```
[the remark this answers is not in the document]
nickofthyme: Up to you, if you think it's worth it and could be used elsewhere, otherwise just leave it as a one-off.

PR #2584, opened 2025-01-15
fix(style): align xy axis style to borealis

markov00:
## Summary

This PR aligns the axis borealis colors for Cartesian charts.

Only ~970 VRT changes 😅, sorry about that but this changes all the axis colors.

**Note for reviewers:**
Is easy to just take a look at the changes in the first commit and just ignore the VRT changes.


markov00:
buildkite update screenshots

packages/charts/src/utils/themes/base_colors.ts
nickofthyme, line 1: Looks fine to me, I think we should clean this up at some point when we can point to an eui package for these values. I don't think that's possible rn.

PS I didn't check any of the screenshots 🤞🏼

nickofthyme (approved):
😱 😱 😱

PR #2588, opened 2025-01-20
refactor(theme): reuse borealis token semantic

markov00:
## Summary

This PR refactor a bit the theme colors by following the same semantic applied by EUI.
This will simplify the refactoring when we are going to use directly the EUI theme library and use their colors.


markov00:
buildkite test this

mbondyra (approved):
Nice cleanup! 🙌🏼

PR #2591, opened 2025-01-21
style(heatmap): align colors to borealis theme

markov00:
## Summary

Aligning the Heatmap style to Borealis theme


markov00:
buildkite update screenshots

packages/charts/src/utils/themes/dark_theme.ts
```diff
@@ -406,8 +407,8 @@ export const DARK_THEME: Theme = {
         useGlobalMinFontSize: true,
       },
       border: {
-        strokeWidth: 1,
-        stroke: 'snow',
+        strokeWidth: 0,
+        stroke: DARK_BORDER_COLORS.borderBaseSubdued,
       },
```
nickofthyme, line 412: No cell border?

nickofthyme (approved):
Looks way less harsh that the before!

packages/charts/src/utils/themes/dark_theme.ts
markov00: No I removed it because it is just one more border around the cell that already have the grid lines

Issue #2121, opened 2023-08-02
[cartesian] improve gridlines z-index


markov00:
If different grays are used on axis line and grid lines, there can be overlaps that don't look nice as shown in the following example:
<img width="589" alt="Screenshot 2023-08-02 at 14 10 49" src="https://github.com/elastic/elastic-charts/assets/1421091/aaca18fd-ac74-43e6-b72c-e65a271309fa">

There could be a few possible solutions to this:
- allow using only one color for grid/axis lines
- render always the axis after rendering the grid but this will not solve the problem if the grid lines have different shades.

When different shades are used, then probably we need to sort the rendering order by color or by importance and render first light colors and then darker colors or more important elements ontop of less important elements.

Here is a demo PR to prove the point https://github.com/elastic/elastic-charts/pull/2122

PR #2606, opened 2025-01-29
fix(axis): remove multi-pass axis line rendering

markov00:
## Summary

To align the current theme design with Borealis (but also reflected in Amsterdam) I've removed the use of the `lumaSteps` when rendering the gridlines in a timeseries, this reduce the color difference between major and minor grid lines in times series as seen here
Screenshot 2025-01-30 at 11 44 27

Now every gridline has the same color, independently from their importance (beginning of day, week,month, year).

The PR also slightly improved the ticks rendering logic, by applying the color and stroke only one time per axis, instead of applying that per tick (improving a bit the rendering performance).

It finally also render the ticks before rendering the axis line, fixing a visual issue where the ticks create slight "cavities" into the axis line
Screenshot 2025-01-30 at 11 54 22

fix https://github.com/elastic/elastic-charts/issues/2121


markov00:
buildkite update screenshots

markov00:
buildkite update screenshots

packages/charts/src/chart_types/xy_chart/renderer/canvas/axes/index.ts
```diff
@@ -47,8 +47,10 @@ export function renderAxis(ctx: CanvasRenderingContext2D, props: AxisProps) {
...
   const showTicks = shouldShowTicks(axisStyle.tickLine, axisSpec.hide);
 
-  renderAxisLine(ctx, props); // render the axis line
-  if (!secondary && showTicks) ticks.forEach((tick) => renderTick(ctx, tick, props));
+  if (!secondary && showTicks) {
+    renderTicks(ctx, ticks, props);
+  }
   if (!secondary && axisStyle.tickLabel.visible)
     ticks.forEach((tick) => renderTickLabel(ctx, tick, showTicks, props, layerGirth ?? 0));
+  renderAxisLine(ctx, props); // render the axis line
```
mariairiartef, line 55: nit: I think we can remove this comment since the method name already explains what it does and the comment doesn't add any value

packages/charts/src/chart_types/xy_chart/renderer/canvas/axes/tick.ts
```diff
@@ -7,26 +7,42 @@
...
-import { HIDE_MINOR_TIME_GRID, HIERARCHICAL_GRID_WIDTH, OUTSIDE_RANGE_TOLERANCE } from '../../../utils/grid_lines';
+import { HIDE_MINOR_TIME_GRID, OUTSIDE_RANGE_TOLERANCE } from '../../../utils/grid_lines';
 import { renderMultiLine } from '../primitives/line';
 
 const BASELINE_CORRECTION = 2; // the bottom of the em is a bit higher than the bottom alignment; todo consider measuring
 
 /** @internal */
-export function renderTick(
-  ctx: CanvasRenderingContext2D,
+export function renderTick(ctx: CanvasRenderingContext2D, tick: AxisTick, axisProps: AxisProps) {
```
mariairiartef, line 21: qq: In which cases this method will be used?

mariairiartef (approved):

packages/charts/src/chart_types/xy_chart/renderer/canvas/axes/tick.ts
markov00: nowhere, you are right. I've deleted it here 62b6eb3a211f8d6cab69b0ff395cac2a862e2009

packages/charts/src/chart_types/xy_chart/renderer/canvas/axes/index.ts
markov00: you are right, removed here 1ee7d9e464b7f09469a810bf228495162bd8d08e

PR #2355, opened 2024-03-11
feat(legend): select legend statistic value

markov00:
## Summary

This PR paved the way for the new legend design.
The following tasks are implemented
- renamed the `showLegendExtra` to be a more generic `legendValues` (breaking change)
- the `legendValues` accept an array of `LegendValue` that is an ordered set of statistics to use (breaking change)
- all the possible legend aggregations (defined in https://github.com/elastic/kibana/issues/176583) are implemented
- the current legend implement still works and will only show the first `LegendValue` defined.

 ### BREAKING CHANGE
The `showLegendExtra` prop in the `Settings` is renamed to `legendValues` and accept an array of `LegendValue`s.

## Details

Current VRT changes are related to the change of the `depth` parameter in the `LegendItem` that by default is now zero indexed, instead of `1` as it was before for partition charts.

## Issues

Fix https://github.com/elastic/elastic-charts/issues/561
Initial part of https://github.com/elastic/kibana/issues/176583

### Checklist

- [x] The proper **chart type** label has been added (e.g. `:xy`, `:partition`)
- [x] The proper **feature** labels have been added (e.g. `:interactions`, `:axis`)
- [ ] All related issues have been linked (i.e. `closes #123`, `fixes #123`)
- [ ] New public API exports have been added to `packages/charts/src/index.ts`
- [ ] Unit tests have been added or updated to match the most common scenarios
- [ ] The proper documentation and/or storybook story has been added or updated
- [ ] The code has been checked for cross-browser compatibility (Chrome, Firefox, Safari, Edge)
- [ ] Visual changes have been tested with `light` and `dark` themes

PR #2607, opened 2025-01-30
chore(deps): minor redux related package upgrades

walterra:
## Summary

Minor updates to `redux` related packages, if possible to bring the versions in line with what is used in Kibana (https://github.com/elastic/kibana/pull/175017).

- Upgrades `@types/react-redux` from `7.1.1` to `7.1.20`.
- Upgrades `react-redux` from `7.1.0` to `7.2.8`.
- Upgrades `redux` from `4.0.4` to `4.2.1`.
- Upgrades `reselect` from `4.0.0` to `4.1.8`.


packages/charts/src/chart_types/xy_chart/state/chart_state.tsx
```diff
@@ -104,7 +104,7 @@ export class XYAxisChartState implements InternalChartState {
   }
 
   getLegendItemsLabels(globalState: GlobalChartState) {
-    return getLegendItemsLabelsSelector(globalState);
+    return getLegendItemsLabelsSelector(globalState as unknown as never);
```
nickofthyme, line 107: It looks like this selector was made useless in #2355 but never removed.

We are really averse to using `as` in charts. In this case, I think we can just remove this selector as it's doing nothing. The only reason the selector types are complaining is because the selector input is empty.

We should also delete this selector here...

https://github.com/elastic/elastic-charts/blob/bfb0d28f77794f7e35fa54381c05e21ef1ccaf58/packages/charts/src/chart_types/xy_chart/state/selectors/get_legend_items_labels.ts#L13

```suggestion
    return [];
```

packages/charts/src/chart_types/heatmap/state/selectors/on_brush_end_caller.ts
```diff
@@ -27,7 +27,7 @@ import { DragCheckProps, hasDragged } from '../../../../utils/events';
 export function createOnBrushEndCaller(): (state: GlobalChartState) => void {
   let prevProps: DragCheckProps | null = null;
   let selector: OutputSelector<
-    GlobalChartState,
+    [(state: GlobalChartState) => GlobalChartState],
```
nickofthyme, line 30: This should be an array of `Selectors` not a tuple of one `Selector`.

```suggestion
    Array<(state: GlobalChartState) => GlobalChartState>,
```

nickofthyme (approved):
LGTM, just 2 nits.

packages/charts/src/chart_types/xy_chart/state/chart_state.tsx
walterra: Thanks for the context, that's helpful! I took the opportunity and consolidated the available constants a bit (`EMPTY_...`) in [2b75cf9](https://github.com/elastic/elastic-charts/pull/2607/commits/2b75cf95270bdeafb941db6b2fbe41da5c5a4dd8).

walterra: Deleted the selector in 6aff1ed.

packages/charts/src/chart_types/heatmap/state/selectors/on_brush_end_caller.ts
walterra: Nice catch, fixed it in [0a2258e](https://github.com/elastic/elastic-charts/pull/2607/commits/0a2258e3d9a8151f3b41e4aca83b6892acf6d622).

walterra:
@nickofthyme this i ready for another look, tests passed again.

nickofthyme (approved):
Thanks! Looks good with the new changes.

Issue #2377, opened 2024-03-25
[deps] Upgrade to Redux v5


markov00:
We should try to follow the effort in Kibana to upgrade our Redux version from 4 to 5.

PR #2609, opened 2025-02-04
chore(deps): migrate to use redux-toolkit

walterra:
## Summary

Part of #2377

Updates state management to make use of redux toolkit's `configureStore` and `createSlice`.

## Details

This is part of the [effort to modernise the usage of redux](https://github.com/elastic/elastic-charts/issues/2377) in Elastic Charts. 

- The deprecated `createStore` has been replaced with `configureStore`.
- Instead of the legacy action and reducer style, this now makes use of `createSlice`. Some details on that:

Usually you could set up actions inline within `createSlice` like this (with the benefit of less boilerplate and inferred types):

```ts
const chartSlice = createSlice({
    name: 'chart',
    initialState: getInitialState(),
    reducers: {
        myAction(state, action) => ...
    },
}

export { myAction } = chartSlice.actions;
```

However, I ran into quite some problems with cyclic dependencies with this, if we'd export the plain actions from `chart_state.ts` like above. Luckily Redux still allows a pattern to define actions in separate files, which allowed me to solve the problem. It includes a bit more boilerplate, but on the other hand I was able to retain the original directory structure related to the existing actions. Compared to above, the alternative approach I followed looks like this:

```ts
// actions.ts
import { createAction } from '@reduxjs/toolkit';
export const myAction = createAction<boolean>('MY_ACTION');

// chart_state.ts
import { createSlice} from '@reduxjs/toolkit';
import { myAction } from './actions';

const handleChartActions = (builder: ActionReducerMapBuilder<ChartSliceState>) => {
  builder.addCase(myAction, (state, payload) => {
    // you're action logic goes here
    state.awesome = payload;
  });
};

const chartSlice = createSlice({
    name: 'chart',
    initialState: getInitialState(),
    reducers: {},
    extraReducers: (builder) => {
        handleChartActions(builder);
        ... // more handlers go here, let's us organise actions
  }
};
```

- There were still some cyclic dependency problems, that's why I split up some files and reorganized some imports in this PR, the updated imports make up the bulk of the changed files.
- Note Redux Toolkit's action logic makes use of `immer`. Instead of updating deeply nested state using nested spread operators, you can now treat state like a mutable object, even if the updates are immutable under hood. It simplifies code in a lot of cases, but there are some caveats (with arrays/collections, for example). Here's a helpful overview of some useful patterns on how to work with `immer`: https://immerjs.github.io/immer/update-patterns/
- Overall I followed the applicable steps in this guide: https://redux.js.org/usage/migrating-to-modern-redux
Note this PR doesn't touch React components using redux' `connect` yet, that's something for a follow up.
- By default, Redux Toolkit is more strict about non-serializable state. I set up the store to allow the existing state as is so we are not forced to refactor this right now. The store setup looks like this:

```ts
export const createChartStore = (chartId: string, title?: string, description?: string) => {
  const initialState = getInitialState(chartId, title, description);
  const chartSlice = createChartSlice(initialState);
  return configureStore({
    reducer: chartSlice.reducer,
    middleware: (getDefaultMiddleware) =>
      getDefaultMiddleware({
        // TODO https://github.com/elastic/elastic-charts/issues/2078
        serializableCheck: false,
      }),
  });
};
```

- The existing redux actions had custom interfaces, Redux Toolkit enforces the same style for all actions. Instead of custom arguments for an action, every action with a payload is expected to have one argument (which can be an options object instead of the previous multiple arguments). This makes up the bulk of the changes in the test files to call the actions in the updated format.
- One thing that took me some time to figure out: To identify selected items in tooltips, we used to compare objects by reference. This no longer works because of `immer`'s immutability. I fixed that by generating ids for the selected items which makes them comparable. (First I tried deep equality checks but that can be another can of worms with `immer` and possibliy non-serializable state.)

### Checklist


packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -1160,7 +1161,8 @@ class FlameComponent extends React.Component<FlameProps> {
         </div>
         <BasicTooltip
           canPinTooltip={canPinTooltip}
-          onPointerMove={() => ({ type: ON_POINTER_MOVE, position: { x: NaN, y: NaN }, time: NaN })}
+          // @ts-ignore redux-toolkit complains about this
+          onPointerMove={() => onPointerMove({ position: { x: NaN, y: NaN }, time: NaN })}
```
nickofthyme, line 1165:
```suggestion
          onPointerMove={() => onPointerMove({ position: { x: NaN, y: NaN }, time: NaN })}
```

I'm not sure why this was typed this way...

```ts
interface TooltipDispatchProps {
  onPointerMove: typeof onPointerMoveAction
  // ...
}
```

 but I think it should be...

```ts
interface TooltipDispatchProps {
  onPointerMove: (...args: Parameters<typeof onPointerMoveAction>) => void;
  // ...
}
```

The `tooltip.tsx` file does not require the return value of the action once passed through `dispatch`.

packages/charts/src/state/create_selector.ts
```diff
@@ -55,13 +68,17 @@ class GlobalSelectorCache {
     return {
       keySelector: GlobalSelectorCache.keySelector,
       cacheObject: this.getCacheObject(),
+      selectorCreator: createSelectorWithMaxSize100,
```
walterra, line 71: @markov00 I was able to restore performance by adding `createSelector` being wrapped with its own max cache size of 100. This means the rest of the code can stay the same and we'll still use `re-reselect`. We could revisit again with the Redux Toolkit V2 update, it comes with `reselect` v5 which has some more options for custom caching like this.

packages/charts/src/chart_types/flame_chart/flame_chart.tsx
walterra: Thanks, fixed it in [869226c](https://github.com/elastic/elastic-charts/pull/2609/commits/869226c0e78d1592061264f00885d677a9565aa6).

packages/charts/api/charts.api.md
```diff
@@ -2897,7 +2897,7 @@ export function sortIndexAccessor(n: ArrayEntry): number;
 export interface Spec {
     chartType: ChartType;
     id: string;
-    specType: string;
+    specType: SpecType;
```
walterra, line 2900: Does this change make it a breaking change?

nickofthyme: I don't think so since this is set internally, even though it may affect jest snapshots in consumer code.

markov00 (approved):
Tested locally and everything seems to work as expected. Great first step here, thanks Walter!

PR #2613, opened 2025-02-17
chore(refactor): refactor chart specific selectors out of state

walterra:
## Summary

WORK IN PROGRESS

### Checklist

- [ ] The proper **chart type** label has been added (e.g. `:xy`, `:partition`)
- [ ] The proper **feature** labels have been added (e.g. `:interactions`, `:axis`)
- [ ] All related issues have been linked (i.e. `closes #123`, `fixes #123`)
- [ ] New public API exports have been added to `packages/charts/src/index.ts`
- [ ] Unit tests have been added or updated to match the most common scenarios
- [ ] The proper documentation and/or storybook story has been added or updated
- [ ] The code has been checked for cross-browser compatibility (Chrome, Firefox, Safari, Edge)
- [ ] Visual changes have been tested with `light` and `dark` themes

PR #2637, opened 2025-03-28
chore(refactor): refactor chart specific selectors out of state (#2613) [69.x]

markov00:
 - chore(refactor): refactor chart specific selectors out of state (#2613)


markov00:
buildkite test this

PR #2638, opened 2025-03-28
build: specify firebase-tools package

markov00:
## Summary

Yesterday the `firebase-tools` package was upgraded to version 14.0.0 requesting node 20 to run.
The CI stops now because we can't do the deploy step because the pipeline is built in a docker container with node 18.

I'm first trying to use the previous used pacakged, than we should probably upgrade the docker machine to run on node 20.


markov00:
buildkite test this

PR #2643, opened 2025-04-04
fix(colors): use borealis colors as default

markov00:
## Summary

The current LIGHT_THEME and DARK_THEME where configured with the legacy amsterdam set of colors.
This PR changes it and upgrades it.

### BREAKING CHANGE
The LIGHT_THEME and DARK_THEME now are aligned with borealis colors.


packages/charts/src/utils/themes/amsterdam_dark_theme.ts
nickofthyme, line 1: So we just forgot about the `palette` when converting to borealis colors?

nickofthyme (approved):

nickofthyme:
buildkite update screenshots

PR #2646, opened 2025-04-10
fix(legend): avoid re-rendering

markov00:
Improve legend rendering efficiency by avoiding unnecessary re-renders when the legend configuration changes. This change enhances performance and maintains the integrity of the displayed legend items.

### DIFF
--- a/packages/charts/src/chart_types/xy_chart/state/selectors/get_legend_item_extra_values.ts
+++ b/packages/charts/src/chart_types/xy_chart/state/selectors/get_legend_item_extra_values.ts
@@ -13,12 +13,15 @@ import { EMPTY_LEGEND_ITEM_EXTRA_VALUES } from '../../../../common/legend';
 import type { SeriesKey } from '../../../../common/series_id';
 import { ScaleType } from '../../../../scales/constants';
 import { createCustomCachedSelector } from '../../../../state/create_selector';
+import { getLegendConfigSelector } from '../../../../state/selectors/get_legend_config_selector';
 import { getLegendItemExtraValues } from '../../tooltip/tooltip';
 
 /** @internal */
 export const getLegendItemExtraValuesSelector = createCustomCachedSelector(
-  [getTooltipInfoAndGeomsSelector, getComputedScalesSelector],
-  ({ tooltip: { values } }, { xScale: { type } }): Map<SeriesKey, LegendItemExtraValues> =>
+  [getTooltipInfoAndGeomsSelector, getComputedScalesSelector, getLegendConfigSelector],
+  ({ tooltip: { values } }, { xScale: { type } }, { legendValues }): Map<SeriesKey, LegendItemExtraValues> =>
     // See https://github.com/elastic/elastic-charts/issues/2050
-    type === ScaleType.Ordinal ? EMPTY_LEGEND_ITEM_EXTRA_VALUES : getLegendItemExtraValues(values),
+    type === ScaleType.Ordinal || legendValues.length === 0
+      ? EMPTY_LEGEND_ITEM_EXTRA_VALUES
+      : getLegendItemExtraValues(values),
 );
--- a/packages/charts/src/state/chart_selectors.ts
+++ b/packages/charts/src/state/chart_selectors.ts
@@ -27,9 +27,6 @@ export interface LegendItemLabel {
   depth: number;
 }
 
-/** @internal */
-export const EMPTY_LEGEND_ITEM_LIST: LegendItemLabel[] = [];
-
 /**
  * A set of chart-type-dependant functions that are required by all chart types
  * @internal
@@ -148,10 +145,19 @@ export interface ChartSelectors {
   canDisplayChartTitles(globalState: GlobalChartState): boolean;
 }
 
-/** @internal */
-export type ChartSelectorsFactory = () => ChartSelectors;
+type ChartSelectorsFactory = () => ChartSelectors;
 
+const EMPTY_LEGEND_ITEM_LIST: LegendItemLabel[] = [];
 const EMPTY_TOOLTIP = Object.freeze({ header: null, values: [] });
+const EMPTY_DIMENSION = Object.freeze({ top: 0, left: 0, width: 0, height: 0 });
+const EMPTY_SM_DOMAINS: SmallMultiplesSeriesDomains = Object.freeze({ smVDomain: [], smHDomain: [] });
+const EMPTY_OBJ = Object.freeze({});
+const EMPTY_TOOLTIP_VISIBILITY: TooltipVisibility = Object.freeze({
+  visible: false,
+  isExternal: false,
+  displayOnly: false,
+  isPinnable: false,
+});
 
 type CallbackCreator = () => (state: GlobalChartState) => void;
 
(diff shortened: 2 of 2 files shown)


packages/charts/src/chart_types/goal_chart/state/selectors/get_screen_reader_data.ts
```diff
@@ -22,7 +22,7 @@ export const getScreenReaderDataSelector = createCustomCachedSelector(
     getA11ySettingsSelector,
     (state: GlobalChartState) => state,
   ],
-  (goalChartData, goalChartLabels, internalChartState, a11ySettings, state): ChartSpecificScreenReaderData => {
+  (goalChartData, goalChartLabels, internalChartState, a11ySettings, state): ScreenReaderItem[] => {
     const screenReaderItems: ScreenReaderItem[] = [];
```
mariairiartef, line 26: Maybe we can try to do something like this at the beginning and do an early return:


```suggestion
    const chartTypeDescription = internalChartState?.getChartTypeDescription(state);
    const hasGoalData = !!goalChartData && Number.isFinite(goalChartData.maximum);
    
    if (!chartTypeDescription && !hasGoalData) return EMPTY_SCREEN_READER_ITEMS;
    
    const screenReaderItems: ScreenReaderItem[] = [];
```

packages/charts/src/chart_types/xy_chart/state/selectors/get_legend_item_extra_values.ts
```diff
@@ -13,12 +13,15 @@ import { EMPTY_LEGEND_ITEM_EXTRA_VALUES } from '../../../../common/legend';
...
 import { createCustomCachedSelector } from '../../../../state/create_selector';
+import { getLegendConfigSelector } from '../../../../state/selectors/get_legend_config_selector';
 import { getLegendItemExtraValues } from '../../tooltip/tooltip';
 
 /** @internal */
 export const getLegendItemExtraValuesSelector = createCustomCachedSelector(
-  [getTooltipInfoAndGeomsSelector, getComputedScalesSelector],
-  ({ tooltip: { values } }, { xScale: { type } }): Map<SeriesKey, LegendItemExtraValues> =>
+  [getTooltipInfoAndGeomsSelector, getComputedScalesSelector, getLegendConfigSelector],
+  ({ tooltip: { values } }, { xScale: { type } }, { legendValues }): Map<SeriesKey, LegendItemExtraValues> =>
```
mariairiartef: qq: do you think that we could benefit from having this as a boolean return by the selector? Something like `hasLegendValues`. I think that we check this many times

packages/charts/src/state/selectors/get_screen_reader_summary.ts
```diff
@@ -39,11 +39,11 @@ export const getScreenReaderSummarySelector = createCustomCachedSelector(
     }
 
     // Get chart-specific screen reader data
```
mariairiartef: nit: do you think we should update the comment?

```suggestion
    // Get chart-specific screen reader items
```

packages/charts/src/state/selectors/get_screenreader_data.ts
```diff
@@ -0,0 +1,34 @@
...
+  /** The label for this part of the summary */
+  label: string;
+  /** Optional ID for referencing this part */
+  id?: string;
+  /** The value for this part of the summary */
+  value: string;
+}
+
+/** @internal */
+export const EMPTY_SCREEN_READER_ITEMS = [];
```
mariairiartef: nit: I saw that we are doing this in other places

```suggestion
export const EMPTY_SCREEN_READER_ITEMS:  ScreenReaderItem[] = [];
```

packages/charts/src/state/chart_selectors.ts
```diff
@@ -170,10 +152,19 @@ export interface ChartSelectors {
   canDisplayChartTitles(globalState: GlobalChartState): boolean;
 }
 
-/** @internal */
-export type ChartSelectorsFactory = () => ChartSelectors;
+type ChartSelectorsFactory = () => ChartSelectors;
 
-const EMPTY_TOOLTIP = Object.freeze({ header: null, values: [] });
+const EMPTY_LEGEND_ITEM_LIST: LegendItemLabel[] = [];
+const EMPTY_TOOLTIP = { header: null, values: [] };
```
mariairiartef: I've seen this defined also in other files, such as in packages/charts/src/chart_types/goal_chart/state/selectors/tooltip.ts or packages/charts/src/chart_types/partition_chart/state/selectors/tooltip.ts

Do you think we should rely on only one and export it?

mariairiartef (commented):
I've left some comments!

packages/charts/src/state/chart_selectors.ts
markov00: done in https://github.com/elastic/elastic-charts/pull/2646/commits/fafec1ac7982a9bbd7c4dad2325b60d4338d726b

packages/charts/src/state/selectors/get_screenreader_data.ts
markov00: done in https://github.com/elastic/elastic-charts/pull/2646/commits/fafec1ac7982a9bbd7c4dad2325b60d4338d726b

packages/charts/src/state/selectors/get_screen_reader_summary.ts
markov00: done in https://github.com/elastic/elastic-charts/pull/2646/commits/fafec1ac7982a9bbd7c4dad2325b60d4338d726b

packages/charts/src/chart_types/xy_chart/state/selectors/get_legend_item_extra_values.ts
markov00: Currently no, it will just increase the memory to keep another selector in memory but with a limited gain

packages/charts/src/chart_types/goal_chart/state/selectors/get_screen_reader_data.ts
markov00: I see what you mean but in this case the gain is limited. Even with a check like this, we still need to check if the description is there to add the related description item and we still need to check again if the we the goal maximum value. I believe from the performance point of view we don't gain much

packages/charts/src/state/chart_selectors.ts
```diff
@@ -170,10 +152,20 @@ export interface ChartSelectors {
   canDisplayChartTitles(globalState: GlobalChartState): boolean;
 }
 
-/** @internal */
-export type ChartSelectorsFactory = () => ChartSelectors;
+type ChartSelectorsFactory = () => ChartSelectors;
 
-const EMPTY_TOOLTIP = Object.freeze({ header: null, values: [] });
+const EMPTY_LEGEND_ITEM_LIST: LegendItemLabel[] = [];
```
nickofthyme:
```suggestion
const EMPTY_LEGEND_ITEM_LIST: LegendItemLabel[] = Object.freeze([]);
```

Do you think it's too much to use `Object.freeze` on all these? It's unlikely they would change based on our aversion to mutations but still you never know. I don't think freeze affects runtime performance in any meaningful way.

Up to you...

markov00: I don't think we need to freeze these, just changing the type to be a readonly array should be enough.

I changed the types here:
https://github.com/elastic/elastic-charts/pull/2646/commits/125a221582007f58e0c13ffe665f3bf41eee7da5

PR #2654, opened 2025-04-30
fix: Update aria-labelledby for Sparkline svg to not use spaces

justinkambic:
## Summary

fix: Sparkline svg `aria-labelledby` now generates a valid ID that will make Metric charts more accessible.

## Details

This patch would modify the `aria-labelledby` attribute that is generated for the svg rendered by the Sparkline component within the Metric chart type. Downstream in Kibana we are failing an accessibility audit because our metrics charts do not contain `alt` or `aria` for this SVG. We can allow the svg to be described by an existing element that contains a proper label for the chart, but the `aria-labelledby` contains a space, which is not valid for this attribute and causes the attempt to link up an element ID to fail.

**Before (from [hosted storybook](https://elastic.github.io/elastic-charts/storybook/?path=/story/metric-alpha--basic&globals=toggles.showHeader:true;toggles.showChartTitle:false;toggles.showChartDescription:false;toggles.showChartBoundary:false;theme:light)):**
image

**After:**
image

I've tested this with the Axe testing plugin by manually changing the `aria-labelledby` attributes on my Sparkline svg elements in the DOM in Kibana, and supplying an element with a matching ID and verified this fixed the accessibility failures:

image

Without these edits, this is the result of the scan, where each Sparkline SVG gets logged as a failure:

image

## Issues

This addresses an accessibility failure occurring in the Synthetics app in Kibana: [#212242](https://github.com/elastic/kibana/issues/212242). Once this change is released and consumed by Kibana, a subsequent PR to the Synthetics app will resolve the issue.

### Checklist

- [X] All related issues have been linked (i.e. `closes #123`, `fixes #123`)

### DIFF
--- a/packages/charts/src/chart_types/metric/renderer/dom/sparkline.tsx
+++ b/packages/charts/src/chart_types/metric/renderer/dom/sparkline.tsx
@@ -82,7 +82,7 @@ export const SparkLine: FunctionComponent<{
         viewBox="0 0 1 1"
         preserveAspectRatio="none"
         role="img"
-        aria-labelledby={`${titleId} ${descriptionId}`}
+        aria-labelledby={`${titleId}_${descriptionId}`}
       >
         <title id={titleId} className="echScreenReaderOnly">
           {trendA11yTitle}


markov00 (approved):

markov00:
buildkite test this

nickofthyme (approved):

PR #2661, opened 2025-05-16
fix(heatmap): update tooltip visibility logic to handle empty tooltip info

### LINKED ISSUE #2546: Heatmap axis tooltip shows even when no value is present
**Describe the issue**

The `Heatmap` tooltip has a unique function used to show tick value labels on hover. This is used by ML to enable showing long axis labels in their full un-truncated form.

The issue is when the value is empty the tooltip is shown anyway, leaving a long skinny white tooltip.

Image

**Expected behaviour**
The empty tooltip should be hidden.

andrimal:
## Summary

The heatmap tooltip is no longer visible if it's empty.

Before the fix: 
Kapture 2025-05-22 at 13 03 27

After the fix: 
Kapture 2025-05-22 at 13 02 14

fix #2546 

### Checklist


andrimal:
buildkite update vrt

mariairiartef (commented):
I've checked and the tooltip has already a property called `isVisible`. For the specific case we are covering, we are setting it to `true`:

https://github.com/elastic/elastic-charts/blob/13d94551480093044ba9826666315e5a01ab9d42/packages/charts/src/chart_types/heatmap/state/selectors/tooltip.ts#L87-L100

Have you considered updating this value and using it in `isTooltipVisibleSelector`? For example:

```
isVisible: Boolean(pickedShapes.text)
```

And then in the selector:

```
visible: tooltipInfo.values.some(({ isVisible }) => isVisible) || pinned,
```

Otherwise, you can try to add a condition like at the beginning of `isTooltipVisibleSelector`

```
if (
  type === TooltipType.None ||
  tooltipInfo.values.length === 0 ||
  (tooltipInfo.values.length === 1 && !tooltipInfo.values[0]?.formattedValue)
) {
  return {
    visible: false,
    isExternal: false,
    displayOnly: false,
    isPinnable: false,
  };
}
```

andrimal:
buildkite update vrt

markov00 (approved):
Implementation wise it LGTM. I've doubts if we need to apply the same logic also to the `Y axis value`  and `X axis value` when the mouse pointer is on the chart and one of the axis has an empty value (see second screenshot that shows an empty Y Value) but I believe we can discuss this offline and and add it in case in a subsequent phase.

mariairiartef (approved):

PR #2663, opened 2025-06-04
fix(accessibility): makes charts accessible by tab

ppisljar:
resolves: https://github.com/elastic/kibana/issues/212625, https://github.com/elastic/kibana/issues/211531, https://github.com/elastic/kibana/issues/211703, https://github.com/elastic/kibana/issues/211444, https://github.com/elastic/kibana/issues/210172, ...


markov00:
As discussed offline, the ask is to make chart element accessible via keyboard, meaning that we need to support all the feature that we support today via mouse interaction, but doing that via keyboard (highlight a geometry, show the tooltip, pin the tooltip).
I'm going to close this for now as it doesn't solves the requested need

Issue #2630, opened 2025-03-20
Isolated points on stacked area not shown.


markov00:
**Describe the issue**
Independently from the configuration of the point visibility, when you are rendering a stacked area chart with isolated points, the isolated points are not shown

**To Reproduce**
1. Go to local storybook story (isolated data points) and add the `stackAccessor` property to the chart
2. [Go to storybook](http://localhost:9001/?path=/story/line-chart--isolated-data-points&globals=toggles.showHeader:true;toggles.showChartTitle:false;toggles.showChartDescription:false;toggles.showChartBoundary:false;theme:light&knob-enable fit function=&knob-series type=area&knob-max data points=60&knob-default point radius=3&knob-point visibility=never&knob-point visibility min distance=40&knob-visible series[0]=A&knob-visible series[1]=B&knob-point shape=circle)
3. See the missing isolated data points

**Expected behaviour**

The isolated data point should always be visible, independentely from the general point visibility.
Only if they are explicly configured to be hidden/disabled we should hide them.

**Screenshots**
<img width="1270" alt="Image" src="https://github.com/user-attachments/assets/b3d5dbe1-c8ae-4458-bf3a-0246129367b7" />

**Version (please complete the following information):**
Probably since 67.0.0

**Additional context**
The fact that in the default theme the visibility of the isolated points is configured as 'auto' doesn't change the bahaviour.

Issue #2460, opened 2024-06-09
Line Chart Point Style Not Applying for First Point of Every Group in LineChart - Kibana Version 8.12.0


mageshsankar:
**Summary**
In Kibana version 8.12.0, the pointStyleAccessor is not correctly applying the specified point styles to the first data point of each group in a line chart using @elastic/charts. This issue did not exist in Kibana version 7.16.3, where the point styles were applied correctly.

**Steps to Reproduce**

1. Set up a line chart using @elastic/charts in Kibana version 8.12.0.
2. Define a pointStyleAccessor to customize the appearance of data points.
3. Populate the chart with multiple series of data points, ensuring each series has at least two data points.
4. Observe that the pointStyleAccessor does not apply the specified styles to the first point of each series.
5. Expected Behavior
6. The pointStyleAccessor should consistently apply the defined styles to all data points, including the first point of each series, as it did in Kibana version 7.16.3.

Actual Behavior
The first data point of each series does not have the styles specified in the pointStyleAccessor applied. Subsequent points in the series correctly reflect the defined styles.

Code Example
Below is a simplified version of the code that demonstrates the issue:

```
  const dummyData = [
        { date: "2024-01-01", codeLines: 800, project: "Project A1", version: "v1.0" },
        { date: "2024-06-02", codeLines: 1100, project: "Project A", version: "v1.1" },
        { date: "2024-02-01", codeLines: 900, project: "Project A", version: "v1.0" },
        { date: "2024-07-02", codeLines: 950, project: "Project B", version: "v1.1" },
        { date: "2024-08-03", codeLines: 1000, project: "Project C", version: "v1.2" }
    ];< 
  return (
           <Chart renderer="canvas">
                        <Settings showLegend legendPosition={Position.Bottom} theme={lineChartTheme} />
                        <Tooltip customTooltip={toolTip} placement={Placement.Right} offset={40} />
                        <Axis
                            id={htmlID("left")}
                            position={Position.Left}
                            title="Code Lines"
                            showOverlappingLabels
                            tickFormat={formatNumber}
                        />
                        <Axis
                            id="bottom"
                            title="Data Capture Date"
                            position={Position.Bottom}
                            tickFormat={tickFormat}
                        />
                        <LineSeries
                            id={htmlID("lines")}
                            xScaleType={ScaleType.Ordinal}
                            yScaleType={ScaleType.Linear}
                            xAccessor="date"
                            yAccessors={["codeLines"]}
[...]

PR #2665, opened 2025-06-11
fix(isolatedPoints): fix visibility and remove custom style for isolated points

markov00:
## Summary

Fix the missing visualization of isolated data points in stacked area charts.

## Details

Take a more opinionated choice around isolated points:
- they are always visible (it is required to correctly interpret the chart, no matter what configuration is used in the theme)
- their style is inherited entirely from the other data point style, no external customization is required, we only increase the radius to be a 0.5px bigger then the line width in the case only the isolated points are visible.

A post-task will be https://github.com/elastic/elastic-charts/issues/2673

## Issues

fix https://github.com/elastic/elastic-charts/issues/2630

### Checklist


packages/charts/src/chart_types/xy_chart/rendering/points.ts
```diff
@@ -244,25 +240,41 @@ export function getRadiusFn(
...
 }
 
-function yAccessorForIsolatedPointCheck(datum: DataSeriesDatum): number | null {
-  return datum.filled?.y1 ? null : datum.y1;
+function isNotDefined(
+  datum: DataSeriesDatum,
+  isBandedSpec: boolean,
+  y1Scale: (d: DataSeriesDatum) => number,
+  y0Scale: (d: DataSeriesDatum) => number,
+) {
```
nickofthyme: I know this is not exported but this naming is really vague.

The negation makes naming weird so maybe could switch to be `isDatumDefined` or `isPointDefined` instead of `isDatumNotDefined`.

Maybe add a comment too with to broad cases this is capturing.

packages/charts/src/chart_types/xy_chart/utils/stacked_series_utils.ts
```diff
@@ -89,7 +89,7 @@ export function formatStackedDataSeriesValues(
       if (!dataSeriesProps) return null;
       const data = stackedSeries
         .map<DataSeriesDatum | null>((row) => {
-          const d = row.data[1].get(stackedSeries.key);
+          const d = row.data[1]?.get(stackedSeries.key);
```
nickofthyme:
```suggestion
          const d = row.data[1].get(stackedSeries.key);
```

This is not strictly needed as the type of `row.data` is a tuple of type `XValueSeriesDatum`.

```ts
export type XValueSeriesDatum = [XValue, SeriesValueMap];
```

I don't think we apply optional chaining to any other tuples in charts. Did you add this because you noticed this was `undefined`?

storybook/stories/line/12_isolated_data_points.story.tsx
```diff
@@ -102,6 +103,7 @@ export const Example: ChartsStory = (_, { title, description }) => {
         xAccessor={0}
         yAccessors={[1]}
         splitSeriesAccessors={[2]}
+        stackAccessors={stackAreas ? [2] : undefined}
```
nickofthyme, line 106: ✌🏼

packages/charts/src/chart_types/xy_chart/rendering/points.ts
```diff
@@ -244,25 +240,41 @@ export function getRadiusFn(
...
-  yDefined: YDefinedFn,
+  current: DataSeriesDatum,
+  isBandedSpec: boolean,
+  y1Scale: (datum: DataSeriesDatum) => number,
+  y0Scale: (datum: DataSeriesDatum) => number,
   prev?: DataSeriesDatum,
   next?: DataSeriesDatum,
 ): boolean {
-  if (index === 0 && (isNil(next) || !yDefined(next, yAccessorForIsolatedPointCheck))) {
+  if (isNotDefined(current, isBandedSpec, y1Scale, y0Scale)) {
```
nickofthyme: What was your reason for not using the `yDefined` function?

packages/charts/src/chart_types/xy_chart/rendering/area.ts
```diff
@@ -74,12 +74,10 @@ export function renderArea(
     panel,
     color,
     style.point,
-    style.isolatedPoint,
-    style.line.strokeWidth,
+    true,
     isBandedSpec,
     markSizeOptions,
     false,
-    style.isolatedPoint.enabled,
```
nickofthyme, line 82: I think we should fully remove `isolatedPoint.enabled` and all the usages. Also update the `13_point_style_overrides.story.tsx` with this removed and delete if no longer needed.

e2e/screenshots/all.test.ts-snapshots/baselines/test-cases/point-style-overrides-chrome-linux.png
nickofthyme, line 1: I double checked with #2460 and it looks like that issue is still good with these changes. 👍🏼

packages/charts/src/chart_types/xy_chart/rendering/area.ts
nickofthyme: Never mind, just saw #2673 but added the point about updating the story.

packages/charts/src/chart_types/xy_chart/rendering/points.ts
markov00: We need to mainly verify that the original datum (aka `initialY`) is there and is valid (initialY === datum.y without any stacking change).
The `yDefined` function checks instead if the y0/y1 values are defined, these values are computed values (is the stacked value for example) that's why we need to check for `initialY` and we can't use `yDefined`.
The two functions are distinct ones: one checks if a value is valid after the stacking logic occured, the other one before that.

packages/charts/src/chart_types/xy_chart/rendering/points.ts
markov00: refactored inverted the logic here https://github.com/elastic/elastic-charts/pull/2665/commits/fbb5b9195e7cf0df673d04a333bb4935a735503c

I hope the reversed logic works

packages/charts/src/chart_types/xy_chart/rendering/area.ts
markov00: yes, I'm trying to avoid breaking changes now, so we can easily patch Kibana

storybook/stories/line/12_isolated_data_points.story.tsx
markov00: we need to remove this https://github.com/elastic/elastic-charts/issues/2677

packages/charts/src/chart_types/xy_chart/utils/stacked_series_utils.ts
markov00: correct sorry I saw `data[1].get`, no warning there but my instinct was self conservative and put a `?` in there.
There is no need about that, you are right.
Removed https://github.com/elastic/elastic-charts/pull/2665/commits/c77109fd66260a482158ebeb8837edd494fcab13

packages/charts/src/chart_types/xy_chart/rendering/points.ts
nickofthyme: Thanks. I don't think the inline code comments are needed as they are pretty self evident. The block you added to the description nicely explains the cases that it's looking for.

Feel free to leave the comments if you want.

packages/charts/src/chart_types/xy_chart/rendering/points.ts
nickofthyme: Ah gotcha, thanks for clarifying.

PR #2666, opened 2025-06-11
feat(metric): add new layout configurations

mariairiartef:
Closes https://github.com/elastic/kibana/issues/220618

## Summary

Updates metric chart with new progress bar styling, distinct text colors for subtitle and extra elements, and configurable value (primary metric) positioning options (top/bottom). Additionally, creates a secondary metric component that can be used instead of a custom component.

Screenshot 2025-07-24 at 18 51 58

### Resources

- [Figma](https://www.figma.com/design/duESbCUc00lcUq1ezF95fv/-Charts--Metric-Chart?node-id=3323-425&p=f&m=dev) with the new metric chart layout designs in **Update (9.0)** tab
- [Document](https://docs.google.com/document/d/1sfdM7IcFQDHdPD__XEDDgN-yuRzulVyTu6RWlypGUy8/edit?tab=t.hmz1fv1j6kc4) containing pre-implementation questions and answers

### BREAKING CHANGE

The `valuesTextAlign` property in `MetricStyle` has been removed. Use the new `valueTextAlign` (for main value alignment) and `extraTextAlign` (for extra/secondary alignment) properties instead.

## Details

- [x] Update metric chart theme settings (done in `theme.ts`)
  - Remove `valuesTextAlign` setting
  - Add `extraTextAlign`, `valueTextAlign`, `valuePosition`, `titleWeight` settings
  - Add `textSubtitleDarkColor`, `textSubtitleLightColor`, `textExtraDarkColor`, `textExtraLightColor`
- [x] Separate the _values block_ into _extra block_ and  _value block_  so both can be handled independently
(done in `text.tsx`)
- [x] Add configuration for moving the value block from bottom to top
  - Includes updating icon handling using **grid** styling
- [x] Add configuration for making the title font weight `'bold'` or `'regular'`
- [x] Update progress bar style to align with the designs
- [x] Support different progress bar thicknesses based on height 
- [x] Support different text colors for subtitle and extra texts
- [x] Add support for secondary metric display with optional badge positioning
- [x] Add new storybook story to test the new layout configuration
- [ ] Update test cases and add new tests for the new layout

### Things that haven't been yet addressed

- [ ] Update target style when progress bar is shown in bullet graph
- [ ] Handle tittle and subtitle separately so metric icon is aligned with title only

## Issues

closes https://github.com/elastic/kibana/issues/220618?reload=1?reload=1

### Checklist


mariairiartef:
buildkite update screenshots

mariairiartef:
buildkite update screenshots

mariairiartef:
buildkite update screenshots

mariairiartef:
buildkite update screenshots

mariairiartef:
buildkite update screenshots

mariairiartef:
buildkite update screenshots

mariairiartef:
buildkite update screenshots

mariairiartef:
buildkite update screenshots

packages/charts/src/chart_types/metric/specs/index.ts
```diff
@@ -18,6 +18,34 @@ import type { LayoutDirection, ValueFormatter } from '../../../utils/common';
...
+   */
+  badgeColor?: string;
+  /**
+   * Determines whether the value appears before or after the label.
+   */
+  valuePosition?: 'before' | 'after';
+  /**
+   * Optional CSS properties to apply to the container element.
+   */
+  style?: CSSProperties;
```
mariairiartef, line 36: add `ariaDescription` property

packages/charts/src/chart_types/metric/renderer/dom/secondary_metric.tsx
```diff
@@ -0,0 +1,47 @@
...
+    <span
+      className="echSecondaryMetric__label"
+      style={isValueBeforeLabel ? { marginLeft: `${LABEL_MARGIN}px` } : { marginRight: `${LABEL_MARGIN}px` }}
+    >
+      {label}
+    </span>
+  );
+
+  return (
+    <span className="echSecondaryMetric" style={style}>
```
mariairiartef: add the ariaDescription here

gvnmagni:
hi all, I wanted to list here a few comments that we had the chance to go through in more detail during the past week but since it happened in different conversation there was no full track of it.

**1. Breakpoints** / We currently apply our breakpoints according to height but we also have a series of checks and fallbacks in case the values are not visible or anything is off. Marco and Maria noticed how introducing a check for width as well would simplify a lot the logic behind all this and we agreed on proceeding with this. The result will be a much simple logic and a better final result for customers edge cases.

**2. Subtitle text color** / Originally my plan was to visually differentiate the title from the subtitle but unfortunately this would raise a few accessibility concerns with custom background colors. In order to solve this, for this iteration, I would suggest to use `colors.textParagraph` from EUI as token for this element

**3. Text colors** / Since we are introducing the possibility to apply color to text (Value) I would suggest to populate the color picker not with the classic datavis palette but its alternative for text [EUI Link](https://eui.elastic.co/docs/getting-started/theming/tokens/colors/#dataVisText)

There are a couple of additional details but they will be on the Lens side, not on the chart one. @mariairiartef apologies for the continuous back and forth on some aspects and thank you for the awesome work!

mariairiartef:
buildkite update screenshots

storybook/stories/metric/1_basic.story.tsx
nickofthyme, line 1: This is more a question for @gvnmagni.

For the value text `center` alignment, we treat the icon as a floating/absolute element such that the text is always centered and the icon is adjacent.

| Previously | New layout |
|--------|--------|
| Zight Recording 2025-07-23 at 09 42 26 AM | Zight Recording 2025-07-23 at 09 38 01 AM | 

Notice the yellow line in the top left, previously we create padding on the opposite side of the title to center the text by accounting for the width of the icon.

Notice we did the same thing for the value text and still do. Do you care one way or another? Do you care if the title text is not true center when there is an icon? I think there are pros and cons to each.

storybook/stories/metric/metric.stories.tsx
```diff
@@ -16,3 +16,4 @@ export { Example as bodyContent } from './3_body.story';
 export { Example as trendMultipleData } from './4_trend_bug.story';
 export { Example as arrayOfValues } from './5_array_of_values.story';
 export { Example as extraBadges } from './6_extra_badge.story';
+export { Example as newLayoutConfig } from './7_new_layout_config.story';
```
nickofthyme: nit: I think we just label this `layout`, new becomes old fast and everything is basically a config.

```suggestion
export { Example as layout } from './7_layout.story';
```

storybook/stories/metric/7_new_layout_config.story.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+  const dataExtra = showExtra
+    ? useCustomExtra
+      ? { extra: <span dangerouslySetInnerHTML={{ __html: extra }}></span> }
+      : { extra: secondaryProps }
+    : {};
+
+  // Colors
+
+  // const metricColor = color('Metric color', '#61A2FF', colorGroup);
```
nickofthyme:
```suggestion
```

packages/charts/src/utils/themes/theme.ts
```diff
@@ -305,15 +301,48 @@ export interface HeatmapStyle {
   maxLegendHeight?: number;
 }
 
-/** @public */
+/**
+ * Horizontal text alignment options.
+ * @public
+ */
+export type TextAlign = 'left' | 'center' | 'right';
```
nickofthyme: The types for all these in the current code is kinda all over the place but mostly in `text_utils.ts`.

https://github.com/elastic/elastic-charts/blob/59e9b6894b6374c113b9472bc9f6f3cf24c8bc43/packages/charts/src/common/text_utils.ts#L21-L49

These do not follow our const enum convention (see example of this [here](https://github.com/elastic/elastic-charts/blob/59e9b6894b6374c113b9472bc9f6f3cf24c8bc43/packages/charts/src/chart_types/index.ts#L15-L28))

I was going to suggest we try to reuse those types in this PR but it's already not great. So I opened https://github.com/elastic/elastic-charts/issues/2692 instead. 🙃 

Iideally we would reuse the types that we have wherever possible. Ideally this would be direct but in some cases you may need to omit some of the options that don't apply. Like in this case we already have a `TextAlign` type defined as...

```ts
type TextAlign = "start" | "end" | "left" | "right" | "center";
```

But your use case only permits a subset of these so you could possibly do something like...

```ts
type MetricTextAlign = Extract<TextAlign, 'left' | 'center' | 'right'>;
```

Redefining the subset type and naming it for the use case (i.e. `Metric*`) makes it easier to assign for multiple properties and parameters.

I'll leave it up to you if you see a need to cleanup this and the `HorizontalSide` types.

packages/charts/src/utils/themes/theme.ts
```diff
@@ -305,15 +301,48 @@ export interface HeatmapStyle {
...
+ * Horizontal side options, typically used for alignment or positioning.
+ * @public
+ */
+export type HorizontalSide = 'left' | 'right';
+
+/**
+ * Font weight options for text styling.
+ * @public
+ */
+export type FontWeight = 'bold' | 'regular';
```
nickofthyme: In this case I do think we should reuse the `FontWeight` from `text_utils.ts`. You could limit it to a subset of just `bold` and `normal` but since this value is just passed as a style to the `span` it does not matter, other than possibly messing up the metric spacing logic.

In either case we should attempt to match our options to css values when there is a direct link. Here you are using `regular` which is not a standard css option which then requires extra work later to align it correctly to `normal`.

So I would say remove this or change it to `MetricFontWeight` if you need to use it in multiple places. In which case I would suggest...

```suggestion
/**
 * Font weight options for text styling.
 * @public
 */
export type MetricFontWeight = Extract<FontWeight, 'bold' | 'normal'>;
```

packages/charts/src/chart_types/metric/renderer/dom/titles.tsx
```diff
@@ -0,0 +1,183 @@
...
+}
+
+const TitleElement: React.FunctionComponent<TitleElementProps> = ({
+  title,
+  titleWeight,
+  fontSize,
+  textAlign,
+  linesLength,
+}) => {
+  const fontWeight = titleWeight === 'regular' ? 'normal' : titleWeight;
```
nickofthyme: From the comment above we should change `regular` to `normal` to align with the [css property](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight).

```suggestion
```

packages/charts/src/chart_types/metric/renderer/dom/titles.tsx
```diff
@@ -0,0 +1,183 @@
...
+      {title}
+    </span>
+  );
+};
+
+interface TitleProps {
+  metricId: string;
+  onElementClick?: () => void;
+  titleElementProps: TitleElementProps;
+}
```
nickofthyme: nit: Here you are passing the `titleElementProps` in a nested prop but the end user of the `Title` component does not care that you are conditionally wrapping the `TitleElement` in a `button`. In both cases these style are applied to the title text.

In that case I would suggest just extending the props and passing whatever you need to `TitleElement`.

```suggestion
type TitleProps = TitleElementProps & {
  metricId: string;
  onElementClick?: () => void;
}
```

packages/charts/src/chart_types/metric/renderer/dom/_badge.scss
```diff
@@ -0,0 +1,34 @@
...
+  border: 1px solid transparent;
+  border-radius: 2px;
+
+  font-size: inherit;
+  line-height: inherit;
+  text-align: start;
+  text-decoration: none;
+
+  background-color: #e3e8f2;
+  color: #07101f;
```
nickofthyme: Can we use a token for these?

packages/charts/src/chart_types/metric/renderer/dom/titles.tsx
```diff
@@ -0,0 +1,183 @@
...
+    titleWeight,
+    linesLength: titleLines.length,
+  };
+
+  return (
+    <div
+      className={classNames('echMetricText__titlesBlock', `echMetricText__titlesBlock--${textAlign}`)}
+      style={titlesBlockStyle}
+    >
+      {showTitle && <Title metricId={metricId} onElementClick={onElementClick} titleElementProps={titleElementProps} />}
```
nickofthyme: From previous comment...

```suggestion
      {showTitle && (
        <Title
          metricId={metricId}
          onElementClick={onElementClick}
          title={title}
          fontSize={titleFontSize}
          textAlign={textAlign}
          titleWeight={titleWeight}
          linesLength={titleLines.length}
        />
      )}
```

storybook/stories/metric/1_basic.story.tsx
gvnmagni: (apologies for answering this only now, didn't receive a notification for some reason)

if we can I would love to keep the previous behavior where we compensate the presence of the icon with the additional padding on the opposite side.

I think it is safer and more elegant, also because it could create very evident misalignments such as the one shown in the following image

Screenshot 2025-07-28 at 14 51 14

packages/charts/src/utils/themes/theme.ts
mariairiartef: Done [here](https://github.com/elastic/elastic-charts/pull/2666/commits/2e788a88fe74f59db4169384fd451dec74422540)

packages/charts/src/chart_types/metric/renderer/dom/titles.tsx
mariairiartef: Done [here](https://github.com/elastic/elastic-charts/pull/2666/commits/2e788a88fe74f59db4169384fd451dec74422540)

packages/charts/src/chart_types/metric/renderer/dom/titles.tsx
mariairiartef: Updated [here](https://github.com/elastic/elastic-charts/pull/2666/commits/c051d232f7c24f2d8a7592ab0d681047cab95983)

packages/charts/src/chart_types/metric/renderer/dom/titles.tsx
mariairiartef: Done [here](https://github.com/elastic/elastic-charts/pull/2666/commits/c051d232f7c24f2d8a7592ab0d681047cab95983)

packages/charts/src/chart_types/metric/renderer/dom/_badge.scss
mariairiartef: Done [here](https://github.com/elastic/elastic-charts/pull/2666/commits/bdd449cee729ccf84e6d83e944478e96dc774414)

packages/charts/src/utils/themes/theme.ts
mariairiartef: Makes sense! I've followed your suggestion [here](https://github.com/elastic/elastic-charts/pull/2666/commits/c64f4e38e7a4ebcb389b7b6042d8162923efa03b) :)

storybook/stories/metric/1_basic.story.tsx
mariairiartef: Makes sense, I tried to solve it [here](https://github.com/elastic/elastic-charts/pull/2666/commits/bb4f5e7935769bf583e4f4a219417839baacb006). Please take a look to see if it is the expected behavior

mariairiartef:
buildkite update screenshots

storybook/stories/metric/1_basic.story.tsx
gvnmagni: seems to work really well!

mariairiartef:
buildkite update screenshots

mariairiartef:
buildkite update screenshots

mariairiartef:
buildkite update screenshots

packages/charts/api/charts.api.md
```diff
@@ -2109,7 +2114,9 @@ export interface MetricStyle {
     // (undocumented)
     emptyBackground: Color;
     // (undocumented)
-    iconAlign: 'left' | 'right';
+    extraTextAlign: MetricTextAlign;
```
markov00: There is no need to have an extra type here, as you are doing more or less the same also for the `iconAlign`. I believe you can just remove the `MetricFontWeight` and just replace it with the `Extract...`

packages/charts/api/charts.api.md
```diff
@@ -2119,15 +2126,32 @@ export interface MetricStyle {
...
+    titlesTextAlign: MetricTextAlign;
+    // (undocumented)
+    titleWeight: MetricFontWeight;
     // (undocumented)
     valueFontSize: 'default' | 'fit' | number;
     // (undocumented)
-    valuesTextAlign: 'left' | 'center' | 'right';
+    valuePosition: 'top' | 'bottom';
+    // (undocumented)
+    valueTextAlign: MetricTextAlign;
```
markov00: Same as above, replace it with just the `Extract`

packages/charts/api/charts.api.md
```diff
@@ -2671,6 +2695,19 @@ export const ScaleType: Readonly<{
 // @public
 export type ScaleType = $Values<typeof ScaleType>;
 
+// @alpha
+export interface SecondaryMetricProps {
+    ariaDescription?: string;
+    badgeBorderColor?: Color;
+    badgeColor?: Color;
+    icon?: string;
+    iconSide?: 'left' | 'right';
```
markov00: This also can use `before` and `after` like the value position, so we don't need to have another different way to describe a position.
Probably also the name can be reconsidered and using the already used `position` like `iconPosition` since it indicates the same type of placement done also for the value in relation to the label

packages/charts/src/chart_types/metric/renderer/_index.scss
```diff
@@ -1,7 +1,13 @@
...
+@import './variables';
+@import 'dom/text';
+@import 'dom/progress';
+@import 'dom/sparkline';
+@import 'dom/secondary_metric';
+@import 'dom/badge';
+
+$progress-bar-target-size: 8px;
+$progress-bar-target-width: $progress-bar-target-size;
+$progress-bar-target-height: $progress-bar-target-size;
```
markov00: you can also move these progress-bar variables into the `variables` file

packages/charts/src/common/color_calcs.ts
```diff
@@ -55,6 +55,13 @@ export interface ColorContrastOptions {
   lightColor?: RgbaTuple;
 }
 
+/** @internal */
+export interface TextContrastOptions {
+  text: ColorContrastOptions;
+  subtitle: ColorContrastOptions;
+  extra: ColorContrastOptions;
+}
```
markov00: This is a specific type for the Metric Chart, is shouldn't be there. Also see the other comment about this

packages/charts/src/common/color_calcs.ts
```diff
@@ -131,3 +138,67 @@ export function highContrastColor(
 ): HighContrastResult {
   return HIGH_CONTRAST_FN[mode](background, options);
 }
+
+/** @internal */
+export function getBorderRecommendation(
+  color1: string,
+  color2: string,
```
markov00: What are color1 and color2? background or foreground colors?
Also can we consider renaming this function and make it more general? Can I apply this function to compute some other reccomendation?

packages/charts/src/chart_types/metric/renderer/dom/progress.tsx
```diff
@@ -49,24 +64,62 @@ export const ProgressBar: React.FunctionComponent<ProgressBarProps> = ({
...
+      ...borderRadius,
+      borderTopLeftRadius: PROGRESS_BAR_BORDER_RADIUS,
+      borderTopRightRadius: PROGRESS_BAR_BORDER_RADIUS,
+    };
+  }
+  if (isVertical && min <= 0) {
+    borderRadius = {
+      ...borderRadius,
+      borderBottomLeftRadius: PROGRESS_BAR_BORDER_RADIUS,
+      borderBottomRightRadius: PROGRESS_BAR_BORDER_RADIUS,
```
markov00, line 108: What is the scope of all this, if the value of the borderRadius is always 8, both here in this code and in the SCSS applied?

packages/charts/src/chart_types/metric/renderer/dom/secondary_metric.tsx
```diff
@@ -0,0 +1,64 @@
...
+  style,
+  ariaDescription,
+  badgeBorderColor,
+  icon,
+  iconSide,
+}) => {
+  const isValueBeforeLabel = valuePosition === 'before';
+  const hasLabel = !!label;
+
+  // TODO: consider icon separated to we give priority when the value is too long
```
markov00: Is this something that needs to be done in this PR or a different one?

packages/charts/src/chart_types/metric/renderer/dom/text.tsx
```diff
@@ -49,130 +91,132 @@ export const MetricText: React.FunctionComponent<{
...
+
+  let currentGridColumns = isIconVisible ? gridColumns[valuePosition][iconAlign] : gridSingleColumn;
+  if (isIconVisible && valuePosition === 'top' && style.valueTextAlign === 'center') {
+    currentGridColumns = { ...currentGridColumns, value: '2' }; // Center value in middle column
+  }
+  if (isIconVisible && valuePosition === 'bottom' && style.titlesTextAlign === 'center') {
+    currentGridColumns = { ...currentGridColumns, titles: '2' }; // Center titles in middle column
+  }
+
+  let extraElement = null;
```
markov00: Better to type this ` React.JSX.Element | null`

packages/charts/src/chart_types/metric/renderer/dom/metric.tsx
```diff
@@ -30,7 +31,83 @@ import type {
...
+
+  if (hasTrend) {
+    const { ratio, color, shade } = fillTextColor(
+      backgroundColor,
+      getSparkLineColor(blendedColor),
+      undefined,
+      contrastOptions,
+    );
+
+    // TODO verify this check is applied correctly
```
markov00, line 85: Is that verified?

packages/charts/api/charts.api.md
mariairiartef: Done [here](https://github.com/elastic/elastic-charts/pull/2666/commits/8b2d00c997e36a141d2677655c206b80b3b67996)

packages/charts/api/charts.api.md
mariairiartef: Done [here](https://github.com/elastic/elastic-charts/pull/2666/commits/8b2d00c997e36a141d2677655c206b80b3b67996)

packages/charts/api/charts.api.md
mariairiartef: Done [here](https://github.com/elastic/elastic-charts/pull/2666/commits/8b2d00c997e36a141d2677655c206b80b3b67996)

packages/charts/src/chart_types/metric/renderer/dom/secondary_metric.tsx
mariairiartef: Is already done! I forgot to remove the comment, my bad

packages/charts/src/chart_types/metric/renderer/_index.scss
mariairiartef: Done [here](https://github.com/elastic/elastic-charts/pull/2666/commits/aed04e7fb6a9d3257b46087ace55610eaf6b2953)

packages/charts/src/common/color_calcs.ts
mariairiartef: Moved [here](https://github.com/elastic/elastic-charts/pull/2666/commits/df3f01fe3ba9380fe4105f6bcdea3026f4cadefe)

packages/charts/src/common/color_calcs.ts
mariairiartef: In the end it can be either a background or foreground color. I added a comment to the function and I renamed it [here](https://github.com/elastic/elastic-charts/pull/2666/commits/79021ee0103c6489cb28b2e9fbb7b975c6f1b42e)

packages/charts/src/chart_types/metric/renderer/dom/progress.tsx
mariairiartef: This logic also applies do the bullet graph, where we can have a zero baseline. That's why I added these checks

packages/charts/src/chart_types/metric/renderer/dom/metric.tsx
mariairiartef: Not really. I copied the logic and extracted it in a function with that comment. I'll check

mariairiartef:
buildkite update screenshots

packages/charts/src/common/color_calcs.ts
markov00: Great, I can understand this function way better now, thanks

mariairiartef:
buildkite update screenshots

mariairiartef:
buildkite update screenshots

mariairiartef:
buildkite update screenshots

Issue #2614, opened 2025-02-19
[highlight] Borealis lighter colors are not so visible


markov00:
**Describe the issue**
With the introduction of the new Borealis color palette, the chart series highlighter does not work effectively. When hovering over a light-colored series, the "highlight" is barely visible because the other series are only slightly dimmed, making it difficult to distinguish the hovered series from the rest.

**Possible Solutions**
- Increase the dimming effect on non-hovered series.
- Introduce an additional outline or glow effect for the hovered series.
- Adjust contrast dynamically based on the hovered series’ color.

**To Reproduce**
1. Use a chart with multiple series in the new Borealis color palette.
1. Hover over a series with a lighter color.
1. Observe that the highlight effect is minimal, making it hard to differentiate the hovered series.

**Expected behaviour**
The hovered series should stand out clearly against the dimmed background, ensuring better visibility and focus.

**Screenshots**

**Version (please complete the following information):**
 - Browser: [e.g. chrome, safari]
 - Elastic Charts: [e.g. 1.0.1]

**Additional context**
This issue is particularly noticeable with lighter colors in the Borealis palette. Improving the highlight mechanism would enhance usability and readability.

**Errors in browser console**
N/A

**Checklist**

- [ ] The proper **chart type** label has been added (e.g. `:xy`, `:partition`)
- [ ] The proper **feature** labels have been added (e.g. `:interactions`, `:axis`)
- [ ] Every related Kibana issue is listed under `Kibana Cross Issues` list and the `kibana cross issue` tag is applied

nickofthyme:
Yeah I noticed it is not easy to always clear what, especially for gradients. Sometimes the highlight is nearly the same color as an adjacent geom.

![Image](https://github.com/user-attachments/assets/b6cb957f-8d16-44da-99c4-73d450894afd)

I think we rethink the highlight strategy altogether.

gvnmagni:
I believe the solution we adopted for linecharts and areachart is basically showing how we should deal with this interaction on all type of charts, with minor tweaks of course.

I think the idea of introducing neutral colors for un-selected data series makes it scalable and more easy to maintain, following this concept I would suggest to have a similar approach for other typologies of charts as well such as Piecharts, Treemaps and Barcharts.

The following image show the concept in action, at hover, all the un-selected series would become neutral while leaving the hovered one its original color.

[...]

PR #2676, opened 2025-06-25
fix(highlight): improve highlight on area/line charts 

markov00:
## Summary

This pull request introduces significant enhancements to the rendering logic for area, line, and point geometries in the XY chart. The changes improve support for dimmed and focused states, enabling more flexible styling and highlighting of chart elements. Additionally, the refactoring simplifies the codebase by removing redundant parameters and consolidating some logic.

### Rendering Enhancements:
* **General improvements**: dimmed area and line charts are now grayed out in the background when another series is focused. This improves the readability of the chart when using Borealis colors in light mode. 

* **Area Rendering**: Added support for dimmed and focused states in area geometries. Refactored `renderAreas` to sort and render areas based on their highlight state (`dimmed`, `focused`, and `default`) and updated the `renderArea` function to use the new highlight state for styling. The `focused` area is now also rendered on top of all the other areas to improve its visibility. The rendering of stacked areas is also improved and avoids rendering areas on top of area-line border like it was happening today (most of the failing tests are about this).

* **Line Rendering**: Incorporated highlight states into line geometries. Refactored `renderLines` to sort lines by highlight state and updated the `renderLine` function to apply dimmed and focused styles. 
* **Point Rendering**: Enhanced point rendering logic to support dimmed styles for fill and stroke colors, based on the highlight state. The `focused` line is now also rendered on top of all the other lines to improve its visibility.

### API Changes:
* **Styling Interfaces**: Updated `AreaSeriesStyle`, `LineSeriesStyle`, and `PointStyle` interfaces to include `dimmed` and `focused` properties for more granular control over opacity, fill, stroke, and texture.

### Codebase Simplification:
* **Shared Style Removal**: Removed `SharedGeometryStateStyle` parameter from rendering functions (`renderAreas`, `renderLines`, `renderPoints`) and replaced it with `GeometryHighlightState` for better state management. 
* **Tests Update**: Adjusted area style tests to use `GeometryHighlightState` instead of `GeometryStateStyle`, ensuring the tests align with the new rendering logic. 

## Issues

Part of https://github.com/elastic/elastic-charts/issues/2614


markov00:
buildkite update screenshots

packages/charts/src/chart_types/xy_chart/renderer/canvas/styles/area.ts
```diff
@@ -11,29 +11,72 @@ import type { Color } from '../../../../../common/colors';
...
+  themeDimmedStyle: AreaStyle['dimmed'],
+  highlightState: GeometryHighlightState,
+): TexturedStyles {
+  if (highlightState !== 'dimmed') {
+    return textureStyle;
+  }
+  if ('opacity' in themeDimmedStyle) {
+    return {
+      ...textureStyle,
+      opacity: textureStyle.opacity ?? 1 * themeDimmedStyle.opacity,
```
mariairiartef: qq: so in here if the highlightState is dimmed, we are still using the opacity from the texture styles?

markov00: Correct, we usually have always two parameters in te theme one is the color the other is the opacity.
The dimmed opacity is always multiplied and does not replace the theme opacity, so that a texture that was rendered by default at 0.5 opacity will be rendered at 0.25 if the dimmed opacity is 0.5, this will respect the original color and reduce the opacity by the dimmed one. If we just replace it there can be situation like the one below that doesn't cause any visible change.

The reason to have an opacity value is that you can associate a `ColorVariant` to a shape color, saying that it will take, for example, the color of the series. with the Opacity parameter you can alter this color by applying a different opacity. The main example is the Area style of an area chart: the area color is the Series color, but we provide an opacity value to reduce by default the series color opacity a bit. If we then need reduce again the opacity when dimmed we need to multiply all these opacities (color own opacity, opacity prop and dimmed opacity).

markov00: fixed the bug here https://github.com/elastic/elastic-charts/pull/2676/commits/c1f344639202bd488c4a9bc2aa25dd437c1f6070

mariairiartef (approved):

packages/charts/src/chart_types/xy_chart/renderer/canvas/areas.ts
```diff
@@ -12,99 +12,123 @@ import { renderLinePaths, renderAreaPath } from './primitives/path';
...
+  type AreaGeometry,
+  type PerPanel,
+} from '../../../../utils/geometry';
 
 /** @internal */
 export function renderAreas(
   ctx: CanvasRenderingContext2D,
   imgCanvas: HTMLCanvasElement,
   areas: Array<PerPanel<AreaGeometry>>,
-  sharedStyle: SharedGeometryStateStyle,
```
nickofthyme, line 32: Are you planning to add a breaking change note for all removals?

e2e/screenshots/all.test.ts-snapshots/baselines/area-chart/stacked-chrome-linux.png
nickofthyme, line 1: Is this due to the sorting changes? Are you now sorting lines separate from areas?

Note the subtle change to the line color on the middle series. I think it's an improvement from before. Now it looks like we are backing off the top series area from occluding the line below. but just wondering.

Zight Recording 2025-07-09 at 12 45 30 PM

e2e/screenshots/stylings_stories.test.ts-snapshots/stylings-stories/fit-function-styling/should-apply-opacity-on-styled-fit-series-chrome-linux.png
nickofthyme, line 1: Not necessarily in this PR, but I think we should mute these custom styles as well, or somehow handle this better.

image

packages/charts/api/charts.api.md
```diff
@@ -209,14 +209,22 @@ export interface AreaSeriesStyle {
...
 // @public (undocumented)
 export interface AreaStyle {
+    dimmed: {
+        opacity: number;
+    } | {
+        fill: Color | ColorVariant;
+        texture: {
+            opacity: number;
+        };
+    };
```
nickofthyme, line 227: Are we planning to add this behavior to other chart types? Now `Area`/`Line`/`Bubble` grey-out the unhighlighted series while `Bar` and `Partition` use the muted simple opacity reduction/white blend (non-grey) approach.

packages/charts/src/chart_types/xy_chart/renderer/canvas/areas.ts
```diff
@@ -12,99 +12,123 @@ import { renderLinePaths, renderAreaPath } from './primitives/path';
...
-  const fitAreaFillColor = style.fit.area.fill === ColorVariant.Series ? color : style.fit.area.fill;
-  const fitAreaFill: Fill = {
-    texture: getTextureStyles(ctx, imgCanvas, fitAreaFillColor, geometryStateStyle.opacity, style.fit.area.texture),
-    color: overrideOpacity(
-      colorToRgba(fitAreaFillColor),
-      (opacity) => opacity * geometryStateStyle.opacity * style.fit.area.opacity,
-    ),
-  };
+  const fitAreaFillThemeStyle = { ...style.fit.area, dimmed: style.area.dimmed };
+  const fitAreaFill = buildAreaStyles(ctx, imgCanvas, color, fitAreaFillThemeStyle, highlightState);
```
nickofthyme, line 131: Per my comment above, we are now removing the `sharedStyle` in favor of explicit states, but these state do not apply to the fitted area/lines. I don't particularly love this.

Why not apply the same `GeometryHighlightState` to  `LineFitStyle`/`AreaFitStyle`? I think that would be better than just showing it as is. It just sticks out like a sore 👍🏼 .

image

e2e/screenshots/stylings_stories.test.ts-snapshots/stylings-stories/fit-function-styling/should-apply-opacity-on-styled-fit-series-chrome-linux.png
nickofthyme: See https://github.com/elastic/elastic-charts/pull/2676#discussion_r2195728167 below

packages/charts/src/chart_types/xy_chart/renderer/canvas/areas.ts
markov00: Not sure, since I was going to backport this in Kibana 9.0/9.1 I was also considering not creating a breaking change.
I typed the new style to be compatible with the old one, so you can achieve the same result by switching the theme configuration.

e2e/screenshots/all.test.ts-snapshots/baselines/area-chart/stacked-chrome-linux.png
markov00: Correct, I've sorted thee areas and their lines to render from the one on the top first, so that the next line will be painted on top of the previous area so that the line always shows with the right color. Previously, the rendering logic was rendering from the bottom stacked area first, causing the area line to be overwritten in half by the area fill above.

packages/charts/api/charts.api.md
markov00: Kind of: for partition chart and bar graying out is not a nice solution because they don't have borders and everything becomes flat. I was considering adding borders or some similar strategies to make it works nicely. 
I've played with something like this in heatmap, but something similar can be applied also to bar and pie.
image
I believe we need different strategies mostly because you can't really add a meaning border to a line or an area.

e2e/screenshots/stylings_stories.test.ts-snapshots/stylings-stories/fit-function-styling/should-apply-opacity-on-styled-fit-series-chrome-linux.png
markov00: it should be applied also to the texture, not sure why is not showing correctly here, let me take a look

packages/charts/src/chart_types/xy_chart/renderer/canvas/areas.ts
markov00: It should be applied, as you can see in the code the `fitAreaFill` color applies it
```
  const fitAreaFill = buildAreaStyles(ctx, imgCanvas, color, fitAreaFillThemeStyle, highlightState);
```
but I'm not sure why is showing like that, let me check the code and the example

packages/charts/api/charts.api.md
nickofthyme: Yeah good point but I think it would be nice to open an issue to discuss the other applications and how to align this behavior across all of them.

e2e/screenshots/stylings_stories.test.ts-snapshots/stylings-stories/fit-function-styling/should-apply-opacity-on-styled-fit-series-chrome-linux.png
nickofthyme: Resolving in favor of other comment

packages/charts/api/charts.api.md
markov00: https://github.com/elastic/elastic-charts/issues/2685

packages/charts/src/chart_types/xy_chart/renderer/canvas/areas.ts
markov00: actually this looks like a bug in the texture or something that we never implemented https://github.com/elastic/elastic-charts/issues/2686
The background is not controlled by the theme itself, the fill/stroke are just used for the texture shape and the alpha applied to the pattern canvas doesn't have effect actually for the canvas background

markov00:
buildkite update screenshots

PR #2681, opened 2025-07-01
fix(axis): prevent sticky label in multi-layer time axis

### LINKED ISSUE #2536: Multilayer Time axis shows confusing time ticks at start of axis
### The Problem

In some cases on a multi-layer time axis where the smallest time granularity is a unit of time, the starting tick shows as the previous nearest bucket. Take the example below, the first bucket is for the interval at `02:00` but the axis value shows `00:00` instead.

Image

This is because the leftmost tick represents the **start** of the first bucket. This works fine for time granularities like months or years that define an implicit broad range and not a precise time/start like hour of the day. You can see this in the demo below where `October` label moves to represent the full month until it is beyond that month, same for the `2020` year label.

Zight Recording 2024-07-31 at 09 49 44 AM

### The Solution

I think the best solution in this case is disable the sticky placement behavior where we only these ticks at their exact accurate placement on the scale.

A more involved solution would be applied later to keep this behavior but show an arrow to indicate the tick is a reference to an off-screen value as described in https://github.com/elastic/elastic-charts/issues/1934#issuecomment-1387334085.

nickofthyme:
## Summary

Omits first tick on first layer if tick is offset (aka sticky) such that it's position is not aligned with the true value.

Area Chart - Timeslip ⋅ Storybook 2025-07-01 at 6 56 12 PM

> Note: This does not impact the `Timeslip` chart type as that chart tick logic is separate from the cartesian chart axis logic.

## Details

- Omit first tick in first layer if offset
- Remove useless dependency on axisSpec
- Add tests for updates cases
- Cleanup tests, remove commented and useless skipped tests

## Issues

Fixes #2536

### Checklist


nickofthyme:
buildkite update screenshots

packages/charts/src/chart_types/xy_chart/utils/axis_utils.test.ts
```diff
@@ -1336,417 +812,79 @@ describe('Axis computational utils', () => {
...
-    ).toEqual([
-      { value: 1547208000000, label: '2019-01-11', position: 25.145833333333332, layer },
-      { value: 1547251200000, label: '2019-01-12', position: 85.49583333333334, layer },
-      { value: 1547337600000, label: '2019-01-13', position: 206.19583333333333, layer },
-      { value: 1547424000000, label: '2019-01-14', position: 326.8958333333333, layer },
-      { value: 1547510400000, label: '2019-01-15', position: 447.59583333333336, layer },
-      { value: 1547596800000, label: '2019-01-16', position: 568.2958333333333, layer },
-    ]);
-  });
-  test('should show unique consecutive ticks if duplicateTicks is set to false', () => {
```
nickofthyme, line 1390: `duplicateTicks` logic was removed from the `generateTicks` method long ago so this was doing nothing.

packages/charts/src/chart_types/xy_chart/utils/axis_utils.test.ts
```diff
@@ -1336,417 +812,79 @@ describe('Axis computational utils', () => {
...
+    expect(result).toHaveLength(8);
+    expect(result).toMatchObject([
+      { label: '00:30:00', value: 1547166600000 },
+      { label: '01:00:00', value: 1547168400000 },
+      { label: '01:30:00', value: 1547170200000 },
+      { label: '02:00:00', value: 1547172000000 },
+      { label: '02:30:00', value: 1547173800000 },
+      { label: '03:00:00', value: 1547175600000 },
+      { label: '03:30:00', value: 1547177400000 },
+      { label: '04:00:00', value: 1547179200000 },
```
nickofthyme, line 850: Simplify test assertions to only what matters.

packages/charts/src/chart_types/xy_chart/utils/axis_utils.test.ts
```diff
@@ -975,109 +554,6 @@ describe('Axis computational utils', () => {
...
-      const axisTicksPosition = getAxesGeometries(
-        {
-          chartDimensions: chartDim,
-          leftMargin: 0,
-        },
-        LIGHT_THEME,
-        { rotation: NO_ROTATION },
-        axisSpecs,
-        axisDims,
-        axisStyles,
```
nickofthyme, line 994: So much of these tests have been commented for years. It's time to say goodbye and cleanup this file.

packages/charts/src/chart_types/xy_chart/state/selectors/visible_ticks.ts
```diff
@@ -70,7 +70,6 @@ function getDirectionFn({ type }: ScaleBand | ScaleContinuous): (label: string)
 
 /** @internal */
 export function generateTicks(
-  axisSpec: AxisSpec,
```
nickofthyme, line 73: This `axisSpec` was never used in this function.

e2e/screenshots/axis_stories.test.ts-snapshots/axis-stories/timeslip-multilayer-time-axis/should-not-render-offset-or-clipped-tick-labels-chrome-linux.png
markov00, line 1: Why this in is 3 layers? aren't we only focusing on a max of 2 layers?

e2e/screenshots/all.test.ts-snapshots/baselines/interactions/brush-selection-tool-on-histogram-time-charts-chrome-linux.png
markov00, line 1: Probably the chart is correct, but I would consider adding an additional test where we are display a one tick at the beginning and we want to see both the gridline for that tick and the label for it, just to be sure about the fact that we are not just removing the first label but that we show only the labels if the tick is present

packages/charts/src/chart_types/xy_chart/state/selectors/visible_ticks.ts
```diff
@@ -82,22 +81,28 @@ export function generateTicks(
...
   const isContinuous = isContinuousScale(scale);
-  return ticks.map<AxisTick>((value) => {
+  return ticks.reduce<AxisTick[]>((acc, value, i) => {
     const domainClampedValue = isContinuous && typeof value === 'number' ? Math.max(value, scale.domain[0]) : value;
     const label = labelFormatter(value);
-    return {
+    const position = (scale.scale(value) || 0) + offset; // todo it doesn't look desirable to convert a NaN into a zero
+    const domainClampedPosition = (scale.scale(domainClampedValue) || 0) + offset; // todo it doesn't look desirable to convert a NaN into a zero
+
+    if (layer === 0 && i === 0 && position < domainClampedPosition) return acc;
```
markov00: If we want to we can fix the todos like:

```suggestion
    const position = scale.scale(value);
    const domainClampedPosition = scale.scale(domainClampedValue
    if(!isFiniteNumber(position) || !isFiniteNumber(domainClamnedPosition) return acc;
    if (layer === 0 && i === 0 && position < domainClampedPosition) return acc;
    
    ...
    position: position + offset,
    domainClampedPosition: domainClampedPosition + offset
```

e2e/screenshots/all.test.ts-snapshots/baselines/interactions/brush-selection-tool-on-histogram-time-charts-chrome-linux.png
nickofthyme: Yeah this image the first bucket is just short of the full bucket.

I can add a test for this if you'd like but this is the default case which is tested [here](https://github.com/elastic/elastic-charts/blob/main/e2e/screenshots/all.test.ts-snapshots/baselines/area-chart/timeslip-chrome-linux.png), shown below.

image

Then I added a case to test the initial offset https://github.com/elastic/elastic-charts/pull/2681/files#r2182033546

e2e/screenshots/axis_stories.test.ts-snapshots/axis-stories/timeslip-multilayer-time-axis/should-not-render-offset-or-clipped-tick-labels-chrome-linux.png
nickofthyme: I don't know this is just the current default for the `Area > Timeslip` story. 

https://ech-e2e-ci--pr-2681-dy10qs5l.web.app/storybook/?path=/story/area-chart--timeslip

packages/charts/src/chart_types/xy_chart/state/selectors/visible_ticks.ts
nickofthyme: Added in ed5ffbdc8622512b844c67d92687fe82926a16fe

I simply did `const position = scale.scale(value) + offset` because it will be `NaN` regardless of the `offset`.

e2e/screenshots/axis_stories.test.ts-snapshots/axis-stories/timeslip-multilayer-time-axis/should-not-render-offset-or-clipped-tick-labels-chrome-linux.png
nickofthyme: I updated the story in 93408987b0dbad13e0aecac93fe01be3b2bf48fc to use 2 by default.

nickofthyme:
buildkite update screenshots

nickofthyme:
buildkite test this

e2e/screenshots/all.test.ts-snapshots/baselines/interactions/brush-selection-tool-on-histogram-time-charts-chrome-linux.png
markov00: ok sounds good

PR #2684, opened 2025-07-09
test(a11y): e2e baseline tests for a11y screen reader descriptions

walterra:
## Summary

This PR adds e2e accessibility tests to assert screen reader summaries for different chart types and configurations. The aim is to get a baseline of the current state of these summaries before we merge the improvements being worked on in #2683. This will allow us to track regressions/improvements in a better way.

## Details

The accessibility e2e tests pick up the existing setup to run visual regression tests. Originally I thought we could run them as part of the visual regression tests, but then it would be tricky to run them locally because the visual ones don't pass there. So I implemented this via a separate command you can run via `yarn test:e2e:a11y`. Like the visual regression tests, the a11y ones also walk through the storybook examples and then assert the screen reader description (what's within `.echScreenReaderOnly`).

I added the tests to the CI setup too, the scripts are mostly copies of the ones used for the visual regression tests.

Open Question: We now run `yarn test:e2e` for visual regression tests and `yarn test:e2e:a11y`. Should we be more explicit about the visual regression test ones (like `yarn test:e2e:vrt`) - and also update descriptive texts to include VRT or Visual Regression Testing for the default e2e tests?

## Issues

Part of [#36386](https://github.com/elastic/kibana/issues/36386).

### Checklist

- [x] Unit/e2e tests have been added or updated to match the most common scenarios


.buildkite/steps/playwright_a11y.ts
```diff
@@ -0,0 +1,60 @@
...
+  return {
+    group: ':playwright: Playwright a11y',
+    key: 'playwright_a11y',
+    skip,
+    steps: [
+      {
+        ...commandStepDefaults,
+        label: ':playwright: Playwright a11y',
+        skip,
+        parallelism: 5,
```
markov00: We don't need all this parallelism, I've checked the runtime and we are actually spending ~20 seconds for the tests and 1.5 minutes for the startup. I believe we can just have a single instance running to save on resources with the same results

e2e/tests_a11y/axis_chart_a11y.test.ts
```diff
@@ -0,0 +1,63 @@
...
...lsnone$13 Bn52%Africa2South AfricaMineral fuels, lubricants and related materials$13 Bn52%Asia1Mineral fuels, lubricants and related materialsnone$717 Bn26%Asia2JapanMineral fuels, lubricants and related materials$177 Bn6%Asia2ChinaMineral fuels, lubricants and related materials$168 Bn6%Asia2South KoreaMineral fuels, lubricants and related materials$108 Bn4%Asia2IndiaMineral fuels, lubricants and related materials$97 Bn4%Asia2SingaporeMineral fuels, lubricants and related materials$67 Bn2%Asia2ThailandMineral fuels, lubricants and related materials$27 Bn1%Asia2IndonesiaMineral fuels, lubricants and related materials$24 Bn1%Asia2TurkeyMineral fuels, lubricants and related materials$22 Bn1%Asia2MalaysiaMineral fuels, lubricants and related materials$13 Bn0%Asia2Hong KongMineral fuels, lubricants and related materials$13 Bn0%Asia1Manufactured goods classified chiefly by materialnone$238 Bn9%Asia2ChinaManufactured goods classified chiefly by material$79 Bn3%Asia2South KoreaManufactured goods classified chiefly by material$33 Bn1%Asia2JapanManufactured goods classified chiefly by material$32 Bn1%Asia2IndiaManufactured goods classified chiefly by material$31 Bn1%Click to show more data',
```
markov00, line 60: I'm interested in knowing why these labels are all attached one after the other without a space. 
This looks like the partition accessibility table

e2e/tests_a11y/flame_chart_a11y.test.ts
```diff
@@ -0,0 +1,32 @@
...
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+import { test, expect } from '@playwright/test';
+
+import { common } from '../page_objects/common';
+
+test.describe('Flame Chart Accessibility', () => {
```
markov00: we can probably skip this test, rather then adding a valid condition that doesn't actually satisfy the test

e2e/tests_a11y/goal_chart_a11y.test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { test } from '@playwright/test';
+
+import { common } from '../page_objects/common';
+
+test.describe('Goal Chart Accessibility', () => {
+  test('should generate correct a11y summary for goal chart', async ({ page }) => {
+    await common.testA11ySummary(page)(
+      'http://localhost:9001/?path=/story/goal-alpha--minimal-goal',
+      'Revenue 2020 YTD  (thousand USD)  Chart type:goal chartMinimum:0Maximum:300Target:260Value:280',
```
markov00, line 17: Again I'm wondering if we are just missing a "space" in the concatenation of the strings or if there is another issue

e2e/tests_a11y/axis_chart_a11y.test.ts
walterra: At the moment it works like that: We select all `.echScreenReaderOnly` elements and join them with `' '`. But within each of the elements we select the text with `element.textContext`, this is what causes the text to end up without spaces when the dom is for example `<div>text1</div><div>test2</div>` -> `text1text2`. Like you suspected for table elements (`<td>...</td>...`) it has the same effect to summarize text content without spaces. Screen readers on the other hand would again consider the DOM structure so it's not an issue on the rendered page.

e2e/tests_a11y/goal_chart_a11y.test.ts
walterra: This is again rendered in different elements in the DOM and `.textContent` joins that without spaces.

.buildkite/steps/playwright_a11y.ts
walterra: Updated parallelism to 1 in 5af5ad6.

markov00:
buildkite test this

walterra:
@nickofthyme as discussed I renamed the existing playwright tests/reports to include a reference to VRT so we have a better VRT/A11Y distinction for the e2e tests.

e2e/tests_a11y/legend_chart_a11y.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    const summaryText = await common.getA11ySummaryText(page)();
+    expect(summaryText).toBe(
+      'Chart type:sunburst chart The table represents only 20 of the 31 data pointsDepthLabelParentValuePercentage1Mineral fuels, lubricants and related materialsnone$779 Bn24%2AsiaMineral fuels, lubricants and related materials$454 Bn14%3JapanAsia$177 Bn5%3ChinaAsia$168 Bn5%3South KoreaAsia$108 Bn3%2North AmericaMineral fuels, lubricants and related materials$325 Bn10%3United StatesNorth America$325 Bn10%1Miscellaneous manufactured articlesnone$227 Bn7%2North AmericaMiscellaneous manufactured articles$227 Bn7%3United StatesNorth America$227 Bn7%1Crude materials, inedible, except fuelsnone$174 Bn5%2AsiaCrude materials, inedible, except fuels$174 Bn5%3ChinaAsia$174 Bn5%1Manufactured goods classified chiefly by materialnone$130 Bn4%2North AmericaManufactured goods classified chiefly by material$130 Bn4%3United StatesNorth America$130 Bn4%1Chemicals and related productsnone$128 Bn4%2North AmericaChemicals and related products$128 Bn4%3United StatesNorth America$128 Bn4%1Machinery and transport equipmentnone$1,854 Bn56%Click to show more data',
+    );
```
nickofthyme, line 67: For these larger more complex ones I could see it becoming very tedious to manually update these strings to reflect the changes.

I wonder if there is something similar to [`jest. toMatchInlineSnapshot `](https://jestjs.io/docs/snapshot-testing#inline-snapshots) that could manually update them. I'm not sure how [`.toMatchSnapshot `](https://playwright.dev/docs/api/class-snapshotassertions#snapshot-assertions-to-match-snapshot-1) in Playwright but could be workable.

Not for this PR but something to note and watch for when working with these new tests going forward.

.buildkite/pipelines/pull_request/pipeline.ts
nickofthyme, line 1: Looks like there is an issue with the `Playwright VRT report` link https://ech-e2e-ci--pr-2684-thu8e4g6.web.app/e2e-report

image

.buildkite/utils/github.ts
```diff
@@ -475,7 +477,8 @@ ${deploymentMsg}`;
 - [Docs](${deploymentUrl})
 - [Storybook](${deploymentUrl}/storybook)
 - [e2e server](${deploymentUrl}/e2e)
-${preDeploy ? '- ⏳ Playwright report - Running e2e tests' : `- [Playwright report](${deploymentUrl}/e2e-report)`}`;
+${preDeploy ? '- ⏳ Playwright VRT report - Running e2e tests' : `- [Playwright VRT report](${deploymentUrl}/e2e-report)`}
```
nickofthyme:
```suggestion
${preDeploy ? '- ⏳ Playwright VRT report - Running e2e tests' : `- [Playwright VRT report](${deploymentUrl}/vrt-report)`}
```

e2e/.gitignore
```diff
@@ -1,5 +1,7 @@
 reports/
 merged_html_report/
```
nickofthyme:
```suggestion
```

.buildkite/utils/github.ts
```diff
@@ -461,10 +461,12 @@ Failure${jobLink ? ` - [failed job](${jobLink})` : ''}${err}
 - [Docs](${deploymentUrl})
 - [Storybook](${deploymentUrl}/storybook)
 - [e2e server](${deploymentUrl}/e2e)
-- ([Playwright report](${deploymentUrl}/e2e-report)`
+- ([Playwright VRT report](${deploymentUrl}/e2e-report)
```
nickofthyme:
```suggestion
- ([Playwright VRT report](${deploymentUrl}/vrt-report)
```

.buildkite/scripts/steps/playwright_a11y.ts
```diff
@@ -0,0 +1,79 @@
...
+
+import { getMetadata, setMetadata } from 'buildkite-agent-node';
+
+import { updateCheckStatus } from './../../utils/github';
+import { exec, downloadArtifacts, startGroup, yarnInstall, getNumber, decompress, compress, bkEnv } from '../../utils';
+import { ENV_URL } from '../../utils/constants';
+
+const jobIndex = getNumber(process.env.BUILDKITE_PARALLEL_JOB);
+const shardIndex = jobIndex ? jobIndex + 1 : 1;
+const jobTotal = getNumber(process.env.BUILDKITE_PARALLEL_JOB_COUNT);
```
nickofthyme, line 19: Much of this logic is probably not necessary as the `parallelism` is set to `1` but I think it's fine, we may want to add more tests in the future.

e2e/scripts/test.sh
```diff
@@ -41,6 +64,12 @@ echo "Connected to e2e server at ${ENV_URL}"
...
 
-export VRT=true
-# Run playwright tests with passed args
-playwright test "$@"
+# Set up environment and run tests based on mode
+if [ "$A11Y_MODE" = true ]; then
+  # Run playwright accessibility tests with passed args
+  playwright test --config=playwright.a11y.config.ts "$@"
+else
+  export VRT=true
```
nickofthyme: This variable is mainly used to "seed" the randomized data of the charts. In some stories we generate random data on refresh/change/etc.

I think you may still want this regardless of the mode, to stabilize testing as the a11y tests could rely on random data. Maybe I'm wrong but in any case I don't think it would hurt.

e2e/scripts/test.sh
```diff
@@ -8,18 +8,41 @@ attempt_counter=0
...
+  case $1 in
+    --a11y)
+      A11Y_MODE=true
+      shift
+      ;;
+    *)
+      break
+      ;;
+  esac
+done
```
nickofthyme, line 23: 🪄 MAGIC ✨

e2e/scripts/start.sh
nickofthyme, line 1: Tested script changes locally and still work fine 👍🏼

nickofthyme (approved):

.buildkite/pipelines/pull_request/pipeline.ts
nickofthyme: This should be fixed by updating...

- https://github.com/elastic/elastic-charts/pull/2684#discussion_r2254715017
- https://github.com/elastic/elastic-charts/pull/2684#discussion_r2254716667

walterra: Updated the links in f16bd3b1f98cbe0a54c2db0992657b683aedf073.

.buildkite/utils/github.ts
walterra: Fixed in f16bd3b1f98cbe0a54c2db0992657b683aedf073.

e2e/.gitignore
walterra: Fixed in f16bd3b1f98cbe0a54c2db0992657b683aedf073.

.buildkite/utils/github.ts
walterra: Fixed in f16bd3b1f98cbe0a54c2db0992657b683aedf073.

e2e/scripts/test.sh
walterra: Updated in f16bd3b1f98cbe0a54c2db0992657b683aedf073.

e2e/tests_a11y/legend_chart_a11y.test.ts
walterra: I'll look into that for the follow up that updates the descriptions!

PR #2699, opened 2025-08-11
chore(refactor): refactors A11Y screen reader description generation into chart selector factories

walterra:
## Summary

In general Elastic Charts has strict separation of concerns and individual charts are not aware of each other. How screen reader summaries were created was a bit of an exemption with chart type checks and customization within the screen reader components. To prepare for more detailed screen reader descriptions for individual charts, this PR adds an optional `getScreenReaderData`selector to the chart selector factories. This allows to create individual chart descriptions using a generic shared screen reader component instead of the current hard coded logic.

## Details

- Moved chart type-specific screen reader data generation to dedicated selectors using the `ChartSelectors` pattern
- Introduces `ChartLabelData` interface to make screen reader components chart-agnostic (used by the goal chart)
- Replaced `div` wrappers with semantic HTML elements like `figcaption`
- Fixed invalid HTML by moving screen reader elements outside `canvas` elements

Refactoring examples: 
- `packages/charts/src/components/accessibility/screen_reader_summary.tsx` no longer includes chart specific logic (like `goalChartData`). Instead this can be now found in the chart specific selector in `packages/charts/src/chart_types/goal_chart/state/selectors/get_screen_reader_data.ts`.
- The `ScreenReaderPartitionTable` (specific to the partition chart) was moved from the common `accessbility` directory to a chart specific place in `partition_chart`.

Note: Now that we have baseline e2e a11y tests on CI, we can make a refactor like this with more confidence. The PR passes CI without any updates necessary to the a11y assertions.

Screencast of Safari with A11Y VoiceOver on a basic line chart with current description:

https://github.com/user-attachments/assets/b20d2af5-4542-4e98-b864-76e6b4433097

## Issues

Part of https://github.com/elastic/kibana/issues/36386.

### Checklist


walterra:
buildkite test this

walterra:
buildkite test this

packages/charts/src/chart_types/goal_chart/state/selectors/get_screen_reader_data.ts
```diff
@@ -0,0 +1,69 @@
...
+        },
+      );
+    }
+
+    const returnValue = {
+      screenReaderTypes,
+      labelData: {
+        majorLabel: goalChartLabels.majorLabel,
+        minorLabel: goalChartLabels.minorLabel,
+      },
```
markov00: if you don't mind I think we can simplify the current logic:
- the goal chart is not used in Kibana (replaced by the bullet), so we can get rid of that `labelData` part and move the majorLabel, minorLabel into 2 screenReaderTypes: with label/value like `Minor Label`:minorLabel, `Major Label`: major label (currently we first write the labels then the Types mixing things up a bit, see screenshot)
- we can completely remove the `labelData` from the `ChartSpecificScreenReaderData` type
- if possible I would rename these `screenReaderTypes` into something more clear, like `screenReaderInfos` or something that makes it a bit more descriptives, the `Types` part is whart I don't fully get when reading this code.
- we can get read completely of the `screen_reader_label.tsx` and its component, we can just render that descriptive list.
- the previous part can probably be replaced by just rendering the Label as header and description as Paragraph when available before the rest of the infos screens.


Screenshot 2025-09-04 at 11 27 50

packages/charts/src/chart_types/goal_chart/state/selectors/get_screen_reader_data.ts
```diff
@@ -0,0 +1,69 @@
...
+    getGoalChartLabelsSelector,
+    getInternalChartStateSelector,
+    getA11ySettingsSelector,
+    (state) => state,
+  ],
+  (goalChartData, goalChartLabels, internalChartState, a11ySettings, state): ChartSpecificScreenReaderData => {
+    const screenReaderTypes: ScreenReaderType[] = [];
+
+    // Add chart type description first
+    const chartTypeDescription = internalChartState?.getChartTypeDescription(state);
```
andrimal: Minor suggestion: here we could maybe add the state type as `state: GlobalChartState` to avoid the `Unsafe argument of type any ...` warning in line 28.

walterra: 👍 fixed the types in [c21d298](https://github.com/elastic/elastic-charts/pull/2699/commits/c21d298b021785ee91c34f632c5405e475f1d50a).

packages/charts/src/chart_types/goal_chart/state/selectors/get_screen_reader_data.ts
walterra: Thanks for the suggestions! I pushed some changes to simplify and addressed all the items (renamed `screenReaderTypes` to `screenReaderItems` as discussed).

Issue #2572, opened 2024-12-04
[Meta] React 18 preparation


markov00:
React 18 in StrictMode has some restriction that forces us to change the way the charting library works.

This meta issue collects the issues around strict mode we currently encountered:
- https://github.com/elastic/elastic-charts/issues/2173
- https://github.com/elastic/elastic-charts/issues/2569
- https://github.com/elastic/elastic-charts/issues/2171

markov00:
Probably blocked by Redux upgrade

PR #2703, opened 2025-08-18
chore: upgrade local `react` to version `18.x`

nickofthyme:
## Summary

Upgrades to `react@^18.x` and `react-redux@^8.x`. Later we can upgrade to `react-redux@^9.x` but this version does not support `react@16` nor `react@17`.

Need to revisit https://github.com/elastic/elastic-charts/issues/2572 after merge.

### Checklist


e2e_server/server/generate/vrt_page_template.js
nickofthyme, line 1: @markov00 the failing e2e test is because the story here...

https://ech-e2e-ci--pr-2703-zhgv47us.web.app/e2e/?path=%2Fstory%2Fcomposable--basic&knob-debug=false

...never sets `data-ech-render-complete` to `true` even though the chart is clearly rendered. Any ideas why it's just on this chart?

nickofthyme:
buildkite update screenshots

e2e_server/server/generate/vrt_page_template.js
nickofthyme: Basically whenever we re-render a story without altering the chart props. The chart call `upsertSpec` again setting the render state to `false` then once rendered it checks `deepEqual` here...

https://github.com/elastic/elastic-charts/blob/ac76ca34e2a6ba4c2dbe3a8cfd8fe38affa76382/packages/charts/src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx#L120-L122

but since the props are the same it never calls `componentDidMount` to update the status render state.

https://github.com/elastic/elastic-charts/blob/ac76ca34e2a6ba4c2dbe3a8cfd8fe38affa76382/packages/charts/src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx#L108-L118

I added a quick fix in 4637ba14c82240457941258701b7ee5ea5adaaa1, to jigger the props to render again but should be looked into further.

nickofthyme:
buildkite update screenshots

e2e_server/server/generate/vrt_page_template.js
markov00: Great thanks, I've created https://github.com/elastic/elastic-charts/issues/2713

PR #2712, opened 2025-08-30
test(enzyme): Migrate Enzyme to RTL

markov00:
## Summary

This PR, (currently on top of https://github.com/elastic/elastic-charts/pull/2703) removes Enzyme and migrate all the Enzyme tests to use RTL.

A couple of snapshot tests are removed, trusting a bit the current VRTs around the similar logic.

A small change was applied to a test configuration, that was actually not testing the real output but just the property passed to a React component.


markov00:
replaced by https://github.com/elastic/elastic-charts/pull/2714

Issue #2499, opened 2024-08-02
[CI] Move to elastic/ci buildkite implementation


nickofthyme:
Currently we use our custom buildkite implementation that was modeled after the initial buildkite implementation in kibana. Since then the elastic/ci team have built out a canonical way of configuring buildkite pipelines across kibana.

Here are the starter docs https://docs.elastic.dev/ci/getting-started-with-buildkite-at-elastic

jbudz:
Checking in on this, are there any updates / do you think we'll be able prioritize this?

nickofthyme:
@jbudz No movement yet but we can prioritize within maybe a month if that works for y'all.

jbudz:
Thanks, sounds good.

PR #2716, opened 2025-09-02
chore(ci): Migrate to Elastic-wide Buildkite infra

delanni:
## Summary
This PR adds a `catalog-info.yaml` through which a new pipeline can be created in the shared elastic infrastructure.

## Details
The PR also adds a branching logic, so that the pipeline will work more or less identical on the new infrastructure. The changes required for this are: 
 - emit to `stdout` instead of uploading steps - the runner will run the pipeline file, and upload the resulting string
 - use different agent targeting rules - there aren't agent-queues on the new infra, like we had in Kibana-buildkite

Once this PR is merged, Terrazzo will pick up the changes, create the pipeline, and the two pipelines can live parallel to each other, while we're testing. If we're happy with the pipeline, we can work on triggering and if we need to keep history of the previous job.

## Issues
Doesn't fix it just yet, but part of the work of: http://github.com/elastic/elastic-charts/issues/2499

### Checklist

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -20,7 +20,7 @@ spec:
       env:
         ELASTIC_SLACK_NOTIFICATIONS_ENABLED: "false"
         ELASTIC_BUILDKITE_INFRA: "true"
-      repository: elastic/kibana
+      repository: elastic/elastic-charts
       provider_settings:
         trigger_mode: none
       pipeline_file: ".buildkite/pipelines/pull_request/pipeline.sh"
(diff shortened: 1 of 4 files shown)


delanni:
buildkite test this

catalog-info.yaml
```diff
@@ -0,0 +1,33 @@
...
+  implementation:
+    apiVersion: buildkite.elastic.dev/v1
+    kind: Pipeline
+    metadata:
+      name: kibana / elastic-charts / build
+    spec:
+      env:
+        ELASTIC_SLACK_NOTIFICATIONS_ENABLED: "false"
+        ELASTIC_BUILDKITE_INFRA: "true"
+      repository: elastic/kibana
```
nickofthyme: Is this supposed to be `kibana` or `elastic-charts`?

nickofthyme (approved):
Nice! Thanks a lot for the help here.

catalog-info.yaml
delanni: you're right!
```suggestion
      repository: elastic/elastic-charts
```

Issue #2728, opened 2025-09-10
[Metric chart] Title disappears when font size it set to fit


mariairiartef:
**Describe the issue**
When the primary value font size is set to true, the title disappears. 

**To Reproduce**
1. Go to [storybook](https://elastic.github.io/elastic-charts/storybook/?path=/story/metric-alpha--layout&globals=toggles.showHeader:true;toggles.showChartTitle:false;toggles.showChartDescription:false;toggles.showChartBoundary:false;theme:light)
2. Change the following settings
  - Remove extra
  - Remove Subtitle 
  - Change value to 5
  - Remove value postfix
  - Set primary metric font size mode to Fit
3. See how title disappears
 
**Expected behaviour**
The elements must be visible.

**Screenshots**

<img width="526" height="421" alt="Image" src="https://github.com/user-attachments/assets/07469338-850e-4026-872c-ac93a8503ef8" />

https://github.com/user-attachments/assets/8eadc552-2fa6-49a9-8b75-ed38d5b220a3

//

<img width="1361" height="625" alt="Image" src="https://github.com/user-attachments/assets/263bf669-4a82-4627-b307-355c7939666b" />

when primary metric font size mode is set to default, the metric text compone

<img width="1404" height="614" alt="Image" src="https://github.com/user-attachments/assets/0b9772bf-e202-47d5-8d27-0332d7e5520c" />

**Checklist**

- [ ] The proper **chart type** label has been added (e.g. `:xy`, `:partition`)
- [ ] The proper **feature** labels have been added (e.g. `:interactions`, `:axis`)
- [ ] Every related Kibana issue is listed under `Kibana Cross Issues` list and the `kibana cross issue` tag is applied

mariairiartef:
I'm checking and it seems that when primary metric font size mode is set to `fit`, we use the first breakpoint. In the following example, for the given panel height, we are using an specific breakpoint. When the setting is set to fit, it uses the [first breakpoint](https://www.figma.com/design/duESbCUc00lcUq1ezF95fv/-Charts--Metric-Chart?node-id=3396-81928&m=dev).

| Default  | Fit |
| ------------- | ------------- |
| <img width="212" height="161" alt="Image" src="https://github.com/user-attachments/assets/16d47d0d-5b50-4061-a757-597b3ea52441" /> | <img width="208" height="159" alt="Image" src="https://github.com/user-attachments/assets/d9ed256e-9a63-4b8b-9e46-406316ecea34" />  |
[...]

Issue #2725, opened 2025-09-08
[Metric Chart] Update introduces unwanted differences in title and subtitle font size


gvnmagni:
Hi @mariairiartef , while making a comparison between before and after our latest update on metric chart, I noticed a behavior that I didn't fully understand.

As you can see in the following images, the Title, Subtitle and Secondary Metric font sizes are smaller in new version compared to previous version (look at the panels on the left, the biggest ones)

**Previous version**
<img width="2742" height="688" alt="Image" src="https://github.com/user-attachments/assets/a2d5215b-b3cb-4143-9978-35f6b46477b0" />

**New version**
<img width="2742" height="688" alt="Image" src="https://github.com/user-attachments/assets/2f2de1ef-876a-4d59-875d-e9b4f9979de5" />

My assumption is that this is related to the fact that now, whenever we need to resize the primary value because it doesn't fit into the panel, we move to a smaller breakpoint but we do this to all elements, while originally we did that only to the primary value. Like if they were more independent 

Did I get this correctly? Is there anything that we can do?

mariairiartef:
@gvnmagni, this is expected behavior introduced with the new layout. When calculating the height based sized, we use the smaller value between the panel’s height and width ([Math.min(panel.height, panel.width)](https://github.com/elastic/elastic-charts/blame/9b4311837413f22e089708d207fd054262805184/packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx#L170)

cc: @markov00 

Should we consider reverting to the previous approach?

gvnmagni:
Question, are the elements still independent right? I can see the title getting smaller but not the value, this is right from my perspective but just wanted to be sure... before solving the main issue

markov00:
@gvnmagni can you share the dashboard? I'd like to verify a couple of things before giving you an answer

PR #2730, opened 2025-09-11
fix(metric): fix `fit` font size mode

mariairiartef:
## Summary

- Fixes `fit` font size mode so it doesn't overflow the panel height.
- Adds **workaround** in `default` font size mode to prevent edge case transitions where tiny height differences cause layout changes due to value font size jumps.
- Reverts a change in height-based size calculation. Previously, the smaller value between the panel's height and width was used, but this update restores the original logic to rely on the panel's height.

### Fit font size mode

https://github.com/user-attachments/assets/c5317c0c-c8ca-4b1d-bf51-bee17dcb1a70

https://github.com/user-attachments/assets/312cdd6f-4526-4aa1-b2d2-c5350f94640e

#### Before

_Smaller height makes visible more elements._

| 149 px  | 151 px |
| ------------- | ------------- |
| Screenshot 2025-09-16 at 10 13 32 | Screenshot 2025-09-16 at 10 21 27 |

#### After

| 149 px  | 151 px |
| ------------- | ------------- |
| Screenshot 2025-09-16 at 10 18 05 | Screenshot 2025-09-16 at 10 20 00 |

| 99 px  | 101 px |
| ------------- | ------------- |
| Screenshot 2025-09-16 at 10 24 48 | Screenshot 2025-09-16 at 10 25 16 |

## Issues

Closes https://github.com/elastic/elastic-charts/issues/2728
Closes https://github.com/elastic/elastic-charts/issues/2725

## Checklist


packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx
mariairiartef, line 1: @gvnmagni 

> [!WARNING]  
> There is a problem regarding edge case transitions. When the panel gets smaller, more elements fit in the metric chart.


https://github.com/user-attachments/assets/f3121c7b-1c07-41e1-9399-c8483ab15693

The larger panel had a larger value font, which consumed more space, leaving less room for title, subtitle, and extra elements. So the responsive logic chose a more restrictive breakpoint.

Meanwhile, the smaller panel had a smaller value font, which consumed less space, leaving more room for other elements. So it chose a more permissive breakpoint.

----

The value font size was affecting the responsive breakpoint decision, where larger panels sometimes got more restrictive layouts than smaller ones.

## Workaround 🩹 

I added a buffer to the `isVisible` function. This prevents these edge case transitions where tiny height differences cause layout changes due to value font size jumps.

mariairiartef: One of the problems of this workaround is that it looks like we are moving to a smaller breakpoint prematurely in the bigger breakpoints 😅 

https://github.com/user-attachments/assets/27c75ae8-1145-40f1-a3cc-c2b645d2cb21

packages/charts/src/chart_types/metric/renderer/dom/_text.scss
```diff
@@ -80,6 +81,8 @@
 
     button {
       width: 95%;
+      outline: none;
+      display: block;
```
mariairiartef, line 85: This removes an extra space added between the title and the subtitle. 🤔  That's why many screenshot tests fail, because with this was removed, making the totalHeight calculation more accurate.


https://github.com/user-attachments/assets/3419ce80-0c63-4270-bf2b-9325870edbc7

packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx
```diff
@@ -106,58 +136,47 @@ const SUBTITLE_FONT: Font = {
...
+  availableHeight: number,
   textParts: TextParts[],
   minValueFontSize: number,
   hasIcon: boolean,
-): number {
-  const maxWidth = (width - 2 * PADDING) * 0.98; // small buffer to prevent clipping
+  isFitMode: boolean,
+) {
+  const maxWidth = (totalWidth - 2 * PADDING) * 0.98; // Buffer to prevent clipping
+  const maxHeight = (availableHeight - 2 * PADDING) * 0.98;
```
markov00, line 158: have you tried to use a pixel bufffer rather then a percentage one? as you already noticed here https://github.com/elastic/elastic-charts/pull/2730/files#r2344124391 the buffer increase with the increase of the chart height/width, so if we want to reduce this effect the buffer should stay the same.

packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx
```diff
@@ -276,69 +295,83 @@ const getResponsiveBreakpoints = (title: boolean, subtitle: boolean, extra: bool
   { titleMaxLines: 1, subtitleMaxLines: 0, title, subtitle: false, extra: false },
 ];
 
-function elementVisibility(
+function getLayoutResult(
```
markov00: Probably we can find a better name then `LayoutResult`? result is a bit too vague

packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx
```diff
@@ -276,69 +295,83 @@ const getResponsiveBreakpoints = (title: boolean, subtitle: boolean, extra: bool
...
+      actualExtraHeight,
+      nonValueElementsHeight,
+      totalHeight,
+    };
+  }
+
+  /** Determines if the given breakpoint should be considered "visible" for the provided text measurement */
+  const isVisible = (breakpoints: ResponsiveBreakpoints, measure: TextMeasure) => {
+    const { totalHeight } = getTextLayoutInfo(breakpoints, measure);
+    const bufferHeight = 26;
```
markov00: Ah probably this is the actual buffer, not the one in my other comment.
Why 26? is something that is determined based on some tests?
How much we can reduce this to avoid the problem you highlighted?

Would be also great to move this into a commented BUFFER_HEIGHT const.

packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx
```diff
@@ -276,69 +295,83 @@ const getResponsiveBreakpoints = (title: boolean, subtitle: boolean, extra: bool
...
-    measure: TextMeasure,
-  ) =>
-    (title && titleMaxLines > 0 ? titleHeight(titleMaxLines, measure) : 0) +
-    (subtitle && subtitleMaxLines > 0 ? subtitleHeight(subtitleMaxLines, measure) : 0) +
-    (extra ? extraHeight : 0) +
-    valueHeight +
-    PADDING +
-    // For the height calculation, take into account when there is an horizontal progress bar
-    progressBarHeight;
+  function getTextLayoutInfo(breakpoints: ResponsiveBreakpoints, measure: TextMeasure) {
```
markov00: `getTextsLayoutConfigs`? `Info` is a bit too vague

packages/charts/src/chart_types/metric/renderer/dom/titles.tsx
```diff
@@ -22,7 +22,7 @@ function lineClamp(maxLines: number): CSSProperties {
     lineClamp: maxLines,
     WebkitBoxOrient: 'vertical',
     overflow: 'hidden',
-    whiteSpace: 'pre-wrap',
+    whiteSpace: 'pre-line',
```
markov00, line 25: have you found some issues with `pre-wrap` when using multiple spaces for example?

packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx
mariairiartef: yes, I was testing manually the transitions and with this value, the issue doesn't happen

packages/charts/src/chart_types/metric/renderer/dom/titles.tsx
mariairiartef: Yes, It was adding some indentation to the subtitle:
Screenshot 2025-09-15 at 16 22 55

packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx
mariairiartef: how about `computeMetricTextSize`?

Renamed [here](https://github.com/elastic/elastic-charts/pull/2730/commits/2edeca575acdf7f554e6b2a98416fb75c8e7cc52)

packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx
mariairiartef: How about `MetricTextLayout`?

packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx
mariairiartef: I set the buffer only for smaller breakpoints, that is when we have the problem with smaller font sizes:
```
const buffer = panel.height <= SMALL_HEIGHT_THRESHOLD ? LAYOUT_PIXEL_BUFFER : 0;
```

packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx
mariairiartef: Renamed [here](https://github.com/elastic/elastic-charts/pull/2730/commits/2edeca575acdf7f554e6b2a98416fb75c8e7cc52)

packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx
mariairiartef: Yes, the pixel buffer is added in the `isVisible` method

markov00 (approved):

mariairiartef:
buildkite update screenshots

mariairiartef:
buildkite update screenshots

packages/charts/src/chart_types/metric/renderer/dom/_text.scss
mariairiartef: Add the buffer only to smaller breakpoints: https://github.com/elastic/elastic-charts/pull/2730#discussion_r2349421056

Issue #1921, opened 2023-01-03
Sort highlighter by geometry zIndex


markov00:
**Describe the issue**
In a cartesian chart, if two data points fall into the same position, the topmost highlighter covers the bottom most.
The problem is that the topmost highlighter is not the one relative to the topmost geometry, but the opposite.
If you check the attached screenshot you can see that:
- Sensor1, in green, is rendered in the background of Sensor 2, in blue.
- Sensor1 is shown on the top in the Legend and in the tooltip.
- Sensor1 highlighter is shown on top of the Sensor2 highlighter.
- 
<img width="850" alt="Screenshot 2023-01-03 at 14 46 57" src="https://user-images.githubusercontent.com/1421091/210371018-a44acb75-7b6b-43b8-8933-fef536699b16.png">

**Expected behaviour**
The rendering order of the highlighter should follow the rendering order of the geometries.

PR #2731, opened 2025-09-12
style(xy): Update area and line chart style

### LINKED ISSUE #2721: [Line and areachart] Style update
Image

### Stroke width

Bring stroke width from 2px to 1.5px (I am assuming, since we are in Canvas, that the half pixel is not an issue but please let me know in case it is)

### Dots style

Apply a 1.5 px stroke width to dots border, and fill the dots with token `colors.backgroundBasePlain`
When hovered, just fill the dot with the same stroke color (we will need to evaluate together if this is enough)

### Activate dots at smaller buckets size

We can consider displaying dots at even smaller buckets size so that they appear more often. This is also linked, I believe, to https://github.com/elastic/kibana/issues/191513

### Areachart opacity

Reduce opacity of areas to 0.2

andrimal:
## Summary

This PR updates the style of area and line charts.

## Details

The updated style:
- **line stroke width**: `1.5px`
- **area opacity**: `0.2`
- **points/ dots**:
  - **stroke width**: `1.5px`
  - **fill color**: `colors.backgroundBasePlain`
  - **isolated points**: keep the same style for isolated points when line/area points are not visible
  - **dimmed fill color**: same as the **fill color**  -> `backgroundBasePlain`
  - **focused stroke width**: same as the line/area **focused stroke width**
  - **highlight logic**: 
    - _hover-highlighting_: keep the same highlighting logic when pointer is over a geometry
    - _bucket-highlighting_: fill all the area/line points of the hovered bucket with the same stroke color
- brush opacity in dark theme: `0.2`

## Screenshots

<details>
<summary> Line Chart Screenshots </summary>

- Light theme (Before):
line_before

- Light theme (After):
line_after

- Dark theme (Before):
line_before_dark

- Dark theme (After):
line_after_dark

</details>

<details>
<summary> Area Chart Screenshots </summary>

- Light theme (Before):
area_before

- Light theme (After):
area_after

- Dark theme (Before):
area_before_dark

- Dark theme (After):
area_after_dark

</details>

## Issues

fix #2721 

### Checklist


andrimal:
buildkite update vrt

andrimal:
buildkite update vrt

andrimal:
buildkite update vrt

andrimal:
buildkite test this

gvnmagni:
A detail that I noticed  when we do NOT display dots: 

isolated data points are now represented with a white dot with stroke colored as the line. It might be better to keep as it was, a filled dot with no stroke

The overall logic for isolated point is: they should be as similar as possible to the line, because they are basically a little point of that line isolated from the rest of the series

How it is with this update:
Screenshot 2025-09-26 at 15 44 53


How it should be:
Screenshot 2025-09-26 at 15 45 12

andrimal:
> A detail that I noticed when we do NOT display dots:
> 
> isolated data points are now represented with a white dot with stroke colored as the line. It might be better to keep as it was, a filled dot with no stroke
> 
> The overall logic for isolated point is: they should be as similar as possible to the line, because they are basically a little point of that line isolated from the rest of the series
> 
Thank you @gvnmagni! Yes I checked that the isolated points actually inherit the normal point style. I'll proceed with the changes to decouple them.

andrimal:
buildkite update vrt

andrimal:
buildkite test this

andrimal:
buildkite update vrt

andrimal:
buildkite test this

gvnmagni:
The update is looking good, thank you @andrimal for the effort! I'll write down a few details that I would love to check (either next release or next days, I'll check what we can do) because there are a few tiny pieces that I would love to improve, but these are NOT a blocker from my perspective

1. Hover state for points. We re-introduced that transparent bigger dot whenever we hover a single point, works perfectly but I will probably explore an alternative where it is even bigger. Especially on light lines it is very hard to be depicted. Again, the PR works perfectly, it is just an improvement that we can do in term of style

2. Interactions: when we isolate a dataseries through the legend, the stroke-width of that line increase of a pixel (or so..), that increase is not reflected on the dot stroke-width (as you can see in the folowing image)

3. Interactions: when we isolate a dataseries, all the other dots get to be in grey colors and transparent. While I would expect the stroke of many overlapped dots to be darker,  I am not understanding why their fill seem to get darker too. Like, if it has a white fill, how does it become darker when overlapped with additional white fills? I am sure I am missing something 🙇‍♂️ 

Screenshot 2025-09-30 at 20 34 26

andrimal:
> 1. We re-introduced that transparent bigger dot whenever we hover a single point, works perfectly but I will probably explore an alternative where it is even bigger.

Yes I agree on this. When re-introduced the transparent bigger point I used the original highlighter style with `radius 5`. But we could definitely increase this easily, unless you want to rethink the overall highlighter style then we could later proceed to a follow up fix. :)   
```typescript
opacity: 0.5,
fill: ColorVariant.Series,
stroke: ColorVariant.None,
strokeWidth: 0,
radius: 5,
``` 
>2. Interactions: when we isolate a dataseries through the legend, the stroke-width of that line increase of a pixel (or so..), that increase is not reflected on the dot stroke-width (as you can see in the folowing image)

Until now, there wasn't a focused style for the dots. I introduced it now -> [7f63cee](https://github.com/elastic/elastic-charts/pull/2731/commits/7f63cee3cc4f7259211c65636b7516a18134af7e)

>3. Interactions: when we isolate a dataseries, all the other dots get to be in grey colors and transparent. While I would expect the stroke of many overlapped dots to be darker, I am not understanding why their fill seem to get darker too. Like, if it has a white fill, how does it become darker when overlapped with additional white fills? I am sure I am missing something 🙇‍♂️

This was actually not transparent. There was a dimmed point style applied to dimmed line/area charts. I changed the fill color from light gray to background base plain here -> [9484300](https://github.com/elastic/elastic-charts/pull/2731/commits/94843008e6648767b63c6ce5842e2c20587526d6)

Screenshot 2025-10-01 at 1 01 07 PM
Screenshot 2025-10-01 at 1 01 15 PM

andrimal:
buildkite update vrt

gvnmagni:
Looking good, I noticed this while playing around (https://github.com/elastic/elastic-charts/issues/2752) but it is not a problem related to this specific PR, it already existed before all this

andrimal:
> Looking good, I noticed this while playing around (#2752) but it is not a problem related to this specific PR, it already existed before all this

Yes I noticed that too. I can definitely take this as a follow up fix along with this https://github.com/elastic/elastic-charts/issues/1921 :)

markov00:
> Looking good, I noticed this while playing around (#2752) but it is not a problem related to this specific PR, it already existed before all this

I believe this is connected: we are now showing all the dots in the highlighter, but they are sorted in a different way then the way its sorted in the rendered. Note the blue series is ontop of the light green one, when hovering the highlighter shows the light green on top of the blue one. I believe is a matter of aligning the rendered point to the actual series order.
can be done also on a subsequent PR

Screenshot 2025-10-03 at 09 42 01

Screenshot 2025-10-03 at 09 42 16

andrimal:
> > Looking good, I noticed this while playing around (#2752) but it is not a problem related to this specific PR, it already existed before all this
> 
> I believe this is connected: we are now showing all the dots in the highlighter, but they are sorted in a different way then the way its sorted in the rendered. Note the blue series is ontop of the light green one, when hovering the highlighter shows the light green on top of the blue one. I believe is a matter of aligning the rendered point to the actual series order. can be done also on a subsequent PR
> 
> Screenshot 2025-10-03 at 09 42 01 Screenshot 2025-10-03 at 09 42 16

Hey @markov00 you are right, this is the other issue #1921  that I mentioned [here](https://github.com/elastic/elastic-charts/pull/2731#issuecomment-3357292306) and would be nice to handle as a fix!

packages/charts/src/utils/themes/light_theme.ts
```diff
@@ -482,6 +484,13 @@ export const LIGHT_THEME: Theme = {
...
       radius: 5,
+      bucketHighlighted: {
+        opacity: 1,
+        fill: ColorVariant.Series,
+        stroke: ColorVariant.Series,
+        strokeWidth: 1.5,
+        radius: 3,
+      },
     },
   },
```
markov00: I believe this can be cleaned up a bit for multiple reasons:
- the term `bucketHighlighted` is a bit strange and doesn't really tell much about what is going on.
- we can quickly revisit the style to make it a bit more clear, by switching the styles: the `bucketHighlighted` style should be moved at the root level (inside `highlighter.point`) because this is the actual default style for a point in the highlighter. Then inside `highlighter.point.onHover` you can have the hover style that was currently at the root level.

In this way is a bit more clear how the style is applied

packages/charts/src/chart_types/xy_chart/state/selectors/get_tooltip_values_highlighted_geoms.ts
```diff
@@ -239,3 +252,9 @@ export const getHighlightedGeomsSelector = createCustomCachedSelector(
   [getHighlightedTooltipTooltipValuesSelector],
   ({ highlightedGeometries }): IndexedGeometry[] => highlightedGeometries,
 );
+
+/** @internal */
+export const getBucketHighlightedPointsSelector = createCustomCachedSelector(
```
markov00: As for the theme, we can keep calling this just `getHighlighterPointsSelector`

packages/charts/src/chart_types/xy_chart/renderer/dom/highlighter.tsx
```diff
@@ -81,21 +90,65 @@ class HighlighterComponent extends React.Component<HighlighterProps> {
...
+        {highlightedGeometries.map((geom) => {
+          const { panel } = geom;
+          const x = geom.x + geom.transform.x;
+          const y = geom.y + geom.transform.y;
+          const geomTransform = getTransformForPanel(panel, chartRotation, chartDimensions);
+          const baseKey = `${geom.seriesIdentifier.key}-${geom.value.x}-${geom.value.y}-hovered`;
 
           if (isPointGeometry(geom)) {
-            const fillColor = getColorFromVariant(RGBATupleToString(geom.style.fill.color), style.point.fill);
+            const seriesSpec = seriesSpecs.find((spec) => spec.id === geom.seriesIdentifier.specId);
```
markov00: I'm not super happy about that, but fine for now (mostly because in some situation we need to execute this a  lot and can cause issues if the number of series specs is high)

packages/charts/src/utils/themes/light_theme.ts
andrimal: ✅  Done here [bfc1d9c](https://github.com/elastic/elastic-charts/pull/2731/commits/bfc1d9c16af0985e33f991235656bbdce77c9639)

Initially I went the the opposite approach in order to leave the highlighter behavior in the legacy themes unchanged, but now there is more consistency in the highlighter functionality across all themes.

packages/charts/src/chart_types/xy_chart/state/selectors/get_tooltip_values_highlighted_geoms.ts
andrimal: ✅  Done here [8dd6bf0](https://github.com/elastic/elastic-charts/pull/2731/commits/8dd6bf0ae49a759e177c23b6612c735ab4f23aff)

packages/charts/src/chart_types/xy_chart/renderer/dom/highlighter.tsx
andrimal: ✅ Fixed here: [1253ae9](https://github.com/elastic/elastic-charts/pull/2731/commits/1253ae9af00975851e476d2e6896a90dcd444b13) and [72c9bd5](https://github.com/elastic/elastic-charts/pull/2731/commits/72c9bd54ab709e09482210da559dd45ba7e97523)

andrimal:
buildkite update vrt

andrimal:
buildkite test this

markov00 (approved):

PR #2717, opened 2025-09-03
fix(partition): fix tooltip nested colors 

andrimal:
## Summary

The tooltip no longer shows wrong color for nested slices. This issue was caused when nested layers shared the same label and value.

Before: 

Kapture 2025-09-03 at 13 26 21

After: 

Kapture 2025-09-03 at 13 29 58

## Issues
fix #2574 

### Checklist

- [ ] The proper **chart type** label has been added (e.g. `:xy`, `:partition`)
- [ ] The proper **feature** labels have been added (e.g. `:interactions`, `:axis`)
- [ ] All related issues have been linked (i.e. `closes #123`, `fixes #123`)
- [ ] New public API exports have been added to `packages/charts/src/index.ts`
- [ ] Unit tests have been added or updated to match the most common scenarios
- [ ] The proper documentation and/or storybook story has been added or updated
- [ ] The code has been checked for cross-browser compatibility (Chrome, Firefox, Safari, Edge)
- [ ] Visual changes have been tested with `light` and `dark` themes

PR #2742, opened 2025-09-23
fix(partition): fix tooltip nested colors  (#2717) [69.x]

andrimal:
 - fix(partition): fix tooltip nested colors  (#2717)


andrimal:
buildkite test this

andrimal:
buildkite update vrt

andrimal:
buildkite test this

Issue #1587, opened 2022-02-10
Performance: Formatter is called for each data point in bar chart even if not used


flash1293:
**Describe the issue**
In https://github.com/elastic/elastic-charts/blob/master/packages/charts/src/chart_types/xy_chart/rendering/bars.ts#L92 the provided formatter is called for each data point in the series, but it's only used in special circumstances:

https://github.com/elastic/elastic-charts/blob/master/packages/charts/src/chart_types/xy_chart/rendering/bars.ts#L96

In case of charts with lots of series and slow formatting functions provided this can have a noticable performance impact.

E.g. in the following chart it's spending 45ms in `getConvertedValue` (out of ~300ms total blocking rendering time):
<img width="1277" alt="Screenshot 2022-02-10 at 12 36 32" src="https://user-images.githubusercontent.com/1508364/153401293-045186ad-9447-4248-9e7b-48ad27d3214e.png">

**To Reproduce**
Configure a bar chart without labels on the bars and observe the provided formatter being called.

**Expected behaviour**
Formatter should only be called if necessary to render the chart correctly

markov00:
Thanks Joe, yep, this can definitely be improved a lot

markov00:
Should be fixed by https://github.com/elastic/elastic-charts/pull/2754

PR #2754, opened 2025-10-15
perf(xy): avoid computing bar value display width

markov00:
## Summary

This PR improves a bit the geometries rendering logic by avoiding some computations.
In particular it avoids computing the bar values to display if the option is not enabled.
Not a huge gain in performance, but under heavy load this can save some cycles.

fix https://github.com/elastic/elastic-charts/issues/1587

### DIFF
--- a/packages/charts/src/chart_types/xy_chart/rendering/bars.ts
+++ b/packages/charts/src/chart_types/xy_chart/rendering/bars.ts
@@ -98,15 +98,7 @@ export function renderBars(
       !(displayValueSettings?.isAlternatingValueLabel && barGeometries.length % 2);
     const barGeometry: BarGeometry = {
       displayValue: shouldDisplayValue
-        ? computeDisplayValue(
-            y1Value,
-            barGeometries,
-            sharedSeriesStyle,
-            measureText,
-            chartRotation,
-            width,
-            displayValueSettings,
-          )
+        ? computeDisplayValue(y1Value, sharedSeriesStyle, measureText, chartRotation, width, displayValueSettings)
         : undefined,
       x,
       y: yScreenSpaceCoord,
@@ -146,7 +138,6 @@ export function renderBars(
 
 function computeDisplayValue(
   y1Value: number | null,
-  barGeometries: BarGeometry[],
   sharedSeriesStyle: BarSeriesStyle,
   measureText: TextMeasure,
   chartRotation: number,


packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -47,11 +47,10 @@ export function renderBars(
   styleAccessor?: BarStyleAccessor,
   stackMode?: StackMode,
 ): BarTuple {
-  const { fontSize, fontFamily } = sharedSeriesStyle.displayValue;
   const initialBarTuple: BarTuple = { barGeometries: [], indexedGeometryMap: new IndexedGeometryMap() } as BarTuple;
   const y1Fn = getY1ScaledValueFn(yScale);
   const y0Fn = getY0ScaledValueFn(yScale);
-
+  const seriesIdentifier: XYChartSeriesIdentifier = getSeriesIdentifierFromDataSeries(dataSeries);
```
mariairiartef: nit:  the explicit type is redundant, I think we can remove so we keep types aligned automatically
```suggestion
  const seriesIdentifier = getSeriesIdentifierFromDataSeries(dataSeries);
```

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -169,6 +134,60 @@ export function renderBars(
...
+    displayValueSettings?.overflowConstraints ?? [
+      LabelOverflowConstraint.ChartEdges,
+      LabelOverflowConstraint.BarGeometry,
+    ],
+  );
+
+  // Based on rotation scale the width of the text box
+  const bboxWidthFactor = isHorizontalRotation ? textScalingFactor : 1;
+
+  return displayValueText && displayValueSettings?.showValueLabel
```
mariairiartef: nit: `displayValueSettings?.showValueLabel` is checked twice, maybe be can remove it from here and leave it only in L90, so we avoid repetition

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
```diff
@@ -169,6 +134,60 @@ export function renderBars(
...
+  measureText: TextMeasure,
+  chartRotation: number,
+  width: number,
+  displayValueSettings?: DisplayValueSpec,
+): BarGeometry['displayValue'] | undefined {
+  const { fontSize, fontFamily } = sharedSeriesStyle.displayValue;
+  const formattedDisplayValue = displayValueSettings?.valueFormatter?.(y1Value);
+
+  // only show displayValue for even bars if showOverlappingValue
+  const displayValueText =
```
mariairiartef: suggestion: this is also checked in L179, so if  `displayValueText` is `undefined` we return `undefined`. Maybe we can add an early return here so we avoid computing  what it is below.

Or, another thing could be extract that part into:

```
const shouldDisplayValue = displayValueSettings?.showValueLabel && !(displayValueSettings?.isAlternatingValueLabel && barGeometries.length % 2)
```

and then:

```
    const barGeometry: BarGeometry = {
      displayValue: shouldDisplayValue
        ? computeDisplayValue(
            y1Value,
            barGeometries,
            sharedSeriesStyle,
            measureText,
            chartRotation,
            width,
            displayValueSettings,
          )
        : undefined,
      x,
```

packages/charts/src/chart_types/xy_chart/rendering/points.ts
```diff
@@ -62,35 +62,34 @@ export function renderPoints(
   const needSorting = !markSizeOptions.enabled;
 
   let style = buildPointGeometryStyles(color, pointStyle);
-  let styleOverrides: RecursivePartial<PointStyle> | undefined = undefined;
+  const seriesIdentifier: XYChartSeriesIdentifier = getSeriesIdentifierFromDataSeries(dataSeries);
```
mariairiartef: same here ♻️ 
```suggestion
  const seriesIdentifier = getSeriesIdentifierFromDataSeries(dataSeries);
```

packages/charts/src/utils/bbox/canvas_text_bbox_calculator.ts
```diff
@@ -23,14 +23,14 @@ export type TextMeasure = (text: string, font: Omit<Font, 'textColor'>, fontSize
 
 /** @internal */
 export function measureText(ctx: CanvasRenderingContext2D): TextMeasure {
-  return (text, font, fontSize, lineHeight = 1) =>
-    withContext(ctx, (ctx): Size => {
-      if (text.length === 0) {
-        // TODO this is a temporary fix to make the multilayer time axis work
```
mariairiartef, line 29: qq: why did you remove this comment, did we fix it?

mariairiartef (commented):
🐼 I added some suggestions

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
markov00: fixed https://github.com/elastic/elastic-charts/pull/2754/commits/0e66ece8148d049a52c02973ed0c74ccbe6f1f13

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
markov00: fixed https://github.com/elastic/elastic-charts/pull/2754/commits/b4d62f7ed8eccc969a33f661cdbb6d70eea16a5c

packages/charts/src/chart_types/xy_chart/rendering/bars.ts
markov00: fixed https://github.com/elastic/elastic-charts/pull/2754/commits/b4d62f7ed8eccc969a33f661cdbb6d70eea16a5c

packages/charts/src/chart_types/xy_chart/rendering/points.ts
markov00: fixed https://github.com/elastic/elastic-charts/pull/2754/commits/fd263b2413ef899fc244997772b1bc64f1d333ad

packages/charts/src/utils/bbox/canvas_text_bbox_calculator.ts
markov00: I don't consider this as a temporary fix anymore, the logic is valid: if the text doesn't have a length then we don't need to compute actual size, and we can just return a zero width measureed text

mariairiartef: makes sense, thanks!

PR #2756, opened 2025-10-24
fix(highlighter): Fix sorting logic for highlighted geometries

### LINKED ISSUE #1921: Sort highlighter by geometry zIndex

### LINKED ISSUE #2752: [Linechart & Areachart] Hovered dots end up having weird colors due to overlapping
This issue happens all the times but it is way more noticeable when we have a big number of lines and therefore a lot of dots potentially overlapping with each other.

**First issue**
When we hover a time bucket there is a new z-index logic that gets applied that changes the order of elements, making dots to change color

**Second issue**
When we hover a single dot, since we have many dots in the same place and we applied a transparent bigger dot on top of them (one for each dots that exists) we end up having a sort of sum of these transparent dots which generate weird colors

Image

This is particularly weird when interacting with the chart because it keeps on gets unexpected changes

Image

andrimal:
## Summary

The highlighter now respects the series sorting order.

## Details
When 2 or more point geometries overlap at the same position (`x, y, radius`):

- **Highlighted points**: They are pre-sorted by `sortOrder` so the topmost point renders last and its color appears on top
- **Hovered geometries**: They are now deduplicated to keep only the topmost point (highest `sortOrder`), preventing semi-transparent highlight colors from mixing

## Screenshots
Kapture 2025-10-31 at 12 04 31

### Multiple overlapping points:
Before:
Screenshot 2025-10-30 at 12 32 40 PM

After: 
Screenshot 2025-10-31 at 12 05 39 PM

Screenshot 2025-10-31 at 12 03 22 PM

## Issues
fix #1921 and fix #2752

### Checklist


andrimal:
buildkite update vrt

packages/charts/src/chart_types/xy_chart/renderer/canvas/lines.ts
```diff
@@ -43,7 +43,7 @@ export function renderLines(
         };
       })
       // sort by dimmed first once are rendered ontop of the non-highlighted ones
-      .sort(({ highlightState }) => (highlightState === 'dimmed' ? -1 : 1))
+      .sort((a, b) => (a.highlightState === b.highlightState ? 0 : a.highlightState === 'dimmed' ? -1 : 1))
```
andrimal, line 46: In line charts, this was causing an inconsistent ordering across browsers (Firefox vs Chrome/Safari/Edge). The line charts in Firefox were not rendered based on the `sortOrder` and the point highlighter color ended up wrong.

Firefox: 
Screenshot 2025-10-30 at 4 34 02 PM

Chrome: 
Screenshot 2025-10-30 at 4 34 33 PM

e2e/tests/interactions.test.ts
```diff
@@ -420,12 +420,23 @@ test.describe('Interactions', () => {
...
     );
   });
-  // currently wrong due to https://github.com/elastic/elastic-charts/issues/1921
-  test('highlighter zIndex should respect geometry zIndex', async ({ page }) => {
-    await common.expectChartWithMouseAtUrlToMatchScreenshot(page)(
-      'http://localhost:9001/?path=/story/test-cases--highlighter-z-index',
-      { left: 247, top: 76 }, // mouse over the second point
-    );
+
+  test.describe('Highlighting', () => {
```
mariairiartef, line 424: nit: I'd suggest not to add this under a describe, so we can compare the screenshots. The screenshots were added [here](https://github.com/elastic/elastic-charts/pull/1922) so after fixing the issue we can asses that it has been fixed. Otherwise, you can remove the previously created screenshot

andrimal: Nice catch @mariairiartef, I forgot to clean the old one 💯

e2e/screenshots/interactions.test.ts-snapshots/interactions/highlighting/should-correctly-highlight-the-points-and-the-tooltip-when-hovering-over-overlapping-point-geometries-chrome-linux.png
mariairiartef, line 1: nit/out of scope:

I've noticed that when we have 10 series (that is the max value that allows to show highlighted values in the tooltip) the last one appears cut.

Image

Do you think we should increase the height or change the max value to only 9 series?

andrimal: From what I checked the tooltip height computation is happening in the `elastic-charts/packages/charts/src/components/tooltip/components/tooltip_table.tsx` file:
```typescript
const TOOLTIP_ITEM_HEIGHT = 20;
const TOOLTIP_HEADER_HEIGHT = 25;
const TOOLTIP_FOOTER_HEIGHT = 25;

/** @internal */
export function computeTableMaxHeight<D extends BaseDatum = Datum, SI extends SeriesIdentifier = SeriesIdentifier>(
  pinned: boolean,
  columns: TooltipTableColumn<D, SI>[],
  maxHeight: CSSProperties['maxHeight'],
  maxItems?: number,
): CSSProperties['maxHeight'] {
  if (pinned || isNil(maxItems)) return maxHeight;
  const headerHeight = +columns.some((c) => c.header) * TOOLTIP_HEADER_HEIGHT;
  const bodyHeight = (Math.max(maxItems, 1) + 0.5) * TOOLTIP_ITEM_HEIGHT;
  const footerHeight = +columns.some((c) => c.footer) * TOOLTIP_FOOTER_HEIGHT;
  return headerHeight + bodyHeight + footerHeight;
}
``` 
The problem is that `TOOLTIP_HEADER_HEIGHT`, `TOOLTIP_ITEM_HEIGHT`, and `TOOLTIP_FOOTER_HEIGHT` are hard-coded.
-  If I update them to match the values I measured in the default tooltip (`22` for items, `26` for the header), the bug goes away, but it still relies on constants that might be wrong for other tooltip variants.

Given that, I’d prefer to track this in a separate issue rather than mix it into the highlighter work. What do you think?

mariairiartef: Sure, thank you!

andrimal:
buildkite update vrt

andrimal:
buildkite test this

PR #2771, opened 2026-01-19
feat(legend): Add option to truncate legend labels in the middle

### LINKED ISSUE #1925: Allow text truncation in the middle of the text or at the beginning
When text truncation is essential (too long label, no possibility to wrap) we use it to truncate labels.
Today, this truncation happens only at the end of the line, but there are cases where the truncation is better suited at the beginning or at the middle of the line.
The reason for this is to provide truncation support for text where the most relevant part of the text is at the end of the line or at both at the beginning and end. 

This should be part of the truncation options we have in the `wrapText` function.

To discuss:
- [ ] should we allow such middle/beginning truncation for multi-line wrappings? There are similar examples in the literature where a `[...]` is added in the middle of the text to "skip" part of a quoted discussion. Sometimes is used at the beginning of the quite, some other at the end of it. I'm not fully sure we need to introduce such complexity in a data visualization.

awahab07:
## Summary

Legend labels can now be truncated in the middle, preserving both the beginning and end of long labels for better readability. This is especially useful for service names, URLs, and file paths where distinguishing information often appears at both ends. ~Note that the middle truncation will be the default behavior going forward.~ (After a [discussion](https://github.com/elastic/elastic-charts/pull/2771#issuecomment-4177482829), it's been decided not to set it by default, so the existing behavior, truncationPosition`'end'` will be set by default)

The new `truncationPosition` option in `theme.legend.labelOptions` supports:
- **`'middle'`**: Truncates in the middle → `enterprise-au…ion-service`
- **`'end'`** (default): Traditional CSS truncation → `enterprise-authentication-an…`

The story **Legend -> Label Truncation** demonstrate the feature:
image

https://github.com/user-attachments/assets/bdfd81b0-edaa-40f4-9733-6e9ba862eeb0

https://github.com/user-attachments/assets/a73ca36c-5c6a-4542-8606-d6f222b10d5c

## Details

- Added `truncationPosition` property to `LegendLabelOptions` interface (`'middle'` | `'end'`)
- **Iterative refinement algorithm**: Uses canvas `measureText` with max 5 iterations for accurate text fitting
- **Cross-browser compatibility**: Tested on Chrome, Firefox, Safari, and Edge
  - Safari: Added `resetWrapText` mixin to fix single-line end truncation
  - All browsers: `line-break: anywhere` for accurate multiline breaking
- The following stories have also been updated to show **truncationPosition** knob
  - Legend -> Piechart
  - Legend -> Actions
  - Legend -> Single Series
  - Legend -> Tabular Data

### Usage

```tsx
<Settings
  showLegend
  theme={{
    legend: {
      labelOptions: {
        maxLines: 1,
        truncationPosition: 'middle', // 'middle' | 'end', default: 'end'
      },
    },
  }}
/>
```

## Issues

Implements the upstream (elastic-charts) side of #1925

### Checklist


delanni:
buildkite test this

packages/charts/src/components/legend/use_truncated_label.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    const element = labelRef.current;
+    if (!element) {
+      setTruncatedLabel(label);
+      return;
+    }
+
+    const computedStyle = window.getComputedStyle(element);
+    const { font, fontSize } = getFontFromComputedStyle(computedStyle);
+    const containerWidth = element.clientWidth;
```
markov00: This triggers page reflows and you don't need these as you already have these infos:
- the font and fontSize are already around
- the legend width is already computed

packages/charts/src/components/legend/use_truncated_label.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    return () => {
+      resizeObserver.disconnect();
+    };
+  }, [computeTruncatedLabel, shouldTruncate, label]);
+
+  return {
+    labelRef: shouldTruncate ? labelRef : undefined,
+    truncatedLabel,
+  };
```
markov00: when computing things with canvas measureText you don't need all that, you can do it upfront

awahab07:
When available width is too narrow (only a few characters can fit), middle truncation won't happen ([see](https://github.com/elastic/elastic-charts/pull/2771/commits/884f1c8e97e42371fc4d74173d5da1e2c3c5ac33)). As otherwise the legend labels end up showing `'...x'` or `...`. 

image

This considers character count rather than actual width. So if the width is narrow, but multiple lines can fit more characters, middle truncation would still occur.
image

awahab07:
buildkite update vrt

awahab07:
buildkite update vrt

awahab07:
This PR now needs to address merge conflicts with https://github.com/elastic/elastic-charts/pull/2784. The functionality may have conflicting behaviors as both PRs target legend rendering, thus the who functionality needs to be tested again.

awahab07:
buildkite update vrt

awahab07:
@maryam-saeidi given the change in https://github.com/elastic/elastic-charts/pull/2784 is of the same nature and had conflicts, could you smoke tests if things work as expected? (I've tested the changes introduced in this PR and they LGTM)

awahab07:
buildkite update vrt

awahab07:
While addressing some edge cases in this PR, found some pre-existing issues which need to fixed in order to move forward with this implementation.

They are reproducible in Lens too e.g. 
1. When large legend labels are present and **Line limit** is 1, and legend **Position** is "Inside", legends go off the chart boundary:
    Lens-multi-line-large-labels-are-incorrect

    This is not the case when **Line limit** is > 1
    Lens-multi-line-large-labels-are-correct


2. There's another behavior (not a bug but the way it's currently implemented) which affects this PR and need to be tackled is, when large legend labels are present, sizing doesn't affect the legend width:

    https://github.com/user-attachments/assets/5e02186d-07e2-4cbd-a5eb-8c15699c8de7

awahab07:
buildkite update vrt

awahab07:
Both issues mentioned in https://github.com/elastic/elastic-charts/pull/2771#issuecomment-4046309209 has been dealt with.

> 1. When large legend labels are present and Line limit is 1, and legend Position is "Inside", legends go off the chart boundary:

Resolved in https://github.com/elastic/elastic-charts/pull/2771/commits/5d6a890cd50419c3a1fc9c6d2fb806f142cc79c9

<table><tr><th>Before</th><th>After</th></tr><tr><td>

1925_large_legends_going_off_the_chart

</td><td>

1925_large_legends_going_off_the_chart_FIX

</td></tr></table>

> 2. There's another behavior (not a bug but the way it's currently implemented) which affects this PR and need to be tackled is, when large legend labels are present, sizing doesn't affect the legend width:

Resolved in https://github.com/elastic/elastic-charts/pull/2771/commits/aa42c29960fbd534cdac7146c6081f5738f15aac

<table><tr><th>Before</th><th>After</th></tr><tr><td>

1925_width_does_not_apply_on_large_legends


</td><td>

1925_width_does_not_apply_on_large_legends_FIX


</td></tr></table>

gvnmagni:
@awahab07 pointed out a very important detail, which is that all existing legends would be affected by this new settings if we apply it by default. Therefore, I would feel safer to keep `truncation at the end` as default option, just for now, and evaluate eventually how and when apply the truncation at the middle (maybe when we recognize very long labels that are similar at the end, something smart), but that should probably happens at Lens level.

Long story short, I suggest to keep `truncation at end` as default while having `truncation in the middle` as an additional option available to user

awahab07:
buildkite test this

awahab07:
buildkite test this

awahab07:
buildkite update vrt

e2e/screenshots/legend_stories.test.ts-snapshots/legend-stories/legend-inside-chart/should-correctly-display-bottom-left-in-multiline-mode-chrome-linux.png
awahab07, line 1: ℹ️ These area chart snapshots have been updated because the PR fixes a behavior where legend labels overflow the chart in floating mode (legend inside chart layout). Notice the legends in the before snapshots extending out the chart borders. ([see](https://github.com/elastic/elastic-charts/pull/2771#issuecomment-4061893086))

packages/charts/src/state/selectors/get_legend_size.ts
```diff
@@ -94,8 +94,10 @@ export const getLegendSizeSelector = createCustomCachedSelector(
       const scrollBarDimension = legendHeight > parentDimensions.height ? SCROLL_BAR_WIDTH : 0;
       const staticWidth = spacingBuffer + actionDimension + scrollBarDimension;
 
+      const minWidth = parentDimensions.width * 0.05;
```
awahab07, line 97: Setting 30% of total chart width as minimum for legend size is quite a large number, and may not be desirable for cases where just a few letters needs to be shown in legends. Hence reducing to 5%. (See for [example](https://github.com/elastic/elastic-charts/pull/2771/changes#diff-fcdb2256006b078e6bebafeda9b9081ec7cafc25a8b8de5c1f40a4149fd2cf81))

packages/charts/src/components/legend/style_utils.ts
awahab07, line 1: [Reason](https://github.com/elastic/elastic-charts/pull/2771#issuecomment-4061893086) for style adjustments in this file (and also in [`.../legend/position_style.ts`](https://github.com/elastic/elastic-charts/pull/2771/changes#diff-6908bd77a2388abd332afe6b7d8463d9ac80df02721818b547ae0e5d1226df8bR41) and [`.../legend/legend.tsx`](https://github.com/elastic/elastic-charts/pull/2771/changes#diff-b280b100797cea8f5e3a10197333b3e947a1fbc5d456e6a9de65db67b0061afeR112) above).

e2e/screenshots/legend_stories.test.ts-snapshots/legend-stories/custom-width/position-left/should-limit-width-to-min-of-30-of-computed-width-chrome-linux.png
awahab07, line 1: This snapshot test has been renamed, hence old snapshot is deleted (also see a similar "../position-right/..." below).

packages/charts/src/components/legend/use_truncated_label.ts
```diff
@@ -0,0 +1,251 @@
...
+ *
+ * Runs on every render with an early bailout when the container width
+ * hasn't changed.
+ *
+ * External resizes (window, chart, knobs) cause React re-renders via
+ * Redux state updates, which should naturally re-trigger this measurement.
+ *
+ * @internal
+ */
+export function useMiddleTruncatedLabel({
```
walterra, line 177: should this also include some debouncing check? or can we check that all external resizes are decounbed already?

packages/charts/src/components/legend/_legend_item.scss
```diff
@@ -110,7 +110,13 @@ $legendItemHeight: #{$euiFontSizeXS * $euiLineHeight};
     @include wrapText;
 
     &--singleline {
+      @include resetWrapText;
       @include euiTextTruncate;
+
+      // When using middle truncation, JS handles the ellipsis. `clip` will disable euiTextTruncate
+      &--middle {
```
walterra, line 117: Nit: This creates `--singleline--middle` which isn't proper BEM terminology. Could we do `--middle` as a separate class and apply both or would that breat things?

e2e/tests/legend_stories.test.ts
```diff
@@ -348,7 +348,21 @@ test.describe('Legend stories', () => {
...
+    });
+
+    test('should correctly render middle truncation with floating legend and multiline', async ({ page }) => {
+      await common.expectChartAtUrlToMatchScreenshot(page)(
+        'http://localhost:9001/?path=/story/legend--label-truncation&knob-Inside chart (floating)_Legend options=true&knob-Horizontal alignment_Legend options=left&knob-maxLines_Label options=2&knob-Hide short labels_Data=true&knob-enable legend size_Legend options=true&knob-legend size_Legend options=600',
+      );
+    });
+  });
+
+  http: test.describe('Legend tabular data', () => {
```
markov00, line 365: interesting fist time seeing a label used in JS. I don't think this is used at all

packages/charts/src/components/legend/use_truncated_label.ts
```diff
@@ -0,0 +1,251 @@
...
+      setIsComputed(false);
+      lastMaxLinesRef.current = maxLines;
+    }
+
+    lastWidthRef.current = width;
+
+    if (width <= 0) return;
+
+    const computedStyle = window.getComputedStyle(element);
+    const { font, fontSize } = getFontFromComputedStyle(computedStyle);
```
markov00, line 231: Why do we need to compute the font size? can we use a static one?
if this is because we depend on EUI and other font configured outside, we can probably bring this font configuration inside charts and be sure we don't need to compute it.
This is specially computationally intestive since it also run inside a useLayoutEffect()

packages/charts/src/state/selectors/get_legend_size.ts
markov00: we can probably set a fixed min in width? instead of a percentage, because for smaller chart this can be very small

packages/charts/src/components/legend/use_truncated_label.ts
markov00: Considering that this also run for each label, it seems too much. If really required this should be moved at an higer level

packages/charts/src/components/legend/use_truncated_label.ts
```diff
@@ -0,0 +1,251 @@
...
+      lastWidthRef.current = 0;
+      lastLabelRef.current = label;
+      lastMaxLinesRef.current = maxLines;
+      return;
+    }
+
+    const element = labelRef.current;
+    if (!element) return;
+
+    const width = getStableContainerWidth(element);
```
markov00, line 207: In a condition where we have many labels this connected with the `getComputedStyle` below and `useLayoutEffect` create a regression in performance for the legend.
Can't we derive this value or at least derive it just one time for the entire legend instead for each?

PR #2079, opened 2023-06-23
fix: remove unused redux dev tools

markov00:
## Summary

The redux dev tools are not actively used in ECH developments and our state needs to be sanitized before using them correctly. So for now I prefer to remove their usage from charts and bring it back when https://github.com/elastic/elastic-charts/issues/2078 is fixed.

### Checklist

- [x] The proper **chart type** label has been added (e.g. `:xy`, `:partition`)
- [x] All related issues have been linked (i.e. `closes #123`, `fixes #123`)

Issue #2078, opened 2023-06-23
Sanitize redux state


markov00:
**Describe the issue**

Our current state is not fully serializable, in particular, the `internalChartState` property is a non-serializable property.

This is against the good practices of redux and can cause issues also for other users using redux-dev-tools.

The cleanest solution is to just map the configured chart type in the state and use a re-reselect selector to return the actual unserializable object.

PR #2772, opened 2026-01-19
fix: disable redux dev tools in code produced by yarn build

kqualters-elastic:
## Summary

This pr disables redux dev tools in code produced via yarn build, but keeps it enabled for running this package locally.

Screenshot of transpiled code:
image
image

## Details

When running kibana locally in dev mode, especially in the security solution plugin on pages with multiple instances of elastic charts, the combination of https://github.com/elastic/elastic-charts/issues/2078 and multiple different redux stores makes dev tools almost completely unusable for development. This pr disables the devtools for yarn build, by adding a new script in scripts/replace_dev_constants.js, that is just a flag that defaults to true to not break any other flow that might exist besides when the code is run from dist output. The dev tools were removed for a while in https://github.com/elastic/elastic-charts/pull/2079 but recently added back in https://github.com/elastic/elastic-charts/pull/2703 and the performance degradation in security solution is very noticeable.

### Checklist


packages/charts/src/state/chart_state.ts
```diff
@@ -176,9 +177,11 @@ export const createChartStore = (chartId: string, title?: string, description?:
...
         serializableCheck: false,
       }),
-    devTools: {
-      name: `@elastic/charts - ${chartId}`,
-    },
+    devTools: ELASTIC_CHARTS_DEV_MODE
+      ? {
+          name: `@elastic/charts - ${chartId}`,
+        }
+      : undefined,
```
markov00: I'd like to suggest a different approach, that drastically remove the need of a post-build step.

Let's enable the `devTools` in this library only on demand by checking if a env variable is available, if not then the `devTools` are not enabled.

```suggestion
    devTools: process.env.ELASTIC_CHARTS_DEV_MODE
      ? {
          name: `@elastic/charts - ${chartId}`,
        }
      : undefined,
```

In this way we can:
- enable the `devTools` on demand, both when working from within the chart repo and from Kibana
- the chart `devTools` are disabled by default
- we don't need a post-build step that just increase possible errors and maintenance work

If you agree we can just document this in the chart wiki

Issue #2444, opened 2024-05-30
[Legend] Show data series icon for filters only at hover


gvnmagni:
I would love to consider if it is possible to hide that icon on the right side of each data series and to show it only at mouse-hover, this way we would clean up the legend from unnecessary and repetitive elements

Current state:
<img width="232" alt="Screenshot 2024-05-30 at 11 53 39" src="https://github.com/elastic/elastic-charts/assets/18484576/ca191cd7-f57d-421a-80fd-359a1c22a0cc">

Suggested state:
<img width="232" alt="Screenshot 2024-05-30 at 11 58 36" src="https://github.com/elastic/elastic-charts/assets/18484576/f0fa4f16-49a9-4268-a139-bd2d6a2af2da">

@1Copenut any accessibility issues that we might encounter?

1Copenut:
I'm okay making these Action menus appear on `::hover` as long as they also appear on `::focus`. Practically speaking it means hiding them visually until hovered or focused, then showing. I'm okay going this direction as long as we have a chance to test for keyboard and screen reader usability before the change is merged, or quickly after merge in case we need to roll back.

`EuiDataGrid` has this behavior, but also allows for users to press `Enter` to open the `[ ... ]` menu. That's a standard data grid interaction pattern that presents similarly for mouse interaction but is ever so nuanced for keyboard interaction.

gvnmagni:
Thank you @1Copenut!

maryam-saeidi:
@gvnmagni I tried a quick implementation to see how it looks in this [PR](https://github.com/elastic/elastic-charts/pull/2770) and I have a question: (I am trying to figure out how to build a storybook link so that you can check this in action)

I was wondering if the goal is to just empty the space and show the action on hover or was the original idea to also show more text in this case (related to the `Legend > Action` storybook shown below, but the experience doesn't look great).

https://github.com/user-attachments/assets/8ae340a3-57c7-4f08-8863-d6a6263a3ac7

Another Q: Do we want to have the same for tabular data legend?

gvnmagni:
After conversation, let's just hide the icon and leave the white/empty space as it is

PR #2773, opened 2026-01-20
feat(legend): add legendActionOnHover option to show actions on hover

maryam-saeidi:
## Summary

The legendActionOnHover prop is now available in the Settings component to control when legend action icons are displayed. When set to true, legend actions (like filter buttons) are hidden by default and only appear when hovering over a legend item, providing a cleaner and less cluttered legend appearance. The action remains visible while interacting with it and supports full keyboard accessibility.

https://github.com/user-attachments/assets/94705ba9-3490-4205-9d58-fd24a9efbb9a

#### Example usage
```
      <Chart>
        <Settings
          showLegend
          legendActionOnHover // <-- new prop
          legendPosition={Position.Bottom}
          legendAction={LegendAction}
        />
        <Axis id="bottom" position={Position.Bottom} />
        <Axis id="left" position={Position.Left} />
        <BarSeries
          id="bars"
          yScaleType={ScaleType.Linear}
          xAccessor={0}
          yAccessors={[1]}
          data={[
            ["A", 10],
            ["B", 20],
            ["C", 15],
          ]}
        />
      </Chart>
```

## Details

- Added legendActionOnHover boolean prop to Settings (default: false)
- Created useActionFocusManagement hook to manage focus state for legend actions
- Added CSS classes echLegendItem--actionOnHover and echLegendItem--isActive to control visibility
- Action visibility is triggered by :hover, :focus-within, or active state (when clicked)
- Handles edge case where modals/popovers restore focus to their trigger after closing
- Works with both regular and tabular legend layouts

|Scenario|Screenshot|
|---|---|
|Regular action|image|
|Tabular action|image|
| Keeping selected action visible|image|

## Issues

Closes https://github.com/elastic/elastic-charts/issues/2444

### Checklist

🤖 This pull request was assisted by Cursor

### DIFF
--- a/packages/charts/src/components/legend/_legend_item.scss
+++ b/packages/charts/src/components/legend/_legend_item.scss
@@ -74,6 +74,8 @@ $legendItemHeight: #{$euiFontSizeXS * $euiLineHeight};
   &--actionOnHover {
     .echLegendItem__action {
       opacity: 0;
+      // Not using pointer-events: none here because it causes a rendering
+      // bug in Firefox. See https://github.com/elastic/elastic-charts/pull/2773#discussion_r2792933172
     }
 
     &:hover,


packages/charts/src/components/legend/legend_table/legend_table_item.tsx
```diff
@@ -97,7 +111,9 @@ export const LegendListItem: React.FC<LegendItemProps> = (props) => {
         })}
         {ActionComponent && (
           <LegendTableCell>
-            <div className="echLegendItem__action">{ActionComponent}</div>
+            <div className="echLegendItem__action" ref={itemRef} onPointerDownCapture={() => setIsActive(true)}>
```
markov00: Is that required? i believe if the pointer-event: none is removed and you have the &:hover css change it should be fine and doesn't need an active class.

packages/charts/src/components/legend/legend_item.tsx
```diff
@@ -126,7 +140,7 @@ export const LegendListItem: React.FC<LegendItemProps> = (props) => {
             )
           : null}
         {Action && (
-          <div className="echLegendItem__action">
+          <div className="echLegendItem__action" ref={itemRef} onPointerDownCapture={() => setIsActive(true)}>
```
markov00: Same comment as the other legend item

packages/charts/src/components/legend/_legend_item.scss
```diff
@@ -60,6 +60,8 @@ $legendItemHeight: #{$euiFontSizeXS * $euiLineHeight};
     align-items: center;
     height: $legendItemHeight;
     max-width: calc(#{$legendItemHeight} + #{$euiSizeXS * 2});
+    opacity: 0;
+    pointer-events: none;
```
markov00: this causes firefox rendering in a weird way, we can probably omit that since is anyway shown on hover

packages/charts/src/components/legend/_legend_item.scss
```diff
@@ -71,6 +73,15 @@ $legendItemHeight: #{$euiFontSizeXS * $euiLineHeight};
     }
   }
 
+  &:hover,
+  &--isActive
```
markov00: we need to make it visible also when the element is on focus, because if not moving through the legend with the keyboard you are going to miss the focus around the legend item

maryam-saeidi: Good point, added `focus-within` as well.

packages/charts/src/components/legend/legend_table/legend_table_item.tsx
maryam-saeidi: I've added the active class to keep the action visible when it is clicked, otherwise it looks like this:

https://github.com/user-attachments/assets/1e55db8f-a9d8-4b5f-af95-65e21a1a5e6a

So the behavior is similar to [EUI data grid](https://eui.elastic.co/docs/components/data-grid/):

image

maryam-saeidi:
@markov00 In the current implementation, when moving between cells with keyboard, the action remains visible afterwards as seen below (when clicked, only that action will be visible):

https://github.com/user-attachments/assets/63ebbc37-9dd5-405a-80ff-a2604ea43ac6

I tried briefly to see if I can make keyboard experience similar to click with mouse, but there are so many edge cases and I didn't find a straightforward way to do so. I think the current implementation is good enough for the keyboard scenario, wdyt?

maryam-saeidi:
buildkite update vrt

packages/charts/src/components/legend/legend_item.tsx
```diff
@@ -61,10 +61,24 @@ export const LegendListItem: React.FC<LegendItemProps> = (props) => {
...
+
+  useEffect(() => {
+    const onPointerDown = (e: PointerEvent) => {
+      if (!itemRef.current?.contains(e.target as Node)) {
+        setIsActive(false);
+      }
+    };
+
+    document.addEventListener('pointerdown', onPointerDown);
+    return () => document.removeEventListener('pointerdown', onPointerDown);
```
markov00: The current strategy doesn't provide a great experience.
The keyboard interaction keeps the icon visible that we should avoid.
The mouse interaction instead keeps the icon visible if I click everywhere on the screen.
I believe this needs to be solved as this degrade the current experience.

If we need to we can request one or two callbacks from the Action event, like a onOpen/onClose so we can correctly detect when the Action event, if any, was triggered and when it is done

delanni:
buildkite build this

delanni:
We're in the process of migrating the build pipeline of this repo to an elastic-wide from the custom one we had before, this PR might require an admin-merge when deemed ready because some checks seem to be stuck.

packages/charts/src/components/legend/legend_item.tsx
maryam-saeidi: Fixed in [a4d8914](https://github.com/elastic/elastic-charts/pull/2773/commits/a4d8914923e4d8c0f6de88319c009626f2a80f2e)

packages/charts/src/components/legend/legend_table/legend_table_item.tsx
```diff
@@ -37,10 +39,13 @@ export const LegendListItem: React.FC<LegendItemProps> = (props) => {
     hiddenItems,
   } = props;
   const { color, isSeriesHidden, isItemHidden, seriesIdentifiers, label, path, isToggleable } = item;
+  const { isActive, actionRef, handlePointerDown, handleKeyDown } = useActionFocusManagement();
 
   const itemClassNames = classNames('echLegendTable__item', 'echLegendTable__item--highlightable', {
     'echLegendTable__item--hidden': isSeriesHidden,
     'echLegendTable__item--vertical': positionConfig.direction === LayoutDirection.Vertical,
+    'echLegendItem--isActive': isActive,
```
markov00: Shouldn't this be `echLegendTable__item--isActive`? 
From the SCSS `packages/charts/src/components/legend/legend_table/_legend_single_item.scss` looks like this is the class to use for active legend table items

packages/charts/src/components/legend/_legend_item.scss
```diff
@@ -71,6 +71,27 @@ $legendItemHeight: #{$euiFontSizeXS * $euiLineHeight};
     }
   }
 
+  &--actionOnHover {
+    .echLegendItem__action {
+      opacity: 0;
```
markov00, line 76: nit: even if we enable it on hover, should we disable the pointer events when the opacity is 0 with `pointer-events: none`?

Both here and in the table scss

packages/charts/src/components/legend/use_action_focus_management.ts
```diff
@@ -0,0 +1,93 @@
...
+      }
+    };
+
+    // Use capture phase to intercept events before they bubble
+    document.addEventListener('pointerdown', onPointerDown, true);
+    document.addEventListener('focusin', onFocusIn, true);
+    return () => {
+      document.removeEventListener('pointerdown', onPointerDown, true);
+      document.removeEventListener('focusin', onFocusIn, true);
+    };
```
markov00: Using this hook on every legend items attaches these listeners to the `document` 2 per legend item.
The `actionElement?.contains(e.target as Node)` becomes costly when the legend is huge.
I would suggest:
-  lift the logic to know "which legend action is active" into the parent legend (or a small provider used by both list + table)
- attach at most 1 pointerdown (and optionally focusin) listener only while needed (conditionally only when an actionItem is active, not everytime)

This reduces to just a couple of listeners  and improves the logic.

Also you can disable entirely this logic in the hook if there is no action/or the `legendActionOnHover` is disabled

packages/charts/src/components/legend/legend_table/legend_table_item.tsx
maryam-saeidi: You have hawk eyes! 🦅 

It seems it was working because somehow it was picking up the list scss for the active state. Will fix it.

packages/charts/src/components/legend/_legend_item.scss
maryam-saeidi: I removed `pointer-events: none` because of the firefox bug that I found and the fix that you suggested ([comment](https://github.com/elastic/elastic-charts/pull/2773#discussion_r2711481357))

markov00: oh you are right, do you mind adding a comment so we don't introduce that pinter-events:none in the future?

maryam-saeidi: Sure, sounds good 👍🏻

packages/charts/src/components/legend/_legend_item.scss
maryam-saeidi: Just to document the issue here, adding `pointer-events: none;` causes the following rendering bug in **FireFox**:


https://github.com/user-attachments/assets/54812540-3f3c-4dd1-9843-e80e1f9dafec

packages/charts/src/components/legend/_legend_item.scss
maryam-saeidi: Added in [85871b4](https://github.com/elastic/elastic-charts/pull/2773/commits/85871b4332e3d67029e5f543173f8b9967c7281c) and [9bcfcf7](https://github.com/elastic/elastic-charts/pull/2773/commits/9bcfcf7f96e934ce8360b2d06e55de50efdb5d07)

packages/charts/src/components/legend/legend_table/legend_table_item.tsx
maryam-saeidi: Fixed in [f069aad](https://github.com/elastic/elastic-charts/pull/2773/commits/f069aadd3d3d042ef7c0068024eb2b5ce1acd5bc)

packages/charts/src/components/legend/use_action_focus_management.ts
maryam-saeidi: [Legend](https://github.com/elastic/elastic-charts/pull/2773/changes#diff-b280b100797cea8f5e3a10197333b3e947a1fbc5d456e6a9de65db67b0061afeR151) has a `ActionFocusProvider` provider now that registers 'pointerdown' and 'focusin' listeners only when they are needed (All the related logic can be found in [packages/charts/src/components/legend/use_action_focus_management.ts](https://github.com/elastic/elastic-charts/pull/2773/changes#diff-de61cf2540ec8837a349a8d9a69da94651cc2a2b91c90a38223f2e74ba136b5bR1)) 

Thanks for the suggestion!

Issue #2514, opened 2024-08-26
Introducing border between adjacent elements


gvnmagni:
Following a wider conversation regarding colors, I would suggest to introduce borders between adjacent elements in order to improve readability and to make it simpler to tell apart visual shapes in charts.

We already do this in partition charts for examples (pies, donuts, treemaps...) but we don't do it in stacked barcharts, it would be great to evaluate the feasibility of this and the potential consequences

![Image](https://github.com/user-attachments/assets/2ff5f0d1-d028-40c7-8415-39ded88cfbff)

gvnmagni:
Related issue: https://github.com/elastic/elastic-charts/issues/2513

nickofthyme:
One thing we might run into with this is similar to the `minBarHeight` that sort of defies the scaled representation of the bar. This would be especially problematic when many bars are cluster in a small height to where the border eats into the height so much that the bar vanishes. I think we cross that bridge when we get there just wanted to make a note.

Is this also applied to the horizontal space between dodged bars?

gvnmagni:
Valid point which has been in my head for a while. One potential approach that we can use is to apply this kind of visual features only when the `high contrast` more is turned-on, in that case I feel we can potentially assume that the main need of users it to tell apart elements, therefore we do our best to facilitate it.

But that's only an option, let's have a proper conversation about all this

PR #2774, opened 2026-01-20
feat(highlight): Add support to configure dimmed (unhighlight) colors for Bar and Partition charts

awahab07:
## Summary

Unhighlighted series in bar and partition charts now display a neutral dimmed color instead of reduced opacity, matching the existing behavior in line and area charts.

When hovering over a legend item or series, unhighlighted elements render with a solid neutral gray color:
- **Light mode**: Shade 15 (`#ECF1F9`) at 100% opacity
- **Dark mode**: Shade 110 (`#384861`) at 100% opacity

This provides better visual distinction and consistency across all chart types.

https://github.com/user-attachments/assets/f4f4ba58-abfa-43d1-9739-5a06c92dbe39

image

### Theme Configuration

- Added `blueGrey110` (`#384861`) to `PRIMITIVE_COLORS` and `shade110` to `SEMANTIC_COLORS` in `base_colors.ts` to align with the latest EUI Borealis palette
- Updated `LIGHT_DIMMED_COLORS.barFill` and `arcFill` to use solid `shade15` (`#ECF1F9`)
- Updated `DARK_DIMMED_COLORS.barFill` and `arcFill` to use solid `shade110` (`#384861`)
- Default theme uses neutral gray dimmed colors; legacy/amsterdam themes retain opacity-only dimming for backward compatibility

### Storybook

- Story **Stylings -> Dimmed Highlight Style** has been added to comprehensively test the feature with ability to switch unhighlight shades.
- Other existing stories, helpful to test the chagne:
  - Legend -> Piechart
  - Interactions -> Sunburst Slice Clicks
  - Interactions -> Line Area Bar Point Clicks And Hovers

## Issues

Implements the feature required by https://github.com/elastic/kibana/issues/245829.

### Checklist


awahab07:
buildkite test this

awahab07:
buildkite update vrt

awahab07:
buildkite test this

delanni:
buildkite build this

gvnmagni:
Thank you @awahab07 for putting this together, I finally got time to look at it and it works great! I think we just have a couple of things to settle before proceeding:

- for Light mode my preference is to use Shade 15 as default (solid, 100% opacity)
- for Dark mode I would use Shade 110 (solid, 100% opacity)
- I believe lines and areas should keep behaving as they already do when they are on their own, it should not cause any bad result I think (will check again)
- one last very minor detail that is not a blocker for the PR itself but could be a nice touch. Grouped barcharts display bars attached to each other, when we hover on the legend on one of these series we should add a border of 1px (same color as panel background so that it adapts to dark and light mode) to the non-selected bars so that they can separated visually 

Thanks again Abdul!

awahab07:
> for Light mode my preference is to use Shade 15 as default (solid, 100% opacity)
for Dark mode I would use Shade 110 (solid, 100% opacity)

Default shades have been updated to reflect these values.
image
image

> I believe lines and areas should keep behaving as they already do when they are on their own, it should not cause any bad result I think (will check again)

I've kept them as before.

> one last very minor detail that is not a blocker for the PR itself but could be a nice touch. Grouped barcharts display bars attached to each other, when we hover on the legend on one of these series we should add a border of 1px (same color as panel background so that it adapts to dark and light mode) to the non-selected bars so that they can separated visually

This should be tackled in https://github.com/elastic/elastic-charts/issues/2514, as a follow-up, as mentioned [in](https://github.com/elastic/kibana/issues/245829#issuecomment-3728294410). But could be evaluated/discussed to see if a separate on-the-fly border should be implemented just for highlight/unhighlight purposes.

awahab07:
buildkite update vrt

awahab07:
buildkite test this

awahab07:
buildkite update vrt

packages/charts/src/chart_types/partition_chart/renderer/canvas/canvas_renderers.ts
```diff
@@ -149,22 +154,53 @@ function renderTaperedBorder(
...
+) {
   withContext(ctx, () => {
     ctx.scale(1, -1); // D3 and Canvas2d use a left-handed coordinate system (+y = down) but the ViewModel uses +y = up, so we must locally invert Y
     quadViewModel.forEach((quad: QuadViewModel) => {
-      if (quad.x0 !== quad.x1) renderTaperedBorder(ctx, quad);
+      // Apply dimmed colors for unhighlighted quads
+      const isUnhighlighted = highlightedQuadSet.size > 0 && !highlightedQuadSet.has(quad);
+      const dimmed = arcSeriesStyle.arc.dimmed;
+      const dimmedFill = dimmed && 'fill' in dimmed ? dimmed.fill : undefined;
+      const useDimmedColor = isUnhighlighted && dimmedFill;
```
markov00: Can we extract this into a function and reuse it?
Probably we have probably a function that is used everytime we need to pick up the a dimmed or non-dimmed color and reuse it elsewhere also on area/line/point charts

packages/charts/src/utils/themes/theme.ts
```diff
@@ -732,6 +732,18 @@ export interface ArcStyle {
   strokeWidth: number;
   /** the opacity of each arc on the theme/series */
   opacity: number;
+  /**
+   * The style applied to the arc when it is dimmed relative to other highlighted elements on the chart.
+   * This is typically used to visually de-emphasize the arc, for example, when another series is highlighted.
+   */
+  dimmed?:
+    | { opacity: number }
+    | {
```
markov00: This style is actually not used for a Partition chart. the Partition has the `PartitionStyle` in the theme

packages/charts/src/utils/themes/base_colors.ts
```diff
@@ -209,6 +211,9 @@ export const LIGHT_DIMMED_COLORS = {
   areaStroke: SEMANTIC_ALPHA_COLORS.shade30RGBAlpha50,
   areaPointStroke: SEMANTIC_ALPHA_COLORS.shade30RGBAlpha15,
   areaPointFill: SEMANTIC_COLORS.plainLight,
+
+  barFill: SEMANTIC_COLORS.shade15,
+  arcFill: SEMANTIC_COLORS.shade15,
```
markov00: I get where this comes from `ArcStyle` but we should call it differently, because the same style will be applied on every partition chart, not just the pie chart

packages/charts/src/chart_types/partition_chart/renderer/canvas/canvas_renderers.ts
```diff
@@ -149,22 +154,53 @@ function renderTaperedBorder(
...
+
+      if (useDimmedColor) {
+        // Temporarily override fillColor for dimmed state
+        const originalFillColor = quad.fillColor;
+        quad.fillColor = dimmedFill;
+        if (quad.x0 !== quad.x1) renderTaperedBorder(ctx, quad);
+        quad.fillColor = originalFillColor; // Restore
```
markov00: if we need to update the color just for that call, I prefer instead passing just the required properties (in case also by revisiting the `renderTaperedBorder` arguments to accept just the required props and passing them, rather then modifing and restore that value.

packages/charts/src/chart_types/partition_chart/renderer/canvas/canvas_renderers.ts
```diff
@@ -277,6 +313,12 @@ export function renderPartitionCanvas2d(
     panel,
     chartDimensions,
   }: ShapeViewModel,
+  _focus: unknown,
+  _animationState: AnimationState,
```
markov00: I see where this function get called, we probably need to fix there and call the relative functions with just the minimal amount of arguments, living these arguments there doesn't have much sense

delanni:
buildkite build this

awahab07:
buildkite update vrt

packages/charts/src/chart_types/partition_chart/renderer/canvas/canvas_renderers.ts
awahab07: Yes, makes sense. I've introduced [`getDimmedColor`](https://github.com/elastic/elastic-charts/pull/2774/changes#diff-083e2e38c1b7e7718e8143b3f09237fe29d376323570720670b224ee65105190R124) and is reused everywhere.

packages/charts/src/utils/themes/theme.ts
awahab07: Thanks for pointing this our. I've address it here https://github.com/elastic/elastic-charts/pull/2774/commits/ce3d5518e5e2c2db7bf5f1926fc8efd7ec8929be.

packages/charts/src/utils/themes/base_colors.ts
awahab07: Renamed it [here](https://github.com/elastic/elastic-charts/pull/2774/changes#diff-e6fed848006a5e62688905be5a40ec8ce946e94b4fe9f92653f953f2d0215821R216).

packages/charts/src/chart_types/partition_chart/renderer/canvas/canvas_renderers.ts
awahab07: `highlightedQuadSet` and `partitionStyle` are now [passed](https://github.com/elastic/elastic-charts/pull/2774/changes#diff-58642e87b781300be17caa312039770f410cf3a151fbbd4910767823bde7b755R161) are args and `quad.fillColor` isn't mutated in-place anymore.

packages/charts/src/chart_types/partition_chart/renderer/canvas/canvas_renderers.ts
awahab07: Makes sense. Addressed and now distinct functions are called with relevant arguments only [here](https://github.com/elastic/elastic-charts/pull/2774/changes#diff-6cab0c8662c3ae08c6d0116473b68e89e26fa1fba887209cad466f2ddd041a15R192).

packages/charts/src/utils/themes/theme.ts
```diff
@@ -748,6 +748,18 @@ export interface RectStyle {
...
+   * This is typically used to visually de-emphasize the rect, for example, when another series is highlighted.
+   */
+  dimmed?:
+    | { opacity: number }
+    | {
+        /** The fill color to use when the rect is dimmed. */
+        fill: Color | ColorVariant;
+        /** The opacity multiplier for the texture color when the rect is dimmed */
+        texture?: { opacity: number };
+      };
```
markov00: Please make this required (both the root dimmed property an the texture one).
This is the Theme object and everything fallback to this. There is no reason to make this optional, is always better to have it so we don't fallback programmatically inside the code, but we have the actual fallback already available

packages/charts/src/utils/themes/dimmed_colors.ts
```diff
@@ -0,0 +1,137 @@
...
+ * @internal
+ */
+export function getDimmedColor<T extends Color | ColorVariant | undefined>(
+  isDimmed: boolean,
+  dimmedConfig: DimmedStyleConfig | undefined,
+  colorKey: 'fill' | 'stroke',
+  defaultColor: T,
+): T {
+  if (hasDimmedColor(isDimmed, dimmedConfig, colorKey)) {
+    const dimmedColor = (dimmedConfig as ColorDimmedConfig)[colorKey];
```
markov00: is better to avoid type casting here. Moving the `dimmedConfig` check in the if and marking `dimmedConfig` required in `hasDimmedColor` should do the trick

packages/charts/src/utils/themes/dimmed_colors.ts
```diff
@@ -0,0 +1,137 @@
...
+    if (dimmedColor !== undefined) {
+      return dimmedColor as T;
```
markov00: I'm not sure this is a valid: if `defaultColor` is of type `Color`, but the `ColorDimmedConfig` is of type `ColorVariant` is not "exactly" the same.
We are also accepting an undefined `defaultColor` that in the `points.ts` file makes the return value being of a `undefined` type:

const dimmedStrokeColor = getDimmedColor(isDimmed, pointStyle.dimmed, 'stroke', undefined);

packages/charts/src/chart_types/xy_chart/renderer/canvas/points.ts
```diff
@@ -42,13 +43,14 @@ export function renderPoints(
...
 
-  const dimmedFill =
-    highlightState === 'dimmed' && 'fill' in pointStyle.dimmed ? colorToRgba(pointStyle.dimmed.fill) : undefined;
-  const dimmedStroke =
-    highlightState === 'dimmed' && 'stroke' in pointStyle.dimmed ? colorToRgba(pointStyle.dimmed.stroke) : undefined;
+  // Pre-compute dimmed colors (converted to RgbaTuple) for use inside the loop
+  const dimmedFillColor = getDimmedColor(isDimmed, pointStyle.dimmed, 'fill', undefined);
+  const dimmedStrokeColor = getDimmedColor(isDimmed, pointStyle.dimmed, 'stroke', undefined);
+  const dimmedFill = dimmedFillColor ? colorToRgba(dimmedFillColor) : undefined;
+  const dimmedStroke = dimmedStrokeColor ? colorToRgba(dimmedStrokeColor) : undefined;
```
markov00: The type returned by `getDimmedColor` is undefined and makes the requirement of this code useless. I strongly believe that the typings here are wrong

packages/charts/src/chart_types/partition_chart/renderer/dom/layered_partition_chart.tsx
```diff
@@ -24,6 +28,5 @@ export const chartRenderer: ChartRenderer = (
     <Tooltip getChartContainerRef={containerRef} />
     <Partition forwardStageRef={forwardStageRef} />
     <HighlighterFromHover />
-    <HighlighterFromLegend />
   </>
```
markov00, line 31: Why this HighlighterFromLegend was removed?

packages/charts/src/chart_types/partition_chart/renderer/canvas/canvas_renderers.ts
```diff
@@ -149,22 +155,40 @@ function renderTaperedBorder(
...
+) {
   withContext(ctx, () => {
     ctx.scale(1, -1); // D3 and Canvas2d use a left-handed coordinate system (+y = down) but the ViewModel uses +y = up, so we must locally invert Y
-    quadViewModel.forEach(({ strokeWidth, fillColor, x0, x1, y0px, y1px }) => {
+    quadViewModel.forEach((quad) => {
+      const { strokeWidth, fillColor, x0, x1, y0px, y1px } = quad;
       // only draw a shape if it would show up at all
       if (x1 - x0 >= 1 && y1px - y0px >= 1) {
-        ctx.fillStyle = fillColor;
+        const isUnhighlighted = highlightedQuadSet.size > 0 && !highlightedQuadSet.has(quad);
```
markov00: Nit picking here (but also in other part of the changes: probably makes sense calling this state `isDimmed` rather then `isUnhighlighted`

packages/charts/src/utils/themes/partition.ts
```diff
@@ -82,4 +92,9 @@ export interface PartitionStyle extends FillFontSizeRange {
   linkLabel: LinkLabelConfig;
   sectorLineWidth: Pixels;
   sectorLineStroke: StrokeStyle;
+  /**
+   * The style applied to partition slices when they are dimmed relative to other highlighted elements.
+   * This is typically used to visually de-emphasize slices when hovering over a legend item.
+   */
+  dimmed?: PartitionDimmedStyle;
```
markov00: Same comment as in the `theme.ts` the property here should be required

packages/charts/src/utils/themes/partition.ts
```diff
@@ -12,6 +12,16 @@ import type { Font, PartialFont, FontFamily } from '../../common/text_utils';
...
 import type { PerSideDistance } from '../dimensions';
 
+/** @public */
+export type PartitionDimmedStyle =
+  | { opacity: number }
+  | {
+      /** The fill color to use when partition slices are dimmed. */
+      fill: Color | ColorVariant;
+      /** The stroke color to use when partition slices are dimmed. */
+      stroke?: Color | ColorVariant;
```
markov00: Let's make stroke required

e2e/tests/stylings_stories.test.ts
```diff
@@ -89,4 +89,28 @@ test.describe('Stylings stories', () => {
...
+    eachTheme.describe(({ theme, urlParam }) => {
+      test(`should dim XY chart series on legend hover - ${theme}`, async ({ page }) => {
+        const action = async () => {
+          await common.moveMouseRelativeToDOMElement(page)({ left: 5, top: 5 }, '.echLegendItem');
+        };
+        await common.expectChartAtUrlToMatchScreenshot(page)(
+          `http://localhost:9001/?path=/story/stylings--dimmed-highlight-style&${urlParam}`,
+          { action },
+        );
+      });
```
markov00: Looking at the resulting images, it looks like these are not XY but partisions

packages/charts/src/utils/themes/theme.ts
awahab07: `dimmed` is required everywhere now.

packages/charts/src/utils/themes/dimmed_colors.ts
awahab07: Thanks for pointing this out, I've improved the typings.

packages/charts/src/chart_types/xy_chart/renderer/canvas/points.ts
awahab07: Addressed as above.

packages/charts/src/chart_types/partition_chart/renderer/dom/layered_partition_chart.tsx
awahab07: Legend hover highlighting for partition charts are [now](https://github.com/elastic/elastic-charts/pull/2774/changes#diff-58642e87b781300be17caa312039770f410cf3a151fbbd4910767823bde7b755R161) based on canvas-based dimming (for consistency with bar/line/area charts), hence `HighlighterFromLegend` is no longer needed. HighlighterFromHover is still SVG based (as is the case with bar/line/area charts).

e2e/tests/stylings_stories.test.ts
awahab07: Thanks for noticing. The relevant state on the story for XY wasn't being applied. I've [addressed](https://github.com/elastic/elastic-charts/pull/2774/changes/715b5c7a63e138dfab3d884ee8e24ff732d34bed) it.

packages/charts/src/chart_types/partition_chart/renderer/canvas/canvas_renderers.ts
```diff
@@ -149,22 +155,40 @@ function renderTaperedBorder(
...
   withContext(ctx, () => {
     ctx.scale(1, -1); // D3 and Canvas2d use a left-handed coordinate system (+y = down) but the ViewModel uses +y = up, so we must locally invert Y
-    quadViewModel.forEach(({ strokeWidth, fillColor, x0, x1, y0px, y1px }) => {
+    quadViewModel.forEach((quad) => {
+      const { strokeWidth, fillColor, x0, x1, y0px, y1px } = quad;
       // only draw a shape if it would show up at all
       if (x1 - x0 >= 1 && y1px - y0px >= 1) {
-        ctx.fillStyle = fillColor;
+        const isDimmed = highlightedQuadSet.size > 0 && !highlightedQuadSet.has(quad);
+        const dimmedFillColor = getDimmedColor(isDimmed, partitionStyle.dimmed, 'fill', fillColor);
```
markov00: the color needs to pass first from `getColorFromVariant` because the color type is `Color | ColorVariant`
Unfortunately we don't have a nice typing that enforce that.

packages/charts/src/chart_types/xy_chart/renderer/canvas/points.ts
```diff
@@ -42,13 +43,14 @@ export function renderPoints(
...
```
markov00: These colors needs to pass by `getColorFromVariant` because their types are `Color | ColorVariant`

packages/charts/src/chart_types/partition_chart/renderer/canvas/canvas_renderers.ts
```diff
@@ -149,22 +155,40 @@ function renderTaperedBorder(
...
+  partitionStyle: PartitionStyle,
+) {
   withContext(ctx, () => {
     ctx.scale(1, -1); // D3 and Canvas2d use a left-handed coordinate system (+y = down) but the ViewModel uses +y = up, so we must locally invert Y
     quadViewModel.forEach((quad: QuadViewModel) => {
-      if (quad.x0 !== quad.x1) renderTaperedBorder(ctx, quad);
+      if (quad.x0 === quad.x1) return;
+
+      const isDimmed = highlightedQuadSet.size > 0 && !highlightedQuadSet.has(quad);
+      const fillColor = getDimmedColor(isDimmed, partitionStyle.dimmed, 'fill', quad.fillColor);
```
markov00: same here

packages/charts/src/utils/themes/partition.ts
```diff
@@ -12,6 +12,16 @@ import type { Font, PartialFont, FontFamily } from '../../common/text_utils';
...
 import type { PerSideDistance } from '../dimensions';
 
+/** @public */
+export type PartitionDimmedStyle =
+  | { opacity: number }
+  | {
+      /** The fill color to use when partition slices are dimmed. */
+      fill: Color | ColorVariant;
+      /** The stroke color to use when partition slices are dimmed. */
+      stroke: Color | ColorVariant;
```
markov00: It looks like `stroke` is not used, we can remove it from the API

packages/charts/src/chart_types/partition_chart/renderer/canvas/partition.tsx
```diff
@@ -182,12 +189,21 @@ class PartitionComponent extends React.Component<PartitionProps> {
...
-          ? renderLinearPartitionCanvas2d
-          : isWaffle(geometries.layout)
-            ? renderWrappedPartitionCanvas2d
-            : renderPartitionCanvas2d;
-        renderer(ctx, devicePixelRatio, geometries, focus, this.animationState);
+        if (isSimpleLinear(geometries.layout, geometries.style.fillLabel, geometries.layers)) {
+          renderLinearPartitionCanvas2d(ctx, devicePixelRatio, geometries, focus, this.animationState);
+        } else if (isWaffle(geometries.layout)) {
+          renderWrappedPartitionCanvas2d(ctx, devicePixelRatio, geometries);
+        } else {
```
markov00, line 214: It looks like the waffle rendering now doesn't include the same dimming capabilities as the other charts.

awahab07:
buildkite update vrt

storybook/stories/stylings/27_dimmed_highlight_style.story.tsx
```diff
@@ -0,0 +1,425 @@
...
+      point: {
+        dimmed: {
+          fill: isDarkTheme ? '#0B1628' : '#FFFFFF',
+          stroke: selectedDimmedColor,
+        },
+      },
+    },
+    partition: {
+      dimmed: {
+        fill: selectedDimmedColor,
```
mariairiartef: `PartitionDimmedStyle` can be `{ opacity: number }`. I was checking the behavior and I think it does not work. Do you think we should implement it?

packages/charts/src/chart_types/partition_chart/renderer/canvas/partition.tsx
awahab07: Thanks for spotting this. This also highlights the fact that there's no e2e legend hover test for waffle chart which would have caught this. I've [added](https://github.com/elastic/elastic-charts/pull/2774/changes/b2985b28382e7d7094c4232239e301605b2ded24) the snapshot tests.

awahab07:
buildkite update vrt

packages/charts/src/chart_types/partition_chart/renderer/canvas/partition.tsx
awahab07: Found that dimming for flame charts (`partition_chart/renderer/canvas/canvas_linear_renderers.ts`) was also not implemented. Did that [here](https://github.com/elastic/elastic-charts/pull/2774/commits/0fd360d21c020f4673d38f97530020a305a20738). [snapshot](https://github.com/elastic/elastic-charts/pull/2774/commits/ff19d783917917ad3c6a1e05313aba7e551eabf1)

storybook/stories/stylings/27_dimmed_highlight_style.story.tsx
awahab07: Good catch, Maria! Thank you for testing. That's right, it should be implemented as legacy/Amsterdam themes use `{ opacity: 0.25 }` for partition charts.

The existing XY chart renderers (line, area, bar, points) all check 'opacity' in dimmedConfig and apply it via `overrideOpacity()`.

I've implemented it [here](https://github.com/elastic/elastic-charts/pull/2774/commits/2fd52f855e3ec0f97f44cbc092a53fcf0ec90898) for partition renderer. And later made sure it's implemented for waffle and flame charts as well (which this PR implements dimming for).

Also added snapshot tests to cover the case when dimming is defined as opacity-only value (e.g. `dimmed: { opacity: 0.25 }`).

e2e/screenshots/metric_stories.test.ts-snapshots/metric/should-render-metric-with-value-in-middle-position-chrome-linux.png
markov00, line 1: Why there is a font size change here?

awahab07: Looks like something has changed either upstream or in the environment as this [PR](https://github.com/elastic/elastic-charts/pull/2808/changes#diff-3dfb94f0c4acd6890a6f26e5918b7623039fe98a70e891038fcd0f12fcce3fe0) is also facing this. 

image

awahab07:
buildkite update vrt

e2e/screenshots/metric_stories.test.ts-snapshots/metric/should-render-metric-with-value-in-middle-position-chrome-linux.png
awahab07: It's now confirmed that it's an upstream issue.

The test was added by [Add ability to vertically align the metrics values to the middle of the chart](https://github.com/elastic/elastic-charts/pull/2783/changes), but got merged after the the PR [feat(theme): set default labelOptions.widthLimit in themes and update metric.titleWeight default to 500 #2802](https://github.com/elastic/elastic-charts/pull/2802) which updated the metric chart's title font weight.

PR #2780, opened 2026-01-23
ci: remove `@elastic/datavis CI` overall check

nickofthyme:
## Summary

Removes the main **@elastic/datavis CI**. Currently. this check get stuck in the **pending** state as the old ci is now disabled.

This was handled partially by the old `github_bot`, but with the new ci changes we no longer really need this. We can instead just rely on the individual job status checks from the build.

If we still wanted this we could add a post-build step but I don't think it's necessary. We could also enable the pr check status from `buildkite-pr-bot` to get a similar check.

Zight Recording 2026-01-23 at 01 45 40 PM


nickofthyme:
Additionally, I removed `@elastic/datavis CI` and `Playwright e2e` as required checks.

image

PR #2783, opened 2026-01-26
feat(metric): Add ability to vertically align the metrics values to the middle of the chart

awahab07:
## Summary

The `valuePosition` prop in the `MetricStyle` theme now supports a new `'middle'` value, allowing the primary metric value to be vertically centered within the chart area. When using `valuePosition: 'middle'`, the title and subtitle remain at the top, while the primary metric value is centered with the secondary metric (extra) displayed directly below it.

https://github.com/user-attachments/assets/bb5ac478-b292-45e7-99db-cedaf80756a3

## Details

- Added `'middle'` to the `valuePosition` union type in `MetricStyle` interface (`'top' | 'middle' | 'bottom'`)
- The following stories could be used to test the change:
  - Metric (@alpha) → Basic
  - Metric (@alpha) → Grid
  - Metric (@alpha) → Body Content
  - Metric (@alpha) → Layout

### Update 01
[see](https://github.com/elastic/elastic-charts/pull/2783#issuecomment-3990419873)

## Issues

Implement the underlying functionality required by https://github.com/elastic/kibana/issues/234956.

### Checklist


awahab07:
buildkite test this

awahab07:
buildkite update vrt

awahab07:
buildkite update vrt

awahab07:
buildkite update vrt

delanni:
buildkite build this

gvnmagni:
Thanks @awahab07 , can I ask you to fix one detail? The Secondary Metric should always be anchored to the bottom, as the Title is always anchored to the top, and then it is Primary Metric that basically forced its position around the component

This image show an extreme version of this concept (font sizes are wrong, it is purely to show the logic):
Screenshot 2026-02-27 at 11 09 21

awahab07:
### Update 01
The extra block (secondary metric) will always be anchored now to the bottom (unless when the primary metric is positioned to the bottom in which case secondary metric will appear above it).

Metric-chart-primary-vertical-alignment

packages/charts/src/chart_types/metric/renderer/dom/_text.scss
```diff
@@ -55,6 +55,11 @@
     &--right {
       justify-content: flex-end;
     }
+
+    // Vertical position modifiers
+    &--middle {
+      padding-top: 0;
+    }
```
mariairiartef, line 62: qq: maybe I missed this but why only removing the padding top from the extra block only when the primary metric position is `middle`?

mariairiartef: Image

Image

Image

packages/charts/api/charts.api.md
mariairiartef, line 1: suggestion: maybe we can add a test case here [`e2e/tests/metric_stories.test.ts`](https://github.com/elastic/elastic-charts/blob/0c96e34112cd8305762e3a09a29e5485770ad794/e2e/tests/metric_stories.test.ts#L145) to test the new primary metric position

gvnmagni (approved):
Looks great! 🚀

packages/charts/src/chart_types/metric/renderer/dom/_text.scss
mariairiartef, line 62: If we want to remove the padding only when is in the middle we might need to update this calculation:

https://github.com/elastic/elastic-charts/blob/0c96e34112cd8305762e3a09a29e5485770ad794/packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx#L346C86-L346C110

packages/charts/api/charts.api.md
awahab07: Thanks for the suggestion, I've added a test [here](https://github.com/elastic/elastic-charts/pull/2783/commits/03b7be400e1e305a856a4b7df586f623a55c75c0) with [snapshot](https://github.com/elastic/elastic-charts/pull/2783/changes/4a3051be199f5d0b33feb67829feea119f38536b).

packages/charts/src/chart_types/metric/renderer/dom/_text.scss
awahab07: > qq: maybe I missed this but why only removing the padding top from the extra block only when the primary metric position is middle?

When the primary metric is positioned to the middle, it has to be in the middle of the available vertical height. With extra block carrying a `padding-top: 5px;`, it nudges the primary metric above a bit. Plus, when primary metric is positioned to the middle, extra black will now always be anchored to the bottom, so it shouldn't need the top padding.

awahab07: That's an excellent point. I've addressed that [here](https://github.com/elastic/elastic-charts/pull/2783/commits/41466ba067007536ad33d7746afdf34a6bb43fee).

mariairiartef: ok! Thanks for the explanation :)

PR #2784, opened 2026-01-27
feat(legend): add "list" layout for horizontal legends

maryam-saeidi:
## Summary

A new **List** legend layout is now available for horizontal legends (top/bottom), offering a more space-efficient alternative to the existing GridList/Table layout. In the list layout each legend item flows inline — label, color marker, and values are arranged in a compact row that wraps naturally, making it easier to scan many series without excessive scrolling.

The list layout supports all standard legend features: series highlighting, filtering via actions, `CurrentAndLastValue` display with stable column widths, and label truncation.

image

#### Example usage

```
<Chart>
    <Settings
        showLegend
        legendPosition={Position.Bottom}
        legendLayout="list"  // <-- new prop
        legendValues={[LegendValue.CurrentAndLastValue]}
    />
    <Axis id="bottom" position={Position.Bottom} />
    <Axis id="left" position={Position.Left} />
    <BarSeries
        id="bars"
        yScaleType={ScaleType.Linear}
        xAccessor={0}
        yAccessors={[1]}
        data={[
            ["A", 10],
            ["B", 20],
            ["C", 15],
        ]}
    />
</Chart>
```

#### Demo

https://github.com/user-attachments/assets/1bddc533-8634-4da1-8ef6-a2dcd0a1d31c

### Layout & rendering
- Added `legendLayout: 'list'` option to the `Settings` spec, enabling the list layout for horizontal legends
- Created `LegendList` component (`legend_list.tsx`) that renders legend items as inline `<li>` elements with label, color, values, and optional action grouped together
- Legend values display a title prefix in list layout (e.g. **VALUE:** 42, **AVG:** 10)
- `CurrentAndLastValue` shows a `—` placeholder when no value is available (e.g. on hover)
- Labels are truncated by pixel width limit in list layout (controlled by `theme.legend.labelOptions.widthLimit`)

### Stable column widths
- Added `getLegendMaxFormattedValue` to the `ChartSelectors` interface, allowing each chart type to provide the longest formatted Y domain value
- Implemented the XY chart selector that formats the max Y domain value through each series' tick formatter and returns the longest result
- The max formatted value width is computed via **canvas text measurement** (`withTextMeasure`) and used as `minWidth` on `CurrentAndLastValue` cells, preventing layout shift on hover

### Legend sizing
- Added `computeHorizontalLegendRowCount` utility that measures items and lays them out greedily row-by-row to determine if the legend fits in 1, 2, or 3+ rows — short-circuiting as soon as a third row is reached
- Legend height is computed based on the row count (single-line, two-line, or multi-line)
- Added `shouldDisplayGridList` and `shouldDisplayTable` helpers to select the correct layout mode based on legend values, position, and layout config

### Other
- Renamed `_legend.scss` → `_legend_grid_list.scss` for clarity
- Added storybook story for the new layout (`2_legend_layout.story.tsx`)
- Added unit tests for `computeHorizontalLegendRowCount` and the `shouldDisplay*` helpers

## Issues

Related to https://github.com/elastic/kibana/issues/238777

### Checklist

🤖 This pull request was assisted by Cursor

### DIFF
--- a/packages/charts/src/state/utils/legend_row_count.ts
+++ b/packages/charts/src/state/utils/legend_row_count.ts
@@ -83,7 +83,7 @@ export function computeHorizontalLegendRowCount(args: HorizontalLegendRowCountAr
     const { width: labelOnlyWidth } = textMeasure(item.label, font, 12, 1.5);
     const cappedLabelWidth = widthLimit > 0 ? Math.min(labelOnlyWidth, widthLimit) : labelOnlyWidth;
 
-    const valueCellWidths = item.values.map(({ type, label }) => {
+    const valuesWidth = item.values.map(({ type, label }) => {
       if (type !== LegendValue.CurrentAndLastValue && !label) return 0;
 
       const valueLabel = type === LegendValue.CurrentAndLastValue ? maxFormattedValue ?? (label || '—') : label;
@@ -91,10 +91,9 @@ export function computeHorizontalLegendRowCount(args: HorizontalLegendRowCountAr
       const valueText = !showValueTitle
         ? valueLabel
         : `${legendValueTitlesMap[type]?.toUpperCase() ?? ''}: ${valueLabel}`;
-      return textMeasure(valueText, font, 12, 1.5).width;
+      return sharedMargin + textMeasure(valueText, font, 12, 1.5).width;
     });
 
-    const valuesWidth = valueCellWidths.map((w) => (w ? sharedMargin + w : 0));
     const itemWidths = [
       sharedMargin +
         markerWidth +


storybook/stories/legend/2_legend_layout.story.tsx
```diff
@@ -0,0 +1,92 @@
...
+        splitSeriesAccessors={['g1', 'g2']}
+        data={TestDatasets.BARCHART_2Y2G.slice(0, dataCount)}
+      />
+    </Chart>
+  );
+};
+
+Example.parameters = {
+  markdown:
+    'The `legendLayout` prop allows you to render the legend in a list or table view.\n\n __Note:__ If this prop is undefined, by default a list view is shown unless there is statistics or CurrentAndLastValue.',
```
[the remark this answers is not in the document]
maryam-saeidi: GridList and List are both considered different variations of list in this context.

packages/charts/src/index.ts
```diff
@@ -106,6 +106,7 @@ export {
 // utilities
 export {
   Datum,
+  Layout,
```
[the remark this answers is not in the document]
maryam-saeidi: The naming matches the rest of the names.

packages/charts/src/components/legend/label.tsx
```diff
@@ -125,16 +129,32 @@ export function NonInteractiveLabel({ label, options }: { label: string; options
...
+
+  if (maxLines > 1) {
+    clampStyles.WebkitLineClamp = maxLines;
+  }
+
+  if (widthLimit > 0) {
+    clampStyles.maxWidth = `${widthLimit}px`;
+    clampStyles.overflow = 'hidden';
+    clampStyles.textOverflow = 'ellipsis';
+  }
```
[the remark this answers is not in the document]
maryam-saeidi: Fixed by adding additional condition.

maryam-saeidi:
buildkite update vrt

packages/charts/src/components/legend/legend_item_utils.ts
```diff
@@ -0,0 +1,26 @@
...
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+import { getExtra } from './utils';
+import type { LegendItem, LegendItemExtraValues } from '../../common/legend';
+import { LegendValue } from '../../common/legend';
+
+/** @internal */
+export const prepareLegendValues = (
```
maryam-saeidi: No change in the functionality, only moved here for better organization.

maryam-saeidi:
buildkite update vrt

maryam-saeidi:
buildkite update vrt

maryam-saeidi:
buildkite update vrt

maryam-saeidi:
buildkite update vrt

storybook/stories/legend/2_legend_layout.story.tsx
```diff
@@ -0,0 +1,90 @@
...
+        data={TestDatasets.BARCHART_2Y2G_VARIED_LEGEND.slice(0, dataCount)}
+      />
+    </Chart>
+  );
+};
+
+Example.parameters = {
+  markdown:
+    'The `legendLayout` prop allows you to render the legend in a list or table view.\n\n` +' +
+    '` __Note:__ When this prop is undefined, the layout is chosen automatically: by default a list view is used unless there is statistics or CurrentAndLastValue.',
```
maryam-saeidi: Will update this to:

```
When this prop is undefined, the layout is chosen automatically: by default a grid list view is used unless there is statistics.
```
Because in case of CurrentAndLastValue, we still show grid list view.

delanni:
buildkite build this

packages/charts/src/utils/common.tsx
```diff
@@ -30,6 +30,14 @@ export const Position = Object.freeze({
 /** @public */
 export type Position = $Values<typeof Position>;
 
+/** @public */
+export const LegendLayout = Object.freeze({
+  List: 'list' as const,
+  Table: 'table' as const,
+});
+/** @public */
+export type Layout = $Values<typeof LegendLayout>;
```
markov00: There is no problem calling the same way as the const. This is a practice we already use across the library and saves us a lot of headaches.

```suggestion
export type LegendLayout = $Values<typeof LegendLayout>;
```

packages/charts/src/state/utils/legend_row_count.ts
```diff
@@ -0,0 +1,139 @@
...
+  let currentRowWidth = 0;
+
+  for (const item of items) {
+    const { width: labelOnlyWidth } = textMeasure(item.label, font, 12, 1.5);
+    const cappedLabelWidth = widthLimit > 0 ? Math.min(labelOnlyWidth, widthLimit) : labelOnlyWidth;
+
+    const valueCellWidths = legendValues
+      .map((type) => {
+        const v = item.values.find((vv) => vv.type === type);
+        if (!v) return;
```
markov00: The values in the `item.values` contains the computed elements coming from the `legendValues`. so this check is basically useless because it is always the same.
`item.values` is an array that keeps the same order as the `legendValues` and contains also label/values for non-computable calculations (like when no currentItem is available).

So instead of looping through the `legendValues` array you can directly loop through the `item.values` and simplify it a bit.

packages/charts/src/state/utils/legend_row_count.ts
```diff
@@ -0,0 +1,139 @@
...
+        if (type !== LegendValue.CurrentAndLastValue && !v.label) return;
+
+        const valueLabel = type === LegendValue.CurrentAndLastValue ? maxFormattedValue ?? (v.label || '—') : v.label;
+
+        const valueText = (() => {
+          if (!showValueTitle) return valueLabel;
+          const title = legendValueTitlesMap[type] ?? '';
+          return `${title.toUpperCase()}: ${valueLabel}`;
+        })();
```
markov00: I'm not a fan of IIFE, If you don't want to modify a parameter, you can use a ternary operation here:

```suggestion
        const valueText = !showValueTitle ? valueLabel : `${(legendValueTitlesMap[type]?.toUpperCase() ?? '')}: ${valueLabel}`;
```

packages/charts/src/state/utils/legend_row_count.ts
```diff
@@ -0,0 +1,139 @@
...
+        return textMeasure(valueText, font, 12, 1.5).width;
+      })
+      .filter(isDefined);
```
markov00: You can return `0` instead of undefined so you don't have to loop again to filter these.
in the next `for of` you can just `continue` when the width is zero.

packages/charts/src/state/selectors/get_legend_max_formatted_value.ts
```diff
@@ -0,0 +1,45 @@
...
+import { DEFAULT_FONT_FAMILY } from '../../common/default_theme_attributes';
+import { LegendValue, legendValueTitlesMap } from '../../common/legend';
+import { withTextMeasure } from '../../utils/bbox/canvas_text_bbox_calculator';
+import type { GlobalChartState } from '../chart_state';
+import { createCustomCachedSelector } from '../create_selector';
+
+const LEGEND_VALUE_FONT = {
+  fontFamily: DEFAULT_FONT_FAMILY,
+  fontVariant: 'normal',
+  fontWeight: 900,
```
markov00: Why 900? the default legend font weight is 400, and the bold part is actually 700.

packages/charts/src/state/chart_selectors.ts
```diff
@@ -95,6 +95,12 @@ export interface ChartSelectors {
    */
   getLegendExtraValues(globalState: GlobalChartState): Map<SeriesKey, LegendItemExtraValues>;
 
+  /**
+   * Returns the formatted max Y domain value used to reserve stable width for CurrentAndLastValue legend values.
+   * @param globalState
+   */
+  getLegendMaxFormattedValue(globalState: GlobalChartState): string | undefined;
```
markov00: Better rename this and the comment to something more general like: `getWidestLegendFormattedValue` or `getLongestLegendFormattedValue`

packages/charts/src/components/legend/legend.tsx
```diff
@@ -13,13 +13,14 @@ import type { Dispatch } from 'redux';
...
-import { LegendListItem } from './legend_item';
+import { LegendList } from './legend_list';
 import { LegendTable } from './legend_table';
 import { getLegendPositionConfig, legendPositionStyle } from './position_style';
 import { getLegendStyle, getLegendListStyle } from './style_utils';
+import type { LegendItemProps } from './types';
 import { ActionFocusProvider } from './use_action_focus_management';
 import type { LegendItem, LegendItemExtraValues } from '../../common/legend';
+import { shouldDisplayGridList } from '../../common/legend';
 import { shouldDisplayTable } from '../../common/legend';
```
markov00: can we group these imports

packages/charts/src/state/selectors/get_legend_max_formatted_value.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Returns the pixel width to reserve for the CurrentAndLastValue legend column,
+ * computed via canvas text measurement including the title prefix (e.g. "VALUE: ").
+ * Returns `undefined` when no max-formatted value is available.
+ * @internal
+ */
+export const getLegendMaxFormattedValueWidthSelector = createCustomCachedSelector(
+  [getLegendMaxFormattedValueSelector],
+  (maxFormattedValue): number | undefined => {
+    if (!maxFormattedValue) return undefined;
+    const title = legendValueTitlesMap[LegendValue.CurrentAndLastValue] ?? '';
```
markov00: no need to check here for undefined, this maps and the way you are accessing it always return a value
```suggestion
    const title = legendValueTitlesMap[LegendValue.CurrentAndLastValue];
```

packages/charts/src/state/selectors/get_longest_legend_formatted_value.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Returns the pixel width to reserve for the CurrentAndLastValue legend column,
+ * computed via canvas text measurement including the title prefix (e.g. "VALUE: ").
+ * Returns `undefined` when no max-formatted value is available.
+ * @internal
+ */
+export const getLegendMaxFormattedValueWidthSelector = createCustomCachedSelector(
+  [getLegendMaxFormattedValueSelector],
+  (maxFormattedValue): number | undefined => {
```
markov00, line 39: Some considerations here:
if we always return a number, we don't have to check down the road for nullity.
This makes the code a bit more clean and doesn't rely on type casting everywhere.
Also in `getLegendMaxFormattedValueSelector` you can consider returning always a string, an `''` empty string will do the same and is anyways an edge case to verify.

packages/charts/src/components/legend/legend_list.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+  const itemClassNames = classNames('echLegendItem', {
+    'echLegendItem--hidden': isSeriesHidden,
+    'echLegendItem--vertical': positionConfig.direction === LayoutDirection.Vertical,
+    'echLegendItem--isActive': isActive,
+    'echLegendItem--actionOnHover': legendActionOnHover,
+  });
+
+  const preparedLegendValues = prepareLegendValues(item, legendValues, totalItems, extraValues);
+  const legendValueItems = preparedLegendValues.filter(isDefined);
```
markov00, line 60: not sure, whould be worth adding a useMemo here?
It wasn't there before, but I'm wondering if this is useful now

packages/charts/src/specs/settings.tsx
```diff
@@ -440,6 +441,11 @@ export interface LegendSpec {
    * @defaultValue Position.Right
    */
   legendPosition: Position | LegendPositionConfig;
+  /**
+   * The layout of the legend
+   * @defaultValue undefined
```
markov00, line 452: We can describe that the default legend layout depends on the current configuration. You can also describe quickly the mechanism

packages/charts/src/components/legend/_legend_grid_list.scss
```diff
@@ -1,7 +1,7 @@
 .echLegend {
   z-index: 1;
 
-  .echLegendList {
+  .echLegendGridList {
```
markov00, line 4: Now that I see it working, I'm wondering if we need to keep the `grid list` or we can just go with the default `list` or the `table`.
Because by default now the `grid list` appears only when I have the `CurrentAndLastValue` configured and no layout applied.
And this differs from the other legend represetnation, so probably we can get rid of it.

packages/charts/src/state/selectors/get_legend_max_formatted_value.ts
maryam-saeidi: So, for this one, I previously tried different options such as:

- using 700 for the bold part and 400 for the regular
- using 700 for the whole text (even the part that is not bold)

And in both cases, there was a layout shift when hovering on the chart. I just tested again and it works with 700 🤔 I'll change it to 700, the calculated value is very close the the actual size of the value.

Commit [1bc906b](https://github.com/elastic/elastic-charts/pull/2784/commits/1bc906b56286d3e25bc9995f684d1cf66569b4b2)

packages/charts/src/state/selectors/get_longest_legend_formatted_value.ts
maryam-saeidi: I am not sure how useful removing the undefined would be in this case, given that in `LegendList`, we only need to provide `maxFormattedValueWidth` when isListLayout is true. If we keep `LegendList` prop as is, do you see benefit in changing the default selectors in this file to return only string and number? (basically removing undefined return value)

packages/charts/src/components/legend/_legend_grid_list.scss
maryam-saeidi: > Because by default now the grid list appears only when I have the CurrentAndLastValue configured and no layout applied.

There are more cases with `grid list` view:
- layout: undefined, no aggregated value
- layout: undefined, only CurrentAndLastValue value
- layout: list, legend position: left/right
- layout: list, legend direction: vertical

packages/charts/src/components/legend/legend_list.tsx
maryam-saeidi: I don't think it would be useful since the number of legend values shouldn't be that many, so there is not much of an overhead here (comparing to the overhead of useMemo).

packages/charts/src/utils/common.tsx
maryam-saeidi: Changed in [5f029ba](https://github.com/elastic/elastic-charts/pull/2784/commits/5f029ba8caa63d89d1f5658cdbdfcc349b75d177)

packages/charts/src/state/utils/legend_row_count.ts
maryam-saeidi: Changed in [dc6be94](https://github.com/elastic/elastic-charts/pull/2784/commits/dc6be94e6e33038cb0760d0247a349aac62639c6)

packages/charts/src/state/utils/legend_row_count.ts
maryam-saeidi: [975a1f2](https://github.com/elastic/elastic-charts/pull/2784/commits/975a1f279c73dbef478ef53a1ee6ff3b660d4c68)

packages/charts/src/state/chart_selectors.ts
maryam-saeidi: Change in [dff9f2b](https://github.com/elastic/elastic-charts/pull/2784/commits/dff9f2bbadb5effb93b69db1143b639fba7e0e06), [7e18b48](https://github.com/elastic/elastic-charts/pull/2784/commits/7e18b48f8874d677a42105c3d11f53c22788f858), and [f1529f9](https://github.com/elastic/elastic-charts/pull/2784/commits/f1529f96466ea0523cf3f1d141098800eeb0ef58)

packages/charts/src/components/legend/legend.tsx
maryam-saeidi: Changed in [de45770](https://github.com/elastic/elastic-charts/pull/2784/commits/de45770f4166660a79b19bcfaa33a0a1758aa336)

packages/charts/src/state/selectors/get_legend_max_formatted_value.ts
maryam-saeidi: Removed in [54fd9f1](https://github.com/elastic/elastic-charts/pull/2784/commits/54fd9f1c0c9161929291f9879c2f1c775790b11b)

packages/charts/src/state/utils/legend_row_count.ts
maryam-saeidi: Good point, thanks for the suggestion!
Changed in [0813aca](https://github.com/elastic/elastic-charts/pull/2784/commits/0813aca7fd0a6a20a84240b04db5c8611ab06e8f)

packages/charts/src/specs/settings.tsx
maryam-saeidi: Adjusted the comment in [c4db612](https://github.com/elastic/elastic-charts/pull/2784/commits/c4db612fa956f72f870c3fc2c9446b8d2a10217d)

packages/charts/src/state/utils/legend_row_count.ts
markov00: At this point, if I'm not missing another situation you can also get rid of 
```
const valuesWidth = valueCellWidths.map((w) => (w ? sharedMargin + w : 0));
```
and directly return the `sharedMargin` together with the valueWidth in the previous `legendValues.map`

packages/charts/src/components/legend/legend_list.tsx
markov00: Sounds good, with keeping it as it is

packages/charts/src/components/legend/_legend_grid_list.scss
markov00: we should probably evaluate with Giovanni if keeping the `grid list` in these cases is required or we can remove it:
```
layout: undefined, no aggregated value
layout: undefined, only CurrentAndLastValue value
```
Not actually sure about `layout: list, legend direction: vertical` 
This will definitely remain like that `layout: list, legend position: left/right` (but this is a special Grid)

packages/charts/src/state/selectors/get_longest_legend_formatted_value.ts
markov00: yea, probably not much benefit anyway, is ok to keep it like that

packages/charts/src/state/selectors/get_legend_max_formatted_value.ts
maryam-saeidi: I noticed locally that the calculated width didn't actually change with the fontWeight: 700 and it seems that's also the case in the storybook screenshots 🤔

packages/charts/src/state/utils/legend_row_count.ts
maryam-saeidi: Good catch! [e4d0498](https://github.com/elastic/elastic-charts/pull/2784/commits/e4d04989ed0e1c61a1759cc3941af0a19fc7e07e)

packages/charts/src/components/legend/_legend_grid_list.scss
maryam-saeidi: Do you mean not having the `layout: undefined` altogether?

Personally, I'd like to keep it, just in case we identify places where the previous behavior makes more sense. If there is no such case, we still just have an extra option matching the previous behavior, which doesn't hurt.

markov00: I will discuss this with @gvnmagni and create a subsequent issue to track it

Issue #682, opened 2020-05-25
Design discussion for composite charts


monfera:
[Heavily WIP; just saving occasionally]

# Intro

## What's a composite chart?

A **single** chart is a familiar thing, eg. a pie chart or a bar chart(*).

![image](https://user-images.githubusercontent.com/1548516/82819828-5919ae80-9ea1-11ea-9ee3-181d3c2217e2.png)

A **composite** chart has multiple constituents, most known for distinct, non-overlapping rectangular areas, typically with their own axes. They look like several charts laid out in some tiling, grid or freeform arrangement. 

The planar composition can be homogeneous, eg. a grid of identically sized rectangles, and a lot of commonality in terms of the axes, their data and the chart type (eg. all line charts), known as [small multiples](https://en.wikipedia.org/wiki/Small_multiple), faceted, lattice, trellis or matrix chart. Some of the name proliferation is historical baggage.
![image](https://user-images.githubusercontent.com/1548516/82819751-312a4b00-9ea1-11ea-9460-ba1066599aaa.png)

Alternatively, the planar composition may be heterogeneous(**), ie. its constituents don't usually share a common width/height, chart type etc., ie. it's not based on repetition but on partly shared projection. A common example is the marginal scatterplot: 

![image](https://user-images.githubusercontent.com/1548516/82820563-beba6a80-9ea2-11ea-86dc-0950aa043ee9.png)

_Side notes:_

_(*) Compositing isn't constrained to non-overlapping rectangles. For example, multilayer plots are also composites. Their data (or at least a common root data) may be shared, as well as their Cartesian projections. In an EDA tool, there might be a single toggle that switches between a multilayer view such as [this one](https://serialmentor.com/dataviz/overlapping-points.html) into a small multiples view, scatterplot on the left, contour plot on the right._
![image](https://user-images.githubusercontent.com/1548516/82821558-a1869b80-9ea4-11ea-8474-24056ce83a38.png)

[...]

PR #2785, opened 2026-02-16
feat(axis): add independentYDomain option for per-panel Y-axis scaling

pmoust:
## Summary

Adds an `independentYDomain` option to `SmallMultiplesSpec` that enables each small multiples panel to compute its own Y-axis domain independently, rather than sharing a single global Y-axis domain across all panels.

This is the upstream prerequisite for [elastic/kibana#2021](https://github.com/elastic/kibana/issues/2021), which requests per-panel Y-axis fitting for split charts (small multiples) in Kibana visualizations.

### Problem

When small multiples panels contain data with vastly different value ranges (e.g., panel A: 0–100, panel B: 0–10,000), the shared global Y-axis domain causes low-value panels to appear visually flat and unreadable.

### Solution

When `independentYDomain` is set to `true` on the `<SmallMultiples>` spec:

1. **Per-panel Y domains** are computed by grouping `DataSeries` by their panel key and calling `mergeYDomain()` on each group independently.
2. **Per-panel Y scales** are derived from these panel-specific domains so that `renderGeometries()` uses the correct scale for each panel's bars, lines, and areas.
3. **Per-panel axis ticks** are regenerated in `computePerPanelAxesGeomsSelector` so that each panel's Y-axis displays tick labels matching its own domain rather than the global one.

### API

```tsx
<SmallMultiples
  splitHorizontally="split_column"
  independentYDomain={true}  // new, @alpha, defaults to false
/>
```

## Changes

- `SmallMultiplesSpec` (`specs/small_multiples.ts`): Added `independentYDomain?: boolean` property.
- `y_domain.ts`: Added `getPanelKey()` and `mergeYDomainPerPanel()` functions.
- `SeriesDomainsAndData` (`state/utils/types.ts`): Added optional `yDomainsPerPanel` field.
- `computeSeriesDomains()` (`state/utils/utils.ts`): Conditionally computes per-panel Y domains.
- `computeSeriesGeometries()` / `renderGeometries()` (`state/utils/utils.ts`): Uses per-panel Y scales when available, falling back to global scales.
- `computeYScalesPerPanel()` (`utils/scales.ts`): New utility to create per-panel scale maps.
- `computePerPanelAxesGeomsSelector` (`state/selectors/compute_per_panel_axes_geoms.ts`): Regenerates Y-axis ticks per panel using panel-specific domains.
- `SmallMultiplesGroupBy` (`common/panel_utils.ts`): Propagates `independentYDomain` flag.
- `getSmallMultiplesIndexOrderSelector`: Passes `independentYDomain` through.

## Test plan

- [x] Unit tests for `mergeYDomainPerPanel()`: verifies per-panel domain computation, and confirms no per-panel domains when the flag is unset or `false`.
- [x] Integration test using `computePerPanelAxesGeomsSelector`: verifies that Y-axis tick labels differ between panels with distinct data ranges when `independentYDomain` is enabled.
- [x] TypeScript type check passes cleanly.
- [x] Full XY chart test suite passes (863 tests, 0 failures).
- [ ] Storybook story (`Small Multiples / Independent Y Domain`) for visual verification -- toggle the knob to compare shared vs. independent Y-axis scaling.

Ref: elastic/elastic-charts#682
Ref: elastic/kibana#2021

## Note to reviewers
This is heavily AI-driven, take care reviewing it


packages/charts/src/chart_types/xy_chart/state/selectors/compute_per_panel_axes_geoms.ts
```diff
@@ -31,23 +40,83 @@ const isPrimaryRowFn =
...
+      desiredTickCount: panelYDomain.desiredTickCount,
+      logBase: panelYDomain.logBase,
+      logMinLimit: panelYDomain.logMinLimit,
+      domainPixelPadding: panelYDomain.domainPixelPadding,
+      constrainDomainPadding: panelYDomain.constrainDomainPadding,
+    },
+  );
+
+  const ticks = scale.ticks();
+  const labelFormatter = axisSpec?.tickFormat ?? defaultTickFormatter;
```
timductive: This bypasses the existing axis tick formatter pipeline (labelFormat, tickFormat, and series formatter fallback) and only uses axisSpec.tickFormat. With independentYDomain enabled, panels can lose custom Y-axis formatting that works today without this flag. Could we reuse the same formatter source used by visible ticks / axis tick formatter selectors here instead of defaulting directly from axisSpec?

packages/charts/src/chart_types/xy_chart/state/selectors/compute_per_panel_axes_geoms.ts
```diff
@@ -31,23 +40,83 @@ const isPrimaryRowFn =
...
+
+        if (!panelYDomain) {
+          const updatedAxis = { ...geom.axis, panelTitle, secondary };
+          return { ...geom, axis: updatedAxis };
+        }
+
+        // When independentYDomain is active, force secondary to false for Y-axes
+        // so that each panel renders its own tick marks and labels
+        const updatedAxis = { ...geom.axis, panelTitle, secondary: false };
+        const panelHeight = vertical.bandwidth;
```
timductive: This hard-codes the Y-scale range to vertical.bandwidth. For rotated charts (90/-90), Y-domain axes are horizontal and should use horizontal.bandwidth, otherwise regenerated tick positions do not match the geometry scale. Can we derive this range from chart rotation (same rule used in computeSeriesGeometries) or reuse axisMinMax behavior?

pmoust:
I would have _never_ caught the 90/-90 rotation case. Thanks for the review, working on it

packages/charts/src/chart_types/xy_chart/state/selectors/compute_per_panel_axes_geoms.ts
pmoust: Fixed in f7ebac72ae — `regenerateYAxisTicks` now receives its formatter from the `getAxisTickLabelFormatter` selector, which resolves the full pipeline (`labelFormat` → `tickFormat` → series formatter fallback → `defaultTickFormatter`). Added a unit test that configures a custom `tickFormat` on the Y-axis spec and asserts all per-panel tick labels are formatted through it.

packages/charts/src/chart_types/xy_chart/state/selectors/compute_per_panel_axes_geoms.ts
pmoust: Fixed in f7ebac72ae — the panel size is now derived from chart rotation via `isHorizontalRotation(chartRotation) ? vertical.bandwidth : horizontal.bandwidth`, matching the same rotation-aware rule used in `computeSeriesGeometries`. Also added a storybook story and VRT baseline for 90°/-90° rotated charts to cover this path.

pmoust:
Pushed a follow-up commit addressing the remaining gaps:

- Replaced the direct `axisSpec.tickFormat` lookup with the `getAxisTickLabelFormatter` selector so the full formatter pipeline is respected (labelFormat → tickFormat → series fallback).
- Made the Y-scale range rotation-aware using `isHorizontalRotation` instead of hard-coding `vertical.bandwidth`.
- Added tests for `splitVertically` and 90° rotation to cover those paths at the unit level.
- Extracted the duplicated story data into a shared module and added a vertical-split story with VRT baseline.

markov00 (changes requested):
Thank you @pmoust for the contribution, but  the team doesn't have right now the time to properly review it.

The process we usually follow is to ask the team first if a specific task/issue can be considered for an external collaboration/contribution, the team will then evaluate and provide guidance, then the contribution could start.
Unfortunately with the ease of PR creation via AI, is faster to just create a PR, but this cause some issues:
- reviewing feature PRs is not the same as going through a simple fix PR, requires a careful review not just code checking, we need to verify if the feature make sense with the current direction the project is going, with the current Kibana/Visualization directions, understand the maintainability of the provided code (fundamental part because we will in the end the owner of that code and we need to support it for the time being) and scalability of the functionality.
- we need to step out suddenlty from our current priorities, sometimes with the time pressure of giving a review even when the author explicitly not trust the PR themself as 95% written by AI (as discussed on slack).

The issue this is trying to fix is very old (2014), we have planned to work and improve small multiples this year, but not in this quarter as we don't have capacity for that.
Small multiples are not used in Kibana/Lens today, nor is used in other Kibana solutions today (quickly searched today in main no trace of that except for the legacy/deprecated Visualize Editor).

I'd recommend to not merge this PR today, I thank you for the contribution and I'm happy to see that someone wants to take care of small multiples (a long time requested feature for Lens). This is a good way to push for this feature in our project board.

pmoust:
I share the same concerns. 
Discussed it w/ @timductive yesterday, around the risk of drive-by PRs overwhelming our engineering teams.

I hope reviewing and merging this gets prioritised soon.

Feel free to keep open or close as you see fit.

PR #2789, opened 2026-02-24
ci: update release workflow to use trusted published instead of npm token

maryam-saeidi:
## Summary

Update release workflow to use [trusted published](https://docs.npmjs.com/trusted-publishers) instead of npm token.

Provided by @markov00 :
https://semantic-release.gitbook.io/semantic-release/recipes/ci-configurations/github-actions#node-project-configuration

🤖 Assisted by Cursor


.github/workflows/release.yaml
```diff
@@ -72,6 +75,7 @@ jobs:
         uses: actions/setup-node@v4.0.2
         with:
           node-version: ${{ env.ECH_NODE_VERSION }}
+          registry-url: 'https://registry.npmjs.org'
```
nickofthyme: Did you find this to be required to work?

It look like the semantic-release docs say not to do this, see [docs](https://semantic-release.gitbook.io/semantic-release/recipes/ci-configurations/github-actions#important-avoid-registry-url-in-setup-node).

> Do not set the `registry-url` option in the `actions/setup-node` step when using semantic-release for npm publishing. The `registry-url` option causes `setup-node` to create an `.npmrc` file that can conflict with semantic-release's npm authentication mechanism, leading to `EINVALIDNPMTOKEN` errors even when your token is valid.

As this is the default I think we should remove this.


```suggestion
```

.github/workflows/release.yaml
```diff
@@ -81,9 +85,11 @@ jobs:
       - name: Build library
         run: yarn build
 
+      - name: Upgrade npm for trusted publishing (OIDC) support
+        run: npm install -g npm@11.5.1
```
nickofthyme, line 92: I don't love doing this separate than the node setup but I understand why.

It seems node `v22.22.0` is bundled with npm `v10.9.x`. But trusted release requires npm `v11.5.1` or later and node `v22.14.0` or higher.

Not for this PR but it would be nice if we use a newer node like `v24.14.0` that bundles with npm `v11.9.0`.

.github/workflows/release.yaml
```diff
@@ -5,6 +5,9 @@ name: Publish a Release
 on:
   workflow_dispatch
 
+permissions:
+  id-token: write
```
markov00: I think we should follow the workflow configuration proposed here https://semantic-release.gitbook.io/semantic-release/recipes/ci-configurations/github-actions#node-project-configuration
and in particular providing this permission just to the job requiring that permission (actually the `release` one)

.github/workflows/release.yaml
maryam-saeidi: Good catch, I will remove it 👍🏻

.github/workflows/release.yaml
maryam-saeidi: Do we need any of the other permissions as well, or just `id-token` one?
```
contents: write # to be able to publish a GitHub release
issues: write # to be able to comment on released issues
pull-requests: write # to be able to comment on released pull requests
```

maryam-saeidi: I added all these permissions for now, we can run the release and see how it goes and adjust it later.

PR #2790, opened 2026-02-24
build: update node version from 20.11.1 to 22.22.0

maryam-saeidi:
## Summary

Update node version from 20.11.1 to 22.22.0


package.json
```diff
@@ -172,7 +172,7 @@
     "webpack-dev-server": "^4.1.0"
   },
   "engines": {
-    "node": ">=10",
+    "node": ">=22",
     "yarn": ">=1"
```
nickofthyme, line 176: @markov00 do you see a need to keep compatibility here? I think it's fine just to force devs to use the latest version but it's not strictly necessary outside of the ci.

I don't think this affects 3rd-party consumers of charts that are on an older version of node as  `packages/charts/package.json` does not have an engine constraints.

packages/charts/src/chart_types/partition_chart/state/selectors/get_chart_type_description.ts
```diff
@@ -13,6 +13,6 @@ import { createCustomCachedSelector } from '../../../../state/create_selector';
 export const getChartTypeDescriptionSelector = createCustomCachedSelector(
   [getPartitionSpec],
   (partitionSpec): string => {
-    return `${partitionSpec?.layout} chart` ?? 'Partition chart';
+    return partitionSpec?.layout ? `${partitionSpec.layout} chart` : 'Partition chart';
```
nickofthyme, line 16: Old nonsensical coalescing.

packages/charts/src/chart_types/timeslip/timeslip/render/raster.ts
```diff
@@ -16,13 +16,13 @@ import type {
...
 } from '../../../xy_chart/axes/timeslip/continuous_time_rasters';
-import { MAX_TIME_GRID_COUNT, notTooDense } from '../../../xy_chart/axes/timeslip/multilayer_ticks';
+import { notTooDense } from '../../../xy_chart/axes/timeslip/multilayer_ticks';
 import type { NumericScale } from '../../projections/scale';
 import { clamp } from '../../utils/math';
 import type { TimeslipConfig } from '../config';
 import type { DataState } from '../data_fetch';
 
-const TIMESLIP_MAX_TIME_GRID_COUNT = 100 || MAX_TIME_GRID_COUNT; // use either
+const TIMESLIP_MAX_TIME_GRID_COUNT = 100; // or MAX_TIME_GRID_COUNT
```
nickofthyme, line 25: Roberts way or testing with different constants 😄

packages/charts/src/chart_types/xy_chart/domains/x_domain.ts
```diff
@@ -170,7 +170,8 @@ export function convertXScaleTypes(
   const timeZone = getZoneFromSpecs(specs);
   const type =
     scaleTypes.size === 1
-      ? scaleTypes.values().next().value // pick the only scaleType present
+      ? // TODO fix this non-null assertion
+        scaleTypes.values().next().value! // pick the only scaleType present
```
nickofthyme, line 174: This is caused by TS `5.6`, shipped with the new version of node, which now enforces stricter types around [`Iterator.next()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator#next) for built-in data types like `Set`, `Map`, etc.

I created https://github.com/elastic/elastic-charts/issues/2791 to fix this later.

package.json
```diff
@@ -171,8 +171,11 @@
     "webpack-cli": "^4.8.0",
     "webpack-dev-server": "^4.1.0"
   },
+  "resolutions": {
+    "@types/glob": null
+  },
```
nickofthyme: For some reason these changes create this error below...

```
Cannot find type definition file for 'glob'.
  The file is in the program because:
    Entry point for implicit type library 'glob'ts
```

The `@types/glob` is a transitive dependency of various storybook packages. I think upgrading these will eventually fix this.

For now this is the best way to make it happy. Alternatively, we could set the type in the TS compiler options to `["node"]` but this creates a waterfall affect causing other issues.

nickofthyme: Never mind....I realized the issue was that the `yarn.lock` was deleted and upon the next `yarn install` it tried to pull in all the latest versions based on the specified ranges for every dependency thus upgrading `typescript`, `lodash`, etc.

I reverted the changes to `yarn.lock` and this issue was resolved. It will likely be an issue when we attempt to upgrade TS next 😅.

nickofthyme:
buildkite update screenshots

package.json
markov00: I'm fine with that

e2e/screenshots/all.test.ts-snapshots/baselines/bar-chart/test-single-histogram-bar-chart-chrome-linux.png
markov00, line 1: It appears that we have some different behavior in the updated screenshots.
Upgrading node shouldn't change any chart visually. Do you mind verify this? probably caused by some of the changes in the x domain file

storybook/stories/bar/44_test_single_histogram.story.tsx
```diff
@@ -37,7 +37,7 @@ export const Example: ChartsStory = (_, { title, description }) => {
 
   const xDomain = {
     min: KIBANA_METRICS.metrics.kibana_os_load.v1.data[0][0] - (oddDomain ? 217839 : 0),
-    max: KIBANA_METRICS.metrics.kibana_os_load.v1.data[0][0] + (oddDomain ? 200000 : 0 ?? binWidthMs - 1),
```
markov00, line 40: This seems the culprit: `0 ?? something` evaluate to `0` not to `something`

maryam-saeidi: Yeah, that was the case.

packages/charts/src/chart_types/xy_chart/domains/x_domain.ts
maryam-saeidi: Thanks for creating a ticket!

PR #2798, opened 2026-03-02
fix(chart): Enhance Canvas based text measurement to account for various font features

awahab07:
## Problem

Text measurement should account for inherited CSS advanced font properties, not easily guessable/enumerable    (e.g. `font-feature-settings`, `font-variant-numeric` and other advanced font props/settings). Text layout calculations using a detached off-screen canvas could not inherit take into account all properties reliably, causing unprecedented text clipping or overlap in charts.

For example, when a consumer application Lens uses a specialized font [`"Elastic UI Numeric"`](https://github.com/elastic/kibana/pull/251576), Metric chart with `valueFontSize: 'fit'` gets the primary metric label clipped.

image

## Solution
An in-Dom hidden canvas should naturally inherit all font features/settings making text measurement accurate. This PR implement that and uses a zero sized hidden canvas element in every chart whose ctx passed on when measuring text for the chart.

## Details

The Canvas 2D `ctx.font` property only accepts CSS [`font` shorthand](https://developer.mozilla.org/en-US/docs/Web/CSS/font) syntax, which does not include `font-feature-settings` or `font-variant-numeric`. The [`CanvasTextDrawingStyles`](https://html.spec.whatwg.org/multipage/canvas.html#canvastextdrawingstyles) interface also has no `fontFeatureSettings` attribute. The only way for `measureText()` to account for these properties is through CSS inheritance — which requires the canvas element to be present in the DOM tree. A detached off-screen canvas has no parent, so inherited font properties default to `normal`, producing measurements that don't match the rendered output.

Additionally, not all chart types use a `<canvas>` for rendering (e.g. Metric and Bullet graph are fully DOM-rendered), yet they still rely on canvas-based `measureText()` for layout calculations — so there is no rendering canvas to reuse.

This change:
- Introduces a `ChartMeasureCanvas` component — a hidden, zero-size, `aria-hidden` `<canvas>` placed inside each Chart's `.echChart` DOM subtree so it inherits computed font properties from ancestors.
- Updates `withTextMeasure` to prefer the registered in-DOM canvas, falling back to an off-screen canvas when no Chart component is mounted.

### Performance

The in-DOM canvas is reused across all `withTextMeasure` calls for the lifetime of the Chart component, whereas the previous approach created a new off-screen canvas (`document.createElement('canvas')`) on every invocation. This reuse eliminates repeated canvas allocation and context creation, resulting in a net performance improvement. A DOM `<span>` + `getBoundingClientRect()` alternative was also evaluated but was significantly slower per call and triggers layout reflow, making canvas-based measurement the better choice. Also `ChartMeasureCanvas` component is kept to not re-render or re-mount against parent updates.

The following snippets demonstrate text measurement and performance different, with and without font features applied.

<table><tr><th>Without font feature</th><th>With font features</th></tr><tr><td>
image
</td><td>
image
</td></tr></table>

## Issues

Related to https://github.com/elastic/kibana/pull/251576

### Checklist

- ~[ ] New public API exports have been added to `packages/charts/src/index.ts`~


awahab07:
This following limitation has been circumvented when Lens introduced a specialized font [`"Elastic UI Numeric"`](https://github.com/elastic/kibana/pull/251576).

> ### Limitation
> Noticed that font features can't be applied to Canvas in **Firefox**.
> image

packages/charts/src/utils/bbox/canvas_text_bbox_calculator.ts
```diff
@@ -11,24 +11,60 @@ import { cssFontShorthand } from '../../common/text_utils';
...
+  let lastFont = '';
   return (text, font, fontSize, lineHeight = 1) => {
     if (text.length === 0) {
       return { width: 0, height: fontSize * lineHeight };
     }
+    const fontString = cssFontShorthand(font, fontSize);
+    // measureCtx doesn't need withContext, so not using it for performance reasons
+    if (isMeasureCtx) { 
+      // Avoid setting the font multiple times if it hasn't changed for performance
+      if (fontString !== lastFont) {
```
awahab07, line 59: This is to save on performance cost as performance trail shows `ctx.font` assignment is the most expensive call when measuring text (it's even expensive than `ctx.measureText()`).

Image

awahab07:
buildkite update vrt

PR #2802, opened 2026-03-04
feat(theme): set default labelOptions.widthLimit in themes and update metric.titleWeight default to 500

### LINKED ISSUE #2801: [Legend] Set default for legend.labelOptions.widthLimit in themes
### Summary

As you can see [here](https://github.com/elastic/elastic-charts/blob/main/packages/charts/src/utils/themes/light_theme.ts#L249), we set default values for the legend properties to eliminate the need for setting a default value in the component and define those properties as required without forcing the consumer to always pass those parameters.

In this [PR](https://github.com/elastic/elastic-charts/pull/2784), we introduced a new label option property called widthLimit and we need to make this field required and set a default for it in the themes. Something like:
```
labelOptions: {
      maxLines: 1,
      widthLimit: 250
}
```

> [!NOTE]
> If no theme is selected on the consumer side, the LIGHT theme will be used as a fallback.
>
> Users can adjust these properties by passing a different theme or adding customization (that will be merged together with the current base theme)

### Acceptance criteria
- Add default `widthLimit: 250` to the themes
- Make maxLines and widthLimit fields required

maryam-saeidi:
## Summary

Legend label truncation options now have sensible defaults in all built-in themes. `legend.labelOptions.widthLimit` defaults to **250px** (alongside `maxLines: 1`), so consumers don’t need to manually set these values to get consistent legend label truncation behavior.
Metric chart titles now default to a medium (500) font weight (instead of bold).

## Details

- Added `legend.labelOptions.widthLimit: 250` to all built-in themes (Light/Dark, Amsterdam, Legacy).
- Made `LegendLabelOptions.maxLines` and `LegendLabelOptions.widthLimit` required, relying on theme defaults instead of inline component fallbacks.
- Updated the API report (`packages/charts/api/charts.api.md`) to reflect the required fields.
- Updated built-in themes to use `metric.titleWeight: 500` instead of 'bold' (Related to https://github.com/elastic/kibana/issues/252834).

## Issues

Closes #2801

### Checklist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/charts/src/utils/themes/theme.ts
+++ b/packages/charts/src/utils/themes/theme.ts
@@ -400,8 +400,6 @@ export interface LegendLabelOptions {
    * Sets maxlines allowable before truncating
    *
    * Setting value to `0` will _never_ truncate the text
-   *
-   * @defaultValue 1
    */
   maxLines: number;
   /**
@@ -409,8 +407,6 @@ export interface LegendLabelOptions {
    * Only applicable if the list layout is chosen for the top and bottom positions
    *
    * Setting value to `0` will _never_ truncate the text
-   *
-   * @defaultValue 250
    */
   widthLimit: number;
 }


maryam-saeidi:
buildkite update vrt

maryam-saeidi:
buildkite build this

maryam-saeidi:
buildkite update vrt

packages/charts/src/utils/themes/theme.ts
```diff
@@ -403,7 +403,7 @@ export interface LegendLabelOptions {
    *
    * @defaultValue 1
```
mariairiartef: Since this property is now required and we moved the “default” behavior into the theme, having the default value comment can be misleading. I think we should remove it 🤔 

```suggestion
```

packages/charts/src/utils/themes/theme.ts
```diff
@@ -412,7 +412,7 @@ export interface LegendLabelOptions {
    *
    * @defaultValue 250
```
mariairiartef: ♻️ Same as before
```suggestion
```

e2e/screenshots/all.test.ts-snapshots/baselines/metric-alpha/array-of-values-chrome-linux.png
mariairiartef, line 1: suggestion: if we are updating the theme for legend and metric, maybe a better PR name would be `feat(theme): ...`, wdyt?

maryam-saeidi: Sounds good, thanks for the suggestion.

packages/charts/src/utils/themes/theme.ts
maryam-saeidi:
```suggestion
```

packages/charts/src/utils/themes/theme.ts
maryam-saeidi:
```suggestion
```

PR #2810, opened 2026-04-01
fix: guard process.env for browser environments without bundler polyfills

### LINKED ISSUE #2809: `ReferenceError: process is not defined` in browser without polyfill
**Describe the issue**

When using `@elastic/charts` in a browser environment where `process` isn't polyfilled by the bundler, the chart crashes on load with:

 ```
 ReferenceError: process is not defined
 ```

The library accesses `process.env` in several places (`chart_state.ts`, `logger.ts`) without checking if `process` exists first. Some other files (`animations/index.ts`, `flame_chart.tsx`) do have a guard but are missing a `null` check.

**Screenshots**

Image

**Version**
 - Elastic Charts: regression in `71.4.1`, work in `71.4.0`

**Errors in browser console**

-`Uncaught ReferenceError: process is not defined`

**Kibana Cross Issues**

might not be an issue in Kibana since it defines a global `process`.

walterra:
## Summary

Adds consistent safe guards to all `process.env` usage in shipped runtime code, since accessing `process.env` throws in browser environments where `process` is not a global (e.g. no bundler polyfill). 

### Changes

- `state/chart_state.ts` — guard `process.env.ELASTIC_CHARTS_DEV_MODE`
- `utils/logger.ts` — guard `process.env.NODE_ENV` in `isDevelopment()` and `isTest()`
- `xy_chart/.../animations/index.ts` — add missing `!== null` check to existing guard
- `flame_chart/flame_chart.tsx` — add missing `!== null` check to existing guard

All sites now use the same pattern consistent with existing POJO checks in the codebase, for example in `packages/charts/src/utils/accessor.ts.`

Fixes #2809.

### DIFF
--- a/packages/charts/src/chart_types/flame_chart/flame_chart.tsx
+++ b/packages/charts/src/chart_types/flame_chart/flame_chart.tsx
@@ -62,7 +62,7 @@ const LEFT_MOUSE_BUTTON = 1;
 const MINIMAP_SIZE_RATIO_X = 3;
 const MINIMAP_SIZE_RATIO_Y = 3;
 const SHOWN_ANCESTOR_COUNT = 2; // how many rows above the focused in node should be shown
-const SHOULD_DISABLE_WOBBLE = (typeof process === 'object' && process.env && process.env.VRT) === 'true';
+const SHOULD_DISABLE_WOBBLE = (typeof process === 'object' && process !== null && process.env && process.env.VRT) === 'true';
 const WOBBLE_DURATION = SHOULD_DISABLE_WOBBLE ? 0 : 1000;
 const WOBBLE_REPEAT_COUNT = 2;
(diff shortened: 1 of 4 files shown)


packages/charts/src/chart_types/flame_chart/flame_chart.tsx
```diff
@@ -62,7 +62,8 @@ const LEFT_MOUSE_BUTTON = 1;
 const MINIMAP_SIZE_RATIO_X = 3;
 const MINIMAP_SIZE_RATIO_Y = 3;
 const SHOWN_ANCESTOR_COUNT = 2; // how many rows above the focused in node should be shown
-const SHOULD_DISABLE_WOBBLE = (typeof process === 'object' && process.env && process.env.VRT) === 'true';
+const SHOULD_DISABLE_WOBBLE =
+  (typeof process === 'object' && process !== null && process.env && process.env.VRT) === 'true';
```
nickofthyme, line 66: Looks fine to me, we could consider abstracting this to a global util like `getProcess` or even `getProcessEnv` but that could be another time.

nickofthyme (approved):

PR #2813, opened 2026-04-02
fix: remove `produce()` from selectors

### LINKED ISSUE #2814: Click on legend item throws with `Uncaught TypeError: Cannot read properties of undefined (reading 'background')`
**Describe the issue**

When using Elastic Charts with `vite` bundler, clicking chart legend items throws with `Uncaught TypeError: Cannot read properties of undefined (reading 'background')`.

Example `package.json`:

```json
{
  "name": "charts-issue",
  "private": true,
  "version": "0.0.0",
  "type": "module",
  "scripts": {
    "dev": "vite",
    "build": "tsc -b && vite build",
    "lint": "eslint charts-issue",
    "preview": "vite preview"
  },
  "dependencies": {
    "@elastic/charts": "^74.0.0",
    "moment": "^2.30.1",
    "react": "^18.3.1",
    "react-dom": "^18.3.1"
  },
  "devDependencies": {
    "@eslint/js": "^9.39.4",
    "@types/node": "^24.12.0",
    "@types/react": "^19.2.14",
    "@types/react-dom": "^19.2.3",
    "@vitejs/plugin-react": "^6.0.1",
    "eslint": "^9.39.4",
    "eslint-plugin-react-hooks": "^7.0.1",
    "eslint-plugin-react-refresh": "^0.5.2",
    "globals": "^17.4.0",
    "typescript": "~5.9.3",
    "typescript-eslint": "^8.57.0",
    "vite": "^8.0.1"
  }
}
```

**Screenshots**

Image

 - Bundler: vite
 - Elastic Charts: 74.0.0

walterra:
## Summary

Fixes a bug where clicking a legend item would throw with `Uncaught TypeError: Cannot read properties of undefined (reading 'background')`.

## Details

The bug is an edge case that surfaced because of a combination of things a play:

1. Using selectors in redux slice reducers is an anti-pattern
2. Using immer's `produce` in selectors is an anti-pattern (we ended up doing it because of the previous anti-pattern)
3. Certain consuming build setups will end up with 2 different instances of `immer`, one within redux toolkit and a separate one when directly importing from `immer`. This in combination with the previous to anti-pattern will cause the throw.

Elastic Chart's own setup was not affected by the third issue. That's why we didn't encounter the bug in Elastic Chart's own jest tests or storybook. The bug surfaced in a project using `vite` as the bundler that consumed Elastic Charts.

The bug fix in this PR addresses 2) by removing the use of `produce` in the selector. This way 3) can no longer happen. In the long run we should address 1) too.

I tried creating a jest test scenario that would fail on `main` and pass with the fix applied in this PR but it's not really possible because the error is tied to certain build setup and therefor cannot be reproduced in our jest setup.

Some more analysis showed the error happened because RTK's `immer` and the directly imported `immer` ended up differently minified. 

- RTK's minified immer state: `{ t, o, k, j, ... }` (base, copy, draft, revoke)
- Charts' unminified immer state: `{ base_, copy_, draft_, revoke_, ... }`
 
So each instance recognizes the other's drafts. But when charts' `produce()` tries to read internals of a draft created by RTK's immer, it accesses `state.copy_` / `state.base_` on an object that only has `state.o` / `state.t` — both `undefined` — causing the crash.

I tested changing `import { produce } from 'immer'` to `import { createNextState as produce } from '@reduxjs/toolkit'` which would make both calls use the same immer instance which worked. But since again using `produce` in a selector is an anti-pattern I removed it's use anyway.

## Issues

Fixes https://github.com/elastic/elastic-charts/issues/2814

### Checklist

### DIFF
--- a/packages/charts/src/state/reducers/interactions.ts
+++ b/packages/charts/src/state/reducers/interactions.ts
@@ -7,7 +7,6 @@
  */
 
 import type { ActionReducerMapBuilder } from '@reduxjs/toolkit';
-import { produce } from 'immer';
 
 import { ChartType } from '../../chart_types';
 import { drilldownActive } from '../../chart_types/partition_chart/state/selectors/drilldown_active';
@@ -229,10 +228,7 @@ export const handleTooltipActions = (builder: ActionReducerMapBuilder<ChartSlice
 
       const index = state.tooltip.selected.findIndex((item) => createItemId(item) === createItemId(action.payload));
       if (index !== -1) {
-        // deleting from the immutable array using immer's produce
-        state.tooltip.selected = produce(state.tooltip.selected, (draft) => {
-          draft.splice(index, 1);
-        });
+        state.tooltip.selected.splice(index, 1);
       } else {
         state.tooltip.selected.push(action.payload);
       }
--- a/packages/charts/src/state/selectors/get_chart_theme.ts
+++ b/packages/charts/src/state/selectors/get_chart_theme.ts
@@ -6,8 +6,6 @@
  * Side Public License, v 1.
  */
 
-import { isDraft, produce } from 'immer';
-
 import { getSettingsSpecSelector } from './get_settings_spec';
 import { colorToRgba, overrideOpacity, RGBATupleToString } from '../../common/color_library_wrappers';
 import { clamp, mergePartial } from '../../utils/common';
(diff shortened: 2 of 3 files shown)


markov00 (approved):
Changes LGTM, thanks

PR #2816, opened 2026-04-08
feat(theme): Add `fontFamily` property to `MetricStyle`

awahab07:
## Summary

Adds `fontFamily` property to `MetricStyle` so that consumers can override fonts for Metric chart titles. The requirement emerged when Lens in Kibana started [introducing](https://github.com/elastic/kibana/issues/249382) the monospaced font`"Elastic UI Numeric"` for all charts in Lens. Since Metric chart was not accepting `fontFamily` prop (which Lens applies into themes of all Lens rendered charts), text measurements in Metric chart were still using the default hardcoded fontFamily string, in contrast to the rendered titles which were adopting the Lens applied "Elastic UI Numeric" font via DOM inheritance, resulting into unexpected text clipping ([see](https://github.com/elastic/kibana/issues/261506#issue-4216798452)).

The story **Test Cases -> Font Measurement Test** first **reproduces** the issue same as Lens

image

then validates that the **fix** would work in scenario's like Lens

image

## Details

This fixes the Metric fit calculation for cases where the rendered DOM inherits a different font family from its wrapper, such as Lens panels using `"Elastic UI Numeric"`.
The Metric DOM renderer now:
- measures titles, subtitles, and primary values with the configured font family instead of assuming the default chart font

## Issues

Fixes [elastic/kibana#261506](https://github.com/elastic/kibana/issues/261506)
Related: [elastic/elastic-charts#2811](https://github.com/elastic/elastic-charts/pull/2811)

### Checklist

- [x] The proper **chart type** label has been added (e.g. `:xy`, `:partition`)
- [x] The proper **feature** labels have been added (e.g. `:interactions`, `:axis`)
- [x] All related issues have been linked (i.e. `closes #123`, `fixes #123`)
- [x] New public API exports have been added to `packages/charts/src/index.ts`
- [x] Unit tests have been added or updated to match the most common scenarios
- [x] The proper documentation and/or storybook story has been added or updated
- [x] The code has been checked for cross-browser compatibility (Chrome, Firefox, Safari, Edge)
- [x] Visual changes have been tested with `light` and `dark` themes

PR #2817, opened 2026-04-10
ci(docs): Upgrade Docusaurus to `3.10.0`, pin `webpack` dep and track `docs/yarn.lock` for deterministic CI runs

awahab07:
## Summary

- upgrade the docs workspace to `@docusaurus/*@^3.10.0`
- track `docs/yarn.lock` so docs installs are deterministic in CI
- pin the transitive `webpack` version to `5.105.1`, which is compatible with the currently resolved `webpackbar@6.0.1`

## Problem

The docs step started failing on PR builds such as [build 223](https://buildkite.com/elastic/elastic-charts-build/builds/223) and [build 228](https://buildkite.com/elastic/elastic-charts-build/builds/228#019d71c9-301b-4cc2-86bd-b47c45500f3f/L397), including on [PR #2816](https://github.com/elastic/elastic-charts/pull/2816), with the following error:
```shell
[ERROR] Error: Unable to build website for locale en.
--
...
[cause]: ValidationError: Invalid options object. Progress Plugin has been initialized using an options object that does not match the API schema.
```

The underlying issue was not simply "Docusaurus 3.10 is broken". The docs workspace did not have a tracked lockfile, so [`yarn install --frozen-lockfile`](https://github.com/elastic/elastic-charts/blob/a23b25ca9ee227d74ccfe4e61dacd90155117259/.buildkite/utils/exec.ts#L115) had nothing to freeze in `docs/` and CI was free to resolve newer transitive versions than local development.

## Solution / Fix

This PR makes the docs install path deterministic and upgrades it intentionally:

- upgrade the docs workspace from `@docusaurus/*@^3.2.1` to `@docusaurus/*@^3.10.0`
- add a tracked `docs/yarn.lock`
- pin `webpack` to `5.105.1` via `resolutions` in `docs/package.json`

## Details / Background

- `yarn install --frozen-lockfile` is only protective when a lockfile is actually present and tracked. Without `docs/yarn.lock`, CI was still resolving fresh versions for the docs workspace.
- The blanket nested-lockfile ignore came from [#1572](https://github.com/elastic/elastic-charts/pull/1572). That PR ignored nested `yarn.lock` files globally, while `e2e/` explicitly re-included its own lockfile.
- `.buildkite/yarn.lock` is tracked because it predates that ignore rule. Tracked files stay tracked even if a later ignore rule would have matched them.
- The docs workspace was introduced later in [#2280](https://github.com/elastic/elastic-charts/pull/2280), after the blanket nested-lock ignore already existed, so `docs/yarn.lock` never entered git.
- The failure boundary lines up with the `webpack` release timing more than with a docs source change on the branch:
  - [build 222](https://buildkite.com/elastic/elastic-charts-build/builds/222) was created before `webpack@5.106.0` was published
  - [build 223](https://buildkite.com/elastic/elastic-charts-build/builds/223) and [build 228](https://buildkite.com/elastic/elastic-charts-build/builds/228) happened after that publish and hit the docs failure
- Local reproduction showed that `webpackbar@6.0.1` works through `webpack@5.105.1` and starts failing with `webpack@5.106.0+`, while Docusaurus `3.10.0` itself still builds successfully when `webpack` is kept on `5.105.1`.

## How to test

1. Run `yarn typedoc`
2. Run `DOCUSAURUS_URL="https://example.com" DOCS_URL="https://example.com" yarn --cwd docs build`
3. Run `yarn typecheck:all`
4. Run `yarn test`

The generated docs on this PR can also be compared with one of earlier build versions [e.g.](https://ech-e2e-ci--pr-2815-9l0pvkz6.web.app/)


awahab07:
buildkite test this

nickofthyme (approved):
Good catch, I didn't realize at the time non-workspace lock files were needed.

PR #2819, opened 2026-04-15
ci: fix on-merge builds

delanni:
## Summary
Fixes `main` branch build.

## Details
From a distance now it looks clear why this didn't work as intended. The PR builds are handled by the PR bot, but the main build is still handled by buildkite (the name: `buildkite-pr-bot` gives it away that it should be concerned with PRs only :))

For this, I enabled branch builds, with a branch filter on `main`. This should mean, we build when a new commit is pushed to main.

### Checklist

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -22,13 +22,14 @@ spec:
         ELASTIC_BUILDKITE_INFRA: "true"
       repository: elastic/elastic-charts
       provider_settings:
-        trigger_mode: none
+        trigger_mode: code
         build_branches: true
-        build_pull_requests: true
+        build_pull_requests: false
         build_tags: false
         prefix_pull_request_fork_branch_names: false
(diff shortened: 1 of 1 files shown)


delanni:
buildkite test this

nickofthyme:
buildkite test this

catalog-info.yaml
nickofthyme, line 1: Not sure what happened to the first build but the second worked.

nickofthyme (approved):
Thanks!

catalog-info.yaml
nickofthyme: Seems like it still doesn't trigger pushes to `main`. I don't see any builds for main other than my forced one. https://buildkite.com/elastic/elastic-charts-build

delanni: I probably shouldn't have set `build_pull_requests: false` now the builds are rejected. I'm going to fix that -.-

PR #2823, opened 2026-04-20
fix(a11y): allow keyboard navigation of tooltips

biamalveiro:
## Summary

Adds support for keyboard navigation of tooltips and `aria-describedby` label to tooltip triggers, aligning our annotation tooltips with the [APG's Tooltip Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/).

## Details

In order to link the line annotation marker buttons to their annotation tooltip in the accessibility tree, we give the tooltip a stable DOM id, and set the `aria-describedby` props on the marker to match. We pass on `chartId` from `<Annotations>` to each `<LineMarker>` to avoid possible collisions when several charts are on the same page.

Additionally, added an event handler for escape to close the tooltip to adhere to [WCAG Success Criterion 1.4.13 Content on Hover or Focus](https://www.w3.org/WAI/WCAG22/Understanding/content-on-hover-or-focus.html).

## Issues

Closes https://github.com/elastic/kibana/issues/148684, closes https://github.com/elastic/kibana/issues/148685.

### DIFF
--- a/packages/charts/src/chart_types/xy_chart/renderer/dom/annotations/line_marker.tsx
+++ b/packages/charts/src/chart_types/xy_chart/renderer/dom/annotations/line_marker.tsx
@@ -11,6 +11,7 @@ import { createPopper } from '@popperjs/core';
 import type { RefObject, CSSProperties } from 'react';
 import React, { useRef, useEffect, useCallback } from 'react';
 
+import { getAnnotationTooltipDomId } from './annotation_tooltip';
 import { DEFAULT_CSS_CURSOR } from '../../../../../common/constants';
 import type {
   onDOMElementEnter as onDOMElementEnterAction,
@@ -25,6 +26,7 @@ import type { AnimationOptions } from '../../canvas/animations/animation';
 import type { GetAnnotationParamsFn } from '../../common/utils';
 
 type LineMarkerProps = Pick<AnnotationLineProps, 'id' | 'specId' | 'datum' | 'panel'> & {
+  chartId: string;
   marker: AnnotationLineProps['markers'][number];
   chartAreaRef: RefObject<HTMLCanvasElement>;
   chartDimensions: Dimensions;
@@ -52,6 +54,7 @@ function getMarkerCentredTransform(alignment: Position, hasMarkerDimensions: boo
  */
 export function LineMarker({
   id,
+  chartId,
   specId,
   datum,
   panel,
@@ -122,6 +125,7 @@ export function LineMarker({
   void popper?.current?.update?.();
 
   const ariaLabel = datum.ariaLabel ?? datum.details ?? datum.header ?? `line annotation ${datum.dataValue}`;
+  const tooltipDomId = getAnnotationTooltipDomId(chartId, id);
 
   const handleEnter = useCallback(() => {
     onDOMElementEnter({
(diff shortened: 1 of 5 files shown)


packages/charts/src/chart_types/xy_chart/renderer/dom/annotations/line_marker.tsx
```diff
@@ -122,25 +125,30 @@ export function LineMarker({
...
-          type: DOMElementType.LineAnnotationMarker,
-          datum,
-        });
-      }}
+      onMouseEnter={handleEnter}
       onMouseLeave={() => onDOMElementLeave()}
-      onClick={() => onDOMElementClick()}
+      onFocus={handleEnter}
+      onBlur={() => onDOMElementLeave()}
+      {...(clickable ? { onClick: () => onDOMElementClick() } : {})}
```
markov00: I believe this is a bit an a11y anti-pattern, having a button with no action.
The browser will announce it as button but there is no action/click associated.
What if we use a div with a proper role instead, making it focusable with `tabindex="0"` ?

biamalveiro: yup, that's a good point, that's also my go-to for making chart elements keyboard accessible, however, I find it sometimes hard to pick an ARIA role that makes sense semantically in charts. 

Here, the closest would be `img` (or maybe `note`, but a bit of a stretch, since tooltip is the content). However, the linter is a bit unhappy with them here because they're not considered interactive and complains ([eslint-plugin-jsx-a11y/no-noninteractive-tabindex](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-noninteractive-tabindex.md))

As also neither were super expressive, that's why I went with button here, as it gave us the keyboard behavior without `tabIndex` on non-interactive role (and removed the conditional render), so it felt like a possible tradeoff. But agree that it's not really accurate, it's interactive but sets wrong expectations for on click. So happy to go with div+role and make those changes (w/ rule ignore), wdyt?

markov00: It is a tricky situation actually, at this point I believe the best solution is:
- add the `tabindex=0` to the `div`
- don't add a role to the `div`: existing roles don't apply very well for this case

packages/charts/src/chart_types/xy_chart/renderer/dom/annotations/line_marker.tsx
```diff
@@ -121,27 +115,65 @@ export function LineMarker({
...
+
+  const elementProps: HTMLAttributes<HTMLDivElement> & HTMLAttributes<HTMLButtonElement> = useMemo(() => {
+    const markerStyle: CSSProperties = {
+      ...style,
+      ...getAnimatedStyles(options, style),
+      color,
+      top: chartDimensions.top + position.top + panel.top,
+      left: chartDimensions.left + position.left + panel.left,
+      cursor: clickable ? 'pointer' : DEFAULT_CSS_CURSOR,
+    };
```
markov00, line 135: nice!

markov00 (approved):
Changes looks good to me!

PR #2825, opened 2026-04-21
ci: separate main build from PR builds

delanni:
## Summary
Separates a branch build pipeline from the PR build pipeline.

## Details
The reason why this is necessary is the way we're set up for our fork builds. 

We're triggering PR builds through the `elastic-buildkite-pr-bot` / `buildkite-pr-bot`. This bot ensures we're not building non-elastic forks. For this to work well, we need `build_pull_requests: true` combined with disabled code-triggers from buildkite (`trigger_mode: none`), this is required in order to prevent double builds (coming both from the bot trigger + buildkite's own trigger). This allows for manual PR builds, even from forks.

We also want `main` branch builds, when new PRs are merged. This requires `build_branches: true` + `branch_configuration: main`. However, this functionality requires triggers to be sent from Buildkite, as the PR-bot is not concerned with branch pushes - so we'd need to set `trigger_mode: code`. This however contradicts the above.

The resolution is to have 2 separate pipelines:
 - PR build, with manual triggers from `buildkite-pr-bot`: `trigger_mode: none` + `build_branches: false` + `build_pull_requests: true`
 - main build, triggered by buildkite: `trigger_mode: node` + `build_branches: true` + `build_pull_requests: false`


nickofthyme (approved):
Looks fine to me

delanni:
@nickofthyme but would you be OK with having another pipeline, or should we try to find a hack to get it working on a single pipeline

markov00 (approved):

nickofthyme:
> would you be OK with having another pipeline, or should we try to find a hack to get it working on a single pipeline

@delanni Sorry, yeah I think having separate pipelines is fine.

Issue #1490, opened 2021-11-19
[xy] Nicely compute the axis label overflow


markov00:
**Describe the bug**
When showing a categorical vertical bar chart and one label is way longer then all the others the following happen:
- we compute the axis label overflow (the space to leave blank on the right/left of an horizontal axis) depending on the maximum label size: we actually reduce the chat width by the maximum label size, putting half of that size to the right and half on the left of the chart
- the previous point happens before computing the visible labels, which could hide some of them due to occlusions.
- the result is that we will render a smaller chart to accommodate this "maximum size" label, but we are not showing that label due to occlusions.

**To Reproduce**
1. Create a vertical bar chart with 3 bars. 
2. Use a very long label as the x value for the bar in the middle of the chart
3. See the overflow is calculated also if the label is not shown due to its overlap

<img width="768" height="520" alt="Image" src="https://github.com/user-attachments/assets/90464bd6-c0d3-4368-ab38-460fa19508b9" />

**Expected behaviour**
We should limit the size of the tick labels, trim or put the text into multiple lines.
The overflow should be always justified: if we are showing a label that overflows one edge of the axis lets apply the overflow and reduce the chart size. Adjust the overflow depending on the actual overflow of the chart. Use the monotonic hill clumber method to compute the exact overflow

**Screenshots**
![image](https://user-images.githubusercontent.com/1421091/142617060-b09862db-9464-477e-9461-3b4964f7b07e.png)

**Additional context**
Previously described in https://github.com/elastic/elastic-charts/issues/758

markov00:
related to https://github.com/elastic/elastic-charts/issues/1179

markov00:
related to https://github.com/elastic/elastic-charts/issues/313

nickofthyme:
A related aspect to this is when long labels _squeeze_ the axis width assuming they are to be shown then hidden due to inability to show full label. This results in a skinny chart for no reason.

![Image](https://github.com/user-attachments/assets/0bc77151-f51b-49e9-b4c2-35b398b0f3e8)

teresaalvarezsoler:
This is also happening in Heatmaps when the screen is too narrow. 

<img width="443" height="292" alt="Image" src="https://github.com/user-attachments/assets/e97db19d-951d-4e21-91d4-ee234a3dad75" />

PR #2829, opened 2026-05-04
feat(axis): add option to truncate axis tick labels

biamalveiro:
## Summary

Adds ability to set a maximum length for axis tick labels, and truncate in case of overflow. Truncation can be configured to happen at start, middle or end of the string.

## Details

`AxisStyle.tickLabel` config has two new optional properties: `maxLength` (Pixels, or a string such as `80%`, in relation to the container's width) and  `truncate` (`end`, `start`, or `middle`). Formatters are wrapped with `withTickLabelTruncation`, so they are truncated if they overflow.

Additionally, this PR extends `fitText` to support truncation at start or middle of string, not only trailing ellipsis (end). This updated implementation searches for fit based on what is drawn (including the `...` when text is clipped), while the previous used the length of the prefix before swapping last character with an ellipsis. This led to [~1 character differences in snapshots that used text truncation](https://ech-e2e-ci--pr-2829-o9cyfg21.web.app/vrt-report/#?q=s:failed), (when `width(slice(0, v))` is different from `width(slice(0, v−1) + …)`). Updated VRT to fix these misses.

Added a [story where this behavior can be tested.](https://ech-e2e-ci--pr-2829-o9cyfg21.web.app/storybook/?path=/story/bar-chart--tick-label-truncation&globals=toggles.showHeader:true;toggles.showChartTitle:false;toggles.showChartDescription:false;toggles.showChartBoundary:false;theme:light) 

### Scope / follow-ups
This is intentionally a first pass: truncation is applied at the formatter using a max length in pixels (and optional ellipsis position), not a full label-box sizing model.

An idea of having `minSize` / `size` / `maxSize` on axis labels for more control was discussed but needs more thought, as it goes into `labelBox` calculations and we need to consider rotated labels and vertical/horizontal axis. Scoped out for this change, but to be revisited later.

## Issues

Related to https://github.com/elastic/kibana/issues/265517
### Checklist


biamalveiro:
buildkite update vrt

packages/charts/api/charts.api.md
nickofthyme, line 1: It's interesting that in most cases the truncation is applied when there is plenty of room. Most likely just padding and now that `...` is accounted for it , idk. But it's also mostly around the Partition chart which makes me wonder if we somehow accounted for this in the logic for the section link labels.

The thing that makes me question it most most is that the top padding is smaller in many cases than the side and it's still truncating.

Image

packages/charts/src/common/text_utils.ts
```diff
@@ -120,22 +121,59 @@ export function cutToLength(s: string, maxLength: number) {
...
+/** @internal */
+export function fitText(
+  measure: TextMeasure,
+  desiredText: string,
+  allottedWidth: number,
+  fontSize: number,
+  font: Font,
+  position: Truncate['position'] = 'end',
+) {
+  const ELLIPSIS = '…';
```
nickofthyme: I think we keep duplicating this definition, maybe we should share from `packages/charts/src/renderers/canvas/primitives/text.ts` or some common constants file.

packages/charts/api/charts.api.md
nickofthyme: After some searching I found this...

https://github.com/elastic/elastic-charts/blob/788058991ff891ed57347ac077959be05e8ecee8/packages/charts/src/chart_types/partition_chart/layout/viewmodel/link_text_layout.ts#L182-L185

So he used `2 * fontSize` to approximate the width of the ellipsis. I think we be remove this approximation with your changes and see less shifting in the labels, likely still a diff but smaller.

packages/charts/src/utils/themes/theme.ts
```diff
@@ -159,6 +159,12 @@ export interface Opacity {
   opacity: number;
 }
 
+/** @public */
+export interface Truncate {
+  width: Pixels;
```
nickofthyme: Did we ever settle on whether to also allow a percentage string relative to the chart width?

This would then allow us to set a default of like `'80%'`. That way we don't end up with charts like this...

Image

nickofthyme (approved):
Sorry for the delay, I wanted to really understand the logic and test it out.

I left a few things to check out and resolve before merging but nothing real major.

Really nice PR!  🚀

packages/charts/src/renderers/canvas/primitives/text.ts
```diff
@@ -60,8 +60,10 @@ export function renderText(
   });
 }
 
+/** @internal */
+export const ELLIPSIS = '…';
+
```
nickofthyme: It looks like this is causing a circular references and causing storybook to fail.

```
ERR! => Failed to build the preview
--
ERR! Detected 3 cycles which exceeds configured limit of 0
ERR! ../packages/charts/src/common/text_utils.ts -> ../packages/charts/src/renderers/canvas/primitives/text.ts -> ../packages/charts/src/common/text_utils.ts
ERR! ../packages/charts/src/renderers/canvas/primitives/text.ts -> ../packages/charts/src/common/text_utils.ts -> ../packages/charts/src/renderers/canvas/primitives/text.ts
ERR! ../packages/charts/src/utils/bbox/canvas_text_bbox_calculator.ts -> ../packages/charts/src/common/text_utils.ts -> ../packages/charts/src/renderers/canvas/primitives/text.ts -> ../packages/charts/src/utils/bbox/canvas_text_bbox_calculator.ts
ERR! asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
ERR! This can impact web performance.
ERR! Assets:
ERR!   main.775bebaf.iframe.bundle.js (8.85 MiB)
ERR!   vendors~main.ca43dca4.iframe.bundle.js (17.2 MiB)
ERR! entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
ERR! Entrypoints:
ERR!   main (26.1 MiB)
ERR!       runtime~main.5a0b8250.iframe.bundle.js
ERR!       vendors~main.ca43dca4.iframe.bundle.js
ERR!       main.775bebaf.iframe.bundle.js
```

You could move the import types from here...

```ts
import type { Font, TextAlign, TextBaseline } from '../../../common/text_utils';
```

To a separate type file or move these text constants their own file.

biamalveiro:
buildkite test this

packages/charts/src/utils/themes/theme.ts
biamalveiro: 👍 Updated, now tickLabel's  `maxLength?: Pixels | string;` can take such a string `x%`

packages/charts/src/renderers/canvas/primitives/text.ts
biamalveiro: Moved the text constants to `common/text_utils`, `common/` feels like the natural place for them. It was just those 3, so didn't move them to a separate file just yet.

biamalveiro:
buildkite update vrt

biamalveiro:
buildkite update vrt

packages/charts/api/charts.api.md
biamalveiro: good catch! that made a difference

packages/charts/src/chart_types/xy_chart/state/selectors/axis_tick_formatter.ts
```diff
@@ -30,24 +30,12 @@ export type AxisLabelFormatters = { x: Map<SpecId, AxisLabelFormatter>; y: Map<S
...
-  if (truncation?.width !== undefined && truncation?.relative === undefined) {
-    maxWidth = truncation.width;
-  }
-  if (truncation?.width !== undefined && truncation?.relative !== undefined) {
-    maxWidth = Math.max(truncation.width, truncation.relative * width);
-  }
-  if (!maxWidth || maxWidth <= 0 || !Number.isFinite(maxWidth)) return (formatter) => formatter;
-
-  const position = truncation?.position ?? 'end';
+  const maxWidth = maxLength ? getPercentageValue(maxLength, containerWidth, NaN) : undefined;
```
nickofthyme: Probably best to use `undefined` for the default as `NaN` will fail to fall into the early return.

```suggestion
  const maxWidth = maxLength ? getPercentageValue(maxLength, containerWidth, undefined) : undefined;
```

packages/charts/src/chart_types/xy_chart/state/selectors/axis_tick_formatter.test.ts
```diff
@@ -1,97 +0,0 @@
...
-
-import type { TextMeasure } from '../../../../utils/bbox/canvas_text_bbox_calculator';
-import { LIGHT_THEME } from '../../../../utils/themes/light_theme';
-import { withTickLabelTruncation } from './axis_tick_formatter';
-import { fitText } from '../../../../common/text_utils';
-
-jest.mock('../../../../common/text_utils', () => ({
-  ...jest.requireActual('../../../../common/text_utils'),
-  fitText: jest.fn(),
-}));
```
nickofthyme: Did these test move or are they obsolete?

packages/charts/src/chart_types/xy_chart/state/selectors/axis_tick_formatter.ts
```diff
@@ -30,24 +30,12 @@ export type AxisLabelFormatters = { x: Map<SpecId, AxisLabelFormatter>; y: Map<S
 export function withTickLabelTruncation(
   measure: TextMeasure,
   tickLabel: AxisStyle['tickLabel'],
-  width: number,
+  containerWidth: number,
```
nickofthyme, line 34: I wonder if we should just limit this to the vertical axis for now.

Currently the truncation is applied to both horizontal and vertical axes but the `maxLength` as a percent of the chart width for truncation of the x axis ordinal labels is a bit strange. The percentage should really be based on the bandwidth of the bar/bucket.

Image

nickofthyme: This also now makes me think it's strange to have the horizontal and vertical axes truncation settings tied to the same them value. I think it might be better to split them like we do with `tickLabel.alignment`.

https://github.com/elastic/elastic-charts/blob/8d96db05634e86916e09552e5f74aadea8d65f7e/packages/charts/src/utils/themes/theme.ts#L177

storybook/stories/axes/16_tick_label_truncation.story.tsx
```diff
@@ -24,33 +23,34 @@ const data = [
...
-  const widthPx = number('Truncation width', 120, { min: 0, max: 400, step: 10 });
-  const relative = number('Truncation relative', 0.3, { min: 0, max: 1, step: 0.05 });
-  const position = select<NonNullable<Truncate['position']>>(
-    'Truncation position',
-    { end: 'end', start: 'start', middle: 'middle' },
-    'middle',
-  );
+  const maxLength = parseThemeSize(text('maxLength', '120'));
+  const truncate = select<EllipsisPosition>('truncate', { end: 'end', start: 'start', middle: 'middle' }, 'middle');
+  const rotation = select('Chart rotation', { '0°': 0, '90°': 90, '-90°': -90 }, 90);
```
nickofthyme: FYI - there's a handful of common knobs we lump into `customKnobs` if you want to use that.

```suggestion
  const rotation = customKnobs.enum.rotation('Chart rotation')
```

storybook/stories/axes/16_tick_label_truncation.story.tsx
```diff
@@ -24,33 +23,34 @@ const data = [
...
+
+  const tickLabel = {
+    ...(maxLength !== undefined ? { maxLength } : {}),
+    truncate,
+  };
```
nickofthyme: Best to type this so if the types change it throws an error.

nickofthyme (commented):
I added a few more comments. I'm a little more doubtful about applying this to both axes instead of just the vertical axis, at least for this first pass to close the kibana issue. Also think it may be better to split the setting to control horizontal and vertical truncation separately.

packages/charts/src/chart_types/xy_chart/state/selectors/axis_tick_formatter.ts
biamalveiro: Like the alignment, they are separate:

https://github.com/biamalveiro/elastic-charts/blob/2d7ce1752fc7452eada83baf165a10a79c5887a2/packages/charts/src/utils/themes/theme.ts#L167-L187

It was only in this story that was being applied to both axis, as in, same knob value being passed on to both Axis components. But they are separate components, so, from an API standpoint, they are separate configs and not tied to the same value. But the story was indeed confusing, I split the knobs so it’s clearer

storybook/stories/axes/16_tick_label_truncation.story.tsx
biamalveiro: ah, good to know!

packages/charts/src/chart_types/xy_chart/state/selectors/axis_tick_formatter.test.ts
biamalveiro: most were obsolete, they were testing handling priorities from the `axisLabel.truncation` object from the first implementation. However, brought it back, this time with a couple targeted tests that make sense in the current state

packages/charts/src/chart_types/xy_chart/state/selectors/axis_tick_formatter.ts
biamalveiro: good catch! updated it to 0 (`getPercentageValue` default should be a number.), so it goes into the early return

storybook/stories/axes/16_tick_label_truncation.story.tsx
biamalveiro: (ended up removing this, in favor of the two clearer knobs for x,y)

packages/charts/src/chart_types/xy_chart/state/selectors/axis_tick_formatter.ts
biamalveiro: re: on horizontal axis it making more sense to be % of bandwidth in vertical bar charts, also agree that that’s more intuitive. However, as this is a general axis style, we’d need to split logic internally, such as % is bandwidth if in vertical bars, something else for other horizontal scales? 

I also noticed that bandwidth for bars can be influenced by the axis tick length, as it can determine the available space for the x axis, making it not linear to specify it in a single direction. here's a bit of an awkward case:

https://github.com/user-attachments/assets/8efc90cf-e1a6-4255-9a90-9f4875dbe82f

I feel like these definitely fall into stuff to explore in more detail in https://github.com/elastic/elastic-charts/issues/1490 (+ wrapping)? 

For the current moment, this PR allows setting a `maxLength` and set a safe soft boundary axis tick label length which can help in cases like these:

"Before"             |  "After"
:-------------------------:|:-------------------------:
Screenshot 2026-05-13 at 10 31 55  | Screenshot 2026-05-13 at 10 33 02

To be honest, because labels are horizontal by default, for either x/y, while not ideal (there are some awkward cases), it doesn’t feel super strange to me that, if there’s a possibility of a relative limit, that it is in relation to the maximum possible width (the chart’s). And also, makes it more predictable and interpretable, as it’s less likely to mean different things depending on other configs, and we can still keep one interface for `AxisStyle`, and not split for `HorizontalAxisStyle`/`VerticalAxisStyle`. 

But I'm totally ok with restricting it, if we feel it's the safest first step.

nickofthyme (approved):
This looks great! Sorry for all the back and forth. In the end, I think it ended up in a good spot.

biamalveiro:
Thanks @nickofthyme for having a thorough look at this + helping think through and improve the solution!

PR #2834, opened 2026-05-15
build: upgrade `uuid` and add cooldown period

nickofthyme:
## Summary

Adds a cooldown period for renovate so we can remove the cooldown for charts in kibana.

We also update `uuid` to the latest version.


nickofthyme:
Tested against kibana locally with tar file and works fine.

PR #2824, opened 2026-04-21
ci: Publish charts package tarball (`.tgz`) for PRs

awahab07:
## Summary

PR previews now publish a downloadable `@elastic/charts` tarball alongside the existing Firebase preview assets. This lets consumers like Lens (Kibana) validate upstream charts fixes in a deployed PR environment without waiting for the next `@elastic/charts` release.

### Problem
Kibana PR deploys like `/oblt-deploy` can only consume a published `@elastic/charts` pacakge, making upstream pre-release testing harder. Also, local dev testing requires manually packing and inline replacing elastic-charts package, which is fragile given Kibana's optimizer.

### Solution

Make the PR preview itself serve a real package tarball built from `packages/charts`, so downstream PRs can temporarily point `@elastic/charts` at that tarball and verify the upstream fix in a deployed Lens environment.

## This PR
1. PR builds now show links to the packaged tarballs

    image

2. `...web.app/packages/` static page will list available packages for the PR <a name=packages-page></a>

    image
    (Note that only three past commit based packages are retained to keep the static assets size in check.)

### Details

- Buildkite publishes a stable `pr.tgz` package and an immutable commit-specific tarball for each successful PR build.
- Firebase preview deployments serve those packages under `/packages/`, alongside `charts-package-manifest.json` and a semantic `packages/index.html` page with the PR reference, development-only disclaimer, and package descriptions.
- The preview retains the last three successful commit-specific packages so different builds can be compared without rebuilding the PR.
- The deployment comment links `Packages`, `pr.tgz`, and the current `<shortSha>.tgz` tarball on the same line.
- The manifest carries retained-package metadata so the `/packages/` page can render built dates for retained commit packages.
- Downstream repos can consume a specific preview build by resolving `@elastic/charts` dep directly to the commit-specific tarball URL, for example:
  ```json
  {
    "resolutions": {
      "@elastic/charts": "<elastic-charts-preview>/packages/elastic-charts-v71.5.0-pr-2824-ec7de70.tgz"
    }
  }
  ```
- The embedded package `@elastic/charts` version is not changed in preview pacakges to facilitate smoother dev testing.

## How to test

1. Run a PR build (assuming this PR's changes are in) and wait for the package step plus Firebase deployment to complete.
2. Confirm the deployment comment includes one package line in the format `Packages - pr.tgz - <shortSha>.tgz`.
[...]

PR #2837, opened 2026-05-20
chore: charts tarball is only published on PRs, not on main

delanni:
## Summary
https://github.com/elastic/elastic-charts/pull/2824 is not compatible with the branch build version of the pipelines, causing it to fail. This PR is expected to fix that.

## Details
https://github.com/elastic/elastic-charts/pull/2824 introduces tarball uploads and consequently checks for it being present, but the tarballs are not uploaded on non-PR builds, causing failed steps in the branch builds.


awahab07 (approved):
LGTM! Thanks for fixing this.

maryam-saeidi (approved):

PR #2845, opened 2026-05-27
feat(axis): improve axis label overflow

### LINKED ISSUE #313: Axes labels should wrap their text
**Describe the bug**
If there isn't enough available space, axes labels will get cut off.

**To Reproduce**
1. Create a chart with a long axes label

**Expected behavior**
Usually I'm a fan of just truncating with an ellipsis and then use `title` attribute to display the full label. The problem with this is that the chart label is too integral to the understanding of the chart to be hiding pieces of information, also I don't think you can add a `title` attribute to parts of a Canvas or even SVG element.

Therefore, I think the solution is just to allow the text to wrap lines.

**Screenshots**
Screen Shot 2019-08-14 at 15 25 26 PM
The label should say `6:00 - 6:14`

 - Elastic Charts: 9.0.4

### LINKED ISSUE #1490: [xy] Nicely compute the axis label overflow

biamalveiro:
+3.9K/ -2.5K, 300 files changed, wait, I know it looks scary! There are a lot of changes, but part of it was code reorg, breaking and moving functions from `axis_sizes` and `axis_utils` into different files, new tests, + screenshots for XY charts had to be updated as we changed overflow calculations leading to slightly bigger charts (see below). Still there are quite some significant changes (It's a lot, so if you're reviewing this, thank you 🙇‍♀️)

## Summary
Improves axis tick label overflow, by improving chart sizing logic and adding label wrapping. 

Previously, the axes reserved space based on the maximum label width before knowing which ticks would render and what available space was there for them. This PR replaces the single-pass sizing with a multi-step layout that:

1. uses each tick's own label size when deciding which ticks fit
2. limits individual tick labels to a budget and wraps/truncates them when they exceed it
3. iterates the chart sizing until the axis margins stabilise.

It also adds `minExtent` / `maxExtent` properties on the axis theme, inspired by Vega Axis API, and removes truncation logic that was previously used in the tick formatter.

[Storybook/overflow labels story](https://ech-e2e-ci--pr-2845-pv5tgqx7.web.app/storybook/?path=/story/axes--overflow-tick-labels&globals=toggles.showHeader:true;toggles.showChartTitle:false;toggles.showChartDescription:false;toggles.showChartBoundary:false;theme:light)
Examples:

Screenshot 2026-05-27 at 08 38 47
(wrapped + bottom aligned)

Screenshot 2026-06-02 at 09 00 49
(angled example)

---- 

### Per-tick label sizing in `visible_ticks`

The overlap check in [`getVisibleTicks`](https://github.com/biamalveiro/elastic-charts/blob/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/ticks/visible_ticks.ts#L109) can now uses each tick's own `tick.layout.bboxWidth`/`bboxHeight` instead of the maximum, and we also consider a `MIN_LABEL_GAP=4` space between labels. Previously, a tick could be culled even when there was enough room for it it. 

### Edge overflow in axis

Tick labels can overflow axis dimensions (e.g., a long tick label somewhere at the start on x axis). Before, half of the along axis size of the maximum label box was considered as as overflow on start/end sides of the axis. We now consider first and last ticks and the alignment for better calculating this overflow (which was previously was marked as a `TODO`). We also account for the existing outer padding + 1/2 bandwidth in band charts. Alternatively, we could pass exact tick label positions to `getAxesDimensions` to get exact measurements, however these are not available at the boostrap step (see "Multi-step chart sizing" below), so we'd need a back up for that first iteration.

**Updated E2E Screenshots**
Because of these new overflow calculations + new layout algorithm, charts end up being tiny bit bigger. **This is the cause of most of the failed VRT tests/update snapshots for all XY charts.**, as 

e.g.,

https://github.com/user-attachments/assets/97b31730-8e5f-46d7-8002-67772697bdbf

### Label overflow (truncation + wrapping)
Each label has an explicit budget resolved per axis calculated from [`resolveTickLabelConstraints`](packages/charts/src/chart_types/xy_chart/axes/dimensions.ts), it combines
- the band width / cross-axis container size,
- the configured `style.tickLabel.limit` and `style.tickLabel.wrapLines`,
- `minExtent` / `maxExtent` (see below)

[`createTickLabelLayout`](https://github.com/biamalveiro/elastic-charts/blob/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/ticks/labels.ts#L27) turns each label into a 2D `TickLabelBox` (`width`, `height`, `bboxWidth`, `bboxHeight`, `lines`) (see wrap+truncation below) that feeds into the rest of the layout pipeline. It uses existing [`wrapText`](https://github.com/biamalveiro/elastic-charts/blob/feat/axis-label-overflow/packages/charts/src/utils/text/wrap.ts#L24), that was extended to support truncation at `'start'`, `'middle'`, and `'end'` (previously only `'end'`). 

### Multi-step chart sizing
Tick label sizes depend on the axis size, but the axis size depends on the tick label sizes, the  🐔 /🥚 problem. 

The new [`computeChartLayout`](https://github.com/biamalveiro/elastic-charts/blob/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/state/selectors/compute_chart_layout.ts#L132) selector runs the sizing in a loop:
1. Bootstrap with initial estimate of tick dimensions ([`computeBaselineAxisTicksDimensionsSelector`](https://github.com/biamalveiro/elastic-charts/blob/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/state/selectors/compute_baseline_axes.ts#L127))
2. Measure axis margins from those dimensions ([`getAxesDimensions`](https://github.com/biamalveiro/elastic-charts/blob/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/dimensions.ts#L161)), then compute the chart area, and project ticks against the new space budget ([`computeVisibleTickSets`](https://github.com/biamalveiro/elastic-charts/blob/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/ticks/visible_ticks.ts#L247)).
3. Measure margins from the projected ticks again. If margins are stable (within `LAYOUT_EPSILON_PX`), return. Otherwise iterate, iterate up to to `MAX_ITERATIONS = 5`.

The previous selectors (`compute_axes_ticks_dimensions`, `visible_ticks`) are replaced by this loop.

### Min and max extent 
Inspired by [Vega-Lite's axis API](https://vega.github.io/vega-lite/docs/axis.html#general), added [`minExtent` and `maxExtent`](https://github.com/biamalveiro/elastic-charts/blob/feat/axis-label-overflow/packages/charts/src/utils/themes/theme.ts#L192-L194) (cross-axis size, expressed in pixels or as a `%` of the container's cross-axis size). These are resolved together with the fixed band (tick line + padding + title) by [`getAxisBand`](https://github.com/biamalveiro/elastic-charts/blob/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/dimensions.ts#L66) and final resolved size is given by [`measureAxisBand`](https://github.com/biamalveiro/elastic-charts/blob/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/dimensions.ts#L134), so the axis can respect these bounds.

This replaces the less intuitive [`tickLabelMaxLength`](https://github.com/biamalveiro/elastic-charts/blob/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/utils/specs.ts#L754) on `AxisSpec`, which is now deprecated (kept for backwards compatibility and resolved into `maxLineLength` inside `resolveTickLabelConstraints`).

### Remove truncation from the formatter
Since `createTickLabelLayout` is now responsible for the overflow, the truncation in `axis_tick_formatter` is removed. The corresponding `'start'` / `'middle'` / `'end'` truncation options were added on [`wrapText`](https://github.com/biamalveiro/elastic-charts/blob/feat/axis-label-overflow/packages/charts/src/utils/text/wrap.ts#L24).

Truncation also has now a readable lower bound instead of cutting labels down to unhelpful strings like "a…y". When a label is truncated (end, start, or middle), it keeps at least a minimum number of visible characters (4) and only shows an ellipsis when it hides a meaningful bit of text (at least 3 characters). If the available width is too narrow to satisfy that minimum, the label is left to overflow rather than being over-truncated ([readableTickLabelOptions](https://github.com/biamalveiro/elastic-charts/blob/50d9d65c6b9bc2867ebeef2ac2cd4aa5f5be63c2/packages/charts/src/chart_types/xy_chart/axes/ticks/labels.ts#L61-L67)). This is fine because overflowing labels are then removed by the existing visible-ticks/overlap logic. This makes truncation predictable regardless of chart width and bandwidth. 

### API Changes

There are new optional `AxisStyle` [theme properties](https://github.com/biamalveiro/elastic-charts/blob/50d9d65c6b9bc2867ebeef2ac2cd4aa5f5be63c2/packages/charts/src/utils/themes/theme.ts#L185-L198) `tickLabel.wrapLines`, `tickLabel.lineHeight`, `tickLabel.limit`, and `tickLabel.truncate`, these last two replace the deprecated `AxisSpec` props `tickLabelMaxLength` and `tickLabelTruncate`, respectively. These are all optional and fallback to defaults rather than being set explicitly in every theme

```
tickLabel.wrapLines: 1 // wrapping disabled, opt-in
tickLabel.lineHeight: 1.2
tickLabel.truncate: false // truncation disabled, opt-in
```

### Suggestions for review

Start by playing with the new story [storybook/stories/axes/16_overflow_tick_labels.story.tsx](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/storybook/stories/axes/16_overflow_tick_labels.story.tsx) to check behaviour and understand relevant parameters/knobs: wrap lines, truncation options, rotation, min/max axes extent, horizontal/vertical

Commit history isn't really helpful to go through this, work started around wrapping and then moved backward into layout, it's long and includes WIP states and approaches that were later removed.

I'd suggest this as the most relevant path: start with [theme.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/utils/themes/theme.ts) + [specs.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/utils/specs.ts) for API changes, then [axes/ticks/](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/ticks)(+[wrap](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/utils/text/wrap.ts)) to understand how tick labels are calculated. Then [axes/dimensions.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/dimensions.ts) (how ticks impacts axes margins). 

Then, [compute_baseline_axes.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/state/selectors/compute_baseline_axes.ts) + [compute_chart_layout.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/state/selectors/compute_chart_layout.ts), to see how the new layout comes together. 

Then check the rendering changes on [renderer/canvas/axes/tick_label.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/renderer/canvas/axes/tick_label.ts)

<details>
<summary>New stuff</summary>

- [chart_types/xy_chart/state/selectors/compute_chart_layout.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/state/selectors/compute_chart_layout.ts) (+ [compute_chart_layout.test.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/state/selectors/compute_chart_layout.test.ts) and snapshot) is the new iterative layout selector. It repeatedly re-measures axes against the chart area until the margins stabilize and is the single source of truth for chart dimensions and visible ticks. `computeChartDimensionsSelector` and `computeAxesGeometriesSelector` now just read from it.

- [chart_types/xy_chart/axes/ticks/labels.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/ticks/labels.ts) (+ [.test.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/ticks/labels.test.ts)), per-tick label layout, constraint resolution `resolveTickLabelConstraints`, the wrap + truncate (also has previous rotated bbox and max dimensions functions)

- [chart_types/xy_chart/axes/dimensions.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/dimensions.ts) (+ [.test.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/dimensions.test.ts)) — new `getAxesDimensions`, `measureAxisBand`, `getAxisBand` used by the layout loop, also the new `minExtent`/`maxExtent`.

- [utils/text/wrap.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/utils/text/wrap.ts) (+ [.test.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/utils/text/wrap.test.ts)),`wrapText` now has truncate: 'start' | 'middle' | 'end' modes.

- [utils/themes/theme.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/utils/themes/theme.ts#L170) (and theme defaults) + the deprecation in [chart_types/xy_chart/utils/specs.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/utils/specs.ts#L745-L753). Would be great to have feedback on this, if makes sense to be on theme or should be scoped to XY charts, as won't apply to charts other than XY?

- [chart_types/xy_chart/renderer/canvas/axes/tick_label.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/renderer/canvas/aces/tick_label.ts) (multi-line + rotation + alignment from tick.layout)
</details>

<details>
<summary>Moved</summary>
Same logic, some changes, different place.

- [chart_types/xy_charts/state/selectors/compute_axis_ticks_dimensions.ts](https://github.com/elastic/elastic-charts/tree/main/packages/charts/src/chart_types/xy_chart/state/selectors/compute_axis_ticks_dimensions.ts) renamed to [chart_types/xy_charts/state/selectors/compute_baseline_axes.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/state/selectors/compute_baseline_axes.ts). Still the initial axis tick measurements, but now feeds the iterative layout as a "bootstrap". 

-  [chart_types/xy_charts/state/selectors/visible_ticks.ts](https://github.com/elastic/elastic-charts/tree/main/packages/charts/src/chart_types/xy_charts/state/selectors/visible_ticks.ts)→ moved to [chart_types/xy_chart/axes/ticks/visible_ticks.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/ticks/visible_ticks.ts) (+ new [.test.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/ticks/visible_ticks.test.ts)), selector wrapper is gone, function is now used directly inside the layout loop.

- From [chart_types/xy_chart/utils/axis_utils.ts](https://github.com/elastic/elastic-charts/tree/main/packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts), split out:
    - `getTickLabelPosition`, to [axes/ticks/geometry.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/ticks/geometry.ts)
    - `computeRotatedLabelDimensions`  to [axes/ticks/labels.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/ticks/labels.ts)
    - `AxisTick`, `TextDirection`, `Projection`, `Projections`, to [axes/ticks/types.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/types.ts)
    - `getTitleDimension`, `getAllAxisLayersGirth`, to [axes/dimensions.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/axes/dimensions.ts)
    - `AxisGeometry`, `getAxesGeometries`, to [axes/geometry.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/geometry.ts) 
- [chart_types/xy_chart/utils/dimensions.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/utils/dimensions.ts),`computeChartDimensions` → `computeChartArea` (simpler).
</details>

<details>
<summary>Gone</summary>
- [chart_types/xy_chart/axes/axes_sizes.ts](https://github.com/elastic/elastic-charts/tree/main/packages/charts/src/chart_types/xy_chart/axes/axes/sizes.ts), replaced by [axes/dimensions.ts](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/packages/charts/src/chart_types/xy_chart/axes/dimensions.ts).
- [storybook/stories/axes/16_tick_label_truncation.story.tsx](https://github.com/elastic/elastic-charts/tree/main/storybook/stories/axes/16_tick_label_truncation.story.tsx), replaced by [16_overflow_tick_labels.story.tsx](https://github.com/biamalveiro/elastic-charts/tree/feat/axis-label-overflow/storybook/stories/axes/16_overflow_tick_labels.story.tsx)
</details>

## Issues

Closes https://github.com/elastic/elastic-charts/issues/1490 and closes https://github.com/elastic/elastic-charts/issues/313. 

Related to https://github.com/elastic/kibana/issues/238779

### Checklist

BREAKING CHANGE: The `AxisSpec` properties `tickLabelMaxLength` and `tickLabelTruncate` have been removed. Tick label sizing, wrapping, and truncation are now configured in the theme's `AxisStyle`, in `tickLabel` (`wrapLines`, `lineHeight`, `minLength`, `maxLength`, `truncate`) and the axis extent constraints `minExtent` / `maxExtent`. Migration: `tickLabelTruncate` to `theme.axes.tickLabel.truncate`; `tickLabelMaxLength` to `theme.axes.maxExtent` (pixels or `%` string,). For a fixed pixel limit on the label line length, use `theme.axes.tickLabel.maxLength` instead.


packages/charts/src/chart_types/xy_chart/state/selectors/merge_y_custom_domains.test.ts
```diff
@@ -0,0 +1,247 @@
+/*
```
biamalveiro, line 1: these tests were in `utils/axis_utils.test.ts`, just moved them to this separate file so it's closer to `merge_y_custom_domains`.

packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts
```diff
@@ -6,71 +6,17 @@
...
-import { innerPad, outerPad } from '../../../utils/dimensions';
+import type { Position, Rotation } from '../../../utils/common';
 import type { Range } from '../../../utils/domain';
-import type { AxisId } from '../../../utils/ids';
-import type { Point } from '../../../utils/point';
-import type { AxisStyle, TextAlignment, TextOffset, Theme } from '../../../utils/themes/theme';
-import type { ScaleConfigs } from '../state/selectors/get_api_scale_configs';
-import type { Projection } from '../state/selectors/visible_ticks';
+import type { AxisStyle } from '../../../utils/themes/theme';
 import type { SeriesDomainsAndData } from '../state/utils/types';
```
biamalveiro, line 19: Much of the logic around calculating axes geometries used to live here, moved it to `axes/` and split it into separate files per responsability.

packages/charts/src/chart_types/xy_chart/state/selectors/axis_tick_formatter.ts
```diff
@@ -26,30 +22,6 @@ export type AxisLabelFormatter<V = unknown> = (value: V) => string;
 /** @internal */
 export type AxisLabelFormatters = { x: Map<SpecId, AxisLabelFormatter>; y: Map<SpecId, AxisLabelFormatter> };
 
-/** @internal */
-export function withTickLabelTruncation(
```
biamalveiro, line 30: Truncation no longer happens in formatting, but during tick layout that handles sizing, wrapping and truncation.

packages/charts/src/chart_types/xy_chart/axes/geometry.ts
```diff
@@ -0,0 +1,126 @@
...
+    panelTitle?: string; // defined later per panel
+    secondary?: boolean; // defined later per panel
+    multilayerTimeAxis: boolean;
+  };
+  dimension: TickLabelBox;
+  visibleTicks: AxisTick[];
+}
+
+/** @internal */
+export function getAxesGeometries(
```
biamalveiro, line 79: previously in https://github.com/elastic/elastic-charts/blob/cce8260b26b7c534ae253299d31f49f1e4c9b163/packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts#L340

Adapted, as some of its logic was moved to separate helpers (`getPosition` now calls `measureAxisBand`)

packages/charts/src/chart_types/xy_chart/state/selectors/visible_ticks.ts
```diff
@@ -1,397 +0,0 @@
-/*
```
biamalveiro, line 1: (moved) it's no longer a state selector, so extracted logic to [axes/ticks/visible_ticks.ts](https://github.com/elastic/elastic-charts/pull/2845/changes#diff-bb3f0fb544e333e2c8f2a067940cb1b58cc41adbe5eb2e5027323cc2dc3e1a6c)

packages/charts/src/chart_types/xy_chart/state/selectors/compute_chart_layout.ts
```diff
@@ -0,0 +1,228 @@
...
+  return [
+    Math.abs(a.top - b.top) <= LAYOUT_EPSILON_PX,
+    Math.abs(a.bottom - b.bottom) <= LAYOUT_EPSILON_PX,
+    Math.abs(a.left - b.left) <= LAYOUT_EPSILON_PX,
+    Math.abs(a.right - b.right) <= LAYOUT_EPSILON_PX,
+  ].every(Boolean);
+};
+
+/** @internal */
+export function computeChartLayout(params: LayoutParameters): ChartLayout {
```
biamalveiro: This is the new iterative layout algorithm, to avoid the circular dependency between axis size and tick-label layout:
- The space an axis reserves depends on the measured tick-label boxes.
- But the tick labels themselves depend on the available space (they wrap, truncate, rotate, and are filtered for visibility based on the chart area)

So we can't compute either one in a single pass, so we we do multiple passes. Each iteration measures margins from the current ticks, recompute the chart area, re-project ticks into that area, re-measure margins. If the margins on all four sides move by less than `LAYOUT_EPSILON_PX` we consider the layout to be stable and we return

packages/charts/src/chart_types/xy_chart/state/selectors/__snapshots__/compute_chart_layout.test.ts.snap
```diff
@@ -0,0 +1,49 @@
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
+
+exports[`computeChartLayoutSelector with deterministic text measurement hidden axes do not reserve any space 1`] = `
```
biamalveiro: Added to cover same intent of the previous `packages/charts/src/chart_types/xy_chart/utils/__snapshots__/dimensions.test.ts.snap` snapshot tests that was removed.

packages/charts/src/chart_types/xy_chart/axes/axes_sizes.ts
```diff
@@ -1,114 +0,0 @@
...
-  smSpec: SmallMultiplesSpec | null,
-  xScaleType: ScaleType,
-  rotation: Rotation,
-): PerSideDistance & { margin: { left: number } } {
-  const sizes = [...axisDimensions].reduce(
-    (acc, [id, tickLabelBounds]) => {
-      const axisSpec = getSpecsById<AxisSpec>(axisSpecs, id);
-      if (tickLabelBounds.isHidden || !axisSpec) return acc;
-      const multilayerTimeAxis = isMultilayerTimeAxis(axisSpec, xScaleType, rotation);
-      // TODO use first and last labels
```
biamalveiro, line 80: The logic in this file now lives in `packages/charts/src/chart_types/xy_chart/axes/dimensions.ts`, and we also implemented what this TODO intended.

packages/charts/src/chart_types/xy_chart/axes/ticks/visible_ticks.test.ts
```diff
@@ -0,0 +1,213 @@
...
+const linearScale = () => {
+  const xDomain = MockXDomain.fromScaleType(ScaleType.Linear, {
+    domain: [0, 100],
+    isBandScale: false,
+  });
+  return computeXScale({ xDomain, totalBarsInCluster: 0, range: [0, 200] });
+};
+
+describe('generateTicks', () => {
+  test('skips NaN tick values', () => {
```
biamalveiro, line 45: These tests were previously in `packages/charts/src/chart_types/xy_chart/utils/axis_utils.test.ts`.

biamalveiro:
buildkite update vrt

biamalveiro:
buildkite update vrt

biamalveiro:
buildkite update vrt

e2e/screenshots/all.test.ts-snapshots/baselines/line-chart/test-path-ordering-chrome-linux.png
biamalveiro, line 1: Same expected width change as others. However, unsure why we see a change in missing data behaviour. New one seems to be the one that matches data though? https://github.com/elastic/elastic-charts/blob/main/storybook/stories/line/10_test_path_ordering.story.tsx#L21-L40

e2e/screenshots/all.test.ts-snapshots/baselines/area-chart/test-stacked-with-missing-values-chrome-linux.png
biamalveiro, line 1: Same expected width change as others. However, unsure why we see a change in missing data behaviour. New one seems to be the one that matches data though? https://github.com/elastic/elastic-charts/blob/main/storybook/stories/area/16_test_stacked_with_missing.story.tsx#L38-L124

(same as https://github.com/elastic/elastic-charts/pull/2845#discussion_r3440858897)

packages/charts/src/chart_types/xy_chart/axes/ticks/visible_ticks.ts
```diff
@@ -0,0 +1,398 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
```
biamalveiro, line 8: previously at https://github.com/elastic/elastic-charts/blob/main/packages/charts/src/chart_types/xy_chart/state/selectors/visible_ticks.ts

packages/charts/src/chart_types/xy_chart/axes/geometry.ts
```diff
@@ -0,0 +1,129 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
```
biamalveiro, line 8: previously in https://github.com/elastic/elastic-charts/blob/main/packages/charts/src/chart_types/xy_chart/utils/axis_utils.ts#L263, small adjustments

packages/charts/src/chart_types/xy_chart/axes/ticks/visible_ticks.ts
```diff
@@ -0,0 +1,397 @@
...
+  const { showOverlappingTicks, showOverlappingLabels, position } = axisSpec;
+  const bypassOverlapCheck = showOverlappingLabels || multilayerTimeAxis;
+  if (bypassOverlapCheck) return allTicks;
+
+  return allTicks
+    .slice()
+    .sort((a: AxisTick, b: AxisTick) => a.position - b.position)
+    .reduce(
+      (prev, tick) => {
+        const requiredSpace = (isVerticalAxis(position) ? tick.layout.bboxHeight : tick.layout.bboxWidth) / 2;
```
biamalveiro, line 185: This is the bit where we use the actual tick size instead of the maximum label box

biamalveiro:
buildkite update vrt

biamalveiro:
buildkite test this

packages/charts/src/chart_types/xy_chart/state/selectors/compute_chart_layout.ts
```diff
@@ -0,0 +1,217 @@
...
+      const nextMargins = measureMargins(nextProjections);
+
+      if (isLayoutStable(margins, nextMargins)) {
+        return {
+          dimensions: computeChartArea(container, nextMargins, theme),
+          ticks: nextProjections,
+          meta: { iterations: i + 1 },
+        };
+      }
+      Logger.warn('Layout did not converge after', i + 1, 'iterations');
```
markov00: We can probably remove this line, as it will appear and start pollute Kibana console.logs.
For now you can just comment this out, we can consider for the future adding a Logger.info level or something similar that is executed only when running elastic-charts locally not bundled

packages/charts/src/utils/themes/light_theme.ts
```diff
@@ -215,6 +215,8 @@ export const LIGHT_THEME: Theme = {
         vertical: 'near',
         horizontal: 'near',
       },
+      wrapLines: 2,
+      lineHeight: 1.2,
```
markov00: This basically can be considered now a breaking change, so we should merge correctly this PR with the right `BREAKING CHANGE:` PR merge description

packages/charts/src/utils/text/wrap.ts
```diff
@@ -28,25 +30,106 @@ export function wrapText(
   measure: TextMeasure,
   locale: string,
   granularity: Granularity = 'word',
+  truncate: Truncate = 'end',
 ): WrapTextLines {
-  const lines: WrapTextLines = [] as any;
-  lines.meta = {
-    truncated: false,
-  };
+  if (maxLines <= 0) return Object.assign([], { meta: { truncated: false } });
```
markov00, line 36: just as a suggestion, nothing too serious, since we are running this up to 5 interations we could consider a max length for the text in characters (something like 500/1000 charts) so "malicious" strings are not passed to the chart library causing performance troubles

packages/charts/src/chart_types/xy_chart/state/selectors/compute_chart_dimensions.ts
```diff
@@ -6,28 +6,11 @@
...
-    computeAxisTicksDimensionsSelector,
-    getAxesStylesSelector,
-    getAxisSpecsSelector,
-    getSmallMultiplesSpec,
-    getScaleConfigsFromSpecsSelector,
-    getSettingsSpecSelector,
-  ],
-  computeChartDimensions,
+  [computeChartLayoutSelector],
+  (layout) => layout.dimensions,
```
markov00: At this point we can replace every call of `computeChartDimensionSelector` with `computeChartLayoutSelector` to have one less layer

packages/charts/src/chart_types/xy_chart/axes/ticks/labels.ts
```diff
@@ -0,0 +1,196 @@
...
+  } else if (isContinuousScale(scale) && scale.bandwidth > 0) {
+    maxLineLength = Math.max(MIN_LABEL_LENGTH, maxLineLength ?? band.maxExtent);
+  } else {
+    const categorySlotWidth = isBandScale(scale) ? scale.step * (1 - scale.barsPadding / 2) : 0;
+    const bandwidthCap = categorySlotWidth > 0 ? categorySlotWidth : band.maxExtent;
+    maxLineLength = Math.max(MIN_LABEL_LENGTH, maxLineLength ?? bandwidthCap);
+  }
+
+  const lineHeightPx = style.tickLabel.lineHeight * style.tickLabel.fontSize;
+  let maxWrapLines = style.tickLabel.wrapLines;
```
markov00: should we add a safety net to avoid 0 and negative values here?
when I set wrapLines to zero it starts behaving wrongly

packages/charts/src/chart_types/xy_chart/axes/ticks/visible_ticks.ts
```diff
@@ -0,0 +1,397 @@
...
+
+  const { showOverlappingTicks, showOverlappingLabels, position } = axisSpec;
+  const bypassOverlapCheck = showOverlappingLabels || multilayerTimeAxis;
+  if (bypassOverlapCheck) return allTicks;
+
+  return allTicks
+    .slice()
+    .sort((a: AxisTick, b: AxisTick) => a.position - b.position)
+    .reduce(
+      (prev, tick) => {
```
markov00, line 184: Now that we wrap labels, should we just return all ticks everytime? since ticks should be bounded to the bar width, there shouldn't be a tick that is longer then the others.
The only possible case is when with `isCompactSingleWord` that we don't wrap lines with words of 10 charts. If that is not a requirement we can probably skip this case and always wrap words so we can consider everytime all ticks without computing who overlaps each other.

markov00: This could be a symptom: 


https://github.com/user-attachments/assets/c204cafa-1ba9-4720-9a52-4325ef403e03

packages/charts/src/chart_types/xy_chart/state/selectors/compute_chart_layout.ts
biamalveiro: Makes sense, [commented it out](https://github.com/elastic/elastic-charts/pull/2845/changes/cf9b344e8734802cfbb346b0340b314b51d4bf8a#diff-481fdb99e06ecdb307cba3768c4fc9c9c53bd1d4652239d2d13d30b20d0816b0).

packages/charts/src/utils/themes/light_theme.ts
biamalveiro: This was to make the new behavior explicit. But was wondering if could make sense to, on first integration to make these optional (+remove from theme) and rely on defaults that match current behavior (wrapLines: 1), and have these parameters as opt in, to smooth the "breaking changes"? 😅

packages/charts/src/utils/text/wrap.ts
biamalveiro: Fair point, in https://github.com/elastic/elastic-charts/pull/2845/changes/cf9b344e8734802cfbb346b0340b314b51d4bf8a

packages/charts/src/chart_types/xy_chart/state/selectors/compute_chart_dimensions.ts
biamalveiro: 👍 , in https://github.com/elastic/elastic-charts/pull/2845/changes/b93a9997e685e398db0614cb3dd9a8aee57466ca

packages/charts/src/chart_types/xy_chart/axes/ticks/labels.ts
biamalveiro: yup, for sure (https://github.com/elastic/elastic-charts/pull/2845/changes/1905d3f901d3280e76ab79d83994d1173492a6eb)

packages/charts/src/chart_types/xy_chart/axes/ticks/visible_ticks.ts
biamalveiro: I think in the example, category flashes in an out as the space for that label changes based on how the label on the left is being truncated + "category" is `isCompactSingleWord` so it won't truncate, either fits fully or is gone. Might be worth it to also have a limit to how much truncation is acceptable, like those labels that are only 3/4 characters + ellipsis aren't that useful. 

Regarding the overlap check. Yup, the `isCompactSingleWord` from the example is an reason, the other is continuous numeric (and single-layer time) axes where there are no bars, so `maxLineLength` falls back to the axis extent and tick positions can overlap. Those axes don't wrap either, so they rely entirely on dropping overlapping ticks rather than wrapping/truncating. But I agree, it might make sense to split logic per axis type, so we avoid running certain logic, like the wrap/truncation, in axes where these concepts don't apply.

Maybe worth it that I open an issue with these followups?

biamalveiro:
buildkite update vrt

biamalveiro:
buildkite test this

packages/charts/src/utils/themes/light_theme.ts
markov00: sure, make sense!

packages/charts/src/chart_types/xy_chart/state/selectors/compute_chart_dimensions.ts
markov00: nice thanks

packages/charts/src/chart_types/xy_chart/axes/ticks/visible_ticks.ts
markov00: sure, please open an issue with followups

markov00 (approved):
Updated changes looks good to me, thanks for this great work!

biamalveiro:
buildkite update vrt

packages/charts/src/utils/themes/light_theme.ts
biamalveiro: updated [here](https://github.com/elastic/elastic-charts/pull/2845/commits/ddcbfb4398f2556cdd6f0f99f1e716bde8749b2c), they are now optional, so removed them from the themes. Current defaults make wraping and truncation opt in features.

packages/charts/src/chart_types/xy_chart/axes/ticks/visible_ticks.ts
biamalveiro: Ended up implementing the safer truncation [here](https://github.com/elastic/elastic-charts/pull/2845/changes/a136a0a1da64f0f2dd5c531d3bb60e5e1da1303d), felt relevant given that default for wrap lines is 1, we could run into a lot of these over truncated strings in (verticla) bar charts whenever truncation would be enabled.

(before)

https://github.com/user-attachments/assets/182bdb33-dffd-4bd8-a615-0d96651fcd0e

(after)

https://github.com/user-attachments/assets/b2777a85-68fe-4e46-92b6-b8e6335a375a

(https://github.com/elastic/elastic-charts/pull/2845/changes/a136a0a1da64f0f2dd5c531d3bb60e5e1da1303d)

biamalveiro:
buildkite test this

biamalveiro:
buildkite update vrt

biamalveiro:
buildkite update vrt

biamalveiro:
buildkite update vrt

biamalveiro:
buildkite test this

packages/charts/api/charts.api.md
```diff
@@ -305,10 +307,11 @@ export interface AxisStyle {
...
         alignment: TextAlignment;
-        wrapLines: number;
-        lineHeight: number;
-        limit?: Pixels;
-        truncate?: Truncate;
+        wrapLines?: number;
+        lineHeight?: number;
+        maxLength?: Pixels;
+        minLength?: Pixels;
+        truncate?: Truncate | false;
```
markov00: These are properties from the Theme and we can consider these as mandatory (because the theme is predefined and clear by moving all the default you have applied into the theme itself).
If there is a specific property that can't be defaulted becuase you need a second semantic: undefined mean not set and a value mean set, then ok, but I don't see the case probably here.

markov00: All these have a default in theme already, except for maxLength, that we can consider as `Infinity` if we don't truncate.wdyt?

markov00: thanks

PR #2846, opened 2026-05-27
feat(components): Add Meter component

awahab07:
## Summary
- add a public `Meter` component with single, solid-palette, and gradient fill modes
- support positive-only, negative-only, and mixed-sign domains, including zero-anchored fills and review-driven clamping/normalization fixes
- refactor Metric progress bars to render through `Meter` without changing the focused Metric progress-bar regressions
- add Meter Storybook coverage and dedicated Playwright VRT coverage

Note that the Meter component is implemented as a refactor to the existing progress bar implementation inside Metric chart's renderer and then exposed publicly as well. Thus, small variations (few bytes change) in Metric and Bullet chart e2e VRT snapshots are expected.

Meter component stories demonstrate the usage:

image

<table><tr><td>
Components/Meter -> Positive Modes
image
</td><td>
Components/Meter -> Signed Domains
image
</td><td>
Components/Meter -> Markers And Vertical
image
</td></tr></table>

## Details

- add a public `Meter` component to `@elastic/charts`
- support `single`, `solid`, and `gradient` fill styles for positive-only, negative-only, and mixed-sign domains
- keep Metric's existing progress-bar API intact while moving its DOM rendering onto `Meter`
- preserve zero-anchored mixed-sign behavior, including a zero-baseline cue for signed domains
- add Meter stories for positive modes, signed domains, markers/vertical layouts, and a grouped signed-range comparison story
- add dedicated Meter Playwright VRT coverage and refresh its snapshots
- clamp and round Metric percentage output used for titles/ARIA, and fix Meter story layout/theme issues

## Issues

Implements the upstream requirement for elastic/kibana#250708.

### Checklist


awahab07:
buildkite update vrt

awahab07:
buildkite update vrt

packages/charts/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,418 @@
...
+  ariaValueText,
+  ariaValueMin,
+  ariaValueMax,
+  ariaValueNow,
+}) => {
+  const isVertical = orientation === LayoutDirection.Vertical;
+  const normalizedDomain = sortNumbers(domain);
+  const [domainMin, domainMax] = normalizedDomain;
+  const geometry = getMeterGeometry(normalizedDomain, value, baseline);
+  const baselinePosition = clamp(geometry.rawBaselinePosition, 0, 100);
```
biamalveiro: nit: wondering about this `clamp`, since `baselinePosition` is only used when the baseline is in the domain, is it redundant? as places that use `baselinePosition` could use `geometry.rawBaselinePosition` directly and avoid the tiny indirection.

packages/charts/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,418 @@
...
+    ? {
+        bottom: `${geometry.fillStart}%`,
+        top: `${100 - geometry.fillEnd}%`,
+      }
+    : {
+        left: `${geometry.fillStart}%`,
+        right: `${100 - geometry.fillEnd}%`,
+      };
+
+  const fillPaintStyle: CSSProperties =
```
biamalveiro: nit: I found the fill paint a bit hard to scan with the intermediate values and ternaries above + there are a few palette-only constants that run even for `Single` path.  I was thinking  something like a single early-return block per fill type could work, either wrapping this logic in a `useMemo` or in a separate helper, wdyt?

packages/charts/src/components/meter/meter.tsx
```diff
@@ -0,0 +1,418 @@
...
+    roundEnd: roundFillEnd,
+  });
+  const trackBorderRadius = getTrackBorderRadius({
+    isVertical,
+    roundStart: roundTrackStart,
+    roundEnd: roundTrackEnd,
+  });
+
+  if (geometry.isBaselineInDomain && geometry.fillSize > 0 && !roundFillStart) {
+    baselineMarkerNudgePx = isFillForward ? BASELINE_MARKER_ADJUSTMENT : -BASELINE_MARKER_ADJUSTMENT;
```
biamalveiro, line 224: just for me to understand this a bit better, why do we need this slight nudge in the non rounded start case?

packages/charts/src/components/meter/_meter.scss
```diff
@@ -0,0 +1,125 @@
+@import './variables';
+
+$meter-border-radius: 8px;
```
biamalveiro, line 3: Was wondering if this style survives? It seems that typescript ends up handling border radius in `meter.tsx`, maybe we could avoid duplication of the 8px constant? But I could be missing something here

packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx
```diff
@@ -234,6 +247,7 @@ function getHeightBasedFontSizes(
   ranges: [number, number, BreakPoint][],
   value: number,
   style: MetricStyle,
+  progressBarSizeOverride?: MetricProgressBarSize,
```
biamalveiro, line 250: super-nit/just an idea: we could generalise this overrides interface, something like `overrides: Partial<HeightBasedSizes>`, so if in the future something comes up, it's easier to add.  (+ maybe renaming this function to `getHeightBasedSizes` since it's no longer just about font sizes.

e2e/screenshots/meter_stories.test.ts-snapshots/meter/dark-theme/should-render-grouped-signed-ranges-chrome-linux.png
biamalveiro, line 1: (not sure if it was an intentional change, but just to follow the same naming pattern of the other screenshots, maybe these could be renamed from `Meter-stories-theme---dark-should-render-grouped-signed-ranges-1-chrome-linux.png‎` to something like `should-render-grouped-signed-ranges-1-dark-theme-chrome-linux.png‎`)

packages/charts/src/chart_types/metric/renderer/dom/text.tsx
```diff
@@ -78,7 +78,7 @@ interface MetricTextprops {
   datum: MetricDatum;
   style: MetricStyle;
   onElementClick?: () => void;
-  progressBarSize: ProgressBarSize;
+  progressBarSize: MeterSize;
```
biamalveiro, line 81: (out of scope) This is pre-existing to this PR, so nothing to be changed here, just food for thought. But was wondering if we still need `progressBarSize` as a separate prop here?

It seems that `textDimensions` already encodes progress layout, `heightBasedSizes.progressBarThickness` (px) (although I didn't find that to be super intuitive, that progress layout things are part of `MetricTextDimensions`). That way, `<MetricText>` could maybe derive size via `getMetricProgressBarSize` and avoid the extra prop?

packages/charts/src/components/meter/utils.ts
```diff
@@ -0,0 +1,168 @@
...
+  return getScalePositionFn(domain)(value);
+}
+
+/** @internal */
+export function getMeterGeometry(domain: ContinuousDomain, value: number, baseline = 0): MeterGeometry {
+  const getScalePosition = getScalePositionFn(domain);
+  const rawBaselinePosition = getScalePosition(baseline);
+  const rawValuePosition = getScalePosition(value);
+  const [fillStart, fillEnd] = sortNumbers([clamp(rawBaselinePosition, 0, 100), clamp(rawValuePosition, 0, 100)]);
+  const [domainMin, domainMax] = sortNumbers(domain);
```
biamalveiro, line 60: nit: I think `<Meter>` sorts the domain before calling this, and `getScalePositionFn` also sorts it again. Nothing against the extra safety, but was wondering if it could make sense to normalise once and let the downstream utils/steps make the assumption that the domain is correctly ordered?

biamalveiro (approved):
Nice work! 🚀  Played around with the stories, and the new Meter component seems to be working great. 

Left a few comments, but nothing major or blocking.

packages/charts/src/components/meter/meter.tsx
awahab07: Good catch. In the visible baseline-marker path the baseline is already in-domain, so the extra clamp was redundant. I simplified that.

fe809dc1e72f709dc73cf1a85fe13447c4396f07

packages/charts/src/components/meter/meter.tsx
awahab07: Agreed. I refactored the fill setup into small helpers so the single, solid-palette, and gradient-palette paths are easier to scan. I kept it out of useMemo since this is a readability improvement rather than a performance hot path.

64e02995567207e238523ddea337d4025b236e40

packages/charts/src/components/meter/meter.tsx
awahab07: It’s intentional: when the fill start is not rounded, the fill edge is squared off, and the slight nudge keeps the baseline marker visually centered against that edge. I’ve added a short comment (1b30bfef0ca896db417f1cff3c67eb1e939cf50e) to make that clearer.

image

packages/charts/src/components/meter/_meter.scss
awahab07: It's more of honoring the [prior art](https://www.elastic.co/docs/extend/kibana/pr-review#_consistency_style_readability) as it's the existing pattern (example [.tsx constant](https://github.com/elastic/elastic-charts/blob/fa5784367d2ed6bbe3a3cba5505f5deca9b96937/packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx#L82) ↔ [.scss var](https://github.com/elastic/elastic-charts/blob/7c444bb66c86dee72c7815ae069e310fee1d0e01/packages/charts/src/chart_types/metric/renderer/_variables.scss#L15)).

But I agree, ideally we should be using [CSS custom props](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascading_variables/Using_custom_properties).

packages/charts/src/chart_types/metric/renderer/dom/text_measurements.tsx
awahab07: Good point. I renamed (fa5784367d2ed6bbe3a3cba5505f5deca9b96937) `getHeightBasedFontSizes()` to `getHeightBasedSizes()`. I am keeping the override narrow for now, though, because only progress bar size is an intentional override today. And future cases should ask first whether loading the a generic override prop, or an explicit prop makes sense.

e2e/screenshots/meter_stories.test.ts-snapshots/meter/dark-theme/should-render-grouped-signed-ranges-chrome-linux.png
awahab07: Agreed. I normalized the Meter snapshot naming to match the Metric-style structure instead of relying on the raw Playwright default titles. They now live under meter/<theme>-theme/... with explicit snapshot names.

25a70b2d4a4c00c2a891f0b257ff2f930d528732

packages/charts/src/chart_types/metric/renderer/dom/text.tsx
awahab07: Reasonable thought, but I think the explicit prop is clearer here because the parent already owns the semantic size resolution. Re-deriving it from thickness would hide that decision behind a reverse mapping.

packages/charts/src/components/meter/utils.ts
awahab07: I see the duplication. I’ve leaned toward keeping the utils defensive because the extra cost is negligible and domain direction has already been a subtle source of regressions.

Addressing those aforementioned regressions bottom up calls for refactoring the entire Meter utility contract, which would make the PR bloat.

packages/charts/src/components/meter/meter.tsx
biamalveiro: makes sense, thanks!

packages/charts/src/components/meter/_meter.scss
biamalveiro: ah, yup, that's a fair point 👍

packages/charts/src/components/meter/meter.tsx
biamalveiro: nice!

packages/charts/src/components/meter/_meter.scss
biamalveiro: I see, fair enough!

packages/charts/src/chart_types/metric/renderer/dom/text.tsx
biamalveiro: Yeah, that's fair 👍

packages/charts/src/components/meter/utils.ts
biamalveiro: Alright, makes sense if there's that historical context. And agreed, would lead to changes that are definitely beyond the scope of this PR.

biamalveiro (approved):

awahab07:
buildkite update vrt

PR #2671, opened 2025-06-19
fix(flame_chart): Fix font size to occupy available space

rockdaboot:
This fixes the font size of text in the flame_chart boxes to occupy the the available space.

This often has been a major hassle when demoing the flamegraph of Universal Profiling on smaller screens, e.g. on a laptop during a conference.
The same issue happens when demoing via zoom on a large screen. The font size just doesn't scale with the browser scaling.

The screenshots show the effect.

**Before**
Screenshot_20250619_153220

**After**
Screenshot_20250619_153141

Issue #2001, opened 2023-03-17
Flamechart click-on-node resets vertical scaling, when user workflow requires only horizontal to be reset


thomasdullien:
**Describe the issue**
When a user zooms into a node on the graph, it is a common workflow to click on a node to make it the "center of attention", e.g. zoom/pan/horizontal-stretch-squash it to be the top of the screen and scale the underlying nodes accordingly.

In the current flamegraph, this does not only perform the zoom/pan/horizontal-stretch-squatch, but also resets *vertical* scaling, leading to the result becoming unreadable, and making it difficult for the user to obtain the information they seek.

**To Reproduce**
1. Load a flamegraph
2. Zoom in a bit to change the vertical size of nodes (CTRL-Wheel)
3. Click on a node, expecting it to become the center of attention but keep the vertical scaling constant.
4. Node is shifted to be the center of attention, but vertical scaling is reset to default.

**Expected behaviour**
Vertical scaling should be constant for this action.

**Screenshots**

https://user-images.githubusercontent.com/4587964/225979962-ba4535b7-9b5e-41a7-9b16-cc6b50aee542.mp4

**Version (please complete the following information):**
 - OS: Debian Linux
 - Browser: Chrome
 - Elastic Charts: current?

PR #2849, opened 2026-06-14
fix: Scale flamechart font with CTRL-Wheel zoom

### LINKED ISSUE #2000: Flamechart does not scale font when CTRL-Wheel zooming in
When using CTRL-Wheel to zoom into a flamechart, the font is not scaled along with the nodes. On many screens, this results in an unreadable flamechart as the font is too small.

**To Reproduce**
1. Load a flamechart.
2. Use CTRL-wheel to zoom in.

**Expected behaviour**
The font size is scaled proportionally with vertical scaling.

**Screenshots**
Screenshot from 2023-03-17 18-36-45

Screenshot from 2023-03-17 18-37-04

 - OS: Linux
 - Browser: Chrome
 - Elastic Charts: (current? ES stack 8.7)

Malaybhai11:
## Summary

Fixes #2000: When using CTRL-Wheel to zoom into a flamechart, the font was not scaled along with the nodes. This resulted in unreadable text on many screens as the nodes became wider but the font stayed the same.

## Changes

Modified `packages/charts/src/chart_types/flame_chart/render/draw_canvas.ts` to scale font size based on both horizontal and vertical zoom levels.

**Before:** Font size was calculated only based on vertical zoom level (`zoomedRowHeight * cssHeight - 2 * BOX_GAP_VERTICAL`).

**After:** Font size is calculated based on the maximum of both horizontal and vertical zoom factors, ensuring readable text when zooming in any direction.

## Technical Details

The issue occurred because:
1. When using CTRL-Wheel zoom, both X and Y axes zoom proportionally initially
2. After a threshold, Y zoom stops (minimum row pitch reached) but X zoom continues
3. Nodes become wider but font stays the same size

The fix computes zoom factors for both axes:
```typescript
const horizontalZoomFactor = 1 / Math.abs(focusHiX - focusLoX);
const verticalZoomFactor = 1 / Math.abs(focusHiY - focusLoY);
const zoomFactor = Math.max(horizontalZoomFactor, verticalZoomFactor);
const fontSize = baseFontSize * zoomFactor;
```

This ensures font size scales with the dominant zoom direction.

## Testing

- Verified the fix works correctly by examining the code logic
- No unit tests were modified as the existing tests are screenshot-based e2e tests
- The change is backward compatible and only affects behavior when zoomed in

### DIFF
--- a/packages/charts/src/chart_types/flame_chart/render/draw_canvas.ts
+++ b/packages/charts/src/chart_types/flame_chart/render/draw_canvas.ts
@@ -38,7 +38,13 @@ export const drawCanvas2d = (
 ) => {
   const zoomedRowHeight = rowHeight / Math.abs(focusHiY - focusLoY);
   const rowHeightPx = zoomedRowHeight * cssHeight;
-  const fontSize = zoomedRowHeight * cssHeight - 2 * BOX_GAP_VERTICAL;
+  // Scale font size with both vertical and horizontal zoom to ensure readability
+  // when using CTRL-Wheel to zoom (which may zoom X more than Y after Y hits minimum)
+  const horizontalZoomFactor = 1 / Math.abs(focusHiX - focusLoX);
+  const verticalZoomFactor = 1 / Math.abs(focusHiY - focusLoY);
+  const zoomFactor = Math.max(horizontalZoomFactor, verticalZoomFactor);
+  const baseFontSize = rowHeight * cssHeight - 2 * BOX_GAP_VERTICAL;
+  const fontSize = baseFontSize * zoomFactor;
   const minTextLengthCssPix = MIN_TEXT_LENGTH * fontSize; // don't render shorter text than this
   const minRectWidthForTextInCssPix = minTextLengthCssPix + TEXT_PAD_LEFT + TEXT_PAD_RIGHT;
   const minRectWidth = minRectWidthForTextInCssPix / cssWidth;


markov00:
buildkite test this

markov00:
hey @Malaybhai11 welcome and thanks for the fix, I will going to review it shortly.
In order to continue and merge the PR, you need to signing the CLA agreement as posted in this PR a couple of messages above.

Malaybhai11:
I have signed the CLA agreement. Ready for the build and review!

markov00 (changes requested):
I've reviewing it locally but it doesn't seems to fix the problem.
I've also checked main, but it seems already resolved by this PR https://github.com/elastic/elastic-charts/pull/2671.
The issue was still open for no actual reason.
Do you see anything wrong in the current main behaviour that should still be fixed and you where trying to fix in your PR?

This is the record from your PR, the CTRL + wheel seems to do some scaling correctly, but zooming in with just the wheel start scaling the font size in the wrong way:  the wheel-in/out will just horizontally enlarges the chart without any vertical zoom as it should work, in this way the fontsize should remain the same.

See the recording of your PR:


https://github.com/user-attachments/assets/ecfab46d-2431-4d0d-8559-33df0c8bd579

This is instead on main (already correct behaviour)


https://github.com/user-attachments/assets/48bae415-5217-449e-8be8-acbc0869e43d

Malaybhai11:
Thanks for the detailed review @markov00! You are right — the issue was already resolved by #2671 and my PR is not adding any value here, in fact it introduces a regression with the wheel-only zoom behavior. I will close this PR. Apologies for the noise!

Malaybhai11:
Makes sense, thanks for clarifying! Since #2671 already took care of it I will move on. I spotted #2001 (click-on-node resets vertical scaling) which seems like a real pain point in the same area — gonna take a proper look at that one instead. Appreciate the review either way!

PR #2856, opened 2026-07-09
fix: reduce selector cache memory leaks and tooltip DOM cleanup

markov00:
## Summary

Fixes memory leaks in the selector caching layer and tooltip portal, and improves the cache implementation.

### Selector cache leak: orphaned caches from array pruning

`GlobalSelectorCache.removeKeyFromAll` was filtering out empty `CustomMapCache` instances from its `selectorCaches` array after removing a chartId. This caused a leak because each cache is shared by closure with a live `re-reselect` selector created at module load time. Pruning the cache from the tracking array orphans it: `re-reselect` still holds a reference and writes new chartId entries into it, but `removeKeyFromAll` can no longer reach it to clean them up. Over repeated mount/unmount cycles, memoized selectors (retaining geometry arrays, scales, tooltip info, etc.) accumulate without ever being freed.

**Fix:** Stop pruning empty caches from the array. The array grows to ~170 entries at startup (one per selector definition) and stays that size permanently.

### Selector cache sizing: `maxSize` reduced from 100 to 2

Each of the ~170 selectors maintained an LRU cache of up to 100 memoized results per chartId. Benchmarking showed that `maxSize=2` captures all hit-rate gains — sizes 2, 10, and 100 produce identical results. With large datasets and multiple charts on a page, `maxSize=100` caused significant memory retention with no performance benefit.

### Cache implementation: plain object replaced with `Map`

`CustomMapCache` used a plain object (`{}`) as its backing store. This has two downsides: `Object.keys().length` for `isEmpty()` is O(n), and `delete` on plain objects can deoptimize V8 hidden classes. Replaced with `Map` which provides O(1) `.size`, no hidden class issues, and no risk of prototype-key collisions.

### Tooltip portal leak: orphaned anchor DOM nodes

`TooltipPortalComponent` creates anchor DOM nodes dynamically but only cleaned up the portal node on unmount, leaving `echAnchor*` elements in the DOM. Added cleanup for anchor nodes that were created by the component.

### Code cleanup

- Renamed `getNewOptions()` → `registerSelector()` to reflect the side-effecting nature of the method (creates and tracks a new cache).
- Inlined `getCacheObject()` into `registerSelector()` — removes unnecessary indirection for a one-call-site private method.
- Extracted `keySelector` from a static class method to a module-level arrow function.
- Renamed `createSelectorWithMaxSize100` → `createSelectorWithLRU`.

### 2 Selector cache tests

1. Verifies that selector cache objects stay registered after their last chart id is removed. This ensures future chart ids written into the same live selector cache can still be cleaned up by `globalSelectorCache.removeKeyFromAll`, avoiding orphaned caches and memory leaks.

2. Verifies that each chart id keeps only two memoized input combinations. After three different inputs are evaluated for the same chart id, the oldest one is evicted, so requesting it again recomputes the selector result.

### DIFF
--- a/packages/charts/src/components/portal/tooltip_portal.tsx
+++ b/packages/charts/src/components/portal/tooltip_portal.tsx
@@ -179,11 +179,17 @@ const TooltipPortalComponent = ({
 
   useEffect(() => {
     setPopper();
-    const nodeCopy = portalNode;
+    const portalCopy = portalNode;
+    const anchorCopy = anchorNode;
+    const anchorWasCreatedByUs = !isHTMLElement((anchor as PortalAnchorRef).current);
 
     return () => {
-      if (nodeCopy.parentNode) {
-        nodeCopy.parentNode.removeChild(nodeCopy);
+      if (portalCopy.parentNode) {
+        portalCopy.parentNode.removeChild(portalCopy);
+      }
+
+      if (anchorWasCreatedByUs && anchorCopy.parentNode) {
+        anchorCopy.parentNode.removeChild(anchorCopy);
       }
 
       destroyPopper();
--- a/packages/charts/src/state/create_selector.test.ts
+++ b/packages/charts/src/state/create_selector.test.ts
@@ -1,0 +1,60 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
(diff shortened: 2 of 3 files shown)


walterra (approved):
All Code LGTM. Only thing I can think of we could use this opportunity to introduce a little unit test coverage for `create_selector.ts` that covers the change in the pruning behavior in `removeKeyFromAll`.

PR #2857, opened 2026-07-10
fix: reduce selector cache memory leaks and tooltip DOM cleanup (#2856) [70.1.x]

markov00:
Backports the following commits to 70.1.x:
 - fix: reduce selector cache memory leaks and tooltip DOM cleanup (#2856)


markov00:
buildkite test this

nickofthyme:
Buildkite broken, still uses old `.buildkite/` files. Tested types, jest and vrt locally on branch.

PR #2864, opened 2026-07-16
fix(metric): increase metric `minHeight` in theme to avoid clipped metric tiles 

biamalveiro:
## Summary

Wide metric tiles in grids no longer shrink height to a point where value is clipped. 

**Before:**

https://github.com/user-attachments/assets/704cf81e-d546-4bab-b6b2-8a18549ab0a4

**After:** 

https://github.com/user-attachments/assets/95f4db3b-ad4a-4162-abef-9689231d9907

## Details

Increase theme's `minHeight` from `64px` to `80px`. 

Given the padding and font sizes for title and value at the smallest breakpoint, we calculate the minimum height needed to render these elements. It currently is 71 (with small spacing) and 79 (with large spacing). Therefore increased `minHeight` to `80px`(smallest value on the 8-point grid system >79).  Added a test to check that themes' `minHeight` are greater or equal to the measured min, to avoid future regressions if we edit breakpoints, paddings or font sizes.

This increase makes compact metrics have a tiny bit more white space in these circumstances. Panels can have many different configurations (title, subtitle, value, spacing ...) which are conditionally rendered, this makes it difficult to expose a static `minHeight` value.

| Before  | After |
| ------------- | ------------- |
|  Screenshot 2026-07-16 at 15 03 06  | Screenshot 2026-07-16 at 15 03 16 |

I also considered exposing a `getMinHeight` function, so that consumers could get actual minHeight for a given Metric style and use that instead. However, it didn't seem like a common pattern in the codebase, so opted for this simpler and static approach, but happy to discuss!

### To test
To test the behavior go to the [metrics/grid story](https://ech-e2e-ci--pr-2864-sebcw6e3.web.app/storybook/?path=/story/metric-alpha--grid&globals=toggles.showHeader:true;toggles.showChartTitle:false;toggles.showChartDescription:false;toggles.showChartBoundary:false;theme:light), edit number of columns from 4 to 2 (so you can can wide metric tiles) and resize height of chart. Check that value is no longer cropped.

## Issues

Fixes https://github.com/elastic/kibana/issues/268123

### Checklist

### DIFF
--- a/packages/charts/src/utils/themes/light_theme.ts
+++ b/packages/charts/src/utils/themes/light_theme.ts
@@ -464,7 +464,7 @@ export const LIGHT_THEME: Theme = {
     barBackground: LIGHT_BACKGROUND_COLORS.backgroundBaseDisabled,
     emptyBackground: Colors.Transparent.keyword,
     nonFiniteText: 'N/A',
-    minHeight: 64,
+    minHeight: 80,
     titleWeight: 500,
   },
   bulletGraph: LIGHT_THEME_BULLET_STYLE,
--- a/packages/charts/src/chart_types/metric/renderer/dom/text_measurements.test.ts
+++ b/packages/charts/src/chart_types/metric/renderer/dom/text_measurements.test.ts
@@ -1,0 +1,40 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+import { getMinimumMetricHeight } from './text_measurements';
+import { AMSTERDAM_DARK_THEME } from '../../../../utils/themes/amsterdam_dark_theme';
+import { AMSTERDAM_LIGHT_THEME } from '../../../../utils/themes/amsterdam_light_theme';
+import { DARK_THEME } from '../../../../utils/themes/dark_theme';
+import { LEGACY_DARK_THEME } from '../../../../utils/themes/legacy_dark_theme';
+import { LEGACY_LIGHT_THEME } from '../../../../utils/themes/legacy_light_theme';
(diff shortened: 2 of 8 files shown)


packages/charts/src/utils/themes/light_theme.ts
```diff
@@ -464,7 +464,7 @@ export const LIGHT_THEME: Theme = {
     barBackground: LIGHT_BACKGROUND_COLORS.backgroundBaseDisabled,
     emptyBackground: Colors.Transparent.keyword,
     nonFiniteText: 'N/A',
-    minHeight: 64,
+    minHeight: 80,
```
markov00, line 467: Thanks it works great here on storybook, but I'm wondering if this is the right calculation to use also for Kibana.
Checking the code, the grid height is 20 but it also need to consider the gutter size into this calculation.
The current Panel pixel height is  `pixelHeight = h * rowHeight + (h - 1) * gutterSize`
where `rowHeight` is 20 
gutterSize is `8`
and `h` is the number of grid points you are considering.
So with 3 vertical grid points we have a height of 76px, with a grid of 4 points we have 104px.
So not sure if we can have this minHeight at 76 for small spacing and 104 for large spacing?

biamalveiro: Here I made the limit be based on content (typography + layout), and have the test assert that the value exported in the theme is accurate. But yeah, I totally see the point of making the value also be appropriate for Kibana's grid, but I'm thinking that that logic should live over on that side and not in elastic-charts, possibly [here in metric_vis.tsx](https://github.com/biamalveiro/kibana/commit/fb824d91ff59a9404fcfcd936778e9bd20f7fcef)? I don't love this implementation in the commit necessarily, because I'm still hardcoding the values ([were these the constants you were referring to for height and gutter?](https://github.com/elastic/kibana/blob/main/src/platform/plugins/shared/dashboard/public/dashboard_renderer/grid/constants.ts)), but, it would be Kibana's specific logic in Kibana. 

Wdyt? Do you have any preference or recommendation?

markov00: You are right, let's keep this values in the charts for now, in Kibana we already have some customization to the theme used by charts and we can update this property to follow what is best for Kibana Grid

markov00 (approved):

PR #2866, opened 2026-07-22
fix(axis): clip tick labels that overflow chart

biamalveiro:
## Summary

Fixes case where if truncation and wrapping are disabled and a very large tick label would overflow in cases where it was larger that the axis `maxExtent`.

| Before | After | 
| -- | -- | 
|  Screenshot 2026-07-22 at 12 10 44  |  Screenshot 2026-07-29 at 10 32 43

## Details

We add a safety clip rectangle around the rendered tick labels, in order to ensure that they do not overflow the axis cross dimension.

### Checklist


biamalveiro:
buildkite update vrt

biamalveiro:
buildkite update vrt

biamalveiro:
buildkite update vrt

biamalveiro:
buildkite test this

e2e/screenshots/axis_stories.test.ts-snapshots/axis-stories/overflow-tick-labels/should-hide-y-axis-tick-labels-overflowing-a-percentage-max-extent-with-90-deg-chart-rotation-chrome-linux.png
markov00, line 1: Instead of hiding the label, do you think we can consider cutting the label anyway to the max extent?
no truncation sign, but just cut it out to the max size? like an overflow:hidden
Without the label it looks more like an error to me then a too long label

biamalveiro: Thanks for having a look and for the suggestion! I'm not sure I agree with the clipping though. I was thinking through the ways a user can configure an axis that has large labels:

- if ok with partial labels, enable truncation and we truncate with an ellipsis to maximise the used space.
- if wrapping enabled, it's possible to increase the wrapped lines to show more of the label and avoid possible truncation
- if no truncation or wrapping (this case), I'd assume that the user doesn't want a partial/ambiguous label. clipping without an ellipsis would still show partial labels, in fact, it's would just create a truncated label without any notation that it's truncated, which I think it's a bit worse than just truncation with ellipsis, and could look like an error?

So the options I see are show the full label if it fits, or hide it.

(there's also a possible issue with clipping when wrapLines > 1, a bounding-box clip can cut off partial content from multiple lines rather than cleanly truncating the end of the text, since the overflow isn't linear the way it is for a single line)

maryam-saeidi: Personal opinion: 
I was also very surprised to see that some of the labels weren't shown due to limited space, and I thought it was a bug. I am more in favour of the option Marco suggested: to cut off the label when it is too long instead of hiding it, as hiding it looks like a bug.

biamalveiro: Fair enough, definitely want to avoid that feeling that the chart is broken, so pushed a commit that does the clipping instead. Tried it out and looks better than I expected, actually:

Here are a few screenshots for comparison:

Clipping
clipping

Truncation 
truncate

Wrapping w/o truncation, forcing clipping: 
clipping and wrapping

There are some VRT failures that I'm having a look, but lmk what you tink of the behavior and if you see any adjusments needed.

biamalveiro:
buildkite test this

e2e/screenshots/axis_stories.test.ts-snapshots/axis-stories/overflow-tick-labels/should-hide-y-axis-tick-labels-overflowing-a-percentage-max-extent-with-90-deg-chart-rotation-chrome-linux.png
markov00: Perfect, in our case (Lens) truncation should be probably always on so clipping never happen

biamalveiro:
buildkite update vrt

biamalveiro:
buildkite update vrt

biamalveiro:
buildkite test this

e2e/screenshots/axis_stories.test.ts-snapshots/axis-stories/overflow-tick-labels/should-hide-y-axis-tick-labels-overflowing-a-percentage-max-extent-with-90-deg-chart-rotation-chrome-linux.png
markov00, line 1: is this screenshot required? I think we updated that test and fixed it, am I wrong?

biamalveiro: Not at all, the code and test was removed, the image was leftover. deleted it

markov00 (approved):

PR #2868, opened 2026-07-29
fix(axis): no longer reserve space for hidden tick labels

biamalveiro:
## Summary

Fixes issue where space was still being reserved for hidden tick labels. 

| Visible | Before | After | 
| -- | -- | -- |
| Screenshot 2026-07-29 at 11 33 34 |  Screenshot 2026-07-29 at 11 34 44 |  Screenshot 2026-07-29 at 11 33 40 |

## Details

Hidden tick labels were still being accounted for during chart layout calculation, which caused empty space to exist in these charts. We now empty the label layout for those ticks with hidden labels to ensure that they do not influence layout. These ticks be just removed because we still want to show grid lines for them.

### Checklist


biamalveiro:
buildkite update vrt

markov00 (commented):
I believe original issue was related to a specific situation where the labels are visible, but hidden due to the "hide overlaps" logic, and overflow labels will take space even when they are hidden due to the overlaps.
this is visible in the overflow story if you reverse the order of the categories:
```
 Categories: [
    'this is the longest category name in this story',
    'this is an even longer category name',
    'this is a longer category name',
    'another category',
    'category name',
    'category',
    'category',
    'ctg',
  ].reverse(),
```

https://github.com/user-attachments/assets/16a1cf1f-6399-4b87-b97a-24e8bdc84f05

notice that as a specific width the last label is hidden, but the chart keep accounting for that label size for its right margin

biamalveiro:
@markov00 You're right, that issue is still present. 

I had a look at it when investigating the overflows over at https://github.com/elastic/elastic-charts/pull/2866 , and tried out a solution https://github.com/elastic/elastic-charts/pull/2866/changes/c98cf4cf851c634e470e665ce19dd0510a9014fc: currently we use first/last ticks and their alignments to calculate margins (<v72 used largest tick label). But, in fact, we should consider the last visible tick for the end margin. However, this last visible tick might be positioned more towards the middle of the axis, so to correctly calculate overflow, ticks need to carry their positions for `getAxesDimensions` (bootstrap layout step uses first/last). Although more correct, it ended up quite complex so I removed it from the scope of that PR that dealt with overflows in the cross axis direction, and planned to open a separate issue for it.  

Because we'll have truncation enabled in Kibana for XY charts with categorical X scales, I do not think the issue will be very visible there, and will improve current state by considering first/last. This PR here fixes a more pressing issue which would be a regression if we land v72 before this fix, that is if ticks labels have are configured `visibile: false`, which users can set in Kibana, then the layout should not reserve space for these labels.

markov00 (approved):
Agree with the decision and scope reduction of the PR!

PR #2870, opened 2026-08-11
fix(bullet): add adaptive bar stroke when color band contrast is low

### LINKED ISSUE #2376: [Bullet] current bar style with higher contrast
Depending on the selected colors, the `value` bar can have a very low contrast.
Here for example, even if the colors are not nice colors, the contrast it pretty low.
I believe we should do something here to improve the visibility of that bar. 
We also should remember that the bar can cover both dark and light colors.

Possible suggestions: 
- adding a border?
- adding a shadow?
- moving the bar to one side to make it stand out of the bullet?

image

biamalveiro:
## Summary

Bullet graph value bars now render with an adaptive outline when the bar color does not have enough contrast against the background color bands. This improves readability in dark mode and with low-contrast palettes, so the value bar stays visible.

| Theme | Before | After | 
| -- | -- |  -- | 
| Light | Screenshot 2026-08-12 at 08 21 51 | Screenshot 2026-08-12 at 08 22 21 |
| Dark | Screenshot 2026-08-12 at 08 23 11 | Screenshot 2026-08-12 at 08 22 50 | 

When contrast is enough, no stroke is rendered: 

| Light | Dark | 
| -- | -- | 
|  Screenshot 2026-08-11 at 20 48 05 | Screenshot 2026-08-11 at 20 49 03 | 

## Details
When the APCA contrast between the bar fill and every color band is below the minimum threshold (20, see [this discussion on non-text semantic elements](https://github.com/Myndex/SAPC-APCA/discussions/117#discussioncomment-9244796) and [Use-Case & Size Ranges in APCA](https://git.apcacontrast.com/documentation/APCA_in_a_Nutshell.html)), the bar is rendered with a 1 px stroke using the chart background color. Otherwise, the no stroke is rendered.

The bullet canvas renderer was also refactored from a class component to a functional React component.

VRT coverage was added for linear and angular bullets across light/dark themes and different palettes.

## Issues

Fixes https://github.com/elastic/elastic-charts/issues/2376
Related to https://github.com/elastic/kibana/issues/262702

### Checklist


biamalveiro:
buildkite update vrt

biamalveiro:
buildkite test this

storybook/stories/bullet_graph/3_color_bands.story.tsx
```diff
@@ -19,20 +19,21 @@ import {
...
 import type { ChartsStory } from '../../types';
-import { useBaseTheme, useThemeId } from '../../use_base_theme';
 import { customKnobs } from '../utils/knobs';
 import { getKnobFromEnum } from '../utils/knobs/utils';
 
 export const Example: ChartsStory = (_, { title, description }) => {
-  const isDarkTheme = useThemeId().includes('dark');
+  const themeMode = select('theme', { Light: 'light', Dark: 'dark' }, 'light');
+  const baseTheme = themeMode === 'dark' ? DARK_THEME : LIGHT_THEME;
+  const isDarkTheme = themeMode === 'dark';
```
nickofthyme: The `useThemeId` taps into the global theme switcher, this change will ignore the global theme and background color which is connected via `useBaseTheme`.

Image

Is there a reason to not use the global theme here?

packages/charts/src/chart_types/bullet_graph/utils/color.ts
```diff
@@ -363,3 +364,11 @@ function getColorBands(
...
   return scaledColorBands;
 }
+
+/** @internal */
+export function checkBarContrast(bands: ColorTick[], bar: Color): boolean {
+  // as per recommendation in https://github.com/Myndex/SAPC-APCA/discussions/117#discussioncomment-9244796 for
+  // semantic non-text elements and considering BAR_SIZE is 12,
+  const contrastThreshold = 20;
+  return bands.every((band) => Math.abs(APCAContrast(colorToRgba(band.color), colorToRgba(bar))) >= contrastThreshold);
+}
```
nickofthyme: Could we use `getContrastRecommendation` here? Seems like similar logic.


```suggestion
export function checkBarContrast(bands: ColorTick[], bar: Color): boolean {
  // as per recommendation in https://github.com/Myndex/SAPC-APCA/discussions/117#discussioncomment-9244796 for
  // semantic non-text elements and considering BAR_SIZE is 12,
  const contrastThreshold = 20;
  return bands.every(
    (band) => !getContrastRecommendation(band.color, bar, { contrastMode: 'WCAG3', contrastThreshold }).needsBorder,
  );
}
```

packages/charts/src/chart_types/bullet_graph/renderer/canvas/index.tsx
```diff
@@ -83,176 +82,267 @@ interface OwnProps {
...
+  backgroundColor: Color;
+  contrastOptions: ColorContrastOptions;
+}
 
-  constructor(props: Readonly<Props>) {
-    super(props);
-    this.ctx = null;
-    this.devicePixelRatio = window.devicePixelRatio;
-  }
+const BulletMetricContent = ({
```
nickofthyme: Love this cleanup! Could we move this to a separate file just for this component?

packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/horizontal.ts
```diff
@@ -54,12 +56,16 @@ export function horizontalBullet(
...
   const confinedValue = clamp(datum.value, min, max);
   const adjustedZero = clamp(0, min, max);
-  ctx.fillStyle = style.barBackground;
-  ctx.fillRect(
-    datum.value > 0 ? scale(adjustedZero) : scale(confinedValue),
-    verticalAlignment - BAR_SIZE / 2,
-    confinedValue > 0 ? scale(confinedValue) - scale(adjustedZero) : scale(adjustedZero) - scale(confinedValue),
-    BAR_SIZE,
+  const x0 = Number(scale(adjustedZero));
+  const x1 = Number(scale(confinedValue));
```
nickofthyme: What's the purpose for using `Number` here?


```suggestion
  const x0 = scale(adjustedZero);
  const x1 = scale(confinedValue);
```

packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/vertical.ts
```diff
@@ -59,13 +64,12 @@ export function verticalBullet(
...
-  ctx.fillRect(
-    graphArea.size.width / 2 - BAR_SIZE / 2,
-    confinedValue > 0 ? graphPaddedHeight - scale(confinedValue) : graphPaddedHeight - scale(adjustedZero),
-    BAR_SIZE,
-    confinedValue > 0 ? scale(confinedValue) - scale(adjustedZero) : scale(adjustedZero) - scale(confinedValue),
-  );
+
+  const x = graphArea.size.width / 2 - BAR_SIZE / 2;
+  const y0 = graphPaddedHeight - Number(scale(adjustedZero));
+  const y1 = graphPaddedHeight - Number(scale(confinedValue));
```
nickofthyme: Also here...


```suggestion
  const y0 = graphPaddedHeight - scale(adjustedZero);
  const y1 = graphPaddedHeight - scale(confinedValue);
```

packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/common.ts
```diff
@@ -0,0 +1,68 @@
...
+    const baselinePosition =
+      bar.start <= bar.end ? Math.min(bar.start, bar.end) : Math.max(bar.start, bar.end) - BAR_STROKE_WIDTH;
+    ctx.fillStyle = style.barBackground;
+
+    if (orientation === 'horizontal') {
+      ctx.fillRect(baselinePosition, y + BAR_STROKE_WIDTH, BAR_STROKE_WIDTH, BAR_SIZE - BAR_STROKE_WIDTH * 2);
+    } else {
+      ctx.fillRect(x + BAR_STROKE_WIDTH, baselinePosition, BAR_SIZE - BAR_STROKE_WIDTH * 2, BAR_STROKE_WIDTH);
+    }
+  }
```
nickofthyme: What do you think about reusing existing logic in `renderRectStroke`? It may also be cleaner to just use `renderRectStroke` inside of `horizontal.ts‎` and `vertical.ts`.

```suggestion
  ctx.fillStyle = style.barBackground;
  ctx.fillRect(x, y, width, height);

  if (barStroke) {
    // bar.start is always the baseline coordinate
    const baselineAtHighEnd = bar.start > bar.end;
    const strokedSides: StrokedSides = {
      top: orientation === 'vertical' ? baselineAtHighEnd : true,
      bottom: orientation === 'vertical' ? !baselineAtHighEnd : true,
      left: orientation === 'horizontal' ? baselineAtHighEnd : true,
      right: orientation === 'horizontal' ? !baselineAtHighEnd : true,
    };
    renderRectStroke(
      ctx,
      { x, y, width, height },
      { color: colorToRgba(barStroke), width: BAR_STROKE_WIDTH },
      strokedSides,
    );
  }
```

packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/angular.ts
```diff
@@ -82,18 +83,35 @@ export function angularBullet(
...
-    radius,
-    confinedValue > 0 ? scale(adjustedZero) : scale(confinedValue),
-    confinedValue > 0 ? scale(confinedValue) : scale(adjustedZero),
-    counterClockwise,
-  );
-  ctx.stroke();
+  const degStart = confinedValue > 0 ? scale(adjustedZero) : scale(confinedValue);
+  const degEnd = confinedValue > 0 ? scale(confinedValue) : scale(adjustedZero);
+  const arcDirection = counterClockwise ? -1 : 1;
+  const innerArcOffset = BAR_STROKE_WIDTH / radius;
```
nickofthyme: nit: you could scope these just to the `if (hasStroke) {` block.

nickofthyme (commented):
This looks great, just leaving a few comments from a first pass.

biamalveiro:
buildkite update vrt

packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/vertical.ts
biamalveiro: 9a2df9b0a039734667ddc0339b70a6af77a6b34a

packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/horizontal.ts
biamalveiro: 9a2df9b0a039734667ddc0339b70a6af77a6b34a

packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/common.ts
biamalveiro: Oh, I didn't know about it, saves a lot of work here thanks, used directly in `horizontal ` and `vertical`.
9a2df9b0a039734667ddc0339b70a6af77a6b34a

packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/angular.ts
biamalveiro: 9a2df9b0a039734667ddc0339b70a6af77a6b34a

packages/charts/src/chart_types/bullet_graph/utils/color.ts
biamalveiro: Yup, since APCA is also used by`getContrastRecommendation`, this was just a lighter version because we're just doing the check here and don't need the path for calculating a border color from the bar+band colors.

e2cebfc21ac09c0970f965c1bffa2b96f3570456

packages/charts/src/chart_types/bullet_graph/renderer/canvas/index.tsx
biamalveiro: 29029a815f9bc647a1b554bf209361fb5af9113f

biamalveiro:
buildkite update vrt

storybook/stories/bullet_graph/3_color_bands.story.tsx
biamalveiro: That wasn't great, you're right, updated to `useThemeId` ([bc471c5](https://github.com/elastic/elastic-charts/pull/2870/commits/bc471c5968612bd321b38abe280f5afe77abf631)). 
Had some issues with validating the VRT urls with the theme query parameter: Like https://elastic.github.io/elastic-charts/storybook/?path=/story/annotations-lines--x-continuous-domain&globals=toggles.showHeader:true;toggles.showChartTitle:false;toggles.showChartDescription:false;toggles.showChartBoundary:false;theme:dark goes to light, but works fine in the VRTs. This fixes that 9eaae5be14319a916411b6aea0397beb96cdd67a.

e2e/screenshots/bullet_stories.test.ts-snapshots/bullet-stories/subtype-circle/color-config-4/renders-color-bands-independent-of-ticks-flipped-domain-chrome-linux.png
nickofthyme, line 1: This looks like a regression from the original changes wrt to counterclockwise arcs. The start now has a border, while the end does not.

Image

e2e/screenshots/bullet_stories.test.ts-snapshots/bullet-stories/subtype-half-circle/color-config-2/should-render-colors-with-config-chrome-linux.png
nickofthyme, line 1: Similar here the inner value arc extends beyond the color bands at the start.

Image

storybook/stories/bullet_graph/3_color_bands.story.tsx
nickofthyme: Strange, I've never seen that before. Glad you found a solution.

I've had a lot of issues with storybook globals with our implementation [storybookjs/storybook#15632](https://redirect.github.com/storybookjs/storybook/issues/15632).

Eventually we won't have to worry.

Also the reason it works in the VRT is that we have a custom parser for the params, to immediate storybook format, but in the `e2e_server` which is a faster custom runner of the stories for VRT.

gvnmagni:
@andreadelrio might be a better fit for design review

biamalveiro:
buildkite update vrt

e2e/screenshots/bullet_stories.test.ts-snapshots/bullet-stories/subtype-circle/color-config-4/renders-color-bands-independent-of-ticks-flipped-domain-chrome-linux.png
biamalveiro: thanks for the catch, fixed! 430f38b2f08e5ecf97413db6890511777978e45e

e2e/screenshots/bullet_stories.test.ts-snapshots/bullet-stories/subtype-half-circle/color-config-2/should-render-colors-with-config-chrome-linux.png
biamalveiro: (also fixed by 430f38b2f08e5ecf97413db6890511777978e45e)

biamalveiro:
buildkite test this

packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/vertical.ts
```diff
@@ -59,13 +73,28 @@ export function verticalBullet(
...
+  const y1 = graphPaddedHeight - scale(confinedValue);
+
   ctx.fillStyle = style.barBackground;
-  ctx.fillRect(
-    graphArea.size.width / 2 - BAR_SIZE / 2,
-    confinedValue > 0 ? graphPaddedHeight - scale(confinedValue) : graphPaddedHeight - scale(adjustedZero),
-    BAR_SIZE,
-    confinedValue > 0 ? scale(confinedValue) - scale(adjustedZero) : scale(adjustedZero) - scale(confinedValue),
-  );
+  ctx.fillRect(x, y0, BAR_SIZE, y1 - y0);
```
nickofthyme, line 82: The previous implementation checked for `confinedValue` polarity, do you know if this is still needed? Or why you removed this here and not in `horizontal.ts`?

```ts
confinedValue > 0 ? scale(confinedValue) - scale(adjustedZero) : scale(adjustedZero) - scale(confinedValue)
```

packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/angular.ts
```diff
@@ -81,33 +80,33 @@ export function angularBullet(
   ctx.stroke();
 
   // Bar
-  const confinedValue = clamp(datum.value, min, max);
-  const adjustedZero = clamp(0, min, max);
-  const degStart = confinedValue > 0 ? scale(adjustedZero) : scale(confinedValue);
-  const degEnd = confinedValue > 0 ? scale(confinedValue) : scale(adjustedZero);
```
nickofthyme: Same question here, used to conditionally set `degStart` and `degEnd` based on `confinedValue` polarity.

packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/angular.ts
```diff
@@ -80,20 +80,37 @@ export function angularBullet(
...
-    center.x,
-    center.y,
-    radius,
-    confinedValue > 0 ? scale(adjustedZero) : scale(confinedValue),
-    confinedValue > 0 ? scale(confinedValue) : scale(adjustedZero),
-    counterClockwise,
-  );
-  ctx.stroke();
+  const value = scale(clamp(datum.value, min, max));
+  const zero = scale(clamp(0, min, max));
```
nickofthyme, line 84: Same question here, used to conditionally set `degStart` and `degEnd` based on `confinedValue` polarity.

packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/angular.ts
nickofthyme, line 1: I'm not sure what caused this but it looks like the inner arc is causing a rendering issue.

If you go to this [story](https://ech-e2e-ci--pr-2870-adavfj89.web.app/storybook/?path=/story/bullet-graph--color-bands&globals=toggles.showHeader:true;toggles.showChartTitle:false;toggles.showChartDescription:false;toggles.showChartBoundary:false;theme:light), set to 2/3 circle, everything looks normal. Set the value to negative such that it wraps back onto itself, like -25. It's hidden as it should be, but when you now set the band color to fail contrast, you see the arc circling.

https://github.com/user-attachments/assets/d520ddaf-f529-46e8-9c69-3d38ec7cebeb

nickofthyme (commented):
Looks great, changes are all good. Just a few last things to check.

packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/angular.ts
biamalveiro: It's because of the updated `counterClockwise` condition, it's now `value < zero`, directly on the scaled angles, instead of the previous one that was on the domain direction, so the flag was only correct if we adjusted degStart/degEnd.

This way, we can just use it as the `ctx.arc` counter clockwise flag + makes it direct that start is origin and end is value, making simpler to know on which side (and direction) to apply the offset.

packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/angular.ts
biamalveiro: It was missing a clamp in the inner bar value (`value - innerArcOffset * arcDirection`) that was being used directly, so it would go all the way around.

it's now this: `clamp(value - innerArcOffset * arcDirection, Math.min(zero, value), Math.max(zero, value))`

f05d363a9387ffecc4b7a3dd720e11fd5d4c8120

packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/vertical.ts
biamalveiro: It was a bit of a cleanup, `canvas.fillRect` is smart enough so that negative width/height values go the opposite way.

Did it for horizontal as well that I'd missed f05d363a9387ffecc4b7a3dd720e11fd5d4c8120

packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/angular.ts
nickofthyme, line 1: One tiny design nit: I think the clamping is causing a border remnant at the extremes. When the value is much less than the band min, such that it should not appear at all, it shows the sliver of the border at the start. Seeing this on only horizontal and vertical subtypes.

Image

Same happens on the end if the value exceed the band max. Seeing this on all subtypes.

Image

nickofthyme, line 1: This improvement makes me now question the contrast of target line. WDYT? I can open an issue for it.

Image

nickofthyme (approved):
Thanks for all the fixes, everything looks and works great now! 🎉

packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/angular.ts
biamalveiro: thanks! fixed, 864ef8e30b87d3e9a2a55e42a0011a4b8e79c89e

biamalveiro: Fair point, yes. I considered only showing it when it needs contrast with the specific color band it lies on ([here](https://github.com/elastic/elastic-charts/pull/2870/commits/ecc58988ff0623218dc5ac8846cee800862651b3), reverted) , but I think it looks more consistent showing it whenever the value bar has it. Here's how it looks like:

**Light**
| | | 
| -- | -- | 
| Screenshot 2026-08-24 at 08 37 40 | Screenshot 2026-08-24 at 08 40 25 | 

**Dark**
|  | | 
| -- | -- | 
| Screenshot 2026-08-24 at 08 41 10 | Screenshot 2026-08-24 at 08 41 37 | 

wdyt? 
In 029adc4328cb9810977ff3984eff1b165591ad4a, happy to revert and tackle in a separate issue if it needs more discussion.

nickofthyme: I think that looks fine. @markov00 WDYT?

markov00: very nice!

biamalveiro:
buildkite update vrt

biamalveiro:
buildkite test this

markov00 (approved):
Nice, works great and looks great!
Thanks also for refactoring to a function component the bullet chart

PR #2872, opened 2026-08-19
chore(deps): upgrade to Redux Toolkit v2 and react-redux v9

walterra:
## Summary

Part of https://github.com/elastic/elastic-charts/issues/2377.

`@elastic/charts` now runs on Redux Toolkit v2, react-redux v9, reselect v5, and immer v11 — matching the exact versions pinned by Kibana. This modernizes the internal state management stack with no changes to the public chart API.

### BREAKING CHANGE

React `^16.12` and `^17.0` are no longer supported as peer dependencies. `react` and `react-dom` peer ranges are now `^18.0`, required by react-redux v9 (which relies on React 18's native `useSyncExternalStore`).

## Details

Dependency changes (exact versions matching Kibana's pins):

| dep | before | after |
| --- | --- | --- |
| `@reduxjs/toolkit` | 1.9.7 | 2.12.0 |
| `redux` (direct dep) | ^4.2.1 | removed (RTK re-exports all APIs) |
| `react-redux` | ^8.1.3 | 9.2.0 |
| `immer` | ^9.0.21 | 11.1.8 |
| `reselect` (transitive) | 4.1.8 | 5.2.0 |
| `react`/`react-dom` peers | ^16.12 \|\| ^17.0 \|\| ^18.0 | ^18.0 |

Implementation notes:

- All direct `redux` imports (`Dispatch`, `Store`, `Unsubscribe`, `bindActionCreators`) replaced with `@reduxjs/toolkit` re-exports (40 files); the direct `redux` dependency was dropped.
- `src/state/create_selector.ts` now explicitly passes `memoize: lruMemoize` with `maxSize: 2`. Reselect 5 changed its default memoizer to `weakMapMemoize`, which has an unbounded reference-keyed cache; keeping the bounded LRU preserves the existing memory characteristics (~170 selectors × N charts retaining large geometry objects).
- `heatmap/state/selectors/on_brush_end_caller.ts`: replaced a reselect-v4-shaped `OutputSelector<...>` annotation with `Selector<GlobalChartState, void>`, matching the pattern used by all other `*_caller.ts` files.
- The codebase already used RTK 2-compatible patterns (builder-callback `extraReducers`, `configureStore` middleware callback, `createAction`), so no codemods were required.
- Public API surface unchanged (api-extractor: zero diff).

Validation:

- full jest suite: 107 suites / 1496 tests pass, 79 snapshots pass
- timezone suites (UTC, America/New_York, Asia/Tokyo): pass
- typecheck (src, nocomments, dist check): pass
- api-extractor: no public API changes

### Checklist


biamalveiro (approved):
looks good!

walterra:
Thanks for the review @biamalveiro! @markov00 @nickofthyme can you please chime in about dropping support for React 16/17 before I merge, are we good with that?

packages/charts/src/state/create_selector.ts
```diff
@@ -7,7 +7,7 @@
  */
 
 // eslint-disable-next-line no-restricted-imports
-import { createSelector } from '@reduxjs/toolkit';
+import { createSelector, lruMemoize } from '@reduxjs/toolkit';
```
markov00, line 9: do you mind fixing the unnecessary `// eslint-disable-next-line no-restricted-imports` above and right below?

markov00 (approved):
It is fine for me creating a breaking change and upgrading the required react to 18.
React 18 was release in 2022 so is safe to remove dependencies on older react versions.

Please remember to put the BREAKING CHANGE: comment line in the final squashed commit description

packages/charts/package.json
```diff
@@ -56,8 +55,8 @@
   "peerDependencies": {
     "moment": "^2.29.1",
     "moment-timezone": "^0.5.32",
-    "react": "^16.12 || ^17.0 || ^18.0",
-    "react-dom": "^16.12 || ^17.0 || ^18.0"
+    "react": "^18.0",
```
markov00, line 58: I've also noticed that Storybook 6 doesn't like much React 18, but I don't see any issue playing around with the existing Storybook version. Have you noticed anything?

packages/charts/src/state/create_selector.ts
walterra: 👍 updated in 328b62e.

packages/charts/package.json
walterra: From what I see it should still work with Storybook 6, it throws some warnings but React 18 still has the `ReactDOM.render` API with a deprecation warning available (even before this PR it was using React 18 already).

PR #2874, opened 2026-08-24
feat(tooltip): highlight selected tooltip items 

biamalveiro:
## Summary

Adds highlight of chart items when selected in the tooltip. This is useful to preview a a tooltip action, like a filter, or to focus on a subset of data while keeping the overall chart context that is lost when data is filtered.

https://github.com/user-attachments/assets/8e1879ab-09b3-451c-98b2-9f0731d490bb

Now, when a tooltip is pinned and the user selects rows in the tooltip table, we now highlight the matching chart elements, similarly to how we do it in the legend. This applies to XY and partition charts. If there's a tooltip selection that is highlighted and the user hovers on the legend, the legend highlight takes precedence.

## Details

Items can now be highlighted from the legend or from the tooltip, so added a new a new state selector, [`getHighlightedPaths`](https://github.com/biamalveiro/elastic-charts/blob/feat/legend-select-highlight/packages/charts/src/state/selectors/get_highlighted_paths.ts), that given a highlighted path from the legend and a active tooltip values, returns the list of paths that should be highlighted at any moment. 

Everything downstream (XY and partition renderers, and the related selectors), read from this, so the rest of the code should no longer need to know whether the highlight came from the legend or the tooltip. 

In XY, [`getGeometryStateStyle`](https://github.com/biamalveiro/elastic-charts/blob/feat/legend-select-highlight/packages/charts/src/chart_types/xy_chart/rendering/utils.ts#L116) + [`getGeometryHighlightState`](https://github.com/biamalveiro/elastic-charts/blob/feat/legend-select-highlight/packages/charts/src/utils/geometry.ts#L179) were updated to take a list of highlighted items instead of a single one, so multiple series or shapes can be focused/dimmed together.

To make tooltip rows map back to chart elements, each [`TooltipValue`](https://github.com/biamalveiro/elastic-charts/blob/feat/legend-select-highlight/packages/charts/src/specs/tooltip.ts#L68) now has a `path`. 

### For discussion

When right-clicking to pin the tooltip, the currently hovered  elements are immediately selected, which now also means they also get highlighted on the chart. The positive is that this makes it obvious that a selection exists. However, a single right-click going from "hovering" to dimmed chart with only these items focused might feel a bit disruptive?

https://github.com/user-attachments/assets/90f134d8-439a-4bf1-afa8-113c04129f36

If we want to avoid this, an option would be to only apply the tooltip highlight after the user explicitly changes the selection. I didn't want to add that extra complexity without discussing this further. Let me know your thoughts on this!

## Issues

Related to https://github.com/elastic/kibana/issues/271413

### Checklist


packages/charts/src/chart_types/partition_chart/state/selectors/get_highlighted_shapes.ts
```diff
@@ -1,27 +0,0 @@
-/*
```
biamalveiro, line 1: This seems dead code, not used anywhere

biamalveiro:
buildkite update vrt

packages/charts/src/state/selectors/get_highlighted_paths.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/** @internal */
+export function resolveHighlightedPaths(
+  selectedTooltipItems: TooltipValue[],
+  legendHoverPath: LegendPath | null,
+): LegendPath[] {
+  if (legendHoverPath) {
+    return [legendHoverPath];
+  }
+  return selectedTooltipItems.flatMap(({ path }) => (path && path.length > 0 ? [path] : []));
```
markov00, line 27: not my finding (AI), but definitely a valid one:
we could exclude series from being shown in the tooltip via the `filterSeriesInTooltip` prop, that is translated to a 'isVisible' property in the TooltipValue.
We should consider this here and skip invisible tooltip items from being accidentally selected in the tooltip.

examples:
```
It can happen on an XY chart when:

Multiple series have values at the same X position.
filterSeriesInTooltip hides one of those series from the tooltip.
The pointer directly hovers that hidden series’ geometry, making it isHighlighted: true.
Other visible series keep the tooltip populated.
The user right-clicks to pin it.
Pinning selects the hovered hidden series because selection checks only isHighlighted. Its tooltip row is absent because isVisible is false, but its path still highlights that series and dims the visible rows’ series.

Example: bars A and B share an X value; B is filtered from the tooltip. Hover B and right-click. The tooltip shows A, but the chart focuses B and dims A.
```

biamalveiro: fair point, in [41d2ae4](https://github.com/elastic/elastic-charts/pull/2874/commits/41d2ae4f5ed1e8436ecb755af3fa95c7d4c4d4ce)

markov00 (approved):
Changes and behaviour looks good to me!

PR #2877, opened 2026-08-25
fix(canvas): chart reacts to `devicePixelRatio` change

### LINKED ISSUE #2875: Canvas charts blurry at high browser zoom levels (devicePixelRatio not updated on zoom)
## Summary

Canvas-based chart renderers capture `window.devicePixelRatio` **once in the class constructor** and store it as a `readonly` field. When the user zooms the browser (e.g. to 400% for WCAG 1.4.10 Reflow compliance), `window.devicePixelRatio` increases but the cached value is never refreshed. The canvas bitmap buffer stays sized at the old ratio while the browser stretches it to fill the larger CSS area, producing a blurry chart.

Reported downstream in Kibana: elastic/kibana#285614

## Affected chart types

| File | Problem |
|------|---------|
| `src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx` | DPR cached in constructor, no change listener |
| `src/chart_types/heatmap/renderer/canvas/connected_component.tsx` | DPR cached in constructor, no change listener |
| `src/chart_types/partition_chart/renderer/canvas/partition.tsx` | DPR cached in constructor, no change listener |
| `src/chart_types/goal_chart/renderer/canvas/connected_component.tsx` | DPR cached in constructor, no change listener |
| `src/chart_types/bullet_graph/renderer/canvas/index.tsx` | DPR cached in constructor, no change listener |

Each of these contains the same pattern (the `// fixme` comment is already in the code):

```ts
private readonly devicePixelRatio: number; // fixme this be no constant: multi-monitor window drag may necessitate modifying the `<canvas>` dimensions

constructor(props) {
  this.devicePixelRatio = window.devicePixelRatio; // never updated
}
```

And uses it to size the canvas buffer:

```tsx
<canvas
  width={width * this.devicePixelRatio}   // stale
  height={height * this.devicePixelRatio} // stale
  style={{ width, height }}               // CSS grows with zoom, buffer does not
/>
```

## Working reference

`src/chart_types/flame_chart/flame_chart.tsx` already does this correctly — it reads `window.devicePixelRatio` live and registers a `matchMedia` listener to re-render when the value changes:

```ts
window.matchMedia(`(resolution: ${window.devicePixelRatio}dppx)`).addEventListener('change', this.handleDPRChange);
```

`src/chart_types/timeslip/timeslip_chart.tsx` also reads `window.devicePixelRatio` fresh on each render.

## Proposed fix

For each of the five affected renderers:

1. Remove `readonly` from `devicePixelRatio` (or drop the field entirely).
2. Add a `matchMedia` change listener in `componentDidMount` that updates the value and forces a re-render when the DPR changes (identical to the flame chart pattern).
3. Clean up the listener in `componentWillUnmount`.

Reading the value fresh in `render()` alone is not sufficient — without a listener, a zoom change won't trigger a re-render because no props or state changed.

The MDN reference for this pattern: https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio#monitoring_screen_resolution_or_zoom_level_changes

nickofthyme:
## Summary

Fixes #2875

Canvas-based chart types (`xy_chart`, `heatmap`, `partition`, `goal`, `bullet`) captured `window.devicePixelRatio` once in their constructors and never updated it, causing blurry charts when the user zoomed the browser or dragged the window to a different-DPI monitor.

You can see the difference for various charts comparing the VRT changes in https://github.com/elastic/elastic-charts/pull/2877/changes/238d854a0ecb18ea1d2f7fb1bd94bccbd1fed440.

image

## Changes

The fix will add:
- **`packages/charts/src/utils/dpr_observer.ts`** — a module-level singleton that shares a single `matchMedia` listener across all chart instances. When DPR changes, all subscribers are notified and the listener re-registers for the next change.
- **Canvas renderers** (`xy_chart`, `heatmap`, `partition`, `goal`, `bullet`) — subscribe to `subscribeToDPRChange` in `componentDidMount`, update `this.devicePixelRatio`, and call `forceUpdate()` (bypassing `shouldComponentUpdate` which only compares props) to re-render the canvas at the new resolution.

## Test plan

Two new Playwright e2e screenshot suites in `e2e/tests/dpr_zoom.test.ts`, both run at 4× device scale factor:

- **Initial 4x DPR** — navigates to each chart type with `deviceScaleFactor: 4` set from the start; verifies the canvas renders crisply.
- **Change DPR to 4x after mount** — mounts at 1×, then uses CDP to raise `deviceScaleFactor` to 4× and fires a synthetic `matchMedia` `change` event (headless Chromium doesn't fire resolution media query events on CDP overrides); verifies the chart re-renders crisply and the screenshot matches the initial-4× baseline.


nickofthyme:
buildkite update screenshots

nickofthyme:
buildkite update screenshots

nickofthyme:
buildkite test this

packages/charts/src/utils/dpr_observer.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Subscribe to device pixel ratio changes (browser zoom, monitor drag).
+ * A single shared matchMedia listener is used regardless of how many
+ * callers subscribe. Returns an unsubscribe function.
+ * @internal
+ */
+export function subscribeToDPRChange(cb: Callback): () => void {
+  if (subscribers.size === 0) setupListener();
+  subscribers.add(cb);
```
biamalveiro: nice way of only needing one listener 👌

packages/charts/src/chart_types/partition_chart/renderer/canvas/partition.tsx
```diff
@@ -107,6 +106,14 @@ class PartitionComponent extends React.Component<PartitionProps> {
       this.drawCanvas();
       this.props.onChartRendered();
     }
+    this.unsubscribeDPR = subscribeToDPRChange(() => {
+      this.devicePixelRatio = window.devicePixelRatio;
+      this.forceUpdate();
```
biamalveiro: maybe fair for consistency with other charts, but since partition doesn't set its own `shouldComponentUpdate` could dpr live in the component state and skip the `forceUpdate`?

(or https://github.com/elastic/elastic-charts/pull/2877/changes#r3851152798)

packages/charts/src/chart_types/bullet_graph/renderer/canvas/index.tsx
```diff
@@ -100,6 +101,14 @@ class Component extends React.Component<Props> {
       this.drawCanvas();
       this.props.onChartRendered();
     }
+    this.unsubscribeDPR = subscribeToDPRChange(() => {
+      this.devicePixelRatio = window.devicePixelRatio;
+      this.forceUpdate();
```
biamalveiro: I was thinking if there was a way to avoid the `forceUpdate` here 💭 

Maybe one pattern that could work would be moving this dpr subscribe/callback logic to a `withDevicePixelRatio` HOC that would pass `dpr` as a prop to the wrapped component? I think then we could drop the `forceUpdate` + boilerplate across the charts, wdyt?

(I might be missing context on why we went this way, curious what's the reason we for using `shouldComponentUpdate` + `deepEqual` on props across some of the chart components? on the surface feels a bit like we're doing work React should normally do for us?)

packages/charts/src/utils/dpr_observer.ts
```diff
@@ -0,0 +1,34 @@
...
+  });
+}
+
+/**
+ * Subscribe to device pixel ratio changes (browser zoom, monitor drag).
+ * A single shared matchMedia listener is used regardless of how many
+ * callers subscribe. Returns an unsubscribe function.
+ * @internal
+ */
+export function subscribeToDPRChange(cb: Callback): () => void {
```
biamalveiro: cool! really nice way of only needing one listener 👍

packages/charts/src/utils/dpr_observer.ts
markov00, line 1: Thanks for fixing this. I'm wondering why we are doing this outside redux by introducing a custom pub/sub method?

I believe this is an hazard that we should avoid specially if we think the the long run to use react concurrent mode, and doesn't follow the patter we use across the library.

I like to suggest to:
- create a redux property to store and update the devicePixelRatio
- create a pure component, like the Chart Resizer that subscribe and update the store
- pull the DPR from the props on each components that read that

in this I would also suggest to evaluate if we can replace the global Set of subscribes and a singleton listener with a local subscribe and local listener to each chart. I get the reason, one listener many callbacks, but I would prefer an isolated approach per chart as the cost of multiple listeners is probably negligible and reduce the possibilities of missing a unsubscribe keeping these things in memory for too long (rare/impossible case probably right now)

markov00 (changes requested):

packages/charts/src/utils/dpr_observer.ts
nickofthyme: Great idea, I applied your suggestions: redux, selector, isolated listener per chart. See 91a0d4139cfd365ffe3184712c5f4911b5fee7c4

packages/charts/src/chart_types/bullet_graph/renderer/canvas/index.tsx
nickofthyme: Refactored to use redux instead, see 91a0d4139cfd365ffe3184712c5f4911b5fee7c4.

packages/charts/src/chart_types/partition_chart/renderer/canvas/partition.tsx
nickofthyme: Refactored to redux, see 91a0d4139cfd365ffe3184712c5f4911b5fee7c4.

packages/charts/src/components/chart_dpr_observer.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+import { updateDevicePixelRatio } from '../state/actions/chart_settings';
+
+interface DispatchProps {
+  updateDevicePixelRatio: typeof updateDevicePixelRatio;
+}
+
+type Props = DispatchProps;
+
+class DPRObserver extends React.Component<Props> {
```
markov00, line 22: Why a class?  can't we use a function component here?

packages/charts/src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
```diff
@@ -138,7 +136,7 @@ class XYChartComponent extends React.Component<XYChartProps> {
 
   private drawCanvas() {
     if (this.ctx) {
-      renderXYChartCanvas2d(this.ctx, this.devicePixelRatio, this.props, this.animationState);
+      renderXYChartCanvas2d(this.ctx, this.props.devicePixelRatio, this.props, this.animationState);
```
markov00, line 139: nit: we can update the `renderXYChartCanvas2d` arguments, remove the dpr since we are already passing in the `this.props`

markov00 (approved):

packages/charts/src/chart_types/xy_chart/renderer/canvas/xy_chart.tsx
nickofthyme: Good point, fixed in 53b464875d569ab6b4bbb60079fb68662d874f75

packages/charts/src/components/chart_dpr_observer.tsx
nickofthyme: No reason, changed to function, see ddd7b3c2f9a2d04b857e2034274bd9c854aca561.

PR #2886, opened 2026-09-08
feat(metric): add tooltip label position for secondary metrics

biamalveiro:
## Summary

In Metric charts, adds support for displaying the secondary metric label in a tooltip on hover/focus, instead of inline.

https://github.com/user-attachments/assets/afe0a6ea-f764-4992-8f19-dc011122324b

## Details

We extend `labelPosition` from `'before' | 'after'` to `'before' | 'after' | 'tooltip'`. When set to the new `tooltip` option, we don't show the inline label and the label appears in a hover tooltip, using the existing`<Tooltip>` component. Tooltip placement is adjusted depending on secondary metric alignment, that is, the tooltip will be shown on the left if it is right aligned, and vice-versa.

To keep it accessible, if `labelPosition` is the new `tooltip` option, we add a `.echScreenReaderOnly` span with the label so that screen readers can still read the name alongside the value. So we mark the visual tooltip as `aria-hidden`. 

Updated respective knob on the [metric layout story](https://github.com/biamalveiro/elastic-charts/blob/feat/metric-secondary-label-tooltip/storybook/stories/metric/7_layout.story.tsx#L200) to include the tooltip option.

## Issues

Related to https://github.com/elastic/kibana/issues/265733. Follow up, wire this in Lens.

### Checklist

### DIFF
--- a/packages/charts/api/charts.api.md
+++ b/packages/charts/api/charts.api.md
@@ -2908,7 +2908,7 @@ export interface SecondaryMetricProps {
     icon?: string;
     iconPosition?: 'before' | 'after';
     label?: string;
-    labelPosition?: 'before' | 'after';
+    labelPosition?: 'before' | 'after' | 'tooltip';
     style?: CSSProperties;
     value: string;
 }
--- a/packages/charts/src/chart_types/metric/specs/index.ts
+++ b/packages/charts/src/chart_types/metric/specs/index.ts
@@ -33,8 +33,8 @@ export interface SecondaryMetricProps {
   badgeColor?: Color;
   /** Optional text color for the value badge. If not provided, it will be computed automatically for contrast */
   badgeTextColor?: Color;
-  /** Determines whether the label appears before or after the value */
-  labelPosition?: 'before' | 'after';
+  /** Determines whether the label appears before or after the value or is displayed in a tooltip */
+  labelPosition?: 'before' | 'after' | 'tooltip';
   /**  Optional CSS properties to apply to the container element */
   style?: CSSProperties;
   /** Optional aria description */
--- a/packages/charts/src/chart_types/metric/renderer/dom/_secondary_metric.scss
+++ b/packages/charts/src/chart_types/metric/renderer/dom/_secondary_metric.scss
@@ -29,4 +29,9 @@
     text-overflow: ellipsis;
     white-space: nowrap;
   }
+
+  &__tooltip {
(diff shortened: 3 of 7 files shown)


packages/charts/api/charts.api.md
mariairiartef, line 1: question: the idea is to not truncate the text, right?

https://github.com/user-attachments/assets/b58943ab-daf6-473b-a7ed-bb0f908c3d1e

packages/charts/src/chart_types/metric/specs/index.ts
```diff
@@ -33,8 +33,8 @@ export interface SecondaryMetricProps {
   badgeColor?: Color;
   /** Optional text color for the value badge. If not provided, it will be computed automatically for contrast */
   badgeTextColor?: Color;
-  /** Determines whether the label appears before or after the value */
-  labelPosition?: 'before' | 'after';
+  /** Determines whether the label appears before or after the value or is displayed in a tooltip */
+  labelPosition?: 'before' | 'after' | 'tooltip';
```
markov00, line 37: Are we fine with a tooltip that is not an EUI tooltip?
Please double check with Teresa and/or EUI/Design folks. 
Because if not we need to consider a different strategy, by asking the consumer to provide a tooltip component that we can use for this scope.
I'm fine with out tooltip, just wondering

packages/charts/api/charts.api.md
biamalveiro: Good question. Yup, we shouldn't truncate in the tooltip (if we truncate then we'd need a tooltip in the tooltip to have a way to show the full label)

packages/charts/src/chart_types/metric/specs/index.ts
biamalveiro: yeah, fair point, it makes sense for this to be to be able to render in a custom tooltip. I originally aligned it with the chart tooltip, but for integration in Kibana this one is closer to the EUI tooltips behaviorally (showing hidden or truncated text) rather than chart data. I'll make the change.

biamalveiro: added in 0a3769a (+ 5ad8d54, 9c82b9c)

biamalveiro:
buildkite update vrt

biamalveiro:
buildkite test this

markov00 (approved):
Changes works for me, thanks!

PR #2887, opened 2026-09-09
perf: Add performance playground

markov00:
## Summary

Playground/Storybook example that follows the same configuration and data used in the KIbana multi-series performance journey.


markov00:
buildkite update screenshots

monfera (approved):
Thanks for the test case!

PR #2892, opened 2026-09-17
perf(chart): speed up charts by lowering the algorithmic complexity

monfera:
## Summary

Speed up charts with lots of series and/or data points by lowering the algorithmic complexity.

## Details

Speedup on `kibanaManySeriesJourney` (added by https://github.com/elastic/elastic-charts/pull/2887):

```
           Before    After    Speedup
Chrome      4.5 s    0.7 s      6.4 x
Firefox     7.4 s    1.3 s      5.7 x
Safari      9.4 s    2.4 s      3.9 x
```

- remove unintended O(N^2) behavior at five sites, such as linear scan or array rebuild inside a per-item loop
- hoist loop invariants out of per-series and per-datum loops
- reduce per-datum allocation
- (drive-by find) keep datum object shapes consistent so V8 can share a hidden class
- build a lookup index only if needed

There's a bunch more we can do but it's worth going granular.
Also, I haven't checked for this PR if there's something holding Safari back.

Based on the runnable storybook test case [here](https://ech-e2e-ci--pr-2892-fw9986r9.web.app/storybook/?path=/story/test-cases--kibana-many-series-journey&globals=toggles.showHeader:true;toggles.showChartTitle:false;toggles.showChartDescription:false;toggles.showChartBoundary:false;theme:light):

Chrome:
image

Safari:
image

## Issues

Partial iteration on https://github.com/elastic/kibana/issues/236399

### Checklist


packages/charts/src/chart_types/xy_chart/utils/fill_series.ts
```diff
@@ -38,14 +37,12 @@ export function fillSeries(
     const missingValuesArray = [...missingValues.values()];
 
     missingValuesArray.forEach((missingValue) => {
-      const index = sortedXValues.indexOf(missingValue);
-
-      filledData.splice(index, 0, {
```
markov00, line 43: Does this mean that the rest of the code doesn't care about the order of `filledData` items?

packages/charts/src/chart_types/xy_chart/utils/series.ts
```diff
@@ -319,11 +319,13 @@ const getSortedDataSeries = (
   dataSeries: DataSeries[],
   xValues: Set<string | number>,
   xScaleType: ScaleType,
-): DataSeries[] =>
-  dataSeries.map(({ data, ...rest }) => ({
+): DataSeries[] => {
+  const xSortPredicate = datumXSortPredicate(xScaleType, [...xValues.values()]);
+  return dataSeries.map(({ data, ...rest }) => ({
     ...rest,
-    data: data.slice().sort(datumXSortPredicate(xScaleType, [...xValues.values()])),
```
markov00, line 325: 🤦  so bad :D

packages/charts/src/chart_types/xy_chart/utils/stacked_series_utils.ts
```diff
@@ -25,14 +25,18 @@ export interface StackedValues {
...
-  (a: { x: number | string }, b: { x: number | string }) => {
+export const datumXSortPredicate = (xScaleType: ScaleType, sortedXValues?: (string | number)[]) => {
+  let xValueIndices: Map<string | number, number> | undefined;
+  return (a: { x: number | string }, b: { x: number | string }) => {
     if (xScaleType === ScaleType.Ordinal || typeof a.x === 'string' || typeof b.x === 'string') {
-      return sortedXValues ? sortedXValues.indexOf(a.x) - sortedXValues.indexOf(b.x) : 0;
+      if (!xValueIndices && sortedXValues) {
+        xValueIndices = new Map(sortedXValues.map((xValue, index): [string | number, number] => [xValue, index]));
+      }
+      return xValueIndices ? (xValueIndices.get(a.x) ?? -1) - (xValueIndices.get(b.x) ?? -1) : 0;
```
markov00: If we want to we probably speed up this a bit, by passing in the Set (not an array spread) and using a for...of loop to create the map

this is "pseudo code" but you get the point, probably move this inside the `if (!xValueIndices && sortedXValues) {` to instanciate it only if needed), but it could be faster then a spread + map in terms of performance and object allocation

```suggestion
export const datumXSortPredicate = (xScaleType: ScaleType, sortedXValues?: Set<string | number> => {
  const xValueIndices:Map<string | number, number> = new Map();
   let i = 0;
   for (const value of set) {
     xValueIndices.set(value, i++);
   }
  return (a: { x: number | string }, b: { x: number | string }) => {
if (xScaleType === ScaleType.Ordinal || typeof a.x === 'string' || typeof b.x === 'string') {
     
      }
      return xValueIndices ? (xValueIndices.get(a.x) ?? -1) - (xValueIndices.get(b.x) ?? -1) : 0;
```

packages/charts/src/chart_types/xy_chart/utils/stacked_series_utils.ts
```diff
@@ -49,19 +53,22 @@ export function formatStackedDataSeriesValues(
 
   // group data series by x values
   const xMap: XValueMap = new Map();
-  [...xValues].forEach((xValue) => {
-    const seriesMap = new Map<SeriesKey, DataSeriesDatum & { isFiltered: boolean }>();
-    dataSeries.forEach(({ key, data, isFiltered }) => {
-      const datum = data.find(({ x }) => x === xValue);
-      if (!datum) return;
+  xValues.forEach((xValue) => {
```
markov00: a `for...of` would change anything here? just curious

packages/charts/src/chart_types/xy_chart/utils/fill_series.ts
monfera: LMK if I overlooked something-

The ultimate downstream user (path generation) cares about order, and it remains ordered.

`fillSeries` is called only by `computeSeriesDomains`

```typescript
  // fill series with missing x values
  const filledDataSeries = fillSeries(dataSeries, xValues, xDomain.type);

  const formattedDataSeries = getFormattedDataSeries(seriesSpecs, filledDataSeries, xValues, xDomain.type);
  ```

`getFormattedDataSeries` in turn calls `getSortedDataSeries`.

Total order at the use site remains unchanged:
- fillers won't tie with real datums or other fillers
- ties are only possible between two real datums with duplicate x, and their relative order is unchanged by `.sort` (and that `.sort` used to be there anyway)

A minor improvement is that previously, an _ordered_ array was passed around by convention only, rather than type-enforced. It was an unenforced invariant. Now we sort closer to its point of use (again, that `sort` was already present).

(Minor style note, we might reconsider `get`, `compute` etc. function name prefixes.)

packages/charts/src/chart_types/xy_chart/utils/series.ts
monfera: With some luck, JS engines hoisted it out

packages/charts/src/chart_types/xy_chart/utils/stacked_series_utils.ts
monfera: It's better indeed, I just didn't bother with the prior pattern as perf-wise neutral here. 
I've now changed three `forEach` calls: af33f1360bf10a6445c9b1125919bca95c82e6e1

packages/charts/src/chart_types/xy_chart/utils/stacked_series_utils.ts
monfera: Good idea, less allocation 3348b499d46657e8229a8177d8b7afb512ef5952

I didn't tweak it much, as I'm looking into splitting the comparator (better yet the `sort`) into three cases:
- ordinal
- strictly numerical and 
- mixed 

—instead of testing and branching on `ScaleType` and `x` type N*logN times (or worst case, N^2 times) in the comparator. A pre-pass is just O(N) and it could even detect if the series is already in the desired order (though `sort` is approx. O(N) if it's preordered).

markov00 (approved):
Everything looks shine and clean! Thanks for these perf improvements!

Issue #2601, opened 2025-01-28
[Colors] Replace colors with @elastic/eui-theme-borealis


markov00:
Replace [Borealis colors](https://github.com/elastic/elastic-charts/blob/35a5b0a49e78a7730d79907d11764229e58fdb27/packages/charts/src/utils/themes/base_colors.ts) with the respective colors from the `@elastic/eui-theme-borealis` library.

This replacement needs to take place whenever the `eui-theme-borealis` package is in GA release and provides a direct access to each individual color value without the need of a React context.

As agreed with @nickofthyme this dependency (and any associated one) should be added as a `peerDependency` in the `charts` package.

PR #2895, opened 2026-09-22
fix(theme): align dark background color to Kibana dark mode

markov00:
## Summary

A small adjustment to align the charts dark background color to the kibana panels.

Updating to the proper `backgroundBasePlain`

```
backgroundBasePlain: SEMANTIC_COLORS.shade140,
```

https://github.com/elastic/eui/blob/5e35ab9097422732d9c84046eab1762addfec7ff/packages/eui-theme-borealis/src/variables/colors/_colors_dark.ts#L97


markov00:
buildkite update screenshots

markov00:
buildkite test this

mariairiartef (approved):

biamalveiro (commented):
Looks good! 

Maybe a follow-up: wondering how we can be sure that this doesn't drift without us realising?

Our `devDependencies` are still pointing at

```json
    "@elastic/eui": "^99.2.0",
    "@elastic/eui-theme-borealis": "^3.3.0",
    "@elastic/eui-theme-common": "^3.1.0",
```

I understand we want to keep EUI as a dev dependency, but maybe after updating these versions we could add either a unit test comparing our base colors to the EUI theme, or a script that generates the static values from it at build time

biamalveiro (approved):

markov00:
@biamalveiro I believe your suggestion is what was covered by https://github.com/elastic/elastic-charts/issues/2601

PR #2896, opened 2026-09-22
fix(bullet): improve angular responsiveness

biamalveiro:
## Summary

This PR improves how angular bullet charts adjust to small and wide/tall panels. We allow them to become smaller before degrading to metric chart, and, in the case of panels that are too tall/side where there would be either too much empty space or not width/height to display the angular, we degrade to linear.

https://github.com/user-attachments/assets/f72650e6-1739-48b9-926c-2f9b55f73096

### Update degrade to metric threshold
The previous minimum area a static 160x160 size for every angular subtype. Now we set the minimum based on the smallest arc we consider acceptable to render, we set that to be [`MIN_ANGULAR_RADIUS`](https://github.com/biamalveiro/elastic-charts/blob/fix/gauge-responsiveness/packages/charts/src/chart_types/bullet_graph/renderer/canvas/constants.ts#L12-L15), 16px for now, that corresponds to ~128x128 (circle), 128x112 (2/3 circle) and 128x72 (1/2 circle). 

### Tick placement + rendering

A consequence of these smaller sizes is that tick labels are now more likely to overlap (they could overlap before but often we'd switch before that). Here's an example of it happening in Kibana, even before this smaller threshold change:

https://github.com/user-attachments/assets/97271de4-9b51-4e77-837a-8f9e6251e3b4

So, we now [skip rendering the tick labels if they would overlap](https://github.com/biamalveiro/elastic-charts/blob/fix/gauge-responsiveness/packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/angular.ts#L163-L195) with each other or with the bullet.

Before each label was pushed inwards by a fixed padding plus the width of the widest label, that is a short label like 0 ended up further from the bullet than a long label like 100 (this can be seen in the video above as well). So overlaps could happen sooner than they needed to. Now [each label is positioned so that its edge nearest the bullet](https://github.com/biamalveiro/elastic-charts/blob/fix/gauge-responsiveness/packages/charts/src/chart_types/bullet_graph/renderer/canvas/sub_types/angular.ts#L25-L37) is at a fixed distance (padding) and the text grows towards the center.  This is why some screenshots needed updating.

### Degrade to linear on wide/tall panels

There's a new [getRenderMode](https://github.com/biamalveiro/elastic-charts/blob/fix/gauge-responsiveness/packages/charts/src/chart_types/bullet_graph/selectors/get_render_mode.ts#L35-L62) selector that decides which subtype is actually rendered or whether to show the metric instead. This last bit used to live in `getLayout`.

If the graph area is much wider than the arc needs we render horizontal, and if it's much taller, we render vertical. 

The resolved subtype is stored on `BulletPanelDimensions.subtype`.

Happy to get more thoughts on this. It can feel a bit odd to pick an angular chart and get a linear one. But we already override the user's choice when we switch to the metric chart, and that's a much bigger change. A linear bullet in between seems more predictable step to me, so that there isn't such a big visual jump to the metric. That said, it's separate from the main fix, so if you'd rather not include it here because it needs more discussion, just lmk.

## Issues
Related to https://github.com/elastic/kibana/issues/280034

### Checklist


biamalveiro:
buildkite update vrt

biamalveiro:
buildkite update vrt

biamalveiro:
buildkit test this

PR #2897, opened 2026-09-23
test: faster, more accurate local image snapshot comparison

### LINKED ISSUE #2898: Local image snapshots don't match CI, fail on some data fetch and are slow to run
1. Currently, ARM64 is specified for local runs, which is marginally faster (12 vs 14 minutes) but lead to diffs due to rasterization differences relative to AMD64
2. `lens-stress-test` and some upcoming tests hang forever on data fetch, due to an incorrect route
3. It's slow, due to video capture running and failing for each of 1500 tests, generally not needed for local runs

monfera:
# Local visual regression tests to match CI and run faster

Running the VRT suite locally used to produce rasterization diffs, and hung on one story. The PR makes local runs match CI-generated images by switching to AMD64, and cuts the AMD64 runtime from 14m to 5m (even the ARM64 runtime used to be 12mins) measured on a 18-core M5 Max with 16 workers (as embodied in the command lines below–feel free to tweak).

No behavior changes to CI.

The below information will be added to the wiki after some test driving.

Closes https://github.com/elastic/elastic-charts/issues/2898

## Details

Changes:
- Run the VRT container as linux/amd64 instead of linux/arm64. CI unfortunately runs on x86, and SwiftShader's NEON produces rasterization differences from its SSE path. For example, alpha compositing makes it fail, due to zero `maxDiffPixels`.
- Serve all of `public/` at the dev server root instead of under `/public`. Stories fetch data.json relative to the page, so locally that 404's, with `lens-stress-test` timing out. Storybook and the built server both serve it at the root already. Only `devServer` deviated.
- Copy the whole `public/` directory into the built e2e server, not just `data.json`. This is needed as we'll want to remove `kibanaManySeriesJourney` from `storiesToSkip`.
- Add a `VRT_FAST` env flag for local runs. Currently it just disables video recording: under Rosetta, ffmpeg crashes. Skipping it saves about a minute.

## Future doc draft

It's not in a md file yet. Let's have it be in use, settle, and _maybe_ add scripts for these before adding to the doc such as wiki.

### Prerequisites

- Node (version per `.nvmrc`) and Yarn 1.x — via nvm, fnm or similar
- Homebrew (or your favorite alternative)
- Elastic Charts is set up for development
  `cd ~/repos/elastic-charts && yarn install`
    - also on dep changes etc.

### One-time machine setup

``` bash
brew install colima docker
colima start --cpu 18 --memory 24 --vz-rosetta --mount ~/repos:w
docker pull --platform linux/amd64 mcr.microsoft.com/playwright:v1.62.1-noble
```

### After a laptop reboot

```bash
colima start
```

### Running the visual regression test

```bash
cd ~/repos/elastic-charts && yarn test:e2e:generate && NODE_ENV=production yarn test:e2e:server:build && { npx http-server e2e_server/.out -p 9002 --silent & SRV=$!; trap "kill $SRV 2>/dev/null" INT; sleep 2; (cd e2e && VRT_FAST=true yarn test --workers=16 --timeout=10000); kill $SRV; trap - INT; }
```

### Showing the report

```bash
cd ~/repos/elastic-charts/e2e && npx playwright show-report reports/html
```

### Checklist


e2e/playwright.config.ts
```diff
@@ -19,14 +19,16 @@ const isCI = process.env.CI === 'true';
...
+
 const config: PlaywrightTestConfig = {
   use: {
     headless: true,
     locale: 'en-us',
     viewport: { width: 785, height: 1000 },
     trace: 'retain-on-failure',
     screenshot: 'off', // already testing screenshots
-    video: 'retain-on-failure',
+    video: fastMode ? 'off' : 'retain-on-failure',
```
nickofthyme, line 31: You say this saves a minute but that it's due to `ffmpeg` crashing under Rosetta. So I imagine this was the main cause of the slowness not as much the video encoding itself, is that right?

And now running with `linux/amd64` via Colima + Rosetta, wouldn't `ffmpeg` crash less?

Just trying to understand this better. I think keeping this is fine but would be curious what the new comparison would be.

e2e/scripts/start.sh
```diff
@@ -44,13 +44,14 @@ fi
 # Run e2e playwright tests inside container
 docker run \
   --ipc host `# recommended by playwright, see https://playwright.dev/docs/docker#end-to-end-tests` \
-  --platform linux/arm64 `# explicitly set platform` \
+  --platform linux/amd64 `# explicitly set platform` \
```
nickofthyme, line 47: So now running e2e locally will match the ci images perfectly?

nickofthyme (approved):
Changes LGTM, very interesting! 🚀 

I wan't able to test cuz I need to configure docker on my new mac.

e2e/playwright.config.ts
monfera: For successful recording on Rosetta: the fastest local testing spreads work to almost all cores. Recording up to 16 videos simultaneously that are meant to be discarded right away anyway (as all test cases are expected to pass) hogs shared resources. 

It's conditionalized on fast mode, so a local recording can still be made for select, or all stories.

e2e/scripts/start.sh
monfera: At least for now, and on my machine, it's a 100% match.

